PR #10, opened 2020-09-11
NOTIF-14 Modify the accept header so that Json is returned

pilhuhn:
when openapi.json is requested.

Same code as in Policies ui-backend.


theute (approved):
lgtm

PR #60, opened 2020-11-26
NOTIF-57 Bump Quarkus to 1.10.1 and revert changes from NOTIF-54.

pilhuhn:


src/main/java/com/redhat/cloud/notifications/routers/ApplicationService.java
```diff
@@ -26,9 +26,9 @@
     ApplicationResources appResources;
 
     @GET
-    public List<Application> getApplications() {
+    public Uni<List<Application>> getApplications() {
         // Return configured with types?
-        return appResources.getApplications().collectItems().asList().await().indefinitely();
+        return appResources.getApplications().collectItems().asList();
```
josejulio, line 31: I think we can go even further and revert @burmanm workaround to return a `Multi<Application>`

josejulio: Ah yes, check: https://gist.github.com/burmanm/ce371a45b78b290e77763dd5c0dfbd62

josejulio (approved):
Didn't test locally, but changes look good.

PR #73, opened 2020-11-30
NOTIF-67 Skip emails if no one was subscribed (and log it)

josejulio:


theute:
With logs :heart_eyes:

PR #77, opened 2020-12-03
NOTIF-71 Silence db warning about "Duplicate key value for email subs…

josejulio:
…cription"


pilhuhn:
Are we sure we just want to ignore this? Or do we want to inform the caller?
I guess it is harmless to just ignore it.

pilhuhn (approved):

josejulio:
> Are we sure we just want to ignore this? Or do we want to inform the caller?

Just ignore. The user wants to subscribe, but is already subscribed, so no point in tell him otherwise.
Before we ignored this on the java level (Catching the exception), but a WARN is logged and sentry is complaining about that.
Now we are ignoring this on db level for these 3 fields.

PR #85, opened 2020-12-09
Ack before processing.

josejulio:
- Clement's suggestion to ack before processing

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
+++ b/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
@@ -8,13 +8,14 @@ import org.apache.avro.io.DatumReader;
 import org.apache.avro.io.DecoderFactory;
 import org.apache.avro.io.JsonDecoder;
 import org.apache.avro.specific.SpecificDatumReader;
+import org.eclipse.microprofile.reactive.messaging.Acknowledgment;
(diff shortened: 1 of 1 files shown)


src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
```diff
@@ -38,7 +39,7 @@ public EventConsumer(MeterRegistry registry) {
     }
 
     @Incoming("ingress")
-//    @Acknowledgment(Acknowledgment.Strategy.MANUAL)
+    @Acknowledgment(Strategy.PRE_PROCESSING)
     // Can be modified to use Multi<Message<String>> input also for more concurrency
     public Uni<Void> processAsync(Message<String> input) {
```
pilhuhn, line 44: Why not just make this "return" `void`?

pilhuhn, line 44: Should we make this `void` instead of `Uni<Void>` now that we don't return anything anymore?

pilhuhn (approved):

src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
pilhuhn: Probably leave as is for now even if we could make the change

PR #86, opened 2020-12-10
Don't try to check the identity header for those 2 endpoints.

pilhuhn:


theute:
Just a suggestion as I don't know how often we might add stuff through turnpike (and forget to change here)... Should we not prefix with `/internal` and have `/internal/admin` and `/internal/applications` ?

pilhuhn:
@theute Good idea. I'd like to tackle that separately. See NOTIF-83

theute (approved):
I approved, feel free to merge ;-)

PR #87, opened 2020-12-11
Close DB-connections after health-checking

pilhuhn:
If we don't close them, we leak connections and deplete the pool.


theute:
Do you want to fix @Readiness -> @Liveness (I guess) in a different PR ?

pilhuhn:
Good catch. I lost my brutal focus on it, while investigating the connection leak.

PR #91, opened 2020-12-18
[NOTIF-94] EmailTest.testEmailSubscriptionInstant fails in non-US locale

pilhuhn:


josejulio:
I rebased this to latest and added the same change for the new formatter.

josejulio (approved):

PR #93, opened 2021-01-05
Test EndpointType order is not changed

josejulio:
Something I noticed, we need to be careful with the EndpointType enum, and always add new entries at the end of it. The ordinal number is the one stored in the db as "endpoint_type".

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/models/Endpoint.java
+++ b/src/main/java/com/redhat/cloud/notifications/models/Endpoint.java
@@ -15,6 +15,9 @@ import java.util.UUID;
 
 public class Endpoint {
 
+    // Add new values to the bottom of the enum
+    // The ordinal order must remain the same as is used internally
+    // Update test com.redhat.cloud.notifications.models.TestEndpointType to reflect any new enum value
     @Schema(enumeration = { "webhook", "email_subscription", "default" })
     public enum EndpointType {
         @JsonProperty("webhook")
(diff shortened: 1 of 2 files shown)


pilhuhn:
I like the idea here, but wonder if we should not rather explicitly add a property to the enum, which gives the sorting/value for storing. That one could be the same value as the ordinal. My thinking goes that if people have to add a new `ENUMVAL(storeOrder)`, they would not change all the existing values. But then perhaps they would add duplicates. 
This can probably never be fool-proof :-)

src/main/java/com/redhat/cloud/notifications/models/Endpoint.java
```diff
@@ -18,11 +18,11 @@
     @Schema(enumeration = { "webhook", "email_subscription", "default" })
     public enum EndpointType {
         @JsonProperty("webhook")
-        WEBHOOK,
+        WEBHOOK, // 0
```
pilhuhn, line 21: Add a comment on top of the enum, that new values must be added at the end.

pilhuhn (approved):

josejulio:
> I like the idea here, but wonder if we should not rather explicitly add a property to the enum, which gives the sorting/value for storing. That one could be the same value as the ordinal. My thinking goes that if people have to add a new `ENUMVAL(storeOrder)`, they would not change all the existing values. But then perhaps they would add duplicates.
> This can probably never be fool-proof :-)

That could be other option, for the duplicates we could do something to enforce it.
Anyhow I suspect that we will be adding the new ones at the bottom to ensure they are on consecutive order so I don't really know which option would be the best.

src/main/java/com/redhat/cloud/notifications/models/Endpoint.java
josejulio: Done

pilhuhn (approved):

PR #96, opened 2021-01-07
NOTIF-37 Create endpoint call now return response code 201

josejulio:


src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -101,22 +101,27 @@
...
                     .toUni()
                     .onItem()
                     .ifNull()
                     .switchTo(resources.createEndpoint(endpoint));
+        } else {
+            endpointUni = resources.createEndpoint(endpoint);
         }
 
-        return resources.createEndpoint(endpoint);
+        return endpointUni.onItem().transform(e -> Response.created(null).entity(e).build());
```
pilhuhn, line 124: I wonder if the single letter `e` could be confused with "Exception" or "Error". I often use it in `... catch (Exception e) ...`

josejulio: That's true. I'll update.

pilhuhn: Also when I look up the semantics of a 201 I am not sure that matches. See e.g. 
https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

josejulio: mmh.. True. It says that I should return an URL.

josejulio: To be fair, I don't see anywhere that we MUST use `201` when creating resources instead of `200`.

gwenneg: This is written here:

https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html (section `9.5 POST`)

> If a resource has been created on the origin server, the response SHOULD be 201 (Created) and contain an entity which describes the status of the request and refers to the new resource, and a Location header (see section 14.30).

pilhuhn: It says SHOULD. Not MUST :-)

And iirc JavaScript clients seem to have larger issues looking at headers, which is why paging is not via header-fields (`X-href rel=next`), but wrapping the real payload.

theute:
What decision do we make on this old one ? Do we change to 201 or we leave 200 ?

(cc @lahavyuv86 )

bennetelli:
@josejulio ping :)

josejulio:
Closing as is very old now without any resolution

PR #102, opened 2021-01-22
Log when the aggregation starts/ends

josejulio:


theute:
Assigned to me :fear: :-)
Anyway, for a review, sounds good to me.

josejulio:
> Assigned to me :fear: :-)

I heard that you like logs!

PR #103, opened 2021-01-27
NOTIF-91 Endpoint for knowing what notifications an integration impacts

josejulio:


src/test/java/com/redhat/cloud/notifications/routers/NotificationServiceTest.java
```diff
@@ -145,4 +146,133 @@ void testNonExistantDefaults() {
...
+        Response response = given()
+                .header(localIdentityHeader)
+                .when()
+                .contentType(ContentType.JSON)
+                .get("/notifications/eventTypes/affectedByRemovalOfEndpoint/" + ep1.toString())
+                .then()
+                .statusCode(200)
+                .extract().response();
+
+        EventType[] eventTypes = Json.decodeValue(response.getBody().asString(), EventType[].class);
```
pilhuhn, line 177: Just cosmetic and I also need to look that up every time:

you could do something like `String jsonBody = ... .extract().response().bodyAsString(); above and then directly decode here.

pilhuhn (approved):

PR #108, opened 2021-02-03
Prevent applications without templates or aggregating methods to brea…

josejulio:
…k the server

 - It will silently ignore applications without aggregating method or email template
 - Logs errors in the template and aborts the email

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/EmailPayloadAggregator.java
+++ b/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/EmailPayloadAggregator.java
@@ -1,0 +1,60 @@
+package com.redhat.cloud.notifications.processors.email.aggregators;
+
+import com.redhat.cloud.notifications.models.EmailAggregation;
+import io.vertx.core.json.JsonObject;
+
+import java.time.LocalDateTime;
+import java.util.Map;
+
+public abstract class EmailPayloadAggregator {
+
+    private static final String START_TIME_KEY = "start_time";
+    private static final String END_TIME_KEY = "end_time";
+
+    private LocalDateTime startTime;
+    private LocalDateTime endTime;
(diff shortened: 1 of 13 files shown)


josejulio:
@pilhuhn I updated this PR to remove the hard coded values of "insights" and "policies". This also prevents breaking if an app that does not have any email template tries to push a notification. (the email is silently ignored).
//cc: @theute

src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/EmailPayloadAggregator.java
```diff
@@ -0,0 +1,60 @@
+package com.redhat.cloud.notifications.processors.email.aggregators;
+
+import com.redhat.cloud.notifications.models.EmailAggregation;
+import io.vertx.core.json.JsonObject;
+
+import java.time.LocalDateTime;
+import java.util.Map;
+
+public abstract class EmailPayloadAggregator {
```
pilhuhn, line 9: Should we put 'Abstract' in the class name to make it more obvious when glancing at files?

src/main/java/com/redhat/cloud/notifications/templates/EmailTemplate.java
```diff
@@ -0,0 +1,13 @@
+package com.redhat.cloud.notifications.templates;
+
+import com.redhat.cloud.notifications.models.EmailSubscription.EmailSubscriptionType;
+import io.quarkus.qute.TemplateInstance;
+
+public abstract class EmailTemplate {
```
pilhuhn, line 6: Likewise here 'AbstractEmailTemplate'

josejulio: Just updated with the changes.

pilhuhn (approved):

PR #109, opened 2021-02-03
NOTIF-116 Introduce Bundles

pilhuhn:
Introduce Bundles as group of applications. 

Bundle->Applications is 1:n
The application now has a field bundle_id.
Application names are now only unique within a bundle.

The PR also adds some python code in helpers that one can use to quickly generate a Bundle+Application+EventType

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/db/BundleResources.java
+++ b/src/main/java/com/redhat/cloud/notifications/db/BundleResources.java
@@ -1,0 +1,143 @@
+package com.redhat.cloud.notifications.db;
+
+import com.redhat.cloud.notifications.models.Application;
+import com.redhat.cloud.notifications.models.Bundle;
+import io.r2dbc.postgresql.api.PostgresqlResult;
+import io.smallrye.mutiny.Multi;
+import io.smallrye.mutiny.Uni;
+import reactor.core.publisher.Flux;
+
+import javax.enterprise.context.ApplicationScoped;
+import java.util.Date;
+import java.util.UUID;
+
+/**
+ * Deal with Bundles.
+ * More or less stupid copy & paste & search & replace of
+ * code in BundleResources
+ */
+@ApplicationScoped
+public class BundleResources extends  AbstractGenericResource {
+
+    public Uni<Bundle> createBundle(Bundle bundle) {
+        String query = "INSERT INTO public.bundles (name, display_name) VALUES ($1, $2)";
+        // Return filled with id
+        return connectionPublisher.get().onItem()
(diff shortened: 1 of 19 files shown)


src/main/java/com/redhat/cloud/notifications/db/BundleResources.java
```diff
@@ -0,0 +1,143 @@
...
+import reactor.core.publisher.Flux;
+
+import javax.enterprise.context.ApplicationScoped;
+import java.util.Date;
+import java.util.UUID;
+
+/**
+ * Deal with Bundles.
+ * More or less stupid copy & paste & search & replace of
+ * code in BundleResources
```
josejulio, line 17: Did you do search & replace on this comment too? :grin:  (Was expecting to see `code in ApplicationResources`)

pilhuhn: I guess I should change this a bit :)
I startet with c&p&s&r, but then also introduced that common base class (for deletes atm only), which could probably also be used elsewhere

pilhuhn:
This currently fails with GH actions having issues to start Docker

```
Caused by: org.testcontainers.containers.ContainerFetchException: Can't get Docker image: RemoteDockerImage(imageName=postgres:latest, imagePullPolicy=DefaultPullPolicy())
69
Caused by: com.github.dockerjava.api.exception.NotFoundException: 
70
{"message":"No such image: testcontainersofficial/ryuk:0.3.0"}
```

It (still) exists on Dockerhub though: https://hub.docker.com/r/testcontainersofficial/ryuk/tags?page=1&ordering=last_updated

josejulio (approved):
Code looks good. Tests passes locally.

pilhuhn:
Still the docker issue in GH Actions

src/main/java/com/redhat/cloud/notifications/models/Application.java
```diff
@@ -31,6 +33,11 @@
     @JsonInclude(JsonInclude.Include.NON_NULL)
     private Set<EventType> eventTypes; // optional
 
+    @NotNull
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @JsonProperty("bundle_id")
+    private UUID bundleId;
```
josejulio, line 39: This one is marked as non null and at least 1 API call doesn't return it. (It breaks a validation in the openapi.json)

pilhuhn: Good catch and thanks for the PR. I need to improve the tests.

PR #110, opened 2021-02-03
NOTIF-121 Allow to subscribe per application/bundle

josejulio:
Screenshot_2021-02-22_16-29-30


theute:
I'm wondering if we need a "All" such as all existing ones but also future event types

josejulio:
> I'm wondering if we need a "All" such as all existing ones but also future event types

Probably could be extended to cover that case. 
Would be there a "All for this bundle" and the global "All"?

src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -212,50 +216,41 @@
...
+                        type
+                ));
     }
 
     @DELETE
-    @Path("/email/subscription/instant")
+    @Path("/email/subscription/{bundleName}/{applicationName}/{type}")
     @RolesAllowed(RbacIdentityProvider.RBAC_WRITE_INTEGRATIONS_ENDPOINTS)
-    public Uni<Boolean> unsubscribeInstantEmail(@Context SecurityContext sec) {
+    public Uni<Boolean> unsubscribeInstantEmail(
```
pilhuhn, line 240: Nit-pick. As the type is passed as parameter, should the method name changed to no longer contain 'instant' ?
Likewise for subscribe above?

pilhuhn (approved):
lgtm

PR #113, opened 2021-02-09
[NOTIF-123] ID of eventTypes is an integer and not a UUID.

pilhuhn:
Also add (back) a description to the event_type (NOTIF-119)

### DIFF
--- a/src/main/resources/db/migration/V1.6.0__NOTIF-123_event_type_uuid.sql
+++ b/src/main/resources/db/migration/V1.6.0__NOTIF-123_event_type_uuid.sql
@@ -1,0 +1,32 @@
+ALTER TABLE event_type ADD COLUMN description CHARACTER VARYING;
+
+-- ALTER TABLE event_type DROP INDEX "PK_event_type_id";
+ALTER TABLE event_type RENAME COLUMN id TO oldId;
+
+
+
+alter table endpoint_targets drop CONSTRAINT "FK_event_type_id_target" ;
+alter table endpoint_targets rename column event_type_id TO eti; -- out of way
+alter table endpoint_targets add column event_type_id UUID;
+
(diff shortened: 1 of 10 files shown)


src/main/resources/db/migration/V1.6.0__NOTIF-123_event_type_uuid.sql
```diff
@@ -0,0 +1,32 @@
...
+alter table endpoint_targets drop column eti;
+alter table event_type drop column oldId;
+
+update event_type set id =  DEFAULT where id is null;
+
+alter table endpoint_targets add CONSTRAINT "FK_event_type_id_target"
+    foreign key (event_type_id)
+    references event_type(id);
+
+-- ALTER TABLE event_type DROP CONSTRAINT "PK_event_type_id";   DOes not exist?
```
josejulio, line 29: :man_shrugging: 
but after the migration, the primary key is still `id` and is of type `uuid`. Looks like the contraint name was updated to be `event_type_pkey`.

```
# SELECT c.column_name, c.data_type, constraint_name
FROM information_schema.table_constraints tc 
JOIN information_schema.constraint_column_usage AS ccu USING (constraint_schema, constraint_name) 
JOIN information_schema.columns AS c ON c.table_schema = tc.constraint_schema
  AND tc.table_name = c.table_name AND ccu.column_name = c.column_name
WHERE constraint_type = 'PRIMARY KEY' and tc.table_name = 'event_type';
 column_name | data_type | constraint_name 
-------------+-----------+-----------------
 id          | uuid      | event_type_pkey
(1 row)

```

pilhuhn: Yeah, that commented out line should have gone. 
I think dropping the column oldId in line 21 may have deleted it.

josejulio (approved):
LGTM

PR #114, opened 2021-02-11
NOTIF-126 add build_deploy script

lahavyuv86:


build_deploy.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+set -exv
+
+IMAGE="quay.io/notifications-backend"
```
theute, line 5: I guess it's missing the group name. Which I guess is still `cloudservices` ?

build_deploy.sh
```diff
@@ -0,0 +1,29 @@
...
+
+mvn clean package \
+    -DskipTests
+
+DOCKER_CONF="$PWD/.docker"
+mkdir -p "$DOCKER_CONF"
+
+docker --config="$DOCKER_CONF" login -u="$QUAY_USER" -p="$QUAY_TOKEN" quay.io
+docker --config="$DOCKER_CONF" login -u="$RH_REGISTRY_USER" -p="$RH_REGISTRY_TOKEN" registry.redhat.io
+docker --config="$DOCKER_CONF" build -t "${IMAGE}:${IMAGE_TAG}" .
```
theute, line 26: If you use `docker` and not benefit from the Quarkus plugin, I guess it won't find the Dockerfile which lives in `src/main/docker/Dockerfile.jvm`

theute (changes requested):

build_deploy.sh
lahavyuv86: right.. will fix it.

build_deploy.sh
lahavyuv86: `src/main/docker/Dockerfile.jvm` was used via Quarkus, when testing it with docker it looks for Dockerfile at the same path the build_script is located.

theute: I missed the file rename/move in the change log, all good

theute (approved):

PR #119, opened 2021-02-17
NOTIF-133 add clowdapp.yaml

lahavyuv86:


bsquizz:
LGTM, should have no impact since this file isn't wired to an actual deployment yet.

deploy/clowdapp.yaml
```diff
@@ -0,0 +1,85 @@
...
+        readinessProbe:
+          failureThreshold: 3
+          httpGet:
+            path: /
+            port: 8000
+            scheme: HTTP
+          initialDelaySeconds: 35
+          periodSeconds: 5
+          successThreshold: 1
+          timeoutSeconds: 120
```
theute, line 38: Liveness and readiness probe seem wrong they don't match the current ones:
https://github.com/RedHatInsights/e2e-deploy/blob/master/templates/notifications/notifications-backend.yml

deploy/clowdapp.yaml
```diff
@@ -0,0 +1,85 @@
...
+        env:
+        - name: CLOWDER_ENABLED
+          value: ${CLOWDER_ENABLED}
+        resources:
+          limits:
+            cpu: ${CPU_LIMIT}
+            memory: ${MEMORY_LIMIT}
+          requests:
```
theute, line 46: Can we also use parameters like it exists today:
https://github.com/RedHatInsights/e2e-deploy/blob/56d3c6990c7ba1250c536ed691cea45b140f65e3/templates/notifications/notifications-backend.yml#L135-L137

deploy/clowdapp.yaml
```diff
@@ -0,0 +1,85 @@
...
+  required: true
+- description: Image name
+  name: IMAGE
+  value: quay.io/cloudservices/notifications-backend
+- description: Determines Clowder deployment
+  name: CLOWDER_ENABLED
+  value: "false"
+- description: ClowdEnv Name
+  name: ENV_NAME
+  requred: false
```
theute, line 85: `requred` ;-)

theute (changes requested):
I can't really tell if it's correct or not, but just compared with the existing and noticed differences (and a typo)

deploy/clowdapp.yaml
lahavyuv86: Fixed

deploy/clowdapp.yaml
lahavyuv86: Fixed

theute: I really meant to use parameters instead of hard coded 500m and 500Mi values

deploy/clowdapp.yaml
```diff
@@ -0,0 +1,85 @@
...
+        readinessProbe:
+          failureThreshold: 3
+          httpGet:
+            path: /health/ready
+            port: 8000
+            scheme: HTTP
+          initialDelaySeconds: 40
+          periodSeconds: 10
+          successThreshold: 1
+          timeoutSeconds: 1
```
theute, line 38: Port numbers still seem wrong

deploy/clowdapp.yaml
lahavyuv86: Fixed

deploy/clowdapp.yaml
lahavyuv86: Fixed

PR #120, opened 2021-02-17
Patches avro schema

josejulio:


josejulio:
@pilhuhn I can update the file in notifications-gw to use a similar strategy. Or maybe we should split the Action in a separate common repo?

pilhuhn:
This is a good question. I think having the patching in notifications-gw as well would be good.
Is the patching plugin intelligent enough to not try to patch multiple times when `mvn install` is called repeatedly?

josejulio:
> Is the patching plugin intelligent enough to not try to patch multiple times when `mvn install` is called repeatedly?

I haven't had any problem when running  `mvn install` multiple times.

Looking more closely, it looks like on every `mvn install` the `avro-maven-plugin:1.10.0:schema` is run and `maven-patch-plugin:1.2:apply` follows.
In the end the code is only added once to the generated code. If the patch was tried to be run after an already patched file, it would fail.

```
[INFO] --- maven-compiler-plugin:3.8.1:compile (generate-avro-extensions) @ notifications-backend ---
[INFO] Nothing to compile - all classes are up to date
[INFO] 
[INFO] --- avro-maven-plugin:1.10.0:schema (default) @ notifications-backend ---
[INFO] 
[INFO] --- maven-patch-plugin:1.2:apply (patch) @ notifications-backend ---
[INFO] 
```

pilhuhn:
On notifications-gw the Action class is not generated (but hand-copied over from here), so patching there would not work.
Perhaps we need to export the Action + other classes in a shared project from where they would be pulled in. That shared project would take care of the patching then.

josejulio:
> That shared project would take care of the patching then.
Sounds good to me.

> On notifications-gw the Action class is not generated (but hand-copied over from here)
We could still just copy the file there (the patched version is still on the same place)

josejulio:
Probably going to be replaced by #123

PR #122, opened 2021-02-18
NOTIF-134 Fix deleting an eventType

theute:


src/main/java/com/redhat/cloud/notifications/routers/ApplicationService.java
```diff
@@ -70,7 +70,7 @@
     }
 
     @DELETE
-    @Path("/{id}/eventType")
+    @Path("/{id}/eventTypes/{eid}")
```
pilhuhn, line 73: This does not work, as the path expects 2 parameters and the method signature only defines one of the two.

This needs a bit more thinking, as the original one was also flawed.

pilhuhn (changes requested):

src/main/java/com/redhat/cloud/notifications/routers/ApplicationService.java
theute: It passed the testsuite :)

theute: But yes the first {id} is not needed, and extra (useless) work for the API user in the current state of things

theute:
Needs more work

src/main/java/com/redhat/cloud/notifications/routers/ApplicationService.java
pilhuhn: I still believe that is wrong.
If we want to make use of the fact that the id (=PK) of an eventType is a uuid and it does thus not matter for which application it is, then we should not require the app-Id on the path (and possibly move the endpoint to somewhere else).
If we want to verify that the eventType belongs to the app, the we need to pass the app-Id into the method body and also verify it in the backend code.

theute: I agree.
I tried to get into the mold here.

I personally like alignment in the API though.
POST look like:
`/applications/{id}/eventTypes`
so I prefer delete to look like:
`/applications/{id}/eventTypes/{eid}`

You are right, the minimum would be to verify that the app id corresponds to the eventType id.

We can change the POST and make a new endpoint as well. This is what OCM team would prefer (no need to know application id, just bundle and app name).
Now for authorizations, it may be clearer to set authorization on application path.

gwenneg: With the current R2DBC implementation, using a nested resources URL like `/applications/{id}/eventTypes/{eid}` means that we need to add a `join` to the SQL query to check the application/eventType pair. This drawback will go away with Hibernate Reactive though since the `eventTypes` table will contain an `application_id` column because of the `@ManyToOne` relationship.

Now the question is: do we _want_ to use nested URLs in the entire notifications-backend API or not? An identifier like `eid` is a primary key, so we don't _need_ the nested URL to access the event type. But if we keep this nested URL for the event type, then why not apply the same logic for the applications which are nested into a bundle. It would mean an application would be accessible using an URL like `/bundles/{id}/applications/{aid}`? But then, the URL for the event type would become `/bundles/{bid}/applications/{aid}/eventTypes/{eid}`.

I have a preference for shorter URLs whenever it's possible, but this approach has pros and cons. For example, it could be easier for an API user to navigate from a nested resource to its parent with a nest URL. It can also help understanding the relations between resources.

Anyway, I think the current notifications-backend API is a bit messy because it contains both approaches. We should chose one and make sure it is correctly applied everywhere in the API.

gwenneg:
@theute @pilhuhn Could one of you please unresolve the conversation above? I added a comment but I can't unresolve the conversation because I'm not part of the org yet.

src/main/java/com/redhat/cloud/notifications/routers/ApplicationService.java
theute: I completely agree.
I don't have a strong preference. I would say that if we wanted to be clear/human readable we would use names (unique in the context and easier to handle since the app doesn't need to "remember" an id) and nesting rather than ids so something like:
`/bundles/insights/applications/policies`
Otherwise, we should use short URLs with UUID.

The platform docs REST API preference is for the flat version:
https://internal.cloud.redhat.com/docs/reference/rest/urls/

I suggest we go with flat everywhere ie:
/eventTypes/{id}
/bundles/{id}
/applications/{id}

theute: Let's hear from @pilhuhn and @josejulio before we move forward.
(It kinda contradict a previous discussion we had, when a team suggested that they don't need to know about our 'ids', which is why we use bundle/app/eventType names in the messages envelop)

pilhuhn: I agree with https://github.com/RedHatInsights/notifications-backend/pull/122#discussion_r594168472.
Flat for usage with (uu)id, nested for names, as names are not unique under different roots
(e.g. we could have `/bundles/A/applications/X` and `/bundles/B/applications/X` )
And while we are at it, we could shorten `applications` to `apps` :)

theute:
I messed up my rebase completely :-/

gwenneg:
The API needs more work and I'll take care of it in a separate PR. This one was merged to quickly fix the event type deletion.

PR #125, opened 2021-02-24
[NOTIF-112] API to delete applications

pilhuhn:


helpers/create_application.py
```diff
@@ -14,7 +13,12 @@
 event_type = "et1"
 event_type_display_name = "First Event Type"
 
+f = open("/Users/hrupp/insights/policies-ui-backend/server/src/test/resources/rhid_heiko.txt", "r")
```
josejulio, line 16: Could you include a sample rhid without the absolute path? Maybe one that could be used if RBAC is set to off (-Drbac.enabled=false)

josejulio: Also, maybe these changes got committed unintentionally? This is `create_application.py` and the rhid is used to create endpoints.

src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
```diff
@@ -48,9 +48,13 @@
     }
 
     public Uni<Boolean> deleteApplication(UUID applicationId) {
-        String query = "DELETE FROM public.applications WHERE id = $1";
+        // We need to first delete endpoints that point to
+        // event types that are linked to this application
 
-        return runDeleteQuery(applicationId, query);
+        String ep_query = "DELETE FROM endpoint_targets et USING event_type ev  WHERE et.event_type_id = ev.id AND ev.application_id = $1";
```
josejulio, line 54: You won't need to do this anymore after: https://github.com/RedHatInsights/notifications-backend/pull/110/files#diff-cfb416495b160128789c37c83b6a95cd34abc5818859735bcb5354492b9fe51cR20-R24

src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
```diff
@@ -154,10 +158,12 @@
     }
 
     public Uni<Boolean> deleteEventTypeById(UUID eventTypeId) {
-        String query = "DELETE FROM public.event_type WHERE id = $1";
 
-        return runDeleteQuery(eventTypeId, query);
+        // first remove endpoints for the eventType, then the type itself
 
+        String ep_query = "DELETE FROM endpoint_targets WHERE event_type_id = $1";
```
josejulio, line 164: You won't need to do this anymore after: https://github.com/RedHatInsights/notifications-backend/pull/110/files#diff-cfb416495b160128789c37c83b6a95cd34abc5818859735bcb5354492b9fe51cR20-R24

helpers/create_application.py
pilhuhn: Well. That is "end2end" so the functionality is intentional, but the naming of the file is meanwhile no longer good enough.

josejulio (approved):
LGTM

PR #126, opened 2021-03-01
Introduce maven-wrapper

gwenneg:
This PR introduces maven-wrapper which is nice to ensure all developers work with the exact same Maven version.

I think it's also a good idea to maintain a Maven version synced with the one used in Quarkus (see [here](https://github.com/quarkusio/quarkus/blob/master/.mvn/wrapper/maven-wrapper.properties)).

The files were added using `mvn -N io.takari:maven:wrapper`.

If that's OK with you, I'll submit the same kind of PR for the other projects.


mvnw
```diff
@@ -0,0 +1,310 @@
+#!/bin/sh
```
josejulio, line 1: This doesn't have execution permissions, could you add that?

josejulio (changes requested):
Tried in fedora, works well for me.
I had to manually give execution permissions to mvnw. Could you include that in the PR?

josejulio:
One last request, could you update the github actions file to use this `mvnw`? 
https://github.com/RedHatInsights/notifications-backend/blob/master/.github/workflows/build.yml#L24

gwenneg:
Thanks for reviewing this! It should be all good now.

josejulio (approved):

PR #128, opened 2021-03-02
NOTIF-130 Return applications instead of Responses

josejulio:


src/main/java/com/redhat/cloud/notifications/routers/ApplicationService.java
```diff
@@ -29,14 +29,14 @@
...
         // Return configured with types?
         if (bundleName == null || bundleName.isBlank()) {
-            return Multi.createFrom().failure(new IllegalArgumentException("there is no bundle name given. Try ?bundleName=xxx"));
+            return Uni.createFrom().failure(new IllegalArgumentException("there is no bundle name given. Try ?bundleName=xxx"));
         }
         // TODO how can we find out there is nothing to send and return a 404 ?
-        return appResources.getApplications(bundleName).onItem()
-                .transform(app -> { return Response.ok(app).build(); });
+        return appResources.getApplications(bundleName)
+                .collectItems().asList().onItem().transform(apps -> Response.ok(apps).build());
```
gwenneg, line 39: Returning a `Uni<List<Application>>` in this method should produce the same output:
```suggestion
        return appResources.getApplications(bundleName).collectItems().asList();
```

src/test/java/com/redhat/cloud/notifications/routers/ApplicationServiceTest.java
```diff
@@ -147,4 +145,53 @@ void testGetApplicationsRequiresBundleName() {
                 .statusCode(500);
 
     }
+
+    @Test
+    void testGetApplicationsReturnApplications() {
```
gwenneg, line 150: Would it be possible to improve `testPoliciesApplicationAddingAndDeletion()` instead of adding this test? It looks like they have a lot in common but maybe I missed a significant difference.

gwenneg (changes requested):

src/test/java/com/redhat/cloud/notifications/routers/ApplicationServiceTest.java
josejulio: They share a very similar setup, but the other is testing adding/deletion. This one focus on verifying that the contents of the API call is indeed an Application (in this case, I added 10 applications). IMO is OK to repeat some code to have a more focused test when possible.

If needed, the supporting setup code could be refactored in some helper method to be more concise.

src/main/java/com/redhat/cloud/notifications/routers/ApplicationService.java
josejulio: Ah yes, that would be better and I think the openapi would get generated more natural this way.

src/test/java/com/redhat/cloud/notifications/routers/ApplicationServiceTest.java
gwenneg: Makes sense, let's keep it that way.

src/main/java/com/redhat/cloud/notifications/routers/ApplicationService.java
```diff
@@ -29,14 +30,17 @@
     ApplicationResources appResources;
 
     @GET
-    public Multi<Response> getApplications(@QueryParam("bundleName") String bundleName) {
+    public Uni<List<Application>> getApplications(@QueryParam("bundleName") String bundleName) {
         // Return configured with types?
         if (bundleName == null || bundleName.isBlank()) {
-            return Multi.createFrom().failure(new IllegalArgumentException("there is no bundle name given. Try ?bundleName=xxx"));
+            return Uni.createFrom().failure(new IllegalArgumentException("there is no bundle name given. Try ?bundleName=xxx"));
```
gwenneg, line 36: While we're at it, throwing a `javax.ws.rs.BadRequestException` here would probably be better from the client perspective. Right now it fails with an HTTP error 500 and the exception message is not sent to the client.

josejulio: +1

src/test/java/com/redhat/cloud/notifications/routers/ApplicationServiceTest.java
gwenneg: I can't resolve my own comment because I don't have the required rights on the repo, but this is resolved 😄

src/main/java/com/redhat/cloud/notifications/routers/ApplicationService.java
josejulio: Done

gwenneg (approved):
LGTM

josejulio:
Thanks @gwenneg !

PR #131, opened 2021-03-09
NOTIF-142 openapi fixes

josejulio:
- Add description to 204 contents
- Remove some types that were strange (JsonObject as array)
- Remove internal types

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
+++ b/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
@@ -135,7 +135,7 @@ public class EndpointService {
     @DELETE
     @Path("/{id}")
     @RolesAllowed(RbacIdentityProvider.RBAC_WRITE_INTEGRATIONS_ENDPOINTS)
-    @APIResponse(responseCode = "204", content = @Content(schema = @Schema(type = SchemaType.STRING)))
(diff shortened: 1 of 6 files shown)


src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -135,7 +135,7 @@
     @DELETE
     @Path("/{id}")
     @RolesAllowed(RbacIdentityProvider.RBAC_WRITE_INTEGRATIONS_ENDPOINTS)
-    @APIResponse(responseCode = "204", content = @Content(schema = @Schema(type = SchemaType.STRING)))
+    @APIResponse(responseCode = "204", description = "No Content", content = @Content(schema = @Schema(type = SchemaType.STRING)))
```
pilhuhn, line 138: Shouldn't we have a better description? No Content is what 204 means already. Does that imply the action was successful? What do we return if not?

josejulio: I agree that we can expand the description to be something more useful now that we are adding it. 
`"No content"` is what Quarkus uses by default if the function returns void (Looks like a required openapi field at least for swagger). In our current situation it just simply skips it, that's why we had to add it manually.

I'll expand the description to something more inline with the function.

josejulio: I updated all the 204's descriptions.

pilhuhn (approved):

PR #133, opened 2021-03-12
Remove Yasson from test dependencies

gwenneg:
While investigating a deserialization issue with the Hibernate Reactive migration branch, I noticed Yasson in the POM. Since Jackson is already available and the app is relying on it, I think we can get rid of Yasson.


josejulio (approved):

gwenneg:
Thanks @josejulio!

PR #134, opened 2021-03-12
Fix broken event type deletion

gwenneg:
This fixes and tests `DELETE /internal/applications/{id}/eventType`.


gwenneg:
I just noticed that #122 took care of the same issue. Closing this one as it is a duplicate.

PR #135, opened 2021-03-15
Multi-step build

theute:
We would have to figure out how to run the testsuite, non trivial here as the testsuite starts docker container, and the build is now running inside a docker container


lahavyuv86:
LGTM

PR #136, opened 2021-03-16
Fix user preferences

josejulio:


src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
```diff
@@ -124,9 +126,9 @@
                                             .onItem().transformToMulti(emailSubscriptionType -> {
                                                 values.bundles.get(bundle.getName()).applications.get(application.getName()).notifications.put(emailSubscriptionType, false);
                                                 return Multi.createFrom().empty();
-                                            }).merge();
-                                }).merge();
-                    }).merge().collectItems().asList().onItem().transformToMulti(objects -> emailSubscriptionResources.getEmailSubscriptionsForUser(account, username))
+                                            }).concatenate();
+                                }).concatenate();
```
gwenneg, line 130: In theory, `merge` should be preferred over `concatenate` for these two lines because there's no concurreny issue (no database access) and the parallel invocations of the mappers with `merge` _could be_ completed faster than the sequential mappers invocations with `concatenate`.

But that's only theory. I don't think it will change anything in terms of performance here since the mappers invocation time is close to zero :)

gwenneg (approved):
LGTM

PR #138, opened 2021-03-17
[RFR] add latest tag to build_deploy.sh

lahavyuv86:


theute:
Thanks ! 
Let's keep it open for now until we transition, the lack of "master" for now helps us figuring out which of the build systems was used when looking at Quay

theute:
Merging now, since CI depends on it, and then @lahavyuv86  you can disable the "old" build mechanism I guess

PR #139, opened 2021-03-18
NOTIF-138 Replace jsonb with text in database

gwenneg:
The `jsonb` native type from PostgreSQL isn't natively supported by Hibernate (Reactive or ORM). Someone from the Quarkus community created a [Quarkiverse extension](https://github.com/quarkiverse/quarkus-hibernate-types) (based on Vlad Mihalcea's [hibernate-types](https://github.com/vladmihalcea/hibernate-types)) to add `jsonb` support to Hibernate ORM with Quarkus, but it does not work with Hibernate Reactive.

@josejulio I tested this locally, everything seemed OK from a backend point of view. My tests included manipulating data saved from `master` with the `jsonb` type and then reused from this branch with the migrated `text` type.

As we discussed earlier, could you please test this with the full notifications app?

cc @pilhuhn

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/db/NotificationResources.java
+++ b/src/main/java/com/redhat/cloud/notifications/db/NotificationResources.java
@@ -4,7 +4,6 @@ import com.redhat.cloud.notifications.models.NotificationHistory;
 import io.r2dbc.postgresql.api.PostgresqlConnection;
 import io.r2dbc.postgresql.api.PostgresqlResult;
 import io.r2dbc.postgresql.api.PostgresqlStatement;
-import io.r2dbc.postgresql.codec.Json;
 import io.smallrye.mutiny.Multi;
 import io.smallrye.mutiny.Uni;
 import io.smallrye.mutiny.converters.multi.MultiReactorConverters;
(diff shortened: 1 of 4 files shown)


josejulio:
I didn't see any change in the backend API. The UI is working fine :+1:

src/main/java/com/redhat/cloud/notifications/db/NotificationResources.java
```diff
@@ -41,9 +40,9 @@
                             .bind("$5", history.getEventId());
 
                     if (history.getDetails() != null) {
-                        st.bind("$6", Json.of(new JsonObject(history.getDetails()).encode()));
+                        st.bind("$6", new JsonObject(history.getDetails()).encode());
```
josejulio, line 43: Just wondering, no change is required to "load" from the database?

josejulio (approved):
LGTM

src/main/java/com/redhat/cloud/notifications/db/NotificationResources.java
gwenneg: The `details` field was already loaded as a String [here](https://github.com/RedHatInsights/notifications-backend/blob/8a5f888055194825c492ec0b2a93d5c623760110/src/main/java/com/redhat/cloud/notifications/db/NotificationResources.java#L101) so there's nothing to change.

gwenneg:
Thanks for testing this! I think we should merge and confirm again everything's fine after it's been deployed.

PR #140, opened 2021-03-19
NOTIF-138 Prepare tests for the Hibernate Reactive migration

gwenneg:

Unlike R2DBC or Hibernate ORM, Hibernate Reactive does not allow us to retrieve values (UUID or timestamp) generated by PostgreSQL during an `INSERT` or an `UPDATE`. Because of that, if we want to keep sending generated values in the REST responses without performing an additional `SELECT` each time we change something in the database, the generation has to happen in the Java code (which is fairly easy with Hibernate). But the Hibernate UUID generator is only run if the `id` attribute is `null` when the data is about to be persisted, so we have to make sure that an `id` will never come from an API request. Therefore, in the Hibernate Reactive migration PR, all API classes identifiers will be annotated with `@JsonProperty(access = READ_ONLY)` to make sure the UUID generation is working fine. Same goes for timestamps which should not come from the API requests. The problem with read-only attributes is that we can't deserialize REST responses into API classes anymore in our tests because Jackson now knows that the attributes are read-only. This is why I had to refactor the tests a bit.

▶️ The review is easier than it looks and the tests coverage is exactly the same as before.

### DIFF
--- a/src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java
+++ b/src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java
@@ -15,7 +15,6 @@ import com.redhat.cloud.notifications.models.EventType;
 import com.redhat.cloud.notifications.models.HttpType;
 import com.redhat.cloud.notifications.models.NotificationHistory;
 import com.redhat.cloud.notifications.models.WebhookAttributes;
-import com.redhat.cloud.notifications.routers.models.EndpointPage;
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.quarkus.test.common.QuarkusTestResource;
@@ -25,6 +24,7 @@ import io.restassured.http.ContentType;
 import io.restassured.http.Header;
 import io.restassured.response.Response;
 import io.vertx.core.json.Json;
+import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import org.eclipse.microprofile.reactive.messaging.Channel;
 import org.eclipse.microprofile.reactive.messaging.Emitter;
@@ -41,6 +41,7 @@ import java.time.LocalDateTime;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
@@ -58,6 +59,13 @@ import static org.mockserver.model.HttpResponse.response;
 @TestMethodOrder(MethodOrderer.Alphanumeric.class)
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 public class LifecycleITest {
+
+    /*
+     * In the tests below, most JSON responses are verified using JsonObject/JsonArray instead of deserializing these
+     * responses into model instances and checking their attributes values. That's because the model classes contain
+     * attributes annotated with @JsonProperty(access = READ_ONLY) which can't be deserialized and therefore verified
+     * here. The deserialization is still performed only to verify that the JSON responses data structure is correct.
+     */
 
     private static final String APP_NAME = "policies-lifecycle-test";
     private static final String BUNDLE_NAME = "my-bundle";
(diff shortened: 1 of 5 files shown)


josejulio:
> so we have to make sure that an `id` will never come from an API request

There are some requests that send the `id`  such as the edit ~Integration~ Endpoint.

> The problem with read-only attributes is that we can't deserialize REST responses into API classes anymore

Wouldn't this make tests harder to understand? 
Does this affect the `openapi.json` file? The UI relies heavily on it to generate the types used for the requests.

gwenneg:
> There are some requests that send the `id` such as the edit ~Integration~ Endpoint.

I don't know why I didn't think about that... I guess I didn't use the UI enough 😃 This PR needs more work, don't bother reviewing the content.

Thanks @josejulio

josejulio:
> I don't know why I didn't think about that... I guess I didn't use the UI enough  This PR needs more work, don't bother reviewing the content.

To be fair,  I'm not yet sure if we send the id as part of the payload or in the url (or both). And even if we only send it on the URL, not sure where the backend is using it up from.

gwenneg:
I checked the code, the Endpoint edit is the only operation where the ID from the payload could have been used, but we're not doing that. The ID is also passed as a `@PathParam` and that's the one we use, so making all IDs read-only in payloads is safe and won't break anything.

I also checked the OpenAPI files with the full Hibernate Reactive migration changes, the read-only IDs have no effect the the OpenAPI definitions.

Finally, let's talk about tests readability. Here are two examples of the kind of changes this PR contains:

```
// Before PR:
Application appResponse = Json.decodeValue(response.getBody().asString(), Application.class);
assertNotNull(appResponse.getId());
assertEquals(theBundle.getId(), appResponse.getBundleId());

// After PR:
JsonObject appResponse = new JsonObject(response.getBody().asString());
appResponse.mapTo(Application.class);
assertNotNull(appResponse.getString("id"));
assertEquals(theBundle.getString("id"), appResponse.getString("bundle_id"));
```

```
// Before PR:
Endpoint[] endpoints = Json.decodeValue(response.getBody().asString(), EndpointPage.class).getData().toArray(new Endpoint[0]);
assertEquals(2, endpoints.length);
for (Endpoint ep: endpoints) {
    // Omitted code...
}

// After PR:
JsonArray endpoints = new JsonObject(response.getBody().asString()).getJsonArray("data");
assertEquals(2, endpoints.size());
for (int i = 0; i < endpoints.size(); i++) {
     JsonObject ep = endpoints.getJsonObject(i);
     ep.mapTo(Endpoint.class);
    // Similar omitted code...
}
```
This sums up pretty much everything there is to know what this PR. I don't think this affects the tests readability.

WDYT?

josejulio:
> This sums up pretty much everything there is to know what this PR. I don't think this affects the tests readability.

+1

josejulio (approved):
lgtm

gwenneg:
Fixing the conflicts...

src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java
```diff
@@ -97,24 +107,25 @@ void setup() {
...
-        theBundle =
-            given()
+        Response response = given()
                 .body(bundle)
                 .contentType(ContentType.JSON)
                 .basePath("/")
-            .when()
-                    .post("/internal/bundles")
-            .then()
+                .when()
```
pilhuhn, line 114: I know IDEs don't really like the mixed indent. I like it when `given()`, `when()`, ` then()` are less indented as this sort of shows the control structure as with normal if/then/else blocks

pilhuhn:
Why are the tests easier to read when we use the more abstract JsonObject than the concrete model class?

pilhuhn (approved):

src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java
gwenneg: Unfortunately I didn't change that one on purpose, that's IntelliJ's automatic formatting.

pilhuhn: I know - this is why I say IDEs don't like that

gwenneg:
> Why are the tests easier to read when we use the more abstract JsonObject than the concrete model class?

I didn't mean that they're easier to read, but only that they're (almost) not harder to read. Nothing can beat the concrete class.

PR #141, opened 2021-03-19
NOTIF-148 Uses a new format for notification preferences.

josejulio:
Supports https://github.com/RedHatInsights/user-preferences-frontend/pull/60

To avoid any possible issue while moving from `user-preferences/email` to `user-preferences/notifications` I'm just adding a copy of the old format that should be removed after adding all the required info in configuration of user-preferences and cloud-services-config.

To support the use case of policies knowing if the user is subscribed to any event (instant/daily); added a new endpoint that receives the bundle and application and returns the instant/daily configuration.

Old* files are a copy of what they were before this PR to prevent breaking anything while we migrate data.


src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
```diff
@@ -89,15 +122,37 @@
...
+        Uni<EmailSubscription> uniDaily = emailSubscriptionResources.getEmailSubscription(account, name, bundleName, applicationName, EmailSubscriptionType.DAILY);
+        Uni<EmailSubscription> uniInstant = emailSubscriptionResources.getEmailSubscription(account, name, bundleName, applicationName, EmailSubscriptionType.INSTANT);
+
+        return Uni.combine().all().unis(uniDaily, uniInstant).combinedWith((daily, instant) -> {
+            UserConfigPreferences preferences = new UserConfigPreferences();
+            preferences.setDailyEmail(daily != null);
+            preferences.setInstantEmail(instant != null);
+            return preferences;
+        });
+    }
```
gwenneg, line 144: This will execute the SQL queries concurrently because of `Uni.combine()`. Could you chain the two calls?

Something like this should work:
```
getEmailSubscription(DAILY)
.onItem().transform(daily -> /* create and return partial preferences */)
.onItem().transformToUni(preferences -> getEmailSubscription(INSTANT)
    .onItem().transform(instant -> /* add missing preferences field and return final object */)
)
```

src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
```diff
@@ -136,14 +192,59 @@
...
+                    }).collectItems().asList().onItem().transform(emailSubscriptions -> values);
+        });
+    }
+
+    // Todo: Remove after all settings are in place for notifications userpreferences section
+    private Uni<String> oldSettingsValuesToJsonForm(Uni<SettingsValues> settingsValuesUni) {
+        return settingsValuesUni.onItem().transform(settingsValues -> OldSettingsValueJsonForm.fromSettingsValue(settingsValues))
+                .onItem().transform(settingsValueJsonForm -> {
+                    try {
+                        return mapper.writeValueAsString(settingsValueJsonForm);
```
gwenneg, line 230: We should inject and use the `ObjectMapper` produced by Quarkus there, but this is not new so it can be fixed later.

src/main/java/com/redhat/cloud/notifications/routers/models/SettingsValueJsonForm.java
```diff
@@ -60,12 +80,12 @@
...
+                    fieldsField.fields.add(field);
                 });
             });
         });
 
-        return forms;
+        return form;
     }
 
     private static Field createLabelField(String label, List<Field> parentContainer) {
```
gwenneg, line 91: Shouldn't this method be removed?

gwenneg: That comment was meant for the new version of the file.

src/main/java/com/redhat/cloud/notifications/routers/models/SettingsValueJsonForm.java
```diff
@@ -29,16 +31,34 @@
...
+            name = "notification-preferences";
+            component = COMPONENT_SECTION;
+        }
+    }
 
-    public static List<SettingsValueJsonForm> fromSettingsValue(SettingsValues values) {
-        ArrayList<SettingsValueJsonForm> forms = new ArrayList<>();
+    public List<Sections> fields = new ArrayList<>();
+
+    public static SettingsValueJsonForm fromSettingsValue(SettingsValues values) {
```
gwenneg, line 46: This part of the code is a bit confusing for me but I suppose that's because I don't have a clear picture of what's going on in the UI. I don't know if that's possible but maybe a few comments would make it clearer.

src/main/java/com/redhat/cloud/notifications/routers/models/UserConfigPreferences.java
```diff
@@ -0,0 +1,22 @@
+package com.redhat.cloud.notifications.routers.models;
+
+public class UserConfigPreferences {
```
gwenneg, line 3: Snake case should probably be enforced here with `@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)`.

src/test/java/com/redhat/cloud/notifications/routers/UserConfigServiceTest.java
```diff
@@ -91,14 +89,14 @@ void testSettings() {
         String bundle = "insights";
         String application = "policies";
 
-        List<SettingsValueJsonForm> jsonForms = given()
+        SettingsValueJsonForm jsonForm = given()
                 .header(identityHeader)
-                .when().get("/user-config/email-preference")
+                .when().get("/user-config/notification-preference?bundleName=insights")
```
gwenneg, line 94:
```suggestion
                .queryParam("bundleName", bundle)
                .when().get("/user-config/notification-preference")
```
Same for the 4 similar calls below.

gwenneg (changes requested):

src/main/java/com/redhat/cloud/notifications/routers/models/SettingsValueJsonForm.java
gwenneg: I know this is quite urgent so comments could also be added later.

src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
josejulio: Sure.

Curious about this, is there any issue on executing both concurrently?

gwenneg: With R2DBC it will work fine, but Hibernate will complain because the same `Session` is used concurrently.

src/main/java/com/redhat/cloud/notifications/routers/models/SettingsValueJsonForm.java
josejulio: good catch, looks like it is not used anymore.

src/main/java/com/redhat/cloud/notifications/routers/models/UserConfigPreferences.java
josejulio: This doesn't seem to update the openapi.json file accordingly.
My tests shows that the received data is on snake_case, but openapi shows camelCase.

josejulio: Not something urgent, as the API is hidden anyway.

gwenneg: I ran into the same issue with the Hibernate Reactive branch. I couldn't find an application-wide setting so I ended up using `@Schema(name = "some_snake_case_name")` on each attribute with a camel case name.

josejulio: I ran into other issue with the `@RestClient` in policies code (it calls this service).  
The return value doesn't seem to honor the `@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)` or the `@Schema(name = "some_snake_case_name")`.

gwenneg: I just found an issue about the snake case strategy support during OpenAPI files generation: https://github.com/smallrye/smallrye-open-api/issues/712.

src/main/java/com/redhat/cloud/notifications/routers/models/SettingsValueJsonForm.java
gwenneg: Thanks for the comments and link!

gwenneg (approved):
LGTM

PR #142, opened 2021-03-22
NOTIF-138 Move static inner classes out of future JPA entities

gwenneg:

Nothing fancy in this one. It only moves static inner classes / enums out of the upcoming JPA entities because the entities will contain both JPA annotations (persistence layer) and Jackson annotations (API layer) and that's more than enough :)

There isn't a single change in the code apart from the move.


gwenneg:
There will be a few conflicts (very easy to solve) with #140.

josejulio (approved):
lgtm

pilhuhn:
I like that.

PR #143, opened 2021-03-22
NOTIF-138 Work around Quarkus flyway/hibernate-reactive incompatibility

gwenneg:

`quarkus-hibernate-reactive` is not currently compatible with `quarkus-flyway` because:

- Flyway requires a JDBC connection
- `quarkus-flyway` creates a JDBC datasource to satisfy Flyway's requirement
- `quarkus-hibernate-reactive` cannot start if there is a JDBC datasource in the application

This incompatibility is in the process of being fixed. More infos in the following Quarkus ticket: https://github.com/quarkusio/quarkus/issues/10716

In the meantime, we can work around this issue with this PR which can be merged into `master` as soon as it has been reviewed.

FYI, the same issue happens with `quarkus-liquibase`.

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/db/FlywayWorkaround.java
+++ b/src/main/java/com/redhat/cloud/notifications/db/FlywayWorkaround.java
@@ -1,0 +1,34 @@
+package com.redhat.cloud.notifications.db;
+
+import io.quarkus.runtime.StartupEvent;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.flywaydb.core.Flyway;
+import org.jboss.logging.Logger;
+
+import javax.enterprise.context.ApplicationScoped;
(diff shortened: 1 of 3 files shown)


gwenneg:
I forgot to add that the workaround adds a WARN log entry at application start:
```
2021-03-22 15:03:36,018 WARN  [com.red.clo.not.db.FlywayWorkaround] (main) Starting Flyway workaround... remove it ASAP!
```

josejulio:
I think we should create a JIRA for removing the workaround.

josejulio (approved):

gwenneg:
> I think we should create a JIRA for removing the workaround.

Good point, thanks! I'll do it right away.

gwenneg:
https://issues.redhat.com/browse/NOTIF-158

PR #144, opened 2021-03-23
Bump Quarkus to 1.13.0.Final

gwenneg:
We need that Quarkus version for many reasons, including Hibernate Reactive.

Tests are failing, it seems to be caused by:

https://github.com/RedHatInsights/notifications-backend/blob/c46eecf5d808081fb7fef1bf7a52acecd6999422/src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java#L84-L85

Smallrye Reactive Messaging is throwing the exception and its version was bumped in the Quarkus release. I'm talking with them to figure out what's wrong and how to fix that.

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/auth/RHIdentityAuthMechanism.java
+++ b/src/main/java/com/redhat/cloud/notifications/auth/RHIdentityAuthMechanism.java
@@ -9,11 +9,9 @@ import io.quarkus.vertx.http.runtime.security.ChallengeData;
 import io.quarkus.vertx.http.runtime.security.HttpAuthenticationMechanism;
 import io.quarkus.vertx.http.runtime.security.HttpCredentialTransport;
 import io.smallrye.mutiny.Uni;
-import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.RoutingContext;
 
 import javax.enterprise.context.ApplicationScoped;
-import java.util.Base64;
 import java.util.Collections;
 import java.util.Set;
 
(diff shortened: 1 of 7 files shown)


gwenneg:
Here's the root cause of the startup failure:
`io.smallrye.reactive.messaging.wiring.TooManyUpstreamCandidatesException: 'SubscriberMethod{method:'com.redhat.cloud.notifications.events.EventConsumer#processAsync', incoming:'ingress'}' supports a single upstream producer, but found 2: [Emitter{channel:'ingress'}, IncomingConnector{channel:'ingress', attribute:'mp.messaging.incoming.ingress'}]. You may want to add the '@Merge' annotation on the method.`

I had to debug the app to find that, it's not printed in the log.

I added a workaround commit but it's not a proper fix, we shouldn't need to `@Merge` multiple producers on the `Ingress` channel. Working on something better...

gwenneg:
All problems solved in a clean way! Let's wait for the final release which is planned for tomorrow before this is merged.

gwenneg:
Quarkus 1.13.0.Final was released today but now we have a lot of tests failing because of an HTTP 401 status being returned. They all worked fine with 1.13.0.CR1... I'll investigate that tomorrow.

gwenneg:
This is ready for a review.

I had to refactor the authentication code because of https://github.com/quarkusio/quarkus/commit/b3378b45a6c904f2d602c467d23232f86c78b5ff. Because of that commit, it is no longer possible to build a `QuarkusSecurityIdentity` with an empty `principal` and an `anonymous` flag set to false. We were building two different `QuarkusSecurityIdentity` during the authentication process and kind of merging them at the end. Now everything is done with a single builder.

src/main/java/com/redhat/cloud/notifications/auth/RHIdentityAuthMechanism.java
```diff
@@ -51,7 +49,7 @@
                     good = true;
                 }
             } else if (path.startsWith("/openapi.json") || path.startsWith("/internal")
-                    || path.startsWith("/admin") || path.startsWith("/health")) {
+                    || path.startsWith("/admin") || path.startsWith("/health") || path.startsWith("/q/health")) {
```
pilhuhn, line 52: Same should apply for /metrics and /q/metrics.
Or we do:

```
# Quarkus since 1.11 redirects non-apps to /q/. We need to prevent this
quarkus.http.redirect-to-non-application-root-path=false
quarkus.http.non-application-root-path=/
```

gwenneg: I wouldn't change the default Quarkus behavior without a good reason, so I'm in favor of adding `/metrics` and `/q/metrics` to the list.

pilhuhn (approved):

src/main/java/com/redhat/cloud/notifications/auth/RHIdentityAuthMechanism.java
pilhuhn: The /q thing can be tricky, as we learned with openapi. If you request /openapi.json, it will return a 302 redirect which clients (sometimes) can't understand

gwenneg: That sounds like the good reason I needed 😄

pilhuhn: See POL-443

PR #146, opened 2021-03-23
NOTIF-138 Fix database concurrency issues

gwenneg:

It prevents exisiting (and yet mostly invisible) concurrent accesses to the database which will become a blocking problem with Hibernate Reactive since the `Session` (and the underlying `EntityManager`) should not be used concurrently.

@cescoffier Could you please review these changes? I know there are many other things we need to fix regarding the way Mutiny is used in this project. We will take care of them soon, but not in this PR :)

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/events/EndpointProcessor.java
+++ b/src/main/java/com/redhat/cloud/notifications/events/EndpointProcessor.java
@@ -62,15 +62,12 @@ public class EndpointProcessor {
                 action.getBundle(),
                 action.getApplication(),
                 action.getEventType())
-                .onItem()
-                .transformToUni(endpoint -> {
+                .onItem().transformToUniAndConcatenate(endpoint -> {
                     endpointTargeted.increment();
                     Notification endpointNotif = new Notification(action, endpoint);
                     return endpointTypeToProcessor(endpoint.getType()).process(endpointNotif);
                 })
-                .concatenate()
-                .onItem().transformToUni(history -> notifResources.createNotificationHistory(history))
-                .concatenate();
+                .onItem().transformToUniAndConcatenate(history -> notifResources.createNotificationHistory(history));
 
         // Should this be a separate endpoint type as well (since it is configurable) ?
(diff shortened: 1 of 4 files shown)


src/main/java/com/redhat/cloud/notifications/events/EndpointProcessor.java
```diff
@@ -68,15 +68,18 @@ void init() {
                     Notification endpointNotif = new Notification(action, endpoint);
                     return endpointTypeToProcessor(endpoint.getType()).process(endpointNotif);
                 })
-                .merge()
+                .concatenate()
```
cescoffier, line 71: You can use transformToUniAndConcatenate()

src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -98,10 +98,9 @@
             count = resources.getEndpointsCount(principal.getAccount());
         }
 
-        return Uni.combine().all().unis(
-                endpoints.collectItems().asList(),
-                count
-        ).asTuple().onItem().transform(t -> new EndpointPage(t.getItem1(), new HashMap<>(), new Meta(t.getItem2())));
+        return endpoints.collectItems().asList()
```
cescoffier, line 101: For information - collectItems is deprecated, use collect instead

cescoffier (approved):
I just made a few suggestions.

src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
gwenneg: We're still using an old Mutiny version where `collect` is not available, but this will change pretty soon.

gwenneg:
Thanks for the review Clement!

gwenneg:
@pilhuhn / @josejulio: Do you want to review this PR before it is merged?

pilhuhn (approved):

gwenneg:
Thanks!

PR #147, opened 2021-03-23
NOTIF-161 Adding internal endpoint to pull data from policies-notifications

josejulio:
This PR adds an internal endpoint to migrate the data from `policies-notification`. 
It subscribes each user to the email type (instant/daily) found and updates the actions of the `policies-triggered` eventType of every account to use "Email" (deletes others if found).

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
+++ b/src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
@@ -71,6 +71,15 @@ public class ApplicationResources {
                 .getSingleResultOrNull();
     }
 
+    public Uni<EventType> getEventType(String bundleName, String applicationName, String eventTypeName) {
+        final String query = "FROM EventType WHERE name = :eventTypeName AND application.name = :applicationName AND application.bundle.name = :bundleName";
+        return session.createQuery(query, EventType.class)
+                .setParameter("bundleName", bundleName)
+                .setParameter("applicationName", applicationName)
+                .setParameter("eventTypeName", eventTypeName)
+                .getSingleResultOrNull();
+    }
+
     public Multi<EventType> getEventTypes(UUID appId) {
         String query = "FROM EventType WHERE application.id = :appId";
         return session.createQuery(query, EventType.class)
--- a/src/main/java/com/redhat/cloud/notifications/migration/policynotification/PoliciesMigrationService.java
+++ b/src/main/java/com/redhat/cloud/notifications/migration/policynotification/PoliciesMigrationService.java
@@ -26,7 +26,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 @Produces(MediaType.APPLICATION_JSON)
 public class PoliciesMigrationService {
 
-    static final String BUNDLE = "insights";
+    static final String BUNDLE = "rhel";
     static final String APPLICATION = "policies";
     static final String EVENT_TYPE = "policy-triggered";
     static final String DAILY_EMAIL_TYPE = "policies-daily-mail";
--- a/src/main/java/com/redhat/cloud/notifications/migration/policynotification/PoliciesMigrationService.java
+++ b/src/main/java/com/redhat/cloud/notifications/migration/policynotification/PoliciesMigrationService.java
@@ -67,7 +67,7 @@ public class PoliciesMigrationService {
             )
                     .invoke(() -> response.eventTypesMigrated.incrementAndGet())
                     .onItem().transform(subscribed -> pes.accountId).toMulti())
-                    .concatenate().collectItems().in(HashSet::new, HashSet::add)
+                    .concatenate().collect().in(HashSet::new, HashSet::add)
                     .onItem().transformToMulti(accountIds -> Multi.createFrom().iterable(accountIds))
                     .onItem().castTo(String.class)
                     .onItem().transformToMultiAndConcatenate(accountId -> {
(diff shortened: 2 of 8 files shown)


src/main/java/com/redhat/cloud/notifications/migration/policynotification/PoliciesMigrationService.java
```diff
@@ -0,0 +1,103 @@
...
+                            endpoint.setDescription("");
+                            endpoint.setName("email-endpoint");
+                            endpoint.setType(EndpointType.EMAIL_SUBSCRIPTION);
+                            endpoint.setProperties(new EmailSubscriptionAttributes());
+                            return endpointResources.createEndpoint(endpoint);
+                        })
+                        .onItem().transformToUni(endpoint -> endpointResources.linkEndpoint(accountId, endpoint.getId(), eventType.getId())).toMulti();
+                    });
+        }).collectItems().asList().onItem().transform(objects -> response);
+    }
```
gwenneg, line 93: What happens if this is interrupted halfway for any reason? This looks indempotent (and therefore safe) to me but I'd like you to confirm I didn't miss anything.

src/main/java/com/redhat/cloud/notifications/migration/policynotification/PoliciesMigrationService.java
```diff
@@ -0,0 +1,103 @@
...
+            )
+                    .invoke(() -> response.eventTypesMigrated.incrementAndGet())
+                    .onItem().transform(subscribed -> pes.accountId).toMulti())
+                    .concatenate().collectItems().in(HashSet::new, HashSet::add)
+                    .onItem().transformToMulti(accountIds -> Multi.createFrom().iterable(accountIds))
+                    .onItem().castTo(String.class)
+                    .onItem().transformToMultiAndConcatenate(accountId -> {
+                        response.accountsMigrated.incrementAndGet();
+                        // Remove existing endpoints
+                        return endpointResources.getLinkedEndpoints(accountId, eventType.getId(), new Query())
```
gwenneg, line 76: Is there a chance this could return more than one linked endpoint per `accountId`?

josejulio: Yes, it could return multiple.
I'm returning a Multi here, I'm assuming that all the linked endpoints will get unlink in the next step, later they will get collected and the result not really used.

src/main/java/com/redhat/cloud/notifications/migration/policynotification/PoliciesMigrationService.java
josejulio: Running again should yield the desired results, but we might have partial results (Assuming this is not a single transaction). This does not delete our "inputs".

src/main/java/com/redhat/cloud/notifications/migration/policynotification/PoliciesMigrationService.java
gwenneg: Yeah I saw the result was ignored below, but if `n` endpoints are unliked here, doesn't it mean that we try to create `n` EMAIL_SUBSCRIPTION endpoints for a single `accountId/eventType` tuple?

gwenneg: Oh no, it's all transformed into a List before creating the endpoint, I missed that :)

gwenneg: I'd rather ask stupid question rather than miss something 😄

josejulio: I'm rather new to this reactive/mutiny stuff, thanks for double checking it!

src/main/java/com/redhat/cloud/notifications/migration/policynotification/PoliciesMigrationService.java
gwenneg: Well it does not delete the email subscriptions but we're safe because of the `INSERT INTO ... ON CONFLICT ... DO NOTHING`.

josejulio: Yes, those are not supposed to fail if it already exists (or if it doesn't exist in the case of removing).

gwenneg (approved):
LGTM

gwenneg:
One last question: will `MigrateResponse` be the only way to verify the migration results or are there other possibilities to compare both databases entries?

josejulio:
> One last question: will `MigrateResponse` be the only way to verify the migration results or are there other possibilities to compare both databases entries?

We will have to compare them manually I think (looking at the databases and checking that they are correct).

We could expose a long list of what was migrated too to make it easier to compare.

src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
```diff
@@ -71,6 +71,15 @@
                 .getSingleResultOrNull();
     }
 
+    public Uni<EventType> getEventType(String bundleName, String applicationName, String eventTypeName) {
```
josejulio, line 74: @gwenneg could you review this method?  I rebased the code and I'm using the hibernate API.

gwenneg: This method LGTM 👍

PR #148, opened 2021-03-23
NOTIF-138 Replace R2DBC with Hibernate Reactive

gwenneg:
I'm starting a self-review now, hence the draft, but you can already have a look at it if you want.

I'll add a lot of comments later which should help with the review.


gwenneg:
Here's the changelog for the OpenAPI definition files (`notifications` and `integrations`):
- The type of all timestamps (`created` and `updated` attributes) was changed from `date` to `date-time`. This is a consequence of Hibernate Reactive only supporting types from `java.time` (see [doc](http://hibernate.org/reactive/documentation/1.0/reference/html_single/) section 1.2.10).
- Some validation rules (e.g. max length for an `Application` name) were added to be consistent with the existing database constraints.
- The `Meta.count` type was changed from `int32` to `int64` because this is the result of a SQL `COUNT` which is returned as a `Long` by Hibernate Reactive.
- The `EventType.endpoints` collection was removed from the OpenAPI definitions (but not from the Java code) because it was never valued in the backend.
- The `Bundle.applications` collection was removed from the OpenAPI definitions (but not from the Java code) because it was never valued in the backend.
- The `Application.eventTypes` collection was removed from the OpenAPI definitions (but not from the Java code) because it was never valued in the backend.

@josejulio I'll need you to confirm that all these changes are reasonnable and won't break the UI.

gwenneg:
Let's talk about REST responses dynamic filtering I added in this PR.

▶️ Why would we need that?

Because sometimes we want to return a `child -> parent` relation (or the opposite) in a REST response and sometimes we don't. For example, a query to `GET /internal/applications/{id}/eventTypes` will NOT return the evenTypes' application (the parent) while a query to `GET /notifications/eventTypes` will return it.

Before Hibernate Reactive, this was easy to deal with because we could value a parent or a children colllection only when we wanted to return it. Now, since all the model classes are JPA entities with `@OneToMany` or `@ManyToOne` relationships, it's no longer possible.

▶️ The solution: dynamic filtering based on Jackson's `@JsonFilter`

I added the `ApiResponseFilter` class which allows us to filter out any attribute (that would have been serialized otherwise) at serialization time. This filtering is logged at `debug` level.

This requires a few things to work:
- The model class must be annotated with `@JsonFilter(ApiResponseFilter.NAME)`.
- A `@Transient` and `@JsonIgnore` boolean attribute must by added in the model class for each attribute that would need to be filtered out (see `EventType.filterOutApplication`).
- This boolean attribute must be set to `true` before the REST response is returned (see `ApplicationResources.getEventTypes(UUID appId)`).
- The `ApiResponseFilter` class must contain all the attributes we want to filter out somewhere in the app.
- The `ObjectMapper` provided by Quarkus must be customized to enable the filter (see `ApiObjectMapperCustomizer`).

There are other ways to filter things with Jackson, but that's the only one I found that seemed both easy to implement and easy to maintain. Feel free to comment and/or suggest changes if this can be improved.

There's no unit test for it but I'll fix that today.

gwenneg:
Here are some important facts about Hibernate Reactive that impacted this PR:
- Unlike Hibernate ORM, Hibernate Reactive does not support (yet) database-generated values retrieval (UUID, timestamp...) at `INSERT` or `UPDATE` time. This means we have to execute an additional `SELECT` to retrieve these values if we want to return them in a REST response. I chose to move the UUID and timestamps creation from PostgreSQL to the Java code instead.
- Hibernate Reactive does not support the `TIMESTAMP WITH TIME ZONE` column type from PostgreSQL. I modified the schema to drop the timezone for all `TIMESTAMP` columns and the timestamps are now created using the `UTC` ZoneId in the Java code. This raises a question: should we update the already existing database timestamps when this dev will be deployed in production? These values will be altered by 1 or 2 hours (not sure which one is the good one) otherwise.

⚠️ This list will be updated multiple times.

gwenneg:
Some random and yet important info about this PR:
- UUIDs and timestamps (which are actually `LocalDateTime` in Java) are all annotated with `@JsonProperty(access = READ_ONLY)` to make sure we don't persist a value coming from a REST request.

⚠️ This list will be updated multiple times.

gwenneg:
Things that could be improved after this PR:
- We certainly need to implement database transactions in a few places.

⚠️ This list will be updated multiple times.

josejulio:
> @josejulio I'll need you to confirm that all these changes are reasonnable and won't break the UI.

Some small fixes are required just to be sure - updating the types and update based on that usage.
It turns out I was assuming we were receiving `eventTypes.endpoint` and later loading it from the API anyway.

josejulio:
While testing the UI, the following call failed: `api/notifications/v1.0/notifications/defaults/a4229794-2fc5-4d02-ae49-60d1c019d7d1`
```
2021-03-24 13:14:59,817 ERROR [org.jbo.res.res.i18n] (vert.x-eventloop-thread-14) RESTEASY002020: Unhandled asynchronous exception, sending back 500: io.vertx.pgclient.PgException: column "accountid" does not exist
```

gwenneg:
Thanks for testing! There must be an untested case, I'll fix that.

gwenneg:
There was a column mapping issue which is now fixed. We also had several untested cases regarding defaults, it should be better now.

gwenneg:
I'm marking this as ready for review but #144 should be merged first.

gwenneg:
Rebased onto master.

gwenneg:
Rebased again after the Quarkus bump merge.

src/main/java/com/redhat/cloud/notifications/db/BundleResources.java
```diff
@@ -2,145 +2,79 @@
...
+                .onItem().transform(a -> addBundleReference(a, bundleId))
+                .onItem().transformToUni(session::persist)
+                .call(session::flush)
+                .replaceWith(app);
     }
 
-
+    /**
+     * Adds to the given {@link Application} a reference to a persistent {@link Bundle} without actually loading its
+     * state from the database. The app will remain unchanged if the given bundle identifier is null.
```
pilhuhn, line 68: What would be a use-case for a null bundle? Should we perhaps throw an IllegalArgumentException?

gwenneg: I wanted this method to be null-proof and considered validation should be performed at an earlier stage, but it might be overkill. I checked the code, `bundleId` can't be `null` with the current version.

src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
```diff
@@ -1,274 +1,150 @@
...
-            eventType.setDescription(row.get("et_description", String.class));
-
-            eventType.setApplication(app);
-
-            return eventType;
-        }));
+    /**
+     * Adds to the given {@link Application} a reference to a persistent {@link Bundle} without actually loading its
+     * state from the database. The app will remain unchanged if it does not contain a non-null bundle identifier.
+     *
```
pilhuhn, line 140: Like above: does bundle==null make sense here?

gwenneg: Same kind of answer: it does not seem to be an existing use-case from what I just saw in the code.

These methods were originally introduced because Hibernate Reactive failed in quite an ugly and non-helping way when `session.getReference(SomeEntity.class, null)` was called. I can't remember the exact exception though.

src/main/java/com/redhat/cloud/notifications/routers/LivenessService.java
```diff
@@ -41,18 +38,10 @@
...
-                            Flux<PostgresqlResult> execute = c2.createStatement("SELECT COUNT(1)").execute();
-
-                            return execute.flatMap(postgresqlResult -> postgresqlResult.map((row, rowMetadata) -> true));
-                        })
-                        .withFinalizer(postgresqlConnection -> {
-                            postgresqlConnection.close().subscribe();
-                        }))
-                .merge().toUni()
-                .onFailure().recoverWithItem(false);
+        return session.createNativeQuery("SELECT 1")
```
pilhuhn, line 41: Fun fact: the Oracle driver used to not go to the DB to answer this, but did it locally, so that this query completely failed as connection test.

pilhuhn (approved):
LGTM

src/main/java/com/redhat/cloud/notifications/routers/LivenessService.java
gwenneg: Oh that's why there was a `COUNT` then. Thanks for the story, I didn't know this could happen.

josejulio:
> @josejulio Do you need to run more tests with this PR and [RedHatInsights/notifications-frontend#95](https://github.com/RedHatInsights/notifications-frontend/pull/95)?

I think is good to go, you can merge together when you are ready.

PR #149, opened 2021-03-26
Introduce CodeQL security scans

gwenneg:
[CodeQL](https://codeql.github.com/docs/) is a free code analysis tool developed by GitHub.

It's very easy to set up (the YAML file is auto-generated, I only changed a few details) and can help us detect security vulnerabilities, so I think we should add it on all our repositories. Let's start with this one and see how it goes.


gwenneg:
This PR will trigger a CodeQL analysis, but we could also schedule a periodic analysis instead of running one for each PR.

pilhuhn (approved):

PR #150, opened 2021-03-29
Fix hidden database concurrency issue

gwenneg:
`flatMap()` is a shortcut for `onItem().transformToMulti(mapper).merge()`.

We shouldn't `merge()` here because it will cause DB queries to be executed concurrently, which is not allowed by Hibernate Reactive within a single unit of work.


josejulio (approved):
Looks good!

josejulio:
@gwenneg Isn't there a good way to prevent this without manual inspection?

gwenneg:
Hmm I don't know, it doesn't seem easy to automatically detect merges of pipelines that access the database. There are many Mutiny operators which perform a merge... and sometimes it's explicit, sometimes it's hidden behind a shortcut like this one.

gwenneg:
Thanks for the review!

josejulio:
> Hmm I don't know, it doesn't seem easy to automatically detect merges of pipelines that access the database. There are many Mutiny operators which perform a merge... and sometimes it's explicit, sometimes it's hidden behind a shortcut like this one.

Maybe a special Mutiny object that doesn't have the merge operators could help?

PR #151, opened 2021-03-29
Add layered EventConsumer test

gwenneg:
When I worked on Hibernate Reactive, everything got broken at some point of the migration (which was expected) and `LifecycleITest` didn't really help identifying things that needed to be fixed.

While `LifecycleITest` is a useful test, I think the app is insufficiently tested and we could easily miss a regression in many areas of the backend. In particular, we need isolated tests for each layer. That's why I wrote `EventConsumerTest`.


src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
```diff
@@ -42,38 +42,32 @@ public EventConsumer(MeterRegistry registry) {
...
                                 .onFailure().invoke(t -> rejectedCount.increment())
-                        // TODO Handle here and set the counters for broken input data and produce empty message?
                 )
                 .stage(self -> self
                                 // Second pipeline stage - enrich from input to destination (webhook) processor format
                                 .onItem()
-                                .transformToUni(action -> destinations.process(action))
-                                .onFailure().invoke(t -> processingErrorCount.increment())
+                                .transformToUni(action -> destinations.process(action)
+                                        .onFailure().invoke(t -> processingErrorCount.increment())
```
gwenneg, line 56: There's an app behavior change here:
- Before: `processingErrorCount` was incremented when the payload extraction failed at the previous stage, which doesn't make sense.
- Now: it's no longer incremented.

src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
```diff
@@ -42,38 +42,32 @@ public EventConsumer(MeterRegistry registry) {
...
+                                )
                         // Receive only notification of completion
                 )
                 // Third pipeline stage - ack the Kafka topic
                 .onItemOrFailure()
                 .transformToUni((unused, t) -> {
-                    // Log the throwable ?
                     if (t != null) {
-                        log.severe("Could not process the payload: " +  t.getMessage());
+                        log.log(Level.SEVERE, "Could not process the payload", t);
```
gwenneg, line 64: The previous log wasn't helpful at all: it told us something failed without any lead about the cause.

src/test/java/com/redhat/cloud/notifications/events/EventConsumerTest.java
```diff
@@ -0,0 +1,90 @@
...
+        return action;
+    }
+
+    private void assertCountersIncrement(int expectedRejectedIncrement, int expectedProcessingErrorIncrement, Runnable runnable) {
+        double initialRejectedCount = rejectedCounter.count();
+        double initialProcessingErrorCount = processingErrorCounter.count();
+        runnable.run();
+        assertEquals(expectedRejectedIncrement, rejectedCounter.count() - initialRejectedCount);
+        assertEquals(expectedProcessingErrorIncrement, processingErrorCounter.count() - initialProcessingErrorCount);
+    }
```
gwenneg, line 89: I don't know how to reset counters between tests (tried a few things which didn't work). If you have a better way to handle this, I'm open to suggestions :)

src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
josejulio: > * Before: `processingErrorCount` was incremented when the payload extraction failed at the previous stage, which doesn't make sense.

Could you elaborate on why it doesn't make sense?

gwenneg: My understanding of `processingErrorCount` is that it was added to count the failures which happen during a `destinations.process(action)` execution.

With the current version of the code, if `this::extractPayload` fails, `destinations.process(action)` will not be executed and `processingErrorCount` will still be incremented.

josejulio: Ah, understood.

gwenneg: Exactly.

src/test/java/com/redhat/cloud/notifications/events/EventConsumerTest.java
pilhuhn: You could try to get the metrics registry and clear it. It may even work by just injecting the registry.

gwenneg: That's one of the thing I tried but it didn't work (or I couldn't find the good way maybe).

gwenneg: Anyway this can be improved later. The current way to assert counters increment is working as intended, I just don't really like how it's done :)

src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
```diff
@@ -42,38 +42,32 @@ public EventConsumer(MeterRegistry registry) {
     @Acknowledgment(Strategy.PRE_PROCESSING)
     // Can be modified to use Multi<Message<String>> input also for more concurrency
     public Uni<Void> processAsync(Message<String> input) {
-        if (log.isLoggable(Level.FINE)) {
-            log.fine("Processing: " + input.getPayload());
-        }
-        return Uni.createFrom()
-                .item(input)
+        return Uni.createFrom().item(() -> input.getPayload())
+                .onItem().invoke(payload -> log.fine(() -> "Processing: " + payload))
```
pilhuhn, line 46: This seems to lose the `if (isLoggable())` and is thus always logging at fine level, which is not desirable.

gwenneg: The `"Processing: " + payload` concatenation won't be performed before the logger checks if fine is loggable (it is done internally) because there's a supplier now: `() -> "Processing: " + payload)`.

pilhuhn: Ok. I was less worried about the concatenation, but more about unmarshalling a potentially large payload to string and then throwing it away. Great, learned a new thing.

gwenneg: Most loggers (if not all) check internally if a level is loggable before doing it. But the concatenation part is important: it will be done before the internal check is done unless we use a Supplier.

PR #153, opened 2021-03-30
NOTIF-163 Rename 'insights' bundle to 'rhel'

theute:
Waiting confirmation for bundle display name

### DIFF
--- a/src/main/resources/db/migration/V1.10.0__NOTIF-163_rename_insights_bundle.sql
+++ b/src/main/resources/db/migration/V1.10.0__NOTIF-163_rename_insights_bundle.sql
@@ -16,3 +16,8 @@ SET bundle = 'rhel'
 WHERE
   bundle = 'insights'
 ;
+UPDATE public.email_aggregation
+SET bundle = 'rhel'
+WHERE
+  bundle = 'insights'
+;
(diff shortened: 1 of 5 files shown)


theute:
Will require PRs in https://github.com/RedHatInsights/cloud-services-config but also waiting for confirmation

src/main/resources/db/migration/V1.10.0__NOTIF-163_rename_insights_bundle.sql
```diff
@@ -0,0 +1,18 @@
...
+SET name = 'rhel', display_name = 'Red Hat Enterprise Linux'
+WHERE
+  name = 'insights'
+;
+
+-- Update email susbscriptions
+UPDATE public.endpoint_email_subscriptions
+SET bundle = 'rhel'
+WHERE
+  bundle = 'insights'
```
gwenneg, line 17: It is out of the scope of this PR, but this change shows of the things that surprised me while working on the database. Why are we using the bundle `name` in `endpoint_email_subscriptions` while we could have a foreign key based on `Bundle.id`?  It would make that kind of changes easier and improve data integrity as well.

theute: `app id` even, but yes, I thought the same, not sure why this is like this, I think things have deviated.

PR #155, opened 2021-03-30
Make the non-application-root-path consistent with policies

gwenneg:
Let's align with https://github.com/RedHatInsights/policies-ui-backend/pull/179, it will make things easier to maintain if all projects have the same behavior.

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/auth/RHIdentityAuthMechanism.java
+++ b/src/main/java/com/redhat/cloud/notifications/auth/RHIdentityAuthMechanism.java
@@ -49,7 +49,7 @@ public class RHIdentityAuthMechanism implements HttpAuthenticationMechanism {
                     good = true;
                 }
             } else if (path.startsWith("/openapi.json") || path.startsWith("/internal")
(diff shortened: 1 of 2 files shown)


gwenneg:
cc @pilhuhn

src/main/java/com/redhat/cloud/notifications/auth/RHIdentityAuthMechanism.java
```diff
@@ -49,7 +49,7 @@
                     good = true;
                 }
             } else if (path.startsWith("/openapi.json") || path.startsWith("/internal")
-                    || path.startsWith("/admin") || path.startsWith("/health") || path.startsWith("/q/health")) {
+                    || path.startsWith("/admin") || path.startsWith("/health")) {
```
pilhuhn, line 52: Can you add /metrics in here as well please?

gwenneg: Sure, it's done.

theute:
I'm merging.
@spinder that might impact you

PR #159, opened 2021-03-31
Update Quarkus JAR path for Docker

gwenneg:
https://github.com/quarkusio/quarkus/wiki/Migration-Guide-1.12

> Starting the application
The biggest change here is that to start your Quarkus application, you should now use:
java -jar target/quarkus-app/quarkus-run.jar
(instead of using the versioned -runner jar)

### DIFF
--- a/src/main/docker/Dockerfile-build.jvm
+++ b/src/main/docker/Dockerfile-build.jvm
@@ -38,7 +38,7 @@ RUN microdnf install openssl curl ca-certificates ${JAVA_PACKAGE} \
 ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
 
 COPY --from=build /home/jboss/target/lib/* /deployments/lib/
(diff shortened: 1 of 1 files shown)


spinder:
I think that looks right, but I don't have it running with these latest commits to check.

gwenneg:
And thanks for the review!

src/main/docker/Dockerfile-build.jvm
```diff
@@ -38,7 +38,7 @@ RUN microdnf install openssl curl ca-certificates ${JAVA_PACKAGE} \
 ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
 
 COPY --from=build /home/jboss/target/lib/* /deployments/lib/
```
theute, line 40: This step will fail

gwenneg: Yeah 😢 We should probably copy the entire `/target/quarkus-app` directory. I just saw it was recommended by the Quarkus team.

PR #164, opened 2021-03-31
NOTIF-145 Copy all

theute:
Copy all (Works on my machine)


gwenneg:
Is that ok to remove the `--chown=1001` copy option?

theute:
> Is that ok to remove the `--chown=1001` copy option?

Probably not ;) Fixed

gwenneg (approved):
Let's see if CI is happy with this!

PR #167, opened 2021-03-31
Clear Hibernate session L1 cache after message processing

gwenneg:
@josejulio found an issue on CI: after he successfully changed a webhook URL (the change was persisted), `EndpointProcessor` was still using the old URL while processing a new payload.

I will confirm that tomorrow, but I'm pretty sure this happened because the Hibernate `session` is a `@RequestScoped` bean and it seems there are no request boundaries for a Smallrye RM inbound connector. The request context seems to be created when the first incoming message is received and is never destroyed after that.

This PR contains a simple fix for that issue: the `session` L1 cache is cleared after each processed message. I'll talk with the SM RM and/or Hibernate Reactive teams to determine if that's the best approach. In the meantime, this can be merged to fix the bug.


josejulio (approved):
lgtm

PR #168, opened 2021-04-01
Change NotificationHistory.id type from Integer to UUID

gwenneg:
This PR changes the notification_history.id type from integer to uuid.

We need that for the delivery-via-camel PoC, but it will also be better if we decide to expose the history through an API in the future (exposing an auto-incremented integer id would be a security flaw).

It's a draft for now because I need to figure out how to update the frontend.


gwenneg:
@pilhuhn This is ready for a review. If you merge this PR, https://github.com/RedHatInsights/notifications-frontend/pull/96 will have to be merged as well.

josejulio (approved):
lgtm

pilhuhn (approved):

PR #170, opened 2021-04-07
NOTIF-150 Introduce behavior groups

gwenneg:
I'm creating this for early discussions about the model. A big part of the code is still on my machine.

Locks are purposely absent from the model for now, we need to clarify a few things (in the JIRA ticket) before they can be included here.


src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -0,0 +1,131 @@
...
+        String query = "UPDATE BehaviorGroup SET name = :name, displayName = :displayName WHERE accountId = :accountId AND id = :id";
+        return session.createQuery(query)
+                .setParameter("name", behaviorGroup.getName())
+                .setParameter("displayName", behaviorGroup.getDisplayName())
+                .setParameter("accountId", behaviorGroup.getAccountId())
+                .setParameter("id", behaviorGroup.getId())
+                .executeUpdate()
+                .call(session::flush)
+                .onItem().transform(rowCount -> rowCount > 0);
+    }
```
josejulio, line 54: Not related to BehaviorGroup but to panache/hibernate. 
Do we need to do this manually? I though that we only had to get the `BehaviorGroup` object and set the values as we wanted for them to be updated.

src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -0,0 +1,131 @@
...
+        String query = "DELETE FROM BehaviorGroup WHERE id = :id";
+        return session.createQuery(query)
+                .setParameter("id", id)
+                .executeUpdate()
+                .call(session::flush)
+                .onItem().transform(rowCount -> rowCount > 0);
+    }
+
+    public Uni<Void> createBehaviorGroupAction(UUID behaviorGroupId, UUID endpointId) {
+        // TODO Check behavior group accountId is the good one
```
josejulio, line 66: This `TODO` is to prevent updating the global behavior groups?

src/main/java/com/redhat/cloud/notifications/models/Bundle.java
```diff
@@ -47,6 +49,14 @@
     @JsonIgnore
     private Set<Application> applications;
 
+    @OneToMany(mappedBy = "bundle", cascade = CascadeType.REMOVE)
+    @JsonIgnore
+    private Set<BehaviorGroup> behaviorGroups;
```
josejulio, line 54: I don't think we need this set. It would be HUGE if we tried to fetch it anyway. Or this has only the ones that are not tied to an specific account?

src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
gwenneg: Yes and no 😃

The `BehaviorGroup` could be retrieved first (with a `SELECT` query or even a `session.find`) and then modified without having to write the `UPDATE` query in our code (changes would be commited at `flush` time). However, this would execute 2 SQL queries (`SELECT` first, `UPDATE` second) which is not ideal for performances. On the other hand, we can write the `UPDATE` query that way and get rid of the `SELECT` query. It's not always possible to do that though...

Hibernate sometimes has a bad reputation because it is said to be slow, but that's usually because there are multiple ways of doing the same thing and it's not always obvious which one is the best for performances.

src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
gwenneg: That was a TODO for myself which has already been fixed on my machine. I wanted to make sure there wouldn't be cross-accounts access to the behavior groups.

src/main/java/com/redhat/cloud/notifications/models/Bundle.java
gwenneg: I'm not sure we can get rid of the `Set` because it's part of a Hibernate relationship. However, it is lazy loaded and if we never force a fetch of this attribute, it should never be loaded and therefore not cause any problem.

josejulio: I'm worried if we accidentally use this. Could we just remove this variable? `BehaviorGroup` class already has the relation that we want (BehaviorGroup -> Bundle) anyway. If not, we need to be careful about it.

src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
josejulio: I don't think it would be that bad to do the `SELECT` first, specially to confirm that:
- User is part of the account (i.e. matches `accountId`)
- We aren't changing the `accountId`. We definitely don't want to change it ever.

src/main/java/com/redhat/cloud/notifications/models/Bundle.java
gwenneg: That's a legitimate concern, I agree with you. I'll try removing it after I'm done fixing all broken APIs and tests.

src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
gwenneg: I'm not sure which version you reviewed since I pushed multiple times since your first comment, but the current code already checks the `accountId` and the update is restricted to the `name` and `displayName` fields. I really don't see how a `SELECT` would improve things here while it would certainly degrade performances.

josejulio: > I'm not sure which version you reviewed

Probably and old version.

src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java
```diff
@@ -18,43 +18,43 @@
...
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 import static org.mockserver.model.HttpResponse.response;
 
 @QuarkusTest
 @QuarkusTestResource(TestLifecycleManager.class)
-@TestInstance(TestInstance.Lifecycle.PER_CLASS)
 public class LifecycleITest extends DbIsolatedTest {
```
gwenneg, line 58: Note for future review: don't even try to compare the old and new versions of the code, I had to rewrite this class entirely. Reviewing the new version should not be too hard, hopefully.

src/main/java/com/redhat/cloud/notifications/models/Bundle.java
gwenneg: `@OneToMany` relation removed.

josejulio: Thanks!

gwenneg:
⭐ API changes

🚧 _Work in progress..._ 🚧 

| Before  | After | Description |
| - | - |
| - | /internal/bundles/{id}/behaviorGroups/{behaviorGroupId}/default  |

gwenneg:
I talked about this with @theute today. We're thinking of pushing the changes to `master` in two phases:
- Phase 1: The behavior groups would be pushed without breaking or affecting any existing REST API operation or the Kafka messages processing. This means no data migration at this point. The app behavior would remain exactly the same in prod, but we would be allowed to create/edit/delete behavior groups on CI. Event types wouldn't be linked to behavior groups. The backend could be pushed before the frontend, no need to push PRs at the same time.

- Phase 2: The link between the behavior groups and the event types would be pushed. This would require a data migration and a synchronized push of the backend and the frontend.

WDYT @josejulio?

josejulio:
> WDYT @josejulio?

As discussed on the sync call, I'm +1 on this.

josejulio:
Regarding `N10` and `N11`, we have an issue when adding actions to a evenType (or the defaults). The actions end up in different order as the specified.
This is because the calls to add/delete are `async` in the UI. Do you think we should add them all in a single call (set-actions) or have a "order-field" ?

`N12` will also retrieve the actions in the `BundleGroup` ?

gwenneg:
> Regarding `N10` and `N11`, we have an issue when adding actions to a evenType (or the defaults). The actions end up in different order as the specified.
> This is because the calls to add/delete are `async` in the UI. Do you think we should add them all in a single call (set-actions) or have a "order-field" ?

Yes, a single REST call would make sense to keep a consistent order. We will need an order field in the SQL schema to sort the actions, but I don't think it has to be sent to the UI. The backend can determine the order based on the list received in the request.

> 
> `N12` will also retrieve the actions in the `BundleGroup` ?

If you need the actions, then they will be included. 😄

josejulio:
> If you need the actions, then they will be included.

I'm always happy to make the less number of calls :smile:  but i was asking because i didn't see any call to get the action of a `BundleGroup`. Is good If they are returned in the `BundleGroup`.

gwenneg:
@josejulio I took care of [your comment](https://github.com/RedHatInsights/notifications-backend/pull/170#issuecomment-820606003) in #178.

gwenneg:
Replaced with #178.

PR #171, opened 2021-04-09
NOTIF-170 Updates instant & daily emails

josejulio:
- Using new action and policies data


src/main/resources/application.properties
```diff
@@ -73,3 +73,6 @@ email.subscription.daily.cron=0 0 2 * * ?
 policynotifications/mp-rest/url=http://localhost:8090
 policynotifications/mp-rest/connectTimeout=2000
 policynotifications/mp-rest/readTimeout=2000
+
+# qute
+quarkus.qute.property-not-found-strategy=throw-exception
```
gwenneg, line 78: If an exception is thrown while an e-mail is being sent, what will be the consequence for the user? Will there be another e-mail attempt later or will the e-mail be lost forever?

I suppose we would catch that kind of exceptions before it goes to production so that's probably a very low risk, but since it's still there I want to make sure of what happens after the exception.

src/main/java/com/redhat/cloud/notifications/transformers/BaseTransformer.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private <T extends SpecificRecord> String serializeAvroSchema(T avroObject) {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        try {
+            JsonEncoder jsonEncoder = EncoderFactory.get().jsonEncoder(avroObject.getSchema(), baos);
+            DatumWriter<T> writer = new SpecificDatumWriter<T>((Class<T>) avroObject.getClass());
+            writer.write(avroObject, jsonEncoder);
+            jsonEncoder.flush();
+        } catch (IOException exception) {
+            throw new IllegalArgumentException("Avro serialization failed", exception);
```
gwenneg, line 52: That's completely minor so don't bother commiting only for that, but you could throw an `UncheckedIOException` here.

src/test/java/com/redhat/cloud/notifications/templates/TestPoliciesTemplate.java
```diff
@@ -185,11 +166,10 @@ public void testDailyEmailBodyOnePoliciesAndOneSystem() {
         payload.put("unique_system_count", 1);
 
         String result = Policies.Templates.dailyEmailBody()
-                .data("payload", payload)
+                .data("action", Map.of("context", payload))
                 .render();
 
         assertTrue(result.contains("<b>1 policy</b> triggered on <b>1 system</b>"));
-        assertFalse(result.contains("NOT_FOUND"), "A replacement was not correctly done");
```
gwenneg, line 192: This was removed because of the `quarkus.qute.property-not-found-strategy`, right?

src/test/java/com/redhat/cloud/notifications/processors/webhook/WebhookTest.java
```diff
@@ -76,14 +79,35 @@ void testWebhook() {
...
         webhookActionMessage.setEventType("testWebhook");
-
-        Map<String, Object> payload = new HashMap<>();
-        payload.put("any", "thing");
-        payload.put("we", 1);
-        payload.put("want", "here");
-        webhookActionMessage.setPayload(payload);
         webhookActionMessage.setAccountId("tenant");
 
+        Map<String, Object> payload1 = new HashMap<>();
```
gwenneg, line 84: Did you had plans for `payload2` and then changed your mind? 😃

src/main/resources/application.properties
gwenneg: After further reading of the code, it doesn't look like the e-mail would be canceled in case of exception.

josejulio: If an exception is throw there, the template or the data is probably wrong, nothing we can do in that case. There could be some enhancements to report this broken request to the user/app.

src/test/java/com/redhat/cloud/notifications/processors/webhook/WebhookTest.java
josejulio: Ah, yes. The request accepts multiple `payloads` in the events :)
payload2 is actually empty.

src/test/java/com/redhat/cloud/notifications/templates/TestPoliciesTemplate.java
josejulio: Yes, I think is better to hard fail so is easier for devs working on the templates to know right away that something is wrong.

PR #172, opened 2021-04-12
Improve tests isolation and maintainability

gwenneg:
Our tests are full of side-effects: many of them depend on / suffer from data persisted by other tests. This makes the tests unreliable and sometimes very hard to maintain.

This PR fixes that issue by resetting the database before and after each test. I wanted to do that using [io.quarkus.test.TestTransaction](https://github.com/gwenneg/quarkus/blob/main/test-framework/common/src/main/java/io/quarkus/test/TestTransaction.java) but unfortunately it does not work with Hibernate Reactive. I'll create a Quarkus issue about this later.

I need this PR to be merged before I can finish the behavior groups PR.

⚠️ Check `Hide whitespace changes` for an easier review.

### DIFF
--- a/src/test/java/com/redhat/cloud/notifications/db/DbCleaner.java
+++ b/src/test/java/com/redhat/cloud/notifications/db/DbCleaner.java
@@ -21,15 +21,29 @@ import javax.inject.Inject;
 @ApplicationScoped
 public class DbCleaner {
 
+    private static final String DEFAULT_BUNDLE_NAME = "rhel";
+    private static final String DEFAULT_BUNDLE_DISPLAY_NAME = "Red Hat Enterprise Linux";
+    private static final String DEFAULT_APP_NAME = "policies";
+    private static final String DEFAULT_APP_DISPLAY_NAME = "Policies";
+    private static final String DEFAULT_EVENT_TYPE_NAME = "policy-triggered";
+    private static final String DEFAULT_EVENT_TYPE_DISPLAY_NAME = "Policy triggered";
+    private static final String DEFAULT_EVENT_TYPE_DESCRIPTION = "Matching policy";
+
     @Inject
     Mutiny.Session session;
 
+    @Inject
+    BundleResources bundleResources;
+
+    @Inject
+    ApplicationResources appResources;
+
     /**
-     * Deletes all records from all database tables (except for flyway_schema_history). This method should be called
-     * from a method annotated with <b>both</b> {@link BeforeEach} and {@link AfterEach} in all test classes that
-     * involve SQL queries to guarantee the tests isolation in terms of stored data. Ideally, we should do that with
-     * {@link io.quarkus.test.TestTransaction} but it doesn't work with Hibernate Reactive, so this is a temporary
-     * workaround to make our tests more reliable and easy to maintain.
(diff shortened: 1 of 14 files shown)


gwenneg:
cc @bennetelli

src/test/java/com/redhat/cloud/notifications/routers/NotificationServiceTest.java
```diff
@@ -50,36 +47,32 @@
...
-        mockServerConfig.addMockRbacAccess(identityHeaderValue, MockServerClientConfig.RbacAccess.FULL_ACCESS);
-
+        mockServerConfig.clearRbac();
     }
 
-    @BeforeAll
-    @ActivateRequestContext
-    void init() {
-        helpers.createTestAppAndEventTypes();
+    private Header initRbacMock(String tenant, String username, RbacAccess access) {
```
gwenneg, line 65: I'll probably put that method into some kind of helper later. This code is duplicated in many test classes. But that's out of that PR scope.

src/test/java/com/redhat/cloud/notifications/db/DbCleaner.java
```diff
@@ -0,0 +1,51 @@
...
+    public void clean() {
+        session.withTransaction(transaction -> deleteAllFrom(EmailAggregation.class)
+                .chain(() -> deleteAllFrom(EmailSubscription.class))
+                .chain(() -> deleteAllFrom(NotificationHistory.class))
+                .chain(() -> deleteAllFrom(EndpointDefault.class))
+                .chain(() -> deleteAllFrom(EndpointTarget.class))
+                .chain(() -> deleteAllFrom(EndpointWebhook.class))
+                .chain(() -> deleteAllFrom(Endpoint.class))
+                .chain(() -> deleteAllFrom(EventType.class))
+                .chain(() -> deleteAllFrom(Application.class))
```
josejulio, line 43: There are some default values that we have in the tests. Shouldn't they be created as part of the clean? (e.g. Application `policies` is created as part of a migration)

gwenneg: Is that what I added in `ResourceHelpers.createRhelBundleAndPoliciesAppAndPolicyTriggeredEventType()`?

josejulio: Yes! Not sure if there is anything else on the database added by the migration. IIUC, once `TestTransaction` starts supporting `Hibernate Reactive` it will have to go as well as this clean ?

gwenneg: I just saw thanks to your comment that a few records were added by the Flyway scripts. I'll update the clean method to restore these records after each clean.

When `TestTransaction` and HibernateReactive will work together, we will probably get rid of `DbCleaner` and simply annotate each `@Test` with `@TestTransaction`. That annotation starts a transaction at the beginning of the test and rollbacks all changes at the end of the test.

gwenneg: I pushed a commit to restore the records after each clean.

josejulio (approved):
lgtm

gwenneg:
Thanks for reviewing this!

PR #174, opened 2021-04-14
ADVISOR-1456 Advisor template

josejulio:
Instant email for `new-recommendation`

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/templates/Advisor.java
+++ b/src/main/java/com/redhat/cloud/notifications/templates/Advisor.java
@@ -13,7 +13,10 @@ public class Advisor extends AbstractEmailTemplate {
             return Templates.newRecommendationInstantEmailTitle();
         }
 
-        return null;
+        throw new UnsupportedOperationException(String.format(
+                "No email title template for Advisor event_type: %s and EmailSubscription: %s found.",
+                eventType, type
+        ));
     }
 
     @Override
@@ -22,7 +25,10 @@ public class Advisor extends AbstractEmailTemplate {
             return Templates.newRecommendationInstantEmailBody();
         }
 
-        return null;
+        throw new UnsupportedOperationException(String.format(
+                "No email body template for Advisor event_type: %s and EmailSubscription: %s found.",
+                eventType, type
+        ));
     }
 
     @Override
--- a/src/main/java/com/redhat/cloud/notifications/templates/Advisor.java
+++ b/src/main/java/com/redhat/cloud/notifications/templates/Advisor.java
@@ -1,0 +1,42 @@
+package com.redhat.cloud.notifications.templates;
+
+import com.redhat.cloud.notifications.models.EmailSubscriptionType;
+import io.quarkus.qute.CheckedTemplate;
+import io.quarkus.qute.TemplateInstance;
+
+// Name needs to be "Advisor" to read templates from resources/templates/Advisor
+public class Advisor extends AbstractEmailTemplate {
+
+    @Override
+    public TemplateInstance getTitle(String eventType, EmailSubscriptionType type) {
+        if (type == EmailSubscriptionType.INSTANT) {
+            return Templates.newRecommendationInstantEmailTitle();
+        }
+
+        return null;
+    }
+
+    @Override
+    public TemplateInstance getBody(String eventType, EmailSubscriptionType type) {
+        if (type == EmailSubscriptionType.INSTANT) {
+            return Templates.newRecommendationInstantEmailBody();
+        }
+
+        return null;
+    }
+
+    @Override
+    public boolean isSupported(String eventType, EmailSubscriptionType type) {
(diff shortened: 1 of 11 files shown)


src/main/java/com/redhat/cloud/notifications/templates/Advisor.java
```diff
@@ -0,0 +1,42 @@
...
+// Name needs to be "Advisor" to read templates from resources/templates/Advisor
+public class Advisor extends AbstractEmailTemplate {
+
+    @Override
+    public TemplateInstance getTitle(String eventType, EmailSubscriptionType type) {
+        if (type == EmailSubscriptionType.INSTANT) {
+            return Templates.newRecommendationInstantEmailTitle();
+        }
+
+        return null;
```
gwenneg, line 16: Is that OK to return `null`? I can see code that could throw a NPE but I guess the `Advisor.isSupported` method will be called first and this `null` value should never be returned.

src/main/java/com/redhat/cloud/notifications/templates/Advisor.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Override
+    public TemplateInstance getBody(String eventType, EmailSubscriptionType type) {
+        if (type == EmailSubscriptionType.INSTANT) {
+            return Templates.newRecommendationInstantEmailBody();
+        }
+
+        return null;
```
gwenneg, line 25: Same question here.

src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateFactory.java
```diff
@@ -9,8 +9,15 @@ private EmailTemplateFactory() {
...
     public static AbstractEmailTemplate get(String bundle, String application) {
-        if (bundle.toLowerCase().equals("rhel") && application.toLowerCase().equals("policies")) {
-            return new Policies();
+        if (bundle.toLowerCase().equals("rhel")) {
+            switch (application.toLowerCase()) {
+                case "policies":
+                    return new Policies();
+                case "advisor":
+                    return new Advisor();
+                default:
```
gwenneg, line 18: Sonar won't like the absence of `break;`.

src/main/resources/templates/Advisor/insightsEmailBody.html
```diff
@@ -0,0 +1,606 @@
...
+		        background: #c9190b;
+		        color: #ffffff;
+		      }
+
+		      .rh-card.rh-m-no-danger {
+		        background-color: #b8bbbe;
+		        background: #b8bbbe;
+		        color: #151515;
+		      }
+		    }  */
```
gwenneg, line 534: There are a few commented CSS rules like this one in the file. Is that OK to commit them?

src/main/java/com/redhat/cloud/notifications/templates/Advisor.java
josejulio: Yep, maybe I should replace it with an exception?

gwenneg: Yes, I would prefer an exception with a message that clearly explains what's wrong rather than a potential NPE in the future.

src/main/resources/templates/Advisor/insightsEmailBody.html
josejulio: Ah... It doesn't look like they would affect the output in any way... I'll ask if these serve any purpose.

gwenneg: That's harmless indeed.

src/main/java/com/redhat/cloud/notifications/templates/Advisor.java
```diff
@@ -0,0 +1,48 @@
...
+public class Advisor extends AbstractEmailTemplate {
+
+    @Override
+    public TemplateInstance getTitle(String eventType, EmailSubscriptionType type) {
+        if (type == EmailSubscriptionType.INSTANT) {
+            return Templates.newRecommendationInstantEmailTitle();
+        }
+
+        throw new UnsupportedOperationException(String.format(
+                "No email title template for Advisor event_type: %s and EmailSubscription: %s found.",
```
josejulio, line 17: @gwenneg Throwing an exception now

src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateFactory.java
josejulio: Added the break;

src/main/resources/templates/Advisor/insightsEmailBody.html
josejulio: I'm going to leave this for now, but I asked if this is going to be enabled or if we can remove it

gwenneg (approved):
LGTM

PR #175, opened 2021-04-15
Introduces timeAgo() function on templates.

josejulio:
- Transforms dates(or date-strings) to a "1 year ago" format
 - Fixes on Advisor templates

I ended up using an implementation based on `Period` to mimic what Advisor has in their [UI](https://cloud.redhat.com/insights/advisor/recommendations?rule_status=enabled&sort=-total_risk&limit=10&offset=0&text=%2Fvar%2Fww).

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/utils/TimeAgoFormatter.java
+++ b/src/main/java/com/redhat/cloud/notifications/utils/TimeAgoFormatter.java
@@ -1,0 +1,73 @@
+package com.redhat.cloud.notifications.utils;
+
+import java.time.LocalDateTime;
+import java.time.Period;
+import java.time.temporal.ChronoField;
+import java.time.temporal.TemporalField;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class TimeAgoFormatter {
+
+    private static class Configuration {
+        private final TemporalField temporalField;
+        private final String singular;
+        private final String pluralFormat;
+
+        Configuration(TemporalField temporalField, String singular, String pluralFormat) {
+            this.temporalField = temporalField;
+            this.singular = singular;
+            this.pluralFormat = pluralFormat;
+        }
+
+    }
+
+    private static final List<Configuration> config = List.of(
+            new Configuration(ChronoField.YEAR, "1 year ago", "%d years ago"),
+            new Configuration(ChronoField.MONTH_OF_YEAR, "1 month ago", "%d months ago"),
+            new Configuration(ChronoField.DAY_OF_MONTH, "1 day ago", "%d days ago"),
+            new Configuration(null, "Today", null)
+    );
+
+    static {
+        List<Configuration> fallbackConfiguration = TimeAgoFormatter.config
+                .stream()
+                .filter(configuration -> configuration.temporalField == null && configuration.singular != null)
+                .collect(Collectors.toList());
+
+        if (fallbackConfiguration.size() == 0) {
+            throw new RuntimeException("Configuration does not have a fallback label");
+        }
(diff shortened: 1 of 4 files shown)


src/main/java/com/redhat/cloud/notifications/utils/TimeAgoFormatter.java
```diff
@@ -0,0 +1,73 @@
...
+            this.singular = singular;
+            this.pluralFormat = pluralFormat;
+        }
+
+    }
+
+    private static final List<Configuration> config = List.of(
+            new Configuration(ChronoField.YEAR, "1 year ago", "%d years ago"),
+            new Configuration(ChronoField.MONTH_OF_YEAR, "1 month ago", "%d months ago"),
+            new Configuration(ChronoField.DAY_OF_MONTH, "1 day ago", "%d days ago"),
```
gwenneg, line 28: I don't think `ChronoField.DAY_OF_MONTH` will work if today is `2021-01-01` and the date to format is `2020-31-12`. The `naturalUnit` variable should then be equal to `1 - 31`, which will lead to `Today` being returned while it should be `1 day ago`. The same kind of issue should happen with `ChronoField.MONTH_OF_YEAR`.

gwenneg (changes requested):

src/main/java/com/redhat/cloud/notifications/utils/TimeAgoFormatter.java
```diff
@@ -0,0 +1,73 @@
...
+        Period period = Period.between(dateTime.toLocalDate(), base.toLocalDate());
+
+        for (Configuration configuration : TimeAgoFormatter.config) {
+            if (configuration.temporalField == null) {
+                return configuration.singular;
+            }
+
+            long units = period.get(configuration.temporalField.getBaseUnit());
+            if (units >= 1) {
+                // Calculate natural units so that 2018-Jun-16 is 3 years ago from 2021-Apr-15 and not 2
```
gwenneg, line 61: It means that `2018-Dec-31` will also be `3 years ago` from `2021-Jan-1`. Does that come from Advisor?

josejulio: I'm unsure how they compute the data, I'll have to check what lib they use (if any). 
On manual testings I saw this behavior: `2018-Jun-16 is 3 years ago from 2021-Apr-15 and not 2`

src/main/java/com/redhat/cloud/notifications/utils/TimeAgoFormatter.java
josejulio: Thanks, I'll add tests for that.

josejulio: So this is what insights do: 
https://github.com/RedHatInsights/frontend-components/blob/86b02e3c206663d640d30b32087061c04db1595d/packages/components/src/DateFormat/helper.js#L12-L19

I'll adjust to match that behavior.

josejulio:
Updated to match https://github.com/RedHatInsights/frontend-components/blob/86b02e3c206663d640d30b32087061c04db1595d/packages/components/src/DateFormat/helper.js#L4-L19

gwenneg (approved):
My daughter would certainly love their algorithm 😄 "Dad, my birthday was 183 days ago so I'm one year older now!". Rounding is a bit weird, I would have truncated instead...

Anyway, we're now aligned with their way of calculating "time ago" so I guess we're good.

PR #176, opened 2021-04-16
Repair Flyway migration

gwenneg:
https://flywaydb.org/documentation/command/repair


gwenneg:
@theute This will repair the Flyway repair the schema history table. We should deploy it once, then redeploy without the repair feature.

josejulio:
> @theute This will repair the Flyway schema history table. We should deploy it once, then redeploy without the repair feature.

We might need this for prod too

josejulio (approved):
lgtm

gwenneg:
We can revert this change after the V1.6.0 is done in prod.

PR #177, opened 2021-04-16
Fix error when email_subscriptions has unknown bundle/apps

josejulio:
@theute This is likely cause of the problem that you see on CI with the user-preferences/notifications


src/main/java/com/redhat/cloud/notifications/db/BundleResources.java
```diff
@@ -38,6 +38,13 @@
         return session.find(Bundle.class, id);
     }
 
+    public Uni<Bundle> getBundle(String name) {
+        String query = "FROM Bundle WHERE name = :name";
+        return session.createQuery(query, Bundle.class)
+                .setParameter("name", name)
+                .getSingleResult();
```
gwenneg, line 45: We probably don't want a `NoResultException` to be thrown if the given name is unknown:
```suggestion
                .getSingleResultOrNull();
```
Since this didn't show up in tests, we certainly have one uncovered case (user config retrieval with an uknown bundle name).

src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
```diff
@@ -195,9 +205,7 @@
         return Uni.createFrom().deferred(() -> {
             final SettingsValues values = new SettingsValues();
 
-            return bundleResources.getBundles()
-                    // Todo: Create a method that gets the bundle by name
-                    .filter(bundle -> bundle.getName().equals(bundleName))
+            return bundleResources.getBundle(bundleName)
                     .onItem().transformToMulti(bundle -> {
```
gwenneg, line 209: If you apply the suggestion from my previous comment, this may return a `Uni` containing a `null` value.

Replacing `.onItem().transformToMulti(bundle -> {` with `.onItem().ifNotNull().transformToMulti(bundle -> {` should be enough to handle the `null` value but it needs to be confirmed.

src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
```diff
@@ -213,9 +221,14 @@
...
-                    }).concatenate().collect().asList().onItem().transformToMulti(objects -> emailSubscriptionResources.getEmailSubscriptionsForUser(account, username))
+                    }).collect().asList().onItem().transformToMulti(objects -> emailSubscriptionResources.getEmailSubscriptionsForUser(account, username))
                     .onItem().transform(emailSubscription -> {
-                        values.bundles.get(emailSubscription.getBundleName()).applications.get(emailSubscription.getApplicationName()).notifications.put(emailSubscription.getType(), true);
+                        if (values.bundles.containsKey(emailSubscription.getBundleName())) {
+                            BundleSettingsValue bundleSettingsValue = values.bundles.get(emailSubscription.getBundleName());
+                            if (bundleSettingsValue.applications.containsKey(emailSubscription.getApplicationName())) {
+                                bundleSettingsValue.applications.get(emailSubscription.getApplicationName()).notifications.put(emailSubscription.getType(), true);
+                            }
+                        }
```
gwenneg, line 231: Things would be much simpler if we had SQL constraints between the `endpoint_email_subscriptions`, `bundles` and `applications` tables. We should take care of that soon.

src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
```diff
@@ -90,16 +90,26 @@
...
+                                applicationResources.getApplication(bundleName, applicationName)
+                                .onItem().transformToUni(application -> {
+                                    if (application != null) {
+                                        return emailSubscriptionResources.subscribe(
+                                                account, name, bundleName, applicationName, emailSubscriptionType
+                                        );
+                                    }
+
+                                    return Uni.createFrom().item(true);
+                                }).toMulti()
```
gwenneg, line 107:
```suggestion
                                .onItem().ifNotNull().transformToUni(application -> {
                                    return emailSubscriptionResources.subscribe(
                                            account, name, bundleName, applicationName, emailSubscriptionType
                                    );
                                }).toMulti()
```

src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
```diff
@@ -90,16 +90,26 @@
...
-                bundleSettingsValue.applications.forEach((application, applicationSettingsValue) ->
+        values.bundles.forEach((bundleName, bundleSettingsValue) ->
+                bundleSettingsValue.applications.forEach((applicationName, applicationSettingsValue) ->
                 applicationSettingsValue.notifications.forEach((emailSubscriptionType, value) -> {
                     if (value) {
-                        subscriptionRequests.add(emailSubscriptionResources.subscribe(
-                                account, name, bundle, application, emailSubscriptionType
-                        ).toMulti());
+                        subscriptionRequests.add(
+                                applicationResources.getApplication(bundleName, applicationName)
```
gwenneg, line 98: Checking the app existence with a `SELECT` before the `INSERT` will work, but we could also do that check directly into the `INSERT` query which would be a bit better in terms of performances.

gwenneg (changes requested):
We have some `old` code in UserConfigService that is supposed to be removed soon. Does it need to be changed as well?

josejulio:
> We have some `old` code in `UserConfigService` that is supposed to be removed soon. Does it need to be changed as well?

Maybe, I'll take a look or remove if we get user-preferences on stable.

src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
josejulio: I'll check

src/main/java/com/redhat/cloud/notifications/db/BundleResources.java
josejulio: I think I added a test with that, will double check.

src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
josejulio: This is just a workaround, when we migrate to using the id's rather than the names we will be able to remove the check.

src/main/java/com/redhat/cloud/notifications/db/BundleResources.java
josejulio: I think there is no test for this case. I'll add one (I'm OK with failing the request)

src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
```diff
@@ -159,45 +137,15 @@
...
      * */
     private Uni<SettingsValues> getSettingsValueForUser(String account, String username, String bundleName) {
         return Uni.createFrom().deferred(() -> {
             final SettingsValues values = new SettingsValues();
 
-            return bundleResources.getBundles()
-                    // Todo: Create a method that gets the bundle by name
-                    .filter(bundle -> bundle.getName().equals(bundleName))
+            return bundleResources.getBundle(bundleName)
+                    .onItem().ifNull().failWith(() -> new BadRequestException("bundleName must have a value"))
```
gwenneg, line 148: The `BadRequestException` will also be thrown if `bundleName` has a value but it is unknown.

gwenneg (approved):
LGTM

src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
josejulio: Ah, good catch. I'll update.

theute:
I'm pressing the big green button and will give it a try

theute:
Worked fine, thanks !

PR #179, opened 2021-04-20
add CLOWDER_FILE var to clowdapp

lahavyuv86:


theute (commented):
AFAIK it won't do anything until you add the environment variable in "env" (Line 48)
Also I'm not sure that $ACG_CONFIG is known here, I think this can be overridden in Bonfire, but here I think you should specify the full path

lahavyuv86:
@theute fixed

theute:
There are changes that are unrelated, I can't review those, but for the CLOWDER_FILE, if you remove it from the property at the bottom then it's not overridable anymore and will always be the same value, which is the same as the default in code now.

So I don't see why this would be useful this way ?

pilhuhn:
Side note: we have just updated he notifications-backend to use the updated clowder config source (#180). In theory that should fix our issue without this PR.

lahavyuv86:
@theute what about now?
And the other changes are some tweaks I saw on the Sources clowdapp file that is needed for the ephemeral deployment:
https://github.com/RedHatInsights/sources-api/blob/master/deploy/clowdapp.yaml#L6

PR #181, opened 2021-04-20
Build applications with mvn verify

bennetelli:
I think we don't need to call "mvn install" here. As far as I understand our code, it is not necessary to put the build artefacts in the local .m2/repository folder. So "mvn verify" should be enough, faster and the better option in this case.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
         with:
           java-version: 11
       - name: Build with Maven
-        run: ./mvnw install
+        run: ./mvnw verify
       - uses: actions/upload-artifact@v2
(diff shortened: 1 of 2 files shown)


pilhuhn:
@josejulio I think the openapi artefacts are still correctly saved away. Could you please check anyway?

gwenneg:
That's what I was checking @pilhuhn. They are correctly uploaded indeed because the upload only requires the file to be present in the `/target` directory. Also, the upload is done by a separate build step which is not impacted by this PR.

gwenneg (approved):

pilhuhn:
@gwenneg I also checked and thought it is good but then .. Ok to merge as you verified it as well.

PR #183, opened 2021-04-21
add separate packages for rhid and rbac auth

bennetelli:
add separate packages for rhid and rbac auth so we have a cleaner code base and don't violate the separation of concerns principle.


gwenneg:
That's a long commit message! Could you shorten it so that it's not cut off by GitHub?

gwenneg (approved):

PR #188, opened 2021-04-22
NOTIF-165 Add API to modify a Bundle

gwenneg:
If a Bundle `name` is modified, the integration with onboarded apps will break. As suggested by @theute, we need some kind of alias for the name to give the apps the opportunity to migrate from the old one to the new one. This will be taken care of in a separate PR.

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/routers/BundlesService.java
+++ b/src/main/java/com/redhat/cloud/notifications/routers/BundlesService.java
@@ -9,7 +9,7 @@ import io.smallrye.mutiny.Uni;
 
 import javax.inject.Inject;
 import javax.validation.Valid;
-import javax.ws.rs.Consumes;
+import javax.validation.constraints.NotNull;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.NotFoundException;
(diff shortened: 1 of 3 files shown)


src/main/java/com/redhat/cloud/notifications/routers/BundlesService.java
```diff
@@ -42,11 +39,23 @@
     }
 
     @POST
-    public Uni<Bundle> addBundle(@Valid Bundle bundle) {
-        // We need to ensure that the x-rh-identity isn't present here
```
gwenneg, line 46: I created a ticket in the backlog about this.

src/main/java/com/redhat/cloud/notifications/routers/BundlesService.java
```diff
@@ -9,24 +9,21 @@
...
-import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
 import java.util.UUID;
 
 /**
  * Deal with bundles
  */
 @Path("/internal/bundles")
-@Consumes(MediaType.APPLICATION_JSON)
-@Produces(MediaType.APPLICATION_JSON)
```
gwenneg, line 29: Quarkus does that automatically.

src/test/java/com/redhat/cloud/notifications/routers/BundleServiceTest.java
```diff
@@ -50,42 +51,28 @@ void testAdd() {
...
     void testNoAddSameBundleTwice() {
-        JsonObject b1 = null;
-        try {
-            b1 = createBundle("b1", 200);
-            createBundle("b1", 500);
-        } finally {
-            if (b1 != null) {
-                deleteBundleById(b1.getString("id"));
-            }
-        }
```
gwenneg, line 88: We no longer need to delete database records after tests because the database is cleaned after each test.

josejulio (approved):
lgtm

PR #189, opened 2021-04-22
NOTIF-166 Add API to modify an Application

gwenneg:
If an Application `name` is modified, the integration with onboarded apps will break. As suggested by @theute, we need some kind of alias for the name to give the apps the opportunity to migrate from the old one to the new one. This will be taken care of in a separate PR.

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/routers/ApplicationService.java
+++ b/src/main/java/com/redhat/cloud/notifications/routers/ApplicationService.java
@@ -8,23 +8,21 @@ import io.smallrye.mutiny.Uni;
 
 import javax.inject.Inject;
 import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
 import javax.ws.rs.BadRequestException;
-import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.NotFoundException;
 import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
-import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
-import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
 import java.util.List;
 import java.util.UUID;
 
 @Path("/internal/applications")
-@Consumes(MediaType.APPLICATION_JSON)
-@Produces(MediaType.APPLICATION_JSON)
 public class ApplicationService {
 
     @Inject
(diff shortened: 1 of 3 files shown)


src/main/java/com/redhat/cloud/notifications/routers/ApplicationService.java
```diff
@@ -8,23 +8,21 @@
...
-import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
-import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
 import java.util.List;
 import java.util.UUID;
 
 @Path("/internal/applications")
-@Consumes(MediaType.APPLICATION_JSON)
-@Produces(MediaType.APPLICATION_JSON)
```
josejulio, line 27: Is this no longer needed? or is it the default?

src/test/java/com/redhat/cloud/notifications/routers/ApplicationServiceTest.java
```diff
@@ -321,4 +325,35 @@ void testGetApplicationsReturn404WhenNotFound() {
...
+                    .get("/internal/applications/{id}")
+                    .then()
+                    .statusCode(200)
+                    .extract().body().asString();
+            JsonObject jsonApp = new JsonObject(responseBody);
+            jsonApp.mapTo(Application.class);
+            assertEquals(updatedName, jsonApp.getString("name"));
+            assertEquals(updatedDisplayName, jsonApp.getString("display_name"));
+        }
+    }
```
josejulio, line 358: nitpick - Should be applicationId

```suggestion
    private void updateApplication(String applicationId, String bundleId, int expectedStatusCode) {
        String updatedName = "updated-name";
        String updatedDisplayName = "updatedDisplayName";

        Application app = new Application();
        app.setName(updatedName);
        app.setDisplayName(updatedDisplayName);
        app.setBundleId(UUID.fromString(bundleId));

        given()
                .contentType(ContentType.JSON)
                .pathParam("id", applicationId)
                .body(Json.encode(app))
                .put("/internal/applications/{id}")
                .then()
                .statusCode(expectedStatusCode);

        if (expectedStatusCode >= 200 && expectedStatusCode < 300) {
            String responseBody = given()
                    .pathParam("id", applicationId)
                    .get("/internal/applications/{id}")
                    .then()
                    .statusCode(200)
                    .extract().body().asString();
            JsonObject jsonApp = new JsonObject(responseBody);
            jsonApp.mapTo(Application.class);
            assertEquals(updatedName, jsonApp.getString("name"));
            assertEquals(updatedDisplayName, jsonApp.getString("display_name"));
        }
    }
```

josejulio (approved):
lgtm

src/main/java/com/redhat/cloud/notifications/routers/ApplicationService.java
josejulio: Saw in the other PR that Quarkus does it for us :)

gwenneg:
Thanks for the reviews!

PR #190, opened 2021-04-22
CCXDEV-4514 Adds missing call in EmailTemplateFactory

josejulio:


src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateFactory.java
```diff
@@ -18,7 +18,10 @@ public static AbstractEmailTemplate get(String bundle, String application) {
                 default:
                     break;
             }
-
+        } else if (bundle.toLowerCase().equals("openshift")) {
+            if (application.toLowerCase().equals("advisor")) {
```
josejulio, line 22: @epapbak Could you take a look? Is the called app "advisor" in the bundle "openshift"? 

// cc: @theute

theute: Correct

epapbak: Thanks guys! Sorry I didn't see this PR until today

PR #191, opened 2021-04-23
NOTIF-150 Introduce behavior groups - Phase 2

gwenneg:
This is phase 2 of #170.

You shall not merge

We're not ready 😄

See the migration plan (Google doc) for more details about this.


src/main/java/com/redhat/cloud/notifications/routers/NotificationService.java
```diff
@@ -131,47 +128,19 @@ private String getAddress(Principal principal) {
...
     /*
      * Called by the UI to build the endpoint removal confirmation screen.
      * That screen shows all the event types (and their application) that will be affected by the endpoint removal.
      */
     @GET
-    @Path("/bg/eventTypes/affectedByRemovalOfEndpoint/{endpointId}") // TODO [BG Phase 2] Remove '/bg' path prefix
+    @Path("/eventTypes/affectedByRemovalOfEndpoint/{endpointId}")
     @Produces(APPLICATION_JSON)
     @Operation(summary = "Retrieve the event types affected by the removal of an integration.")
     @Tag(name = OApiService.PRIVATE)
```
josejulio, line 139: Could you remove these too?

gwenneg: Do you mean that `GET /eventTypes/affectedByRemovalOfEndpoint/{endpointId}` is no longer needed after the behavior group migration?

josejulio (approved):
lgtm

src/main/java/com/redhat/cloud/notifications/routers/NotificationService.java
```diff
@@ -131,47 +127,18 @@ private String getAddress(Principal principal) {
...
-    @Path("/bg/eventTypes/affectedByRemovalOfEndpoint/{endpointId}") // TODO [BG Phase 2] Remove '/bg' path prefix
+    @Path("/eventTypes/affectedByRemovalOfEndpoint/{endpointId}")
     @Produces(APPLICATION_JSON)
     @Operation(summary = "Retrieve the event types affected by the removal of an integration.")
-    @Tag(name = OApiService.PRIVATE)
     @RolesAllowed(RbacIdentityProvider.RBAC_READ_NOTIFICATIONS)
     public Uni<List<EventType>> getEventTypesAffectedByRemovalOfEndpoint(@Context SecurityContext sec, @PathParam("endpointId") UUID endpointId) {
         return getAccountId(sec)
-                .onItem().transformToUni(accountId -> apps.getEventTypesByEndpointId_BG(accountId, endpointId));
+                .onItem().transformToUni(accountId -> apps.getEventTypesByEndpointId(accountId, endpointId));
```
josejulio, line 141: I missed this before, but I think we would like to get the behavior groups affected by the removal of the endpoint id.
I'll check with UX.

gwenneg: In this PR we have:
- `GET /eventTypes/affectedByRemovalOfEndpoint/{endpointId}` [1]
- `GET /eventTypes/affectedByRemovalOfBehaviorGroup/{behaviorGroupId}` [2]

Are you saying we will remove [1] and [2] and also introduce `GET /behaviorGroups/affectedByRemovalOfEndpoint/{endpointId}`? These APIs are so similar that I want to make sure I understood correctly 😃

josejulio: We will remove [1].
We want to know what behavior groups are affected by the removal of an endpoint.

We also need [2], so we don't need to touch it :).

gwenneg: I pushed a commit with the `affectByRemoval*` APIs changes.

The new API returns `BehaviorGroup` including their actions, but I'm not sure actions are needed there. Please let me know if I should remove the actions from the REST response.

josejulio (approved):

gwenneg:
Squashed commits, no code change.

PR #192, opened 2021-04-23
NOTIF-273 Bump Quarkus to 2.2.3.Final

gwenneg:
I didn't test the app locally so far, only checked that is starts without any error in dev mode.


gwenneg:
The build is no longer successful because of recent changes on `master`. The failure is caused by an unknown property during a Jackson serialization using an `ObjectMapper` instance produced by Quarkus and managed by Vert.x. This should be easy fo fix with `DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES` but the `ObjectMapper` instance produced by Quarkus is not currently configurable.

I created a Quarkus PR to fix that: https://github.com/quarkusio/quarkus/pull/16944

gwenneg:
This is finally ready for a review! As usual, checking `Hide whitespace changes` is strongly recommended for a muuuuuuch easier review.

⚠️⚠️⚠️ **DO NOT MERGE AFTER THE APPROVAL** ⚠️⚠️⚠️ I'll take care of it 😄 

Tests are still not completely stable (I'm working on it) but it shouldn't be a problem for the review as long as you mark the files as `Viewed`. I will probably make a few minor commits before the merge to stabilize everything.

---

What's new? (in our code, not in Quarkus 2)
==

⭐ **Hibernate Reactive sessions injection**

The way the Hibernate Reactive sessions are injected has changed.

Before Quarkus 2:
```
@Inject
Mutiny.Session session;

@Inject
Mutiny.StatelessSession statelessSession;
```

Now:
```
@Inject
Mutiny.SessionFactory sessionFactory;

public void foo() {
    sessionFactory.withSession(session -> {
        // do something with the session
    })
}

public void bar() {
    sessionFactory.withStatelessSession(statelessSession-> {
        // do something with the stateless session
    })
}
```

⚠️ You will see many `sessionFactory.withSession(session -> {` in the code while the session instance isn't actually used afterwards. This is intentional.

When `withSession` is called:
- If the current request context (CDI) does not contain any `Session`, then a new session is created and automatically closed at the end of the block.
- Otherwise, the existing session is used and the open/close operations are skipped (the enclosing code will do it).

As explained in the Hibernate Reactive doc, we `should reuse the same session across multiple persistence operations within a single reactive stream representing a certain transaction or unit of work`. Calling `sessionFactory.withSession(session -> {` at the entry point of a REST API or a Kafka message processing method guarantees than no more than one session will be used no matter how many DB queries will be executed.

⭐ **Threads management in reactive tests**

With Quarkus 2, Hibernate Reactive queries _have to_ be executed from the Vert.x event loop thread. That thread is where all reactive non-blocking operations are supposed to happen. That change had very little impact on the runtime code, but a huge impact on our tests because they were written in an imperative way (tons of `.await().indefinitely()` everywhere). I rewrote the backend tests entirely and they are now proper reactive tests.

What does this have to do with thread management? Well, now that the tests are reactive and executed on the event loop thread, it is no longer allowed to write blocking code in tests. For example, all RestAssured calls are blocking and can cause tests failures. That's why you will see many `.chain(runOnWorkerThread(() -> {` in the new code. That instruction moves the RestAssured code execution to a worker thread to avoid blocking the event loop. But if the RestAssured code is followed by an Hibernate Reactive query, the reactive stream needs to switch back to the event loop thread. That's what `.emitOn(MutinyHelper.executor(vertx.getOrCreateContext()))` is used for.

⭐ **`.await().indefinitely()` is back, but don't abuse it!**

It's OK to call `.await().indefinitely()` at the very end of a test. It shouldn't happen more than once per `@Test`.

⭐ **Hibernate Reactive tests now work from IntelliJ**

cc @bennetelli 😄

⭐ **`quarkus-cache` now properly caches `Uni`**

All workarounds should be gone from the code.

gwenneg:
YOU SHALL NOT MERGE

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessorTest.java
```diff
@@ -89,6 +89,9 @@ void shouldSuccessfullySendEmail() {
         inMemoryConnector.source(AGGREGATION_CHANNEL).send(Json.encode(aggregationCommand1));
         inMemoryConnector.source(AGGREGATION_CHANNEL).send(Json.encode(aggregationCommand2));
 
+        // TODO Replace fixed delay with a dynamic one (wait until the message has been processed).
+        Thread.sleep(5000L);
```
josejulio, line 93: Could we use mutiny calls with retry.atMost for this?

josejulio (approved):
looks good to me

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessorTest.java
gwenneg: In this particular case, there's nothing we could test to interrupt the retry/wait.

But we could probably get rid of awaitability and replace it with Mutiny retries in other places.

gwenneg:
Rebased and fixed conflicts.

gwenneg:
Rebased and fixed conflicts again.

gwenneg:
And again.

bennetelli:
Sorry.. was it me? Should we merge this PR?

gwenneg:
No the last conflict came from one of my PRs 😃

I am running one last test locally before hitting the merge button.

bennetelli:
great :)

PR #193, opened 2021-04-23
[DRFT-334] Add Drift instant notification email template

eherget:
This is largely cut-n-paste from Policies email template.  I am unable to test it locally, so any and all help (and forgiveness) is appreciated!


josejulio (approved):
lgtm, we can test it on ci

PR #196, opened 2021-04-26
Insert, update and delete behavior group actions with a single REST call

gwenneg:
@josejulio: Here's what we talked about last week.

`DELETE /api/notifications/v1.0/notifications/behaviorGroups/{behaviorGroupId}/actions` no longer exists.

`PUT /api/notifications/v1.0/notifications/behaviorGroups/{behaviorGroupId}/actions` now takes care of actions addition, removal and order in the UI. I couldn't use `order` since it's a SQL reserved word so I used `BehaviorGroupAction.position` instead, but I hesitated with `rank`. This is super easy to rename so don't hesitate to suggest a different name.

### DIFF
--- a/src/test/java/com/redhat/cloud/notifications/db/BehaviorGroupResourcesTest.java
+++ b/src/test/java/com/redhat/cloud/notifications/db/BehaviorGroupResourcesTest.java
@@ -17,6 +17,7 @@ import org.junit.jupiter.api.Test;
 import javax.inject.Inject;
 import javax.persistence.PersistenceException;
 import javax.validation.ConstraintViolationException;
+import java.util.Arrays;
 import java.util.List;
 import java.util.UUID;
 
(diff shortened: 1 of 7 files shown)


gwenneg:
`BehaviorGroupResources#findByBundleId` is not loading the actions while it should be. This is not related to this PR since it also happens with `master`. I'm working on a new unit test and a fix for that issue.

gwenneg:
Problem solved: the Hibernate L1 session cache needed to be cleared before executing one of the tests `SELECT`.

This is ready for a review.

src/test/java/com/redhat/cloud/notifications/db/BehaviorGroupResourcesTest.java
```diff
@@ -349,19 +326,24 @@ private Boolean muteEventType(String accountId, UUID eventTypeId) {
...
-        return session.createQuery(query, BehaviorGroupAction.class)
-                .setParameter("behaviorGroupId", behaviorGroupId)
-                .getResultList()
-                .await().indefinitely();
+    private void updateAndCheckBehaviorGroupActions(String accountId, UUID bundleId, UUID behaviorGroupId, boolean expectedResult, UUID... endpointIds) {
+        Boolean updated = behaviorGroupResources.updateBehaviorGroupActions(accountId, behaviorGroupId, Arrays.asList(endpointIds)).await().indefinitely();
+        // Is the update result the one we expected?
+        assertEquals(expectedResult, updated);
+        if (expectedResult) {
+            session.clear(); // We need to clear the session L1 cache before checking the update result.
```
josejulio, line 334: This looks like something that keeps appearing - At least is the second time we have a problem with the cache IIRC.
In this case, is it because we are using a single session for the whole suite and it doesn't "pick" the changes?

josejulio (approved):
lgtm

src/test/java/com/redhat/cloud/notifications/db/BehaviorGroupResourcesTest.java
gwenneg: Yes, that's exactly the reason. Here's what's happening:
- First we `persist` or `select` an entity which is usually (not always) cached into the L1 cache.
- Then we update it with an HQL or a native query which does not affect the L1 cache content.
- Finally we select the entity again and Hibernate returns the L1 cache entity state because it's not aware that the entity was updated in the database with the HQL/native query.

It only happens in tests where `*Resources` are injected and used to execute multiple queries. It can be a bit annoying while writing tests, but I don't think it's a problem in the end. Testing all database operations directly (not through the app REST APIs) is important for the tests coverage and reliability. Clearing the Hibernate session is a small price to pay to reach this goal 😃

gwenneg:
Thanks for the multiple reviews!

PR #197, opened 2021-04-26
Stop transforming Uni to Multi when not relevant

gwenneg:
Hibernate Reactive result sets are emitted as a `Uni` item. We should not transform them to `Multi` when we don't need to.

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
+++ b/src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
@@ -108,7 +108,7 @@ public class ApplicationResources {
                 .onItem().transform(rowCount -> rowCount > 0);
     }
 
-    public Multi<EventType> getEventTypes(Query limiter, Set<UUID> appIds, UUID bundleId) {
+    public Uni<List<EventType>> getEventTypes(Query limiter, Set<UUID> appIds, UUID bundleId) {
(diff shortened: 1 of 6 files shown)


src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
```diff
@@ -139,8 +139,7 @@
                     .setFirstResult(limiter.getLimit().getOffset());
         }
 
-        return mutinyQuery.getResultList()
-                .onItem().transformToMulti(Multi.createFrom()::iterable);
+        return mutinyQuery.getResultList();
```
josejulio, line 142: What's the preference for returning a `Multi` vs a `Uni<List>` ? Is this because of sql session?

gwenneg: `Uni` is a simpler and lightweight version of `Multi`. We should always favor `Uni` over `Multi` when we don't need any of the `Multi` features, which is the case in all modified classes in this PR. Hibernate always returns `Uni` for all of its asynchronous operations, which is one more reason to stick with `Uni` whenever possible. There are of course many cases when we need a `Multi` and it's perfectly fine to use them, even though some things we do with `Multi` could be done with simple Java streams in notifications-backend.

josejulio (approved):
lgtm

PR #198, opened 2021-04-26
Add EmailAggregationResourcesTest

gwenneg:
This PR fixes a small Hibernate migration oversight (which is luckily harmless) in an HQL query.

Since it scared me for a second, I added a few tests to better detect `EmailAggregationResources` regressions.

### DIFF
--- a/src/test/java/com/redhat/cloud/notifications/db/EmailAggregationResourcesTest.java
+++ b/src/test/java/com/redhat/cloud/notifications/db/EmailAggregationResourcesTest.java
@@ -1,0 +1,112 @@
+package com.redhat.cloud.notifications.db;
+
+import com.redhat.cloud.notifications.TestLifecycleManager;
+import com.redhat.cloud.notifications.models.EmailAggregation;
+import com.redhat.cloud.notifications.models.EmailAggregationKey;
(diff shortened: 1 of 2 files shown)


josejulio:
> Since it scared me for a second, I added a few tests to better detect `EmailAggregationResources` regressions.

I think there is at least one test that touches (indirectly) that part too, but the more tests the better :)

src/test/java/com/redhat/cloud/notifications/db/EmailAggregationResourcesTest.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@QuarkusTest
+@QuarkusTestResource(TestLifecycleManager.class)
+public class EmailAggregationResourcesTest extends DbIsolatedTest {
+
+    private static final ZoneId UTC = ZoneId.of("UTC");
```
josejulio, line 23: nitpick, we could use [ZoneOffset.UTC](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneOffset.html#UTC) which is more typesafe

PR #202, opened 2021-04-27
NOTIF-55 Migrate to resteasy-reactive and rest-client-reactive

gwenneg:
Blocked by https://github.com/quarkusio/quarkus/issues/16577: the `x-rh-identity` header is not sent because of a Quarkus bug, which leads to a failed RBAC call.

It should be fixed in Quarkus `1.13.3.Final`.

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/routers/InternalService.java
+++ b/src/main/java/com/redhat/cloud/notifications/routers/InternalService.java
@@ -11,6 +11,7 @@ import io.smallrye.mutiny.Uni;
 import javax.inject.Inject;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
+import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.NotFoundException;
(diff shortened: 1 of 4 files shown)


gwenneg:
Quarkus `1.13.3.Final` was released yesterday.

https://github.com/quarkusio/quarkus/issues/16577 was fixed as expected, but I found what seems to be a new bug: https://github.com/quarkusio/quarkus/issues/16897

gwenneg:
I worked around the bug and the build is now successful. Unfortunately, the response time is really bad (500 to 600ms now vs 10 to 100ms before) so I'm trying to figure out why the app is so slow.

gwenneg:
⚠️ Please contact me before merging this PR. I'll need to monitor prod after this is deployed.

gwenneg:
Rebased on master and tested with the merged preparation work (#235). It looks good.

This is ready for a review/merge.

josejulio (approved):
Looks good!

src/main/java/com/redhat/cloud/notifications/routers/InternalService.java
```diff
@@ -11,18 +11,24 @@
...
+import javax.ws.rs.Produces;
 import javax.ws.rs.core.Response;
 import java.util.List;
 import java.util.UUID;
 
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
 @Path("/internal")
+@Consumes(APPLICATION_JSON)
+@Produces(APPLICATION_JSON)
```
josejulio, line 31: I thought this was the default and we didn't need to specify it?

gwenneg: That is true for `quarkus-resteasy` but apparently we have to add these annotations back with `quarkus-reasteasy-reactive`. I suspect the reactive extension is missing some bytecode enhancement. I'll investigate that later (or talk with the resteasy-reactive team).

PR #203, opened 2021-04-28
Clean internal APIs and improve tests coverage

gwenneg:
This is the first part of the flat/nested APIs changes, focused only on internal APIs.

- `BundleService` and `ApplicationService` have been replaced with `InternalService`. Why do that? Because dealing the `bundle > app > event type` hierarchy APIs in separate classes created a few problems (APIs overlap, flawed paths, unconsistent ways to access resources...). These issues have been fixed and maintenance should be easier now, hopefully.

- `GET /internal/applications?bundleName=foo` was removed because it was redundant with `GET /internal/bundles/{bundleId}/applications`. It was the only internal API based on a resource name. Now all internal APIs are based on resources ID. Applications can still be retrieved from the bundle name in the public APIs.

- `POST /internal/applications` [1] was removed because it was redundant with `POST /internal/bundles/{bundleId}/applications` [2]. Why keep [2] instead of [1]? Because event types are created with an API similar to [2] so I favored APIs consistency.

- `POST /internal/bundles/{bundleId}/applications` now returns `404` (instead of `500` before) when the bundle is not found.

- `GET /internal/bundles/{bundleId}/applications` now returns `404` (instead of `200` before) when the bundle is not found.

- `GET /internal/applications/{appId}` now returns `404` (instead of `204` before) when the app is not found.

- `POST /internal/applications/{appId}/eventTypes` now returns `404` (instead of `500` before) when the app is not found.

- `GET /internal/applications/{appId}/eventTypes` now returns `404` (instead of `200` before) when the app is not found.

- `BundleServiceTest` and `ApplicationServiceTest` have been replaced with `InternalServiceTest` which includes all previously existing tests and much more!


src/main/java/com/redhat/cloud/notifications/routers/InternalService.java
```diff
@@ -0,0 +1,145 @@
...
+                });
+    }
+
+    @PUT
+    @Path("/bundles/{bundleId}/behaviorGroups/{behaviorGroupId}/default")
+    public Uni<Response> setDefaultBehaviorGroup(@PathParam("bundleId") UUID bundleId, @PathParam("behaviorGroupId") UUID behaviorGroupId) {
+        return behaviorGroupResources.setDefaultBehaviorGroup(bundleId, behaviorGroupId)
+                .onItem().transform(rowCount -> {
+                    if (rowCount == 0) {
+                        return Response.status(Response.Status.NOT_FOUND).build();
```
josejulio, line 77: I think this would return `NOT_FOUND` if the `behaviorGroupId` is already the default behavior group.

src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
```diff
@@ -92,9 +84,13 @@
 
     public Multi<EventType> getEventTypes(UUID appId) {
```
josejulio, line 85: This probably has been changed already in master to `Uni<List<EventType>>`, but leaving a note in case is not.

src/main/java/com/redhat/cloud/notifications/routers/NotificationService.java
```diff
@@ -274,7 +274,9 @@ private String getAddress(Principal principal) {
     @Path("/facets/bundles")
     @Operation(summary = "Return a thin list of configured bundles. This can be used to configure a filter in the UI")
     public Multi<Facet> getBundleFacets(@Context SecurityContext sec) {
```
josejulio, line 276: I think this could also be a `Uni<List<Facet>>`

src/test/java/com/redhat/cloud/notifications/TestLifecycleManager.java
```diff
@@ -54,6 +56,7 @@ private void configureObjectMapper() {
         ObjectMapper mapper = DatabindCodec.mapper();
         mapper.setFilterProvider(filterProvider);
         mapper.registerModule(new JavaTimeModule());
+        mapper.configure(FAIL_ON_UNKNOWN_PROPERTIES, false);
```
josejulio, line 59: Do we have extra properties in some cases? I think we should clean them if possible. Not really important for the internal case, but we should try to ensure what we are returning in the rest APIs.

src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
gwenneg: This one is still a `Multi` on master. I don't remember why, but this is one of the few cases where we really needed the `Multi`.

josejulio: We should leave a note that we require the `Multi` there and the reason if possible.

src/main/java/com/redhat/cloud/notifications/routers/NotificationService.java
gwenneg: The `Uni` is transformed to a `Multi` because we have to transform each item of the `List` individually.

We could transform the `Multi` back to a `Uni` before returning it, or never transform to a `Multi` and transform the `Uni` items using old school Java, or leave it that way. I honestly don't know what's best for performances (because all these solutions are equally maintainable).

src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
gwenneg: I think we're using too many `Multi` in the app while we rarely need them. `Multi` is initially supposed to be used to stream items, not to return and/or transform a `List` of items that is completely available at the beginning of the code execution. That could be the subject of a dedicated PR to clean the code.

josejulio (approved):
lgtm, but @pilhuhn will be happier if python helper scripts gets updated (if any is needed)

src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
josejulio: +1

We should clean and ensure we are following some guidelines of when to use Uni/Multi vs stream

gwenneg:
Thanks for the review! I updated the helper scripts and tested them locally, everything seemed OK.

gwenneg:
@lahavyuv86 The changes listed in the first comment of this PR may affect your tests.

PR #204, opened 2021-04-29
Enforce UTF_8 for all Base64 encoding/decoding

gwenneg:
`UTF_8` should always be enforced when the `String` constructor is called. Same for `String.getBytes`.


josejulio:
> `UTF_8` should always be enforced when the `String` constructor is called. Same for `String.getBytes`.

Is there any linter/rule that could enforce things like this?

PR #206, opened 2021-04-29
NOTIF-182 Stop instantiating WebClient for each webhook call

gwenneg:


gwenneg:
@cescoffier Could you please confirm that it is safe to inject a `@Singleton`-scoped `io.vertx.mutiny.ext.web.client.WebClient` instance and use it to send mutliple HTTP requests that may be executed concurrently?

cescoffier:
Yes it's fine. The WebClient is thread-safe.

gwenneg:
Thanks!

gwenneg:
We may want to configure the WebClient pool size now that it's shared. Don't review/merge yet, I'll make that configurable.

gwenneg:
This is ready for a review.

The `WebClient` max pool size can now be configured with the `webclient.max_pool_size` config property. If that property is missing in the app configuration, then the [default max pool size from Vert.x](https://github.com/eclipse-vertx/vert.x/blob/4fcd51fe01db1581b2fc97aae9843b582daf737c/src/main/java/io/vertx/core/http/HttpClientOptions.java#L35-L38) (5) will be used.

josejulio:
> [default max pool size from Vert.x](https://github.com/eclipse-vertx/vert.x/blob/3.9/src/main/java/io/vertx/core/http/HttpClientOptions.java#L34-L37) (5)

Is there any recommendation on what value to use for prod environments?

src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
```diff
@@ -32,7 +31,12 @@
     private static final String TOKEN_HEADER = "X-Insight-Token";
 
     @Inject
-    Vertx vertx;
+    @SslVerificationEnabled
+    WebClient securedWebClient;
+
+    @Inject
+    @SslVerificationDisabled
+    WebClient unsecuredWebClient;
```
josejulio, line 39: How does this work? What mechanism ensures that this Webclient uses `no SslVerification` ? 
I see that the annotation is new (I though that it was something from vertx)

gwenneg: This is a CDI injection with a qualifier. The magic happens in `WebClientProducer`. This producer can produce two `@Singleton` instances of `WebClient` depending on the qualifier used at the injection point:
- if `@SslVerificationDisabled` is present, then the injected `WebClient` will _trust_ all certificates
- if `@SslVerificationEnabled` is present, then the injected `WebClient` will _verifiy_ all certificates

gwenneg:
> Is there any recommendation on what value to use for prod environments?

That's a very good and important question. Actually, that's why I made it configurable. I don't have the answer yet but I think the default pool size will be fine for our current prod load.

src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
josejulio: Ah OK, so it will match the qualifier used in the producer. Thanks!

josejulio (approved):
lgtm

PR #209, opened 2021-04-30
Hide behavior groups APIs in OpenAPI doc

gwenneg:
cc @theute 

@josejulio Could you please confirm this will not cause any issue with the `notifications-frontend` code generation?

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/routers/NotificationService.java
+++ b/src/main/java/com/redhat/cloud/notifications/routers/NotificationService.java
@@ -147,7 +147,7 @@ public class NotificationService {
      */
     @GET
     @Path("/bg/eventTypes/affectedByRemovalOfEndpoint/{endpointId}") // TODO [BG Phase 2] Remove '/bg' path prefix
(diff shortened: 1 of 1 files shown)


josejulio:
> @josejulio Could you please confirm this will not cause any issue with the `notifications-frontend` code generation?

It will cause an issue that i might be able to workaround.
I think maybe (as a separate PR) we could have a "private-openapi.json" or so?

josejulio (approved):

gwenneg:
FYI, this is only temporary, all the operations will be shown again in the doc when behavior groups go live in prod.

gwenneg:
I don't know if that's possible to generate two OpenAPI files from the same operations, but if that works I'm OK with that.

josejulio:
> I don't know if that's possible to generate two OpenAPI files from the same operations, but if that works I'm OK with that.

We do [that](https://github.com/RedHatInsights/notifications-backend/blob/master/src/main/java/com/redhat/cloud/notifications/routers/OApiService.java#L72) already :) 

The API is split into notifications and integrations.

PR #210, opened 2021-04-30
Fixing some details found while using behavior groups

josejulio:
Found some details while connecting to the `get-behaviors-with-bundle`.

@gwenneg Could you add some tests for these cases? 
Adding comments inline with the fix to explain the problem.

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
+++ b/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
@@ -17,7 +17,6 @@ import java.util.List;
 import java.util.UUID;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.stream.Collectors;
 
 @ApplicationScoped
 public class BehaviorGroupResources {
@@ -42,14 +41,10 @@ public class BehaviorGroupResources {
     }
 
     public Uni<List<BehaviorGroup>> findByBundleId(String accountId, UUID bundleId) {
-        // When PostgreSQL sorts a BOOLEAN column in DESC order, true comes first. That's not true for all DBMS.
-        String query = "FROM BehaviorGroup b LEFT JOIN FETCH b.actions a WHERE b.accountId = :accountId AND b.bundle.id = :bundleId " +
-                "ORDER BY b.defaultBehavior DESC, b.created DESC, a.position ASC";
-        return session.createQuery(query, BehaviorGroup.class)
+        return session.createNamedQuery("findByBundleId", BehaviorGroup.class)
                 .setParameter("accountId", accountId)
                 .setParameter("bundleId", bundleId)
-                .getResultList()
-                .onItem().transform(behaviorGroups -> behaviorGroups.stream().distinct().collect(Collectors.toList()));
+                .getResultList();
     }
 
     // TODO Should this be forbidden for default behavior groups?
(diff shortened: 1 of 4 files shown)


src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -47,7 +48,8 @@
         return session.createQuery(query, BehaviorGroup.class)
                 .setParameter("accountId", accountId)
                 .setParameter("bundleId", bundleId)
-                .getResultList();
+                .getResultList()
+                .onItem().transform(behaviorGroups -> behaviorGroups.stream().distinct().collect(Collectors.toList()));
```
josejulio, line 52: I was getting the `BehaviorGroup` repeated, as many times as it has a `BehaviorGroupAction`

src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
```diff
@@ -51,6 +51,7 @@
     private String name;
 
     @NotNull
+    @JsonProperty(value = "display_name")
```
josejulio, line 54: openapi was reporting it as "displayName", but returned as display_name

src/main/java/com/redhat/cloud/notifications/models/BehaviorGroupAction.java
```diff
@@ -25,6 +25,7 @@
     @ManyToOne
     @MapsId("behaviorGroupId")
     @JoinColumn(name = "behavior_group_id")
+    @JsonIgnore
```
josejulio, line 28: This was causing a stack overflow when trying to serialize the `BehaviorGroup` ( BehaviorGroup -> BehaviorGroupAction -> BehaviorGroup -> ...)

src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
gwenneg: Both `@JsonProperty` and `@Schema` can fix that, but I would favor `@Schema` here because smallrye-open-api will soon [support a global naming strategy](https://github.com/smallrye/smallrye-open-api/pull/788) and we will remove all `@Schema` annotations when that happens.

```suggestion
    @Schema(name = "display_name")
```

src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
gwenneg: Oops, sorry about that one, it should have been tested 😃

I took the liberty of pushing a commit to slightly change the fix. Your version did the job, thanks for finding out that we needed to use `distinct` there! I moved this at a lower level, during the Hibernate query execution. Unfortunately Hibernate Reactive does not support query hints right now so I had to use a `@NamedQuery` but I'll talk with the Hibernate team about that.

gwenneg (approved):
The behavior groups classes hierarchy needs more (de)serialization tests in `NotificationServiceTest`. I'll take care of that in a separate PR if that's OK with you.

josejulio:
> The behavior groups classes hierarchy needs more (de)serialization tests in `NotificationServiceTest`. I'll take care of that in a separate PR if that's OK with you.

Thanks!

gwenneg (approved):

PR #211, opened 2021-05-03
Shorten Application and EventType POST URLs

gwenneg:
@josejulio Sorry, I need your help again reviewing an internal APIs PR 😃

This one shortens:

- `POST /applications/{appId}/eventTypes` to `POST /eventTypes`
- `POST /bundles/{bundleId}/applications` to `POST /applications`

Helper scripts were updated.

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/models/EventType.java
+++ b/src/main/java/com/redhat/cloud/notifications/models/EventType.java
@@ -52,6 +52,11 @@ public class EventType {
     @JsonInclude(NON_NULL)
     private String description;
 
+    @NotNull
+    @Transient
+    @Schema(name = "application_id")
+    private UUID applicationId;
+
     @ManyToOne(fetch = FetchType.LAZY, optional = false)
     @JoinColumn(name = "application_id")
     @JsonInclude(NON_NULL)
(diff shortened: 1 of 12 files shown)


src/main/java/com/redhat/cloud/notifications/models/EventType.java
```diff
@@ -102,6 +107,17 @@ public void setDescription(String description) {
...
 
+    public UUID getApplicationId() {
+        if (applicationId == null && application != null) {
+            applicationId = application.getId();
+        }
+        return applicationId;
+    }
+
+    public void setApplicationId(UUID applicationId) {
+        this.applicationId = applicationId;
```
josejulio, line 118: Just to be safe, should we set this `applicationId` in `setApplication` too (to keep them sync'ed)? WDYT? or maybe move the `Schema` annotations to `getApplicationId` and remove the variable altogether?

josejulio (approved):
lgtm

src/main/java/com/redhat/cloud/notifications/models/EventType.java
gwenneg: `setApplication` is actually a useless setter that could be removed. Hibernate uses the field access strategy because we have an `@Id` annotation at field level.

I had a low priority task on my todo list for a while: remove all useless getters and setters. There are a few of them in the Hibernate entities. I'll take care of that later.

gwenneg: We still need the variable `applicationId` field for tests BTW, so `getApplicationId` and `setApplicationId` can't be removed.

gwenneg:
Thanks for the review!

PR #215, opened 2021-05-04
Limit what we show in logs

theute:
This is a proposal to not log potentially private info but the info that would be useful to us.
A log line would look like:
`Processing received payload: (123456) rhel/policies/triggered-policy`

Additional I changed from DEBUG to INFO

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
+++ b/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
@@ -65,7 +65,7 @@ public class EventConsumer {
                 .onItemOrFailure()
                 .transformToUni((unused, t) -> {
                     if (t != null) {
-                        log.log(Level.INFO, "Could not process the payload", t);
+                        log.log(Level.INFO, "Could not process the payload: " + input.getPayload(), t);
                     }
                     return Uni.createFrom().voidItem();
                 });
--- a/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
+++ b/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
@@ -48,10 +48,9 @@ public class EventConsumer {
     // Can be modified to use Multi<Message<String>> input also for more concurrency
     public Uni<Void> processAsync(Message<String> input) {
         return Uni.createFrom().item(() -> input.getPayload())
-                .onItem().invoke(payload -> log.fine(() -> "Processing: " + payload))
                 .stage(self -> self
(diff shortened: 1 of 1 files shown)


gwenneg:
The new message LGTM but I wonder if we should keep logging the full payload at `fine` level in case of failure (`Could not process the payload`). When an `AvroTypeException` is thrown, the failure cause is clear, but it might now be the case if another kind of exception is thrown.

gwenneg:
If we keep logging the payload in case of failure, we should probably do that at the `onItemOrFailure()` step.

gwenneg:
Can a payload contain info that should never be logged?

josejulio:
> Can a payload contain info that should never be logged?

I guess it could, up to the application, the payloads contain free-form data on `payload` and `context`

theute:
In case of failure we should still know why it failed though (maybe strip the data and context sections)

josejulio:
> In case of failure we should still know why it failed though (maybe strip the data and context sections)

What about (in the future) if we store last X failing payloads to be able to show them in the internal frontend to follow up?

theute:
Once we have the admin UI we should keep some in memory yes.
In the meantime, the log here is what help us help them

Let's improve on the existing, if we are ok with the positive case. We can move where we log for failures but I'm not sure I understand how to do it (this mutiny thing is new to me)

josejulio:
You can change https://github.com/RedHatInsights/notifications-backend/pull/215/files#diff-0e31a1d690e7134bc5653f78759c63ef9bf30220941dc854fdf5296cad9db840R68 to something like:
```
log.log(Level.INFO, "Could not process the payload: " + input.getPayload(), t);
```
That will log if there is any error in the whole process (is that correct @gwenneg ?)

gwenneg:
That's exactly what I had in mind.

theute:
Sorry, I'm still unclear where exactly or just change line 68 ?

gwenneg:
You need to replace line 68 from:
`log.log(Level.INFO, "Could not process the payload", t);`
to:
`log.log(Level.INFO, "Could not process the payload: " + input.getPayload(), t);`

Now, do we want to log that at `INFO` or `FINE` level? I'd say `FINE` but I'm not sure.

theute:
Ok I get why I was confused at first, the log line already shows the payload in the error message (you couldn't know), so I don't think it adds anything here.

theute:
Scratch that, I'm stupid / need sleep...

pilhuhn (approved):

PR #218, opened 2021-05-05
add dependabot

bennetelli:


gwenneg (changes requested):
Could you please add an additional `package-ecosystem` for GitHub Actions?

See https://docs.github.com/en/code-security/supply-chain-security/keeping-your-actions-up-to-date-with-dependabot.

.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
...
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    ignore:
+      - dependency-name: "org.apache.maven.plugins:maven-compiler-plugin"
```
gwenneg, line 13: Why should this be ignored?

bennetelli: got it from the quarkus splunk extension (https://github.com/quarkiverse/quarkus-logging-splunk/blob/main/.github/dependabot.yml) and thought it could happen that dependabot wants to update transitive dependencies of compiler plugin, but I didn't challenge it. We can delete it and see if dependabot erroneously wants to update transitive dependencies of compiler plugin.

gwenneg: Let's try without this and see how it goes. AFAIK, dependabot only deals with direct dependencies and does not check transitive one.

gwenneg (approved):
LGTM

PR #221, opened 2021-05-05
NOTIF-187 Return 200 only if behavior group actions update is successful

gwenneg:


src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -143,8 +149,7 @@
             return session.createQuery(checkQuery, UUID.class)
                     .setParameter("accountId", accountId)
                     .setParameter("id", behaviorGroupId)
-                    .getSingleResultOrNull()
-                    .onItem().ifNull().failWith(new NotFoundException("Behavior group not found: " + behaviorGroupId))
+                    .getSingleResult()
```
gwenneg, line 152: A `NoResultException` is thrown here if the behavior group was not found.

josejulio (approved):
lgtm

josejulio:
Thanks!

PR #230, opened 2021-05-06
add cloudwatch logging

bennetelli:


gwenneg (commented):
Will that cause the issue we discussed (Vert.x blocked thread)?

bennetelli:
No it won't. It's working locally. Even in combination with the DbCleaner thing which did some strange things with the vert.x even loop. I would like to commit it (it's disabled by default), enabled it on a pod and test it on ci. It's disabled on every other stage anyway.

gwenneg (approved):

PR #235, opened 2021-05-10
NOTIF-55 Use Multi when relevant, replace with Uni<List> otherwise

gwenneg:
This is preparation work for #202.

I found something about the response time issue with `quarkus-resteasy-reactive`: the response time can be bad only with APIs that return `Multi`. There's no problem with methods returning `Uni<List>`, which should always be the type we use anyway because we never need to stream data back to the client. I created https://github.com/quarkusio/quarkus/issues/17094 to report the issue with `Multi` to the Quarkus team.

Since we're using `Multi` in many places for wrong reasons, I did some spring cleaning and replaced `Multi` with `Uni<List>`. A few methods could be cleaned further but I chose not to because they will be removed with phase 2 of behavior groups.

Here are some ground rules for future developments:
- The persistence layer (`*Resources` classes) should not return `Multi` unless there's a _very good_ reason to do so (usually, there's not).
- The API layer (`*Service` classes) should not return `Multi` unless we need to stream data back to the client (that's very unlikely in this app).
- We should refrain from using `Multi` to transform items from `Uni<List>`. There are situations when it makes sense but most of the time it doesn't. In particular, side-effects should be implemented with `onItem().invoke()` (synchronous, for code that will not block the I/O thread) or `onItem().call()` (asynchronous, for DB queries for example).

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
+++ b/src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
@@ -6,6 +6,8 @@ import com.redhat.cloud.notifications.auth.rhid.RhIdPrincipal;
 import com.redhat.cloud.notifications.db.ApplicationResources;
 import com.redhat.cloud.notifications.db.BundleResources;
 import com.redhat.cloud.notifications.db.EndpointEmailSubscriptionResources;
+import com.redhat.cloud.notifications.models.Application;
+import com.redhat.cloud.notifications.models.EmailSubscription;
 import com.redhat.cloud.notifications.models.EmailSubscriptionType;
 import com.redhat.cloud.notifications.routers.models.SettingsValueJsonForm;
 import com.redhat.cloud.notifications.routers.models.SettingsValues;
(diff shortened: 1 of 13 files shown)


src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
```diff
@@ -147,48 +149,57 @@
...
+            throw new BadRequestException("bundleName must have a value");
+        }
+
+        /*
+         * The following stream creates a `SettingsValues` instance and then populates it by using side-effects.
+         * It is done by invoking to kind of functions:
+         * - database queries are executed in an asynchronous way using `.onItem().call(...)`
+         * - side-effects are executed in a synchronous way using `.onItem().invoke(...)`
+         */
+        return Uni.createFrom().item(() -> new SettingsValues())
```
gwenneg, line 162: This is roughly the same code than before but we're no longer using `Multi` in this method.

gwenneg:
Unlike #202 (see [this](https://github.com/RedHatInsights/notifications-backend/pull/202#issuecomment-836453183)), this PR can be merged as soon as the review is done.

gwenneg:
Rebased and fixed conflicts.

josejulio:
> Rebased and fixed conflicts.

sorry, merged something and there is one conflict  now

josejulio (approved):

gwenneg:
I solved the conflict. Thanks for the review!

PR #238, opened 2021-05-11
NOTIF-193 Support multiple endpoint properties types

gwenneg:
With this PR, we can now persist multiple types of endpoint properties.

⚠️ All `LEFT JOIN FETCH e.webhook` HQL instructions have been replaced with a separate HQL query that will fetch the endpoint properties. This will impact performances: if we retrieved `n` endpoints with a single query before, we may have up to `1+n` queries executed now. Keep in mind that the previous implementation was flawed and not usable for more than one endpoint properties type.

❗ As a consequence, in a separate PR, we should review all APIs that involve an execution of `EndpointResources#loadProperties` and try to avoid that execution whenever possible. There are probably APIs which return the endpoint properties while the UI does not need them.

💡 If the impact on performances it too bad, we could consider using a single table for all endpoint properties types. That table would contain all columns for all endpoint properties types (each record would contain lots of empty columns) and we could restore the `LEFT JOIN FETCH` logic instead of executing a separate query. The data migration from this PR would still be required so let's see how it goes with this version and maybe improve performances later.


src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -292,46 +316,67 @@
...
+                                        .setParameter("secretToken", attr.getSecretToken())
+                                        .setParameter("endpointId", endpoint.getId())
+                                        .executeUpdate()
+                                        .call(session::flush)
+                                        .onItem().transform(rowCount -> rowCount > 0);
+                            case EMAIL_SUBSCRIPTION:
+                            case DEFAULT:
+                            default:
+                                return Uni.createFrom().item(Boolean.TRUE);
+                        }
```
gwenneg, line 335: Modify this to _update_ a new type of endpoint properties.

src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -292,46 +316,67 @@
...
+                        .onItem().ifNotNull().invoke(webhook -> {
+                            WebhookAttributes properties = buildWebhookAttributes(webhook);
+                            endpoint.setProperties(properties);
+                        })
+                        .replaceWith(endpoint);
+            case EMAIL_SUBSCRIPTION:
+            case DEFAULT: // TODO [BG Phase 2] Delete this case
+            default:
+                // Properties loading is not supported for the endpoint type.
+                return Uni.createFrom().item(endpoint);
```
gwenneg, line 357: Modify this to _load_ a new type of endpoint properties.

pilhuhn (approved):

pilhuhn:
Thank you!

PR #242, opened 2021-05-14
NOTIF-150 Fix behavior group actions

josejulio:
With NOTIF-193, we now have to load manually the endpoints properties

This fixes an issue seen in ci/qa

PR #243, opened 2021-05-14
Batch "properties endpoint" SQL queries

josejulio:
@gwenneg This is an idea I had while working on https://github.com/RedHatInsights/notifications-backend/pull/242 to reduce the N-required queries to only 1 per type.

Could you review it does what I think it does? :) - Using only 1 extra SQL call per found type -

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
+++ b/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
@@ -1,8 +1,8 @@
 package com.redhat.cloud.notifications.db;
 
-import com.redhat.cloud.notifications.models.EndpointProperties;
 import com.redhat.cloud.notifications.models.Endpoint;
 import com.redhat.cloud.notifications.models.EndpointDefault;
+import com.redhat.cloud.notifications.models.EndpointProperties;
 import com.redhat.cloud.notifications.models.EndpointTarget;
 import com.redhat.cloud.notifications.models.EndpointType;
 import com.redhat.cloud.notifications.models.EventType;
--- a/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
+++ b/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
@@ -1,5 +1,6 @@
 package com.redhat.cloud.notifications.db;
 
+import com.redhat.cloud.notifications.models.EndpointProperties;
 import com.redhat.cloud.notifications.models.Endpoint;
 import com.redhat.cloud.notifications.models.EndpointDefault;
 import com.redhat.cloud.notifications.models.EndpointTarget;
@@ -15,8 +16,10 @@ import javax.inject.Inject;
 import javax.persistence.PersistenceException;
 import javax.ws.rs.BadRequestException;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.UUID;
 import java.util.function.Function;
 import java.util.logging.Logger;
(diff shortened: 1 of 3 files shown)


src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -337,25 +342,46 @@
...
+                                endpoint.setProperties(properties);
+                                return endpoint;
+                            })
+                            .collect(Collectors.toList())
+                    );
+        } else {
+            webhookUnis = Uni.createFrom().item(Collections.emptyList());
+        }
+
+        return webhookUnis.replaceWith(endpoints);
```
gwenneg, line 374: It should be possible to get rid of that returned value and return `Uni<Void>` instead. There's only one case where this is not possible but it could be transformed to an `onItem().call()` execution:

https://github.com/RedHatInsights/notifications-backend/blob/3ac94434be11f81ebd3b2ec2916c17c60361e2a6/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java#L172

src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -337,25 +342,46 @@
                 });
     }
 
+    public Uni<List<Endpoint>> loadProperties(List<Endpoint> endpoints) {
+        // Group endpoints in types and load in batches each type.
+        // Currently only webhook types have properties.
+        endpoints = endpoints.stream().distinct().collect(Collectors.toList());
```
gwenneg, line 348: `Endpoint` has `equals` and `hashCode` implementations. Creating a `Set` from the `List` would probably be faster than the dealing with the duplicates with a stream.

src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -78,8 +83,8 @@
         }
 
         return mutinyQuery.getResultList()
-                .onItem().transformToMulti(Multi.createFrom()::iterable)
-                .onItem().transformToUniAndConcatenate(this::loadProperties);
+                .onItem().call(this::loadProperties)
+                .onItem().transformToMulti(Multi.createFrom()::iterable);
```
gwenneg, line 87: The `Uni` -> `Multi` transformation can probably be avoided in most cases thanks to your idea.

gwenneg (commented):
That's an excellent idea!

I added a few comments about some details that could be improved a bit.

src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
josejulio: You mean this whole function to return Uni<Void>?

gwenneg: Yes, because Mutiny discards the returned value during an `onItem().call()` execution.

gwenneg: It only needs to know when the execution is done before proceeding downstream.

josejulio:
@gwenneg Rebased, addressed comments and made it simpler to add more types into the `loadProperties`

src/main/java/com/redhat/cloud/notifications/routers/NotificationService.java
```diff
@@ -365,17 +366,14 @@ private String getAddress(Principal principal) {
...
                                 .map(BehaviorGroup::getActions)
-                                .map(bga -> bga.size() == 0 ? Uni.createFrom().voidItem() : Uni.combine().all().unis(
-                                        bga.stream()
-                                                .map(BehaviorGroupAction::getEndpoint)
-                                                .map(endpoint -> endpointResources.loadProperties(endpoint))
-                                                .collect(Collectors.toList())
-                                        )
-                                                .discardItems()
-                                ).collect(Collectors.toList())
-                ).discardItems());
```
gwenneg, line 379: I just saw while working on something else that there is a concurrency issue here. `Uni.combine()` does not work well with the Hibernate Reactive session because `combine` will cause concurrent execution while the Hibernate session must not be used concurrently.

I'm not fixing it since it will soon be replaced with something else.

josejulio:
Rebased @gwenneg

src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -347,21 +346,39 @@
                 });
     }
 
+    public Uni<Void> loadProperties(List<Endpoint> endpoints) {
+        // Group endpoints in types and load in batches for each type.
+        Set<Endpoint> endpointSet = new HashSet<>(endpoints);
+
+        return Uni.combine().all().unis(
+                this.loadTypedProperties(WebhookProperties.class, endpointSet, EndpointType.WEBHOOK)
+        ).discardItems();
```
gwenneg, line 355: This will become a problem when the `EndpointType.CAMEL` is added because `Uni.combine()` may execute SQL queries concurrently with the same `Mutiny.Session`.

Chaining the `loadTypedProperties` calls will produce the expected result without any concurrency issue:

```
return loadTypedProperties(WebhookProperties.class, endpointSet, EndpointType.WEBHOOK)
        .chain(() -> loadTypedProperties(WebhookProperties.class, endpointSet, EndpointType.CAMEL));
```

src/main/java/com/redhat/cloud/notifications/routers/NotificationService.java
```diff
@@ -365,17 +366,14 @@ private String getAddress(Principal principal) {
...
+                                .flatMap(Collection::stream)
```
gwenneg, line 373: I would add `.filter(Objects::nonNull)` before `.flatMap(Collection::stream)` just to be safe. There's no NPE now but the code may evolve in a way that will cause a NPE later.

gwenneg (changes requested):

src/main/java/com/redhat/cloud/notifications/routers/NotificationService.java
josejulio: Do we have null endpoints? I wondering if adding this filter would potentially hide other issue.

gwenneg: That would filter out behavior groups when `getActions() == null` and prevent a NPE throw when `.stream()` is collected on the `null` collection. We don't have that right now, but who knows in the future... :)

gwenneg (approved):
LGTM

PR #248, opened 2021-05-18
Merge endpoint properties API and persistence classes

gwenneg:
- `Attributes` has been replaced with `EndpointProperties`.
- `WebhookAttributes` and `EndpointWebhook` have been merged, the resulting class is `WebhookProperties`.
- `EmailSubscriptionAttributes` is now `EmailSubscriptionProperties`. It is ready for JPA persistence (it wasn't before) but still not persisted.


gwenneg:
cc @josejulio

I think that's one step towards the right direction and yet I'm not completely happy with this.

We could (should?) get rid of the `Endpoint.properties` deserialization with Jackson's `@JsonSubTypes` and use separate `Endpoint.webhookProperties` and `Endpoint.emailSubscriptionProperties` instead. But that would have an impact on the API and maybe make things harder from an API consumer point of view.

I'm running a few manual tests even if the GitHub build passed to make sure I didn't miss anything.

gwenneg:
Local tests successful, I'm promoting the PR.

josejulio:
> We could (should?) get rid of the `Endpoint.properties` deserialization with Jackson's `@JsonSubTypes` and use separate `Endpoint.webhookProperties` and `Endpoint.emailSubscriptionProperties` attributes instead. But that would have an impact on the API and maybe make things harder from an API consumer point of view.

Yes, maybe it would complicate things more.

src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -33,9 +32,15 @@
                 .onItem().call(() -> {
                     // If the endpoint properties are null, they won't be persisted.
                     if (endpoint.getProperties() != null) {
+                        /*
+                         * As weird as it seems, we need the following line because the Endpoint instance was
+                         * deserialized from JSON and that JSON did not contain any information about the
+                         * @OneToOne relation from EndpointProperties to Endpoint.
+                         */
+                        endpoint.getProperties().setEndpoint(endpoint);
```
josejulio, line 40: Is this the only method that receives the `Endpoint` ?  (I'm actually talking about the REST API that reaches up to this point)
or maybe we need to add this logic as part of the de-serialization process?

gwenneg: That's the only one where we have to do that. `PUT /endpoints/{id}` also receives an `Endpoint` but in that case the relation between `EndpointProperties` and `Endpoint` already exists.

We could do something during the deserialization, but that would probably mean writing multiple custom Jackson deserializers (one for each type of property) and duplicating the code that establishes the missing relation in each deserializer. That would be a lot of code to get rid of that single line :)

josejulio: Then it's fine to have it here I think.

josejulio (approved):
lgtm

PR #252, opened 2021-05-19
NOTIF-150 Lazily create a default behavior group

gwenneg:
With this PR, if an account calls `GET /notifications/bundles/{bundleId}/behaviorGroups` for the first time and the database does not contain any behavior group for that account, then a default behavior group will be created.

From an UI point of view, this means we will always have a default behavior group available.

The display name of that default group is currently hardcoded, we'll need to change that later when i18n is added to the app.

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/routers/NotificationService.java
+++ b/src/main/java/com/redhat/cloud/notifications/routers/NotificationService.java
@@ -9,7 +9,6 @@ import com.redhat.cloud.notifications.db.BundleResources;
 import com.redhat.cloud.notifications.db.EndpointResources;
 import com.redhat.cloud.notifications.db.Query;
 import com.redhat.cloud.notifications.models.BehaviorGroup;
-import com.redhat.cloud.notifications.models.BehaviorGroupAction;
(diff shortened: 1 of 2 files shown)


src/main/java/com/redhat/cloud/notifications/routers/NotificationService.java
```diff
@@ -364,18 +366,51 @@ private String getAddress(Principal principal) {
...
-                                .map(BehaviorGroup::getActions)
-                                .map(bga -> bga.size() == 0 ? Uni.createFrom().voidItem() : Uni.combine().all().unis(
-                                        bga.stream()
-                                                .map(BehaviorGroupAction::getEndpoint)
-                                                .map(endpoint -> endpointResources.loadProperties(endpoint))
-                                                .collect(Collectors.toList())
-                                        )
-                                                .discardItems()
-                                ).collect(Collectors.toList())
-                ).discardItems());
```
gwenneg, line 379: I didn't plan to change this at first but then a NPE showed up during tests and I ended up fixing other issues. This will be replaced soon by @josejulio's batch loading idea.

PR #253, opened 2021-05-19
Mark endpoints that return a string as text/plain

josejulio:
So this is probably be cause of the recent migration to reasteasy-reactive-*.

The UI code started failing because it was receiving empty/string responses with the "application/json" content-type in the response. It looks like we need to annotate each method. I think that the previous implementation used the `@Schema` annotation to decide.

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
+++ b/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
@@ -154,6 +154,7 @@ public class EndpointService {
     @Path("/{id}")
     @RolesAllowed(RbacIdentityProvider.RBAC_WRITE_INTEGRATIONS_ENDPOINTS)
     @APIResponse(responseCode = "204", description = "The integration has been deleted", content = @Content(schema = @Schema(type = SchemaType.STRING)))
+    @Produces(MediaType.TEXT_PLAIN)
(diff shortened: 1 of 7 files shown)


bennetelli (approved):

gwenneg:
Thanks for finding the cause of the UI failures!

We should probably stop declaring `@Produces` and `@Consumes` at class level and only use explicit annotations at method level.

bennetelli:
Yes, I would do it on method level as well. We will probably have to do it anyway when we want to provide different response formats like JSON, XML, text/plain,... for the same resoruce url. 
Putting it on the methods instead of class gives us more flexibility IMO.

josejulio:
> We should probably stop declaring `@Produces` and `@Consumes` at class level and only use explicit annotations at method level.

Most of our endpoints are `application/json`, won't that be too verbose? 
Either way, we could start testing the `content-type` on the calls.

josejulio:
I'm going to go ahead and merge this one to fix ci/qa

gwenneg:
> Most of our endpoints are `application/json`, won't that be too verbose?
> Either way, we could start testing the `content-type` on the calls.

It would be a bit verbose indeed, but also give us full control over what's going on. We just had a bug because we were relying on a Quarkus implicit behavior and that behavior changed without us noticing. Checking the content-type in tests in a very good way to prevent that from happening again in the future, but I would also declare content-types explicitely on methods.

PR #254, opened 2021-05-20
Remove EventBusTypeProcessor

gwenneg:
`EventBusTypeProcessor` does absolutely nothing and yet it is called each time a Kafka message is processed. We should remove it.


josejulio (approved):
IIRC, this was going to be used to notify about incoming notifications. We can add it back if/when we need it.

PR #256, opened 2021-05-20
Test the content-type header returned by all REST APIs

gwenneg:
Also use method-level `@Consumes` and `@Produces`.

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
+++ b/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
@@ -41,16 +41,16 @@ import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.Context;
-import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.SecurityContext;
 import java.util.HashMap;
 import java.util.List;
 import java.util.UUID;
 
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static javax.ws.rs.core.MediaType.TEXT_PLAIN;
+
 @Path(Constants.API_INTEGRATIONS_V_1_0 + "/endpoints")
-@Consumes(MediaType.APPLICATION_JSON)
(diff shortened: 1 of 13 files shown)


gwenneg:
@josejulio This should help us prevent any future regression because of a content-type change caused by Quarkus.

src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -222,7 +231,6 @@
             )
     })
     @APIResponse(responseCode = "200", content = @Content(schema = @Schema(type = SchemaType.STRING)))
-    @Produces(MediaType.TEXT_PLAIN)
```
josejulio, line 225: The schema is `SchemaType.String` but it says it produces an `APPLICATION_JSON`. Which one is correct?

gwenneg: Is that incompatible? I don't know `SchemaType` very well but I don't see any specific value for JSON content and `STRING` seems to be the closest thing available.

`NotificationResources#getNotificationDetails` returns a `JsonObject` instance, that's why I changed the `MediaType` to `APPLICATION_JSON`.

gwenneg: It seems we could (or should?) replace with `@Content(schema = @Schema(type = SchemaType.STRING)` with `@Content(mediaType = MediaType.APPLICATION_JSON)` when returning JSON. Now I'm wondering why we would return `SchemaType` at all instead of always return `MediaType` 😄

gwenneg: With `SchemaType.STRING`:

image

With `MediaType.APPLICATION_JSON`:

image

I guess the second one a bit cleaner but there's no breaking change.

josejulio: > Now I'm wondering why we would return `SchemaType` at all instead of always returning `MediaType`

`SchemaType` helps the openapi creating when we have a custom object; in this case is a string. I'm not sure if it should be plain text. But you are right, in this case application/json is OK vs plaintext.

josejulio (approved):
LGTM,

PR #261, opened 2021-05-25
NOTIF-199 Create email template for RHEL Advisor / resolved recommend…

bennetelli:
…ation


bennetelli:
Would be great if you could take a look at it. Especially at the TestRHELAdvisorTemplate class. 

I am not sure if I should take the OpenShift Advisor template or the existing plain advisor one.
One example:

that Advisor OpenShift one has this content "RHEL - Advisor Instant Notification - 3 Oct 2020" 
whereas the plain Advisor one had this one: "Advisor instant notification - 03 Oct 2020 15:22 UTC - 4 recommendations"

theute:
This is the daily version of RHEL Advisor Instant, so the RHEL Advisor Instant is the right one

src/main/resources/templates/AdvisorRHEL/insightsEmailBody.html
```diff
@@ -580,7 +580,7 @@ <h1><b>{#insert content-title}{/}</b></h1>
                                 <tbody>
                                 <tr>
                                     <td class="rh-footer__text">
-                                        This email was sent by Red Hat Insights |<a href="https://cloud.redhat.com/user-preferences/notification/rhel" target="_blank">&nbsp;Manage notification preferences</a>
+                                        This email was sent by Red Hat Insights |<a href="https://cloud.redhat.com/user-preferences/notifications/openshift" target="_blank">&nbsp;Manage email preferences</a>
```
theute, line 583: The URL was correct here should be https://cloud.redhat.com/user-preferences/notifications/rhel I think ?
The text of the link was correct https://marvelapp.com/prototype/5gdjb12/screen/76689545

src/main/resources/templates/AdvisorRHEL/resolvedRecommendationInstantEmailBody.html
```diff
@@ -1,5 +1,5 @@
-{#include Advisor/insightsEmailBody}
-{#content-title}Advisor instant notification - {action.timestamp.toUtcFormat()}{/content-title}
+{#include AdvisorRHEL/insightsEmailBody}
+{#content-title}RHEL - Advisor Instant Notification - {action.timestamp.format('d MMM uuuu')}{/content-title}
```
theute, line 2: We would need UXD approval if this change, and then applied everywhere, we need UTC time

src/main/resources/templates/AdvisorRHEL/resolvedRecommendationInstantEmailBody.html
```diff
@@ -31,7 +31,7 @@
                                                     <tr>
                                                         <td class="rh-card__header">
                                                             <h2>
-                                                                New recommendations
+                                                                Resolved recommendations
```
theute, line 34: I see that this templates *replaces* the "New recommendations" email, we need both, the 'new' and 'resolved'

theute (changes requested):
Looks like there were confusions:
 - No change to the UXD should happen unless asked/confirmed with UXD team (the date is particularly problematic
 - If I read the changelog right, it seems that the existing "new recommendation" template is gone, we need both like for Policies.

src/main/resources/templates/AdvisorRHEL/resolvedRecommendationInstantEmailBody.html
bennetelli: It actually doesn't replace it. It's another file (AdvisorRHEL/resolvedRecommendationInstantEmailBody.html ) I copy pasted the files from new recommendations first, committed it then to let josejulio take a look at it and then renamed it afterwards.

src/main/resources/templates/AdvisorRHEL/resolvedRecommendationInstantEmailBody.html
bennetelli: in the last commit I took the old text from the plain advisor.

bennetelli:
I just changed the expected test output back to the former expectations and changed the html template and css back as well.

PR #262, opened 2021-05-25
NOTIF-150 Remove account-level default behavior group

gwenneg:
@josejulio Let's get rid of that account-level default behavior group idea. We will probably reintroduce something similar but at app or bundle-level later.


josejulio (approved):
lgtm

PR #265, opened 2021-05-26
NOTIF-199 Add Resolved Recommendations

bennetelli:


src/main/resources/templates/Advisor/resolvedRecommendationInstantEmailTitle.txt
```diff
@@ -0,0 +1 @@
+Advisor instant notification - {action.timestamp.toUtcFormat()} - {action.events.size()} recommendation{#if action.events.size() > 1}s{/if}
```
josejulio, line 1: I think we should update it to say resolved recommendation(s)  (and also update the newRecommendation~ to say new recommendation(s))
WDYT @theute  ?

```suggestion
Advisor instant notification - {action.timestamp.toUtcFormat()} - {action.events.size()} resolved recommendation{#if action.events.size() > 1}s{/if}
```

bennetelli: Wow! That suggested change feature is great! 

If we wouldn't go for your suggested fix @josejulio I would also like to know your opinion about using only one template for both.

josejulio (approved):
lgtm, one minor comment - but I'm unsure if the designs cover that - need to check.

src/main/resources/templates/Advisor/resolvedRecommendationInstantEmailTitle.txt
josejulio: If we are going to use the same text for both, it would be OK to use the same file.
Something else would be to use `qute include section` to parametrize the templates, but I'm unsure, in the future we are going to have these templates elsewhere (maybe in the database) and not sure how is that going to work.

theute: I agree I think we should distinguish the email title. (I'm letting Jerome know)

bennetelli: Wouldn't it be more intuitive and clearer if we change the existing "recommendation" title to "new recommendation" as well?

theute: yes

theute: And Josejulio said so in his initial comment ;-)

bennetelli: Oh. Read it the wrong way. Thoguht we wouldn't touch the existing one and only change the new one after talking to Jerome :-)

PR #274, opened 2021-05-27
Spinder more notif fixes

spinder:
Spinder more notif fixes


bennetelli:
we don't need this anymore.

PR #287, opened 2021-06-03
Skip loading endpoint properties when there's no endpoint

gwenneg:


josejulio (approved):
lgtm

src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -347,6 +347,10 @@
     }
 
     public Uni<Void> loadProperties(List<Endpoint> endpoints) {
+        if (endpoints.isEmpty()) {
+            return Uni.createFrom().voidItem();
+        }
+
```
josejulio, line 353: Just curious, was there any issue related to this one? or it is for performance?

gwenneg: There's no issue, we can just skip a few stream operations execution for email endpoints.

PR #288, opened 2021-06-03
Process Kafka messages with a StatelessSession

gwenneg:
This PR replaces the Hibernate Reactive regular session with a `StatelessSession` for all database queries executed during a Kafka message processing.

> A stateless session:
>- doesn’t have a first-level cache (persistence context), nor does it interact with any second-level caches, and
>- doesn’t implement transactional write-behind or automatic dirty checking, so all operations are executed immediately when they’re explicitly called.

https://hibernate.org/reactive/documentation/1.0/reference/html_single/#_stateless_sessions

⚠️ Persistence operations never cascade to associated instances with a `StatelessSession` (it's not a problem in this PR).

Why do that? Because:

1. The regular session is supposed to be short-lived and discarded at the end of a single logical unit of work. In other words, it should be discarded each time a Kafka message has been processed. We couldn't do that automatically in `EndpointProcessor` because the session produced by Quarkus is `@RequestScoped` and SM Reactive Messaging uses a single request to process all messages received by a method annotated with `@Incoming`. That's why we had the `session.clear()` workaround in `EndpointProcessor` until now.

2. We have a lot of persistence issues on stage at the end of Kafka messages processing. I'm not 100% sure of the cause but I think that's related to the `session.clear()` workaround which may not be working very well when endpoints are deleted.

This may (I hope so) or may not fix the stage issues. Using a `StatelessSession` is a much better practice than clearing the session so this is an improvement anyway.

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
+++ b/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
@@ -14,12 +14,11 @@ import java.util.List;
 public class EmailAggregationResources {
 
     @Inject
-    Mutiny.Session session;
+    Mutiny.StatelessSession statelessSession;
 
     public Uni<Boolean> addEmailAggregation(EmailAggregation aggregation) {
-        return Uni.createFrom().item(aggregation)
-                .onItem().transformToUni(session::persist)
-                .call(session::flush)
+        aggregation.prePersist(); // This method must be called manually while using a StatelessSession.
+        return statelessSession.insert(aggregation)
                 .replaceWith(Boolean.TRUE)
                 .onFailure().recoverWithItem(Boolean.FALSE);
     }
@@ -27,7 +26,7 @@ public class EmailAggregationResources {
     public Uni<List<EmailAggregationKey>> getApplicationsWithPendingAggregation(LocalDateTime start, LocalDateTime end) {
         String query = "SELECT DISTINCT NEW com.redhat.cloud.notifications.models.EmailAggregationKey(ea.accountId, ea.bundleName, ea.applicationName) " +
                 "FROM EmailAggregation ea WHERE ea.created > :start AND ea.created <= :end";
-        return session.createQuery(query, EmailAggregationKey.class)
+        return statelessSession.createQuery(query, EmailAggregationKey.class)
                 .setParameter("start", start)
                 .setParameter("end", end)
                 .getResultList();
@@ -35,7 +34,7 @@ public class EmailAggregationResources {
 
     public Uni<List<EmailAggregation>> getEmailAggregation(EmailAggregationKey key, LocalDateTime start, LocalDateTime end) {
         String query = "FROM EmailAggregation WHERE accountId = :accountId AND bundleName = :bundleName AND applicationName = :applicationName AND created > :start AND created <= :end ORDER BY created";
-        return session.createQuery(query, EmailAggregation.class)
+        return statelessSession.createQuery(query, EmailAggregation.class)
                 .setParameter("accountId", key.getAccountId())
                 .setParameter("bundleName", key.getBundle())
                 .setParameter("applicationName", key.getApplication())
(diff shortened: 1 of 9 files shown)


src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
```diff
@@ -14,28 +14,27 @@
 public class EmailAggregationResources {
 
     @Inject
-    Mutiny.Session session;
+    Mutiny.StatelessSession statelessSession;
```
josejulio, line 17: IIRC, one of these methods is called in REST API calls - but maybe not - . Would that have a bad effect?

josejulio: After checking, all our resources are now using stateless sessions and not just the kafka ones?

src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -268,6 +271,16 @@
...
+    // TODO [BG Phase 2] Delete this method
+    public Uni<List<Endpoint>> getDefaultEndpointsStateless(String tenant) {
+        String query = "SELECT e FROM Endpoint e JOIN e.defaults d WHERE d.id.accountId = :accountId";
+
+        return statelessSession.createQuery(query, Endpoint.class)
+                .setParameter("accountId", tenant)
+                .getResultList()
+                .onItem().call(this::loadProperties);
+    }
+
```
josejulio, line 283: This method had to be duplicated, are there recommendations when mixing stateless and stateful sessions?

src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
gwenneg: I checked that earlier and there was no call from a REST API unless I missed something. I'm in the middle of a huge rebase with tons on conflicts so I can't really check again now.

src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
gwenneg: I didn't want to use the `StatelessSession` for any REST API call because the stateful session includes tools that are not available with the stateless one (L1 and L2 caches, persistence context, cascading...). So I duplicated the method to run the same query with the stateless session for messages processing and with the stateful session otherwise.

There's no issue with mixing both kinds of sessions, but the stateful one should be favored whenever possible because of its additional features.

josejulio: I've some worries here:
- We end up duplicating more methods (although it could be fixed with a base method that takes the session?) - In this particular case, this method will go away soon anyway.
- We use a stateless method in a rest call without noticing (or the other way around) - Is there some way to pick the right session depending on the context instead of specifying in each *Resources method call?

We could continue with this PR to test if it fixes our issue and give more though on how to follow up.

gwenneg: I understand your concerns, we can certainly think about a better way to pick the right session depending on the context. I have something in mind, I'll create another PR for that.

Regarding the duplication, I don't think that's a problem since there's only one case and it will be deleted as soon as behavior groups go live.

gwenneg: The duplication will actually also be removed by the following PR.

josejulio (approved):
lgtm

src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
gwenneg: I just confirmed `EmailSubscriptionTypeProcessor` is the only run time class calling `EmailAggregationResources` methods.

src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
gwenneg: What I had in mind doesn't actually work because `Session` and `StatelessSession` are completely different objects so there's no room for generics there.

We could isolate `StatelessSession` usages in a dedicated `*Resources` class that would only be used during Kafka messages processing and also add comments explaining when `StatelessSession` should or should not be used.

PR #296, opened 2021-06-09
NOTIF-192 Revert test commits

gwenneg:
We need to deploy `master` without the NOTIF-192 test commits.


theute:
I assume @bennetelli you are good with that ?

bennetelli:
Yes, all good.

PR #297, opened 2021-06-09
NOTIF-192 Filter successful health and metric access logs

bennetelli:


bennetelli:
I needed to force push to get rid of the blank (healt h) in my commit message.

bennetelli:
Please take a look at the RegEx and corresponding tests. I tried to test as much as possibile, but it could be that I misses a case.

If anyone has got a good idea of how to integrationtest it, I would be interested in it :-)

pilhuhn:
I think you could/should also test that it does not trigger when a health check does not respond with a 200 ok

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
```diff
@@ -5,17 +5,33 @@
...
+        filterAccessLogs();
+
         showVersionInfo();
     }
 
+    private void filterAccessLogs() {
+        java.util.logging.Logger accessLog = java.util.logging.Logger.getLogger("access_log");
+        accessLog.setFilter(record -> {
+            final String logMessage = record.getMessage().trim();
+            return !Pattern.matches(REGEX_FILTER, logMessage);
```
gwenneg, line 31: This will be called a lot, so the `Pattern` should be compiled and stored in a static variable for better performances.

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
```diff
@@ -5,17 +5,33 @@
...
     void init(@Observes StartupEvent ev) {
+        filterAccessLogs();
+
         showVersionInfo();
     }
 
+    private void filterAccessLogs() {
+        java.util.logging.Logger accessLog = java.util.logging.Logger.getLogger("access_log");
+        accessLog.setFilter(record -> {
+            final String logMessage = record.getMessage().trim();
```
gwenneg, line 30: Is `trim()` really necessary? If possible, it would be better for performances to remove it.

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
```diff
@@ -5,17 +5,33 @@
...
 
+import java.util.regex.Pattern;
+
 public class NotificationsApp {
+
     private static final String BUILD_COMMIT_ENV_NAME = "OPENSHIFT_BUILD_COMMIT";
     private static final String BUILD_REFERENCE_ENV_NAME = "OPENSHIFT_BUILD_REFERENCE";
     private static final String BUILD_NAME_ENV_NAME = "OPENSHIFT_BUILD_NAME";
 
+    public static final String REGEX_FILTER = ".*(/health|/metrics) HTTP/[0-9].[0-9]\\\" 200.*";
```
gwenneg, line 16: There's always something after `/health`: `/live` or `/ready` so the regex needs a minor change.

gwenneg (changes requested):

src/test/java/com/redhat/cloud/notifications/NotificationsAppTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import java.util.regex.Pattern;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class NotificationsAppTest {
+
+    final String regex = ".*(/health|/metrics) HTTP/[0-9].[0-9]\\\" 200.*";
```
gwenneg, line 11: The tests should use `NotificationsApp.REGEX_FILTER` rather than declaring a duplicate of the regex here.

gwenneg (changes requested):

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
bennetelli: Didn't have enough log messages for creating the regex. In my cases there wasn't anything between health/metrics and http. 
Will search for more message examples.

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
bennetelli: At the moment it is. I don't know how to change the regex to get rid of the trim() call.

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
bennetelli: I don't know how to change the RegEx so it also matches endpoints like /health/live

gwenneg: What makes it necessary? The current regex starts and ends with `.*` so it should match even with leading or trailing spaces.

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
gwenneg: What makes it necessary? The current regex starts and ends with `.*` so it should match even with leading or trailing spaces.

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
gwenneg: `.*(/health/(live|ready)|/metrics) HTTP/[0-9].[0-9]\\\" 200.*` should work.

gwenneg: `.*(/health/(live|ready)|/metrics) HTTP/[0-9].[0-9]\\\" 200.*` should work.

src/test/java/com/redhat/cloud/notifications/NotificationsAppTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.regex.Pattern;
+
+import static com.redhat.cloud.notifications.NotificationsApp.*;
+import static org.junit.jupiter.api.Assertions.*;
+
+class NotificationsAppTest {
+
+    final String logMessage = "127.0.0.1 - - 09/Jun/2021:16:07:07 +0200 \"GET /q/health HTTP/1.1\" 200 46 \"-\" \"Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:88.0) Gecko/20100101 Firefox/88.0\"\n";
```
pilhuhn, line 12: Cosmetic question: do we need the `final` here?

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
```diff
@@ -5,17 +5,36 @@
...
+import java.util.regex.Pattern;
+
 public class NotificationsApp {
+
     private static final String BUILD_COMMIT_ENV_NAME = "OPENSHIFT_BUILD_COMMIT";
     private static final String BUILD_REFERENCE_ENV_NAME = "OPENSHIFT_BUILD_REFERENCE";
     private static final String BUILD_NAME_ENV_NAME = "OPENSHIFT_BUILD_NAME";
 
+    public static final String FILTER_REGEX = ".*(/health|/metrics) HTTP/[0-9].[0-9]\\\" 200.*";
+    private final Pattern pattern = Pattern.compile(FILTER_REGEX);
```
gwenneg, line 18: Because `NotificationsApp` uses `@Observes StartupEvent`, Quarkus automatically transforms it to a `@Dependent` CDI bean which is eligible for garbage collection after the event has been fired.

Using an instance attribute instead of a `static` one will prevent the garbage collection so this would be better:
```suggestion
    private static final Pattern pattern = Pattern.compile(FILTER_REGEX);
```

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
pilhuhn: The `\n` at the end makes it necessary, as `.*` only matches until end of line, but not the `\n`.

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
pilhuhn: That assumes that no one just calls `/health` endpoint

gwenneg: Indeed, but does such a call exist apart from us making a manual test?

bennetelli: .*(/health|/health/live|health/ready|/metrics) HTTP/[0-9].[0-9]\" 200.*

works. With /live as well as with health/live and /health/ready.

The version above wouldn't work with logs like this one:
2021-06-10 07:30:59,164 INFO  [access_log] (vert.x-worker-thread-19) 10.130.16.1 - -noauth- 10/Jun/2021:07:30:59 +0000 "GET /health/ HTTP/1.1" 200 892 "-" "Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:88.0) Gecko/20100101 Firefox/88.0"

src/test/java/com/redhat/cloud/notifications/NotificationsAppTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Test
+    void shouldNotMachWhenInputEndsWithNewLine() {
```
pilhuhn, line 15: What is the semantic here?

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
gwenneg: Oh yes, I forgot about that. We could use the [DOTALL](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html#DOTALL) flag.

src/test/java/com/redhat/cloud/notifications/NotificationsAppTest.java
bennetelli: No we don't :-)

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
gwenneg: Yes, but I don't think this will ever be called by OpenShift, which is what we're trying to silence here.

OpenShift calls:
```
2021-06-09 09:46:37,520 INFO  [access_log] (vert.x-worker-thread-11) 127.0.0.1 - -noauth- 09/Jun/2021:09:46:37 +0200 "GET /health/live HTTP/1.1" 200 412 "-" "PostmanRuntime/7.28.0"
2021-06-09 09:46:42,568 INFO  [access_log] (vert.x-worker-thread-13) 127.0.0.1 - -noauth- 09/Jun/2021:09:46:42 +0200 "GET /health/ready HTTP/1.1" 200 525 "-" "PostmanRuntime/7.28.0"
2021-06-09 09:46:48,497 INFO  [access_log] (vert.x-eventloop-thread-15) 127.0.0.1 - -noauth- 09/Jun/2021:09:46:48 +0200 "GET /metrics HTTP/1.1" 200 40500 "-" "PostmanRuntime/7.28.0"
```

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
bennetelli: This hint is interesting. Thanks :) I didn't know that :)

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
bennetelli: I tried the DOTALL in Pattern.compile(FILTER_REGEX); already. It didn't work..

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
pilhuhn: Kube calls whatever we put into the deployment descriptor. No one prevents us from just calling `/health`. Allowing for both is more stable

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
bennetelli: Heiko fixed it by doing fancy stuff :-D thanks :)

src/test/java/com/redhat/cloud/notifications/NotificationsAppTest.java
bennetelli: You mean the test? I just wanted to test the status quo so I know when I am breaking the pattern somehow.

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
```diff
@@ -5,17 +5,36 @@
...
     private static final Logger LOG = Logger.getLogger(NotificationsApp.class);
 
+    // we do need a event as parameter here, otherwise the init method won't get called.
     void init(@Observes StartupEvent ev) {
+        filterAccessLogs();
+
         showVersionInfo();
     }
 
+    private void filterAccessLogs() {
```
pilhuhn, line 29: Perhaps rename to something like "init access log filter"   ?

pilhuhn (approved):

gwenneg (approved):
LGTM

src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
bennetelli: I don't have a strong opinion about the name. I am fine with both, so I will take your suggestion.

PR #302, opened 2021-06-10
NOTIF-255 Return 400 when request body JSON parsing fails

gwenneg:


src/main/java/com/redhat/cloud/notifications/routers/JsonParseExceptionMapper.java
```diff
@@ -0,0 +1,18 @@
...
+import javax.ws.rs.ext.Provider;
+
+import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
+
+@Provider
+public class JsonParseExceptionMapper implements ExceptionMapper<JsonParseException> {
+
+    @Override
+    public Response toResponse(JsonParseException exception) {
+        return Response.status(BAD_REQUEST).entity(exception.getMessage()).build();
```
gwenneg, line 16: @theute Returning the exception message will help the caller fix the request, but also show we're deserializing the body with Jackson. This could be considered a security weakness, but since we're talking about an open-source project, I'm not sure this is a problem here.

theute: And you need to be Red Hat employee to have access to this API

gwenneg: The mapper will be used for all APIs, internal and external. But I still think it's ok.

gwenneg:
Oops, this was NOTIF-225, not 255 😄

PR #303, opened 2021-06-11
NOTIF-226 Introduce maintenance mode

gwenneg:
⚠️ This does NOT include a maintenance mode for the Kafka messages processing. It only affects REST APIs calls.

Get the current status: `GET /api/notifications/v1.0/status`
Change the current status: `PUT /internal/status`

`GET`/`PUT` will respectively return/accept this data structure:

```
{
    "status": "UP",
    "start_time": "2021-06-11T13:09:31.213141",
    "end_time": "2021-06-11T13:09:31.213141"
}
```

The status can be set to:

- `UP`: all APIs are available
- `MAINTENANCE`: `/internal/*`, `/health/*`, `/metrics*` and `/api/notifications/v1.0/status` are available, all other APIs return 503

**❓ How does maintenance mode work?**

I introduced `MaintenanceModeRequestFilter` which performs two checks before a REST API call is processed:

- First it checks whether the API should be affected by maintenance mode or not.
- Then if the API should be affected, it retrieves the current status from the database. If that status is `MAINTENANCE`, then the API will return 503 with the `Maintenance in progress` message.

**❓ Why persist the status in the database?**

Because in-memory status management won't work if we deploy more than one pod. Maintenance mode has to be enabled on all pods at once.

**❓ Won't the filter execute too many database queries?**

No because the queries result will be cached for a certain amount of time defined in `application.properties` (60 seconds in this PR) . It means that one pod will query the database to retrieve the status at most once per minute. ⚠️ It also means that when we enable the maintenance mode, we should wait for at least one minute before considering the app is in maintenance for real.

### DIFF
--- a/src/main/java/com/redhat/cloud/notifications/models/CurrentStatus.java
+++ b/src/main/java/com/redhat/cloud/notifications/models/CurrentStatus.java
@@ -1,0 +1,76 @@
+package com.redhat.cloud.notifications.models;
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.redhat.cloud.notifications.db.converters.StatusConverter;
+import com.redhat.cloud.notifications.models.validation.MaintenanceWithTimeInterval;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+import javax.persistence.Column;
+import javax.persistence.Convert;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Table;
+import javax.validation.constraints.NotNull;
+import java.time.LocalDateTime;
+
+import static com.fasterxml.jackson.annotation.JsonFormat.Shape.STRING;
+import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
+import static com.fasterxml.jackson.databind.PropertyNamingStrategies.SnakeCaseStrategy;
+
+@Entity
+@Table(name = "status")
+@JsonNaming(SnakeCaseStrategy.class)
+@MaintenanceWithTimeInterval
+public class CurrentStatus {
+
+    /*
+     * We don't need this field in the Java code but Hibernate needs each @Entity to have a field annotated with @Id.
+     * This field has a real purpose in the database though: it is used to guarantee that the status table will never
+     * contain more than one row. This is done with a combination of PK and CHECK SQL constraints.
+     */
+    @Id
+    @JsonIgnore
+    private boolean preventMultipleRows = true;
+
+    @NotNull
+    @Convert(converter = StatusConverter.class)
+    @Column(name = "value")
+    public Status status;
+
+    @JsonInclude(NON_NULL)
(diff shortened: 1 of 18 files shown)


gwenneg:
Fun fact: caching a `Uni` returned by Hibernate Reactive with `quarkus-cache` doesn't work with Quarkus `1.13.7.Final`. I fixed that in Quarkus two months ago but the fix was only included in `2.0`.

I'm working on a temporary solution with `1.13.7.Final` here.

gwenneg:
I found a workaround for the `quarkus-cache` issue, that's fixed.

src/main/java/com/redhat/cloud/notifications/routers/StatusService.java
```diff
@@ -0,0 +1,39 @@
...
+@Path(INTERNAL + "/status")
+public class StatusService {
+
+    @Inject
+    StatusResources statusResources;
+
+    @GET
+    @Produces(APPLICATION_JSON)
+    @Tag(name = OApiService.PRIVATE)
+    public Uni<CurrentStatus> getCurrentStatus() {
```
josejulio, line 29: UI will need to access this endpoint. Currently `/internal` is not accessible unless the proxy is used (which is not the case).

gwenneg: Good point, thanks for noticing this.

gwenneg: That's fixed.

src/main/java/com/redhat/cloud/notifications/models/CurrentStatus.java
```diff
@@ -0,0 +1,71 @@
...
+    private boolean preventMultipleRows = true;
+
+    @NotNull
+    @Convert(converter = StatusConverter.class)
+    @Column(name = "value")
+    public Status status;
+
+    @JsonInclude(NON_NULL)
+    @JsonFormat(shape = STRING)
+    private LocalDateTime startTime;
```
josejulio, line 42: This is being returned as `start_time` but the openapi reports it as `startTime` .

src/main/java/com/redhat/cloud/notifications/models/CurrentStatus.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @JsonInclude(NON_NULL)
+    @JsonFormat(shape = STRING)
+    private LocalDateTime endTime;
```
josejulio, line 46: The same goes for `end_time` being reported as `endTime` in the openapi.

src/main/java/com/redhat/cloud/notifications/models/CurrentStatus.java
```diff
@@ -0,0 +1,71 @@
...
+     * contain more than one row. This is done with a combination of PK and CHECK SQL constraints.
+     */
+    @Id
+    @JsonIgnore
+    private boolean preventMultipleRows = true;
+
+    @NotNull
+    @Convert(converter = StatusConverter.class)
+    @Column(name = "value")
+    public Status status;
```
josejulio, line 38: Should we validate (somewhere) that when `status = MAINTENANCE` is set, startTime and endTime are required?

gwenneg:
@josejulio I fixed all issues.

josejulio (approved):
looks good to me

PR #307, opened 2021-06-17
NOTIF-233 Forbid multiple email actions for a single behavior group

gwenneg:


src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -157,19 +163,36 @@
...
     /*
-     * Returns Boolean.TRUE if the behavior group was found and successfully updated.
-     * Returns Boolean.FALSE if the behavior group was not found.
+     * Returns Status.OK if the behavior group was found and successfully updated.
+     * Returns Status.NOT_FOUND if the behavior group was not found.
+     * Returns Status.BAD_REQUEST if the behavior group contains more than one EMAIL_SUBSCRIPTION action.
      * If an exception other than NoResultException is thrown during the update, the DB transaction will be rolled back.
      */
-    public Uni<Boolean> updateBehaviorGroupActions(String accountId, UUID behaviorGroupId, List<UUID> endpointIds) {
+    public Uni<Status> updateBehaviorGroupActions(String accountId, UUID behaviorGroupId, List<UUID> endpointIds) {
```
josejulio, line 171: We could also check that there are not repeated `endpointId` in the list.

gwenneg: Yes. I am wondering what we should do in case of duplicate: return 400 or simply ignore the duplicates?

gwenneg:
I only solved conflicts with the latest force-push. Nothing new to review.

src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
gwenneg: I guess 400 would be better to be consistent with the other part of this PR.

josejulio: A 400 seems fine to me, we care about the order and is not immediately obvious which one we will be ignoring.

gwenneg: Good point, we can't ignore because of the order.

gwenneg: Duplicates are now forbidden.

josejulio (approved):
lgtm

PR #309, opened 2021-06-17
NOTIF-150 Disable reactive messaging health check

gwenneg:


src/main/resources/application.properties
```diff
@@ -7,14 +7,15 @@ quarkus.http.port=8085
 %test.quarkus.http.port=9085
 %test.quarkus.http.test-port=9085
 
+%prod.quarkus.health.extensions.enabled=false # TODO [BG Phase 2.1] Remove this line
```
josejulio, line 10: We would still show as `UP` in the live check?

gwenneg: I'm not sure what impact it will have on OpenShift. On my machine it removes all DOWN health checks. Only that one remains:

```
{
    "status": "UP",
    "checks": [
        {
            "name": "Notifications readiness check",
            "status": "UP",
            "data": {
                "reactive-db-check": true
            }
        }
    ]
}
```

josejulio (approved):
lgtm

PR #311, opened 2021-06-18
NOTIF-150 Fix behavior groups naming

gwenneg:


src/main/java/com/redhat/cloud/notifications/routers/BehaviorGroupMigrationService.java
```diff
@@ -49,7 +50,7 @@
      * This is used to dynamically create behavior groups names at persistence time.
      * See calling point for more details.
      */
-    private final Map<String, AtomicInteger> accountBehaviorGroupIndexes = new ConcurrentHashMap<>();
+    private final Map<BehaviorGroupIndexKey, AtomicInteger> accountBehaviorGroupIndexes = new ConcurrentHashMap<>();
```
gwenneg, line 53: During the migration, behavior groups are automatically named using the following pattern: `"Behavior group " + index`. The index comes from this map and we should use a different `AtomicInteger` instance from each account <ins>and each bundle</ins>. Right now, it's only isolated for accounts and all bundles share the same index, which leads to inconsistent names for accounts with several bundles. This PR fixes that.

src/main/java/com/redhat/cloud/notifications/routers/BehaviorGroupMigrationService.java
```diff
@@ -67,6 +68,7 @@
         return session.withTransaction(tx -> {
             LOGGER.debug("[BG MIGRATION] Start");
             return Uni.createFrom().item(new AccountBehaviorGroupsAggregator())
+                    .onItem().call(ignored -> session.createQuery("DELETE FROM BehaviorGroup").executeUpdate())
```
gwenneg, line 71: Unrelated to the issue mentioned above: we are now deleting all records from the `behavior_group` table before the migration starts. The delete is cascaded to the `behavior_group_action` and `event_type_behavior` tables which are both new. This does not trigger any cascade delete of existing production data.

pilhuhn (approved):
It makes sense to me what you describe, but obviously did not try the code.

gwenneg:
Deployed and tested on CI. Fixed confirmed ✔️ 

cc @josejulio

PR #313, opened 2021-06-18
NOTIF-150 Enable Kafka ingress channel

gwenneg:
Based on #312.

The Docker image resulting from this PR build should be deployed on prod:
- After the data migration is done.
- Before the maintenance mode is disabled.


josejulio (approved):
lgtm

PR #315, opened 2021-06-18
NOTIF-237 Load endpoint properties with stateless session

gwenneg:


src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -159,7 +159,7 @@
                 .setParameter("id", id)
                 .setParameter("accountId", tenant)
                 .getSingleResultOrNull()
-                .onItem().ifNotNull().transformToUni(this::loadProperties);
+                .onItem().ifNotNull().transformToUni(endpoint -> loadProperties(endpoint));
```
gwenneg, line 162: This may look unnecessary but the compiler wasn't happy without it:

```
reference to transformToUni is ambiguous
  both method <R>transformToUni(java.util.function.Function<? super T,io.smallrye.mutiny.Uni<? extends R>>) in io.smallrye.mutiny.groups.UniOnNotNull and method <R>transformToUni(java.util.function.BiConsumer<? super T,io.smallrye.mutiny.subscription.UniEmitter<? super R>>) in io.smallrye.mutiny.groups.UniOnNotNull match
```

src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -281,6 +284,17 @@
         return Uni.createFrom().voidItem();
     }
 
+    private <T extends EndpointProperties> Uni<List<T>> find(Class<T> typedEndpointClass, Set<UUID> endpointIds, boolean useStatelessSession) {
+        if (useStatelessSession) {
+            String query = "FROM " + typedEndpointClass.getSimpleName() + " WHERE id IN (:endpointIds)";
```
josejulio, line 289: I didn't notice before that we have to use HQL always in the stateless sessions.

josejulio (approved):

src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
gwenneg: Not always, there are a [few methods](https://hibernate.org/reactive/documentation/1.0/reference/html_single/#_stateless_sessions) to persist/retrieve entities without HQL. But unfortunately, there is no stateless version of `session.find(typedEndpointClass, endpointIds.toArray())` so this one has to be replaced with HQL.

PR #316, opened 2021-06-21
Fix test race condition

gwenneg:
This fixes a test race condition:

1. The test produces a Kafka message in the `ingress` channel and waits for a webhook call.
2. The message is processed and the webhook call is done.
3. Then in an unpredictable order, an history entry is persisted and the test tries to retrieve that entry. Sometimes, the retrieval happens before the entry was persisted.

cc @josejulio

### DIFF
--- a/src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java
+++ b/src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java
@@ -31,6 +31,7 @@ import org.mockserver.model.HttpRequest;
 import javax.enterprise.inject.Any;
 import javax.inject.Inject;
 import java.io.IOException;
+import java.time.Duration;
 import java.time.LocalDateTime;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


gwenneg:
Check `Hide whitespace changes` for an easier review.

src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java
```diff
@@ -493,45 +496,60 @@ private void checkEventTypeBehaviorGroups(Header identityHeader, String eventTyp
...
-                .get("/endpoints/{endpointId}/history")
-                .then()
-                .statusCode(200)
-                .contentType(JSON)
-                .extract().body().asString();
+    private void retry(Supplier<Boolean> checkEndpointHistoryResult) {
+        await()
+                .pollInterval(Duration.ofSeconds(1L))
+                .atMost(Duration.ofSeconds(5L))
+                .until(() -> checkEndpointHistoryResult.get());
```
josejulio, line 503: Would this fail after the 5 seconds without success?

gwenneg: Yes that's the idea

gwenneg: The test will try to retrieve the history each second and stop (with an exception) after 5 seconds.

josejulio (approved):

PR #317, opened 2021-06-21
Lower ApiResponseFilter log level to reduce noise

gwenneg:


src/main/java/com/redhat/cloud/notifications/models/filter/ApiResponseFilter.java
```diff
@@ -70,6 +70,6 @@ public void serializeAsField(Object pojo, JsonGenerator jgen, SerializerProvider
     }
 
     private void logFilterOut(String className, String fieldName) {
-        LOGGER.debugf("Filtering out %s#%s from a JSON response", className, fieldName);
+        LOGGER.tracef("Filtering out %s#%s from a JSON response", className, fieldName);
```
gwenneg, line 73: Now that behavior groups are enabled, this is making to much noise in logs.

josejulio (approved):
lgtm

PR #318, opened 2021-06-22
NOTIF-236 Lay some groundwork for email duplicates prevention

gwenneg:
This ticket will be divided into two PRs (reviews will be easier). This one changes the way actions are processed:

| Before | After |
| - | - |
| Each target endpoint is processed separately by the relevant processor. There is no way to detect multiple `EMAIL_SUBSCRIPTION` endpoints and prevent email duplicates. | Each processor receives all the endpoints it has to notify at once for a single action. Email duplicates prevention becomes possible. |

Next step: email duplicates prevention implementation.


josejulio (approved):
lgtm

PR #322, opened 2021-06-26
Adding internal UI

josejulio:
Screenshot_2021-06-25_19-11-11


gwenneg (changes requested):
Since this PR introduces frontend dependencies and code, the `dependabot.yml` and `codeql-analysis.yml` files will need to be updated.

backend/pom.xml
```diff
@@ -259,6 +259,54 @@
...
+                <groupId>com.github.eirslett</groupId>
+                <artifactId>frontend-maven-plugin</artifactId>
+                <version>1.11.2</version>
+                <configuration>
+                    <workingDirectory>${project.basedir}/src/main/webapp</workingDirectory>
+                    <installDirectory>target</installDirectory>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>install node and npm</id>
```
gwenneg, line 272:
```suggestion
                        <id>install node and yarn</id>
```

backend/pom.xml
```diff
@@ -259,6 +259,54 @@
...
+                        <goals>
+                            <goal>install-node-and-yarn</goal>
+                        </goals>
+                        <configuration>
+                            <nodeVersion>v14.17.1</nodeVersion>
+                            <yarnVersion>v1.22.10</yarnVersion>
```
gwenneg, line 278: Can't we use a more recent yarn version? This one is pretty old.

backend/src/main/java/com/redhat/cloud/notifications/routers/InternalService.java
```diff
@@ -39,8 +41,15 @@
     ApplicationResources appResources;
 
     @Inject
+
```
gwenneg, line 44:
```suggestion
```

backend/src/main/webapp/.gitignore
```diff
@@ -0,0 +1,23 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
gwenneg, line 1: Is that possible to move this to the root `.gitignore` file?

backend/src/main/webapp/public/index.html
```diff
@@ -0,0 +1,44 @@
...
+    <base href="%PUBLIC_URL%/">
+    <meta charset="utf-8" />
+    <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta name="theme-color" content="#000000" />
+    <meta
+      name="description"
+      content="Web site created using create-react-app"
+    />
+    <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
```
gwenneg, line 13: Does this file exists?

backend/src/main/webapp/public/robots.txt
```diff
@@ -0,0 +1,3 @@
+# https://www.robotstxt.org/robotstxt.html
```
gwenneg, line 1: Do we have robots internally? 😃

.gitignore
```diff
@@ -41,3 +41,6 @@ release.properties
 # Antlr
 *.tokens
 *.interp
+
+# webapp build
+backend/src/main/resources/META-INF
```
gwenneg, line 46: When a new Quarkus app is generated with https://code.quarkus.io/, the `/src/main/resources/META-INF/resources` contains an `index.html` file which can be displayed in a browser.

This raises two questions:
- Is that OK to add this path to `.gitignore`?
- What is difference between `/src/main/webapp` and `/src/main/resources/META-INF/resources`? Why use one or another?

backend/src/main/java/com/redhat/cloud/notifications/routers/InternalService.java
josejulio: I wonder how that got in there :)

backend/src/main/webapp/public/robots.txt
josejulio: Ah, I don't think so. Some of this code is autogenerated by [create-react-app](https://github.com/facebook/create-react-app)

.gitignore
josejulio: `/src/main/webapp` is holding the source code of the app, while `/src/main/resources/META-INF/resources` will hold the compiled and ready to use html/js/css code which gets automatically served by quarkus.

gwenneg: I see, thanks for the explanation!

backend/pom.xml
josejulio: I read that there is some migration effort into moving to yarn 2.0.0 (but didn't really look into the details).
I'll try to update to the last minor version and take some time to migrate across all our frontend apps to 2.0 if it makes sense and not too difficult.

backend/src/main/webapp/public/index.html
josejulio: Not anymore :D (i was a react app logo)

backend/src/main/webapp/.gitignore
josejulio: It's possible, but I'm not sure if it would make sense in the context of the whole app (maybe) as this is mostly for the UI. I'll check.

gwenneg: From my experience, it's easier to manage a single `.gitignore` file at root level in a multi-modules Maven project rather than multiple files. But I don't have a strong opinion on that, I can live with multiple files 😃

backend/pom.xml
josejulio: 1.22.10 is actually the latest stable version. The `berry`  (aka 2.0) is a bit older and the last published tag is `rc`.
I'm not really sure if it's stable or not :)

backend/src/main/webapp/.gitignore
josejulio: I think it's a little awkward to mix them together, but can be  done:

from:
```
# dependencies
/node_modules
/.pnp
.pnp.js

```

to (1):

```
## dependencies
/backend/src/main/webapp/node_modules
/backend/src/main/webapp/.pnp
/backend/src/main/webapp/**/.pnp.js
```

or (2)

```
node_modules
.pnp
.pnp.js
```

(1) is more exact on what we want, while (2) is less awkward, different to the main intent but similar. 
WDYT?

josejulio:
@gwenneg The only remaining task is to decide how we want to have the `.gitignore`. IMO it makes sense to have a separate one for the UI right now, to avoid mixing it with the base one.

gwenneg (approved):
I'm fine with separate `.gitignore` files if that makes more sense to you.

LGTM

PR #325, opened 2021-06-28
NOTIF-244 Return 400 on behavior group creation with blank display name

gwenneg:
@josejulio Maybe we'll need an additional frontend check for that one.


josejulio:
> @josejulio Maybe we'll need an additional frontend check for that one.

There is already in the code, I'll deploy to prod later.

josejulio (approved):

PR #326, opened 2021-06-28
Bump mockserver-client-java to 5.11.2

gwenneg:
While adding mockserver to policies-engine, I realized we're not actually stuck with `mockserver-client-java:5.5.4`.


src/test/java/com/redhat/cloud/notifications/TestLifecycleManager.java
```diff
@@ -24,9 +24,11 @@
 
 public class TestLifecycleManager implements QuarkusTestResourceLifecycleManager {
 
+    // Keep the version synced with pom.xml.
+    private static final DockerImageName MOCK_SERVER_DOCKER_IMAGE = DockerImageName.parse("jamesdbloom/mockserver").withTag("mockserver-5.11.2");
```
gwenneg, line 28: FYI `jamesdbloom/mockserver` is not new. We were already using that image before but with a much older version.

bennetelli (approved):

PR #327, opened 2021-06-29
Add new multi module project

bennetelli:


gwenneg (commented):
The `.github` folder should stay at the repository root level.

theute:
It doesn't seem to be a `git mv` but rather a new directory within the existing with duplicate content ?

notifications-backend/LICENSE.txt
```diff
@@ -0,0 +1,202 @@
+
+                                 Apache License
```
gwenneg, line 2: This file should stay at the repository root level.

notifications-backend/mvnw
```diff
@@ -0,0 +1,310 @@
+#!/bin/sh
```
gwenneg, line 1: This file should stay at the repository root level.

notifications-backend/mvnw.cmd
```diff
@@ -0,0 +1,182 @@
+@REM ----------------------------------------------------------------------------
```
gwenneg, line 1: This file should stay at the repository root level.

bennetelli:
@theute I didn't know that the git mv feature exist :(

theute:
Git is smart enough that git add / git rm is equivalent (and then also keeps history), but I don't see the rm in the changelog.
Now in this PR the whole history is lost since the old files are still in `src`

bennetelli:
Don't know what to do now. Didn't know there is a git mv and also didn't know that there is something that would not lose the history.
Does anyone know how to keep the history?

theute:
You need something like:
```
[theute@localhost notifications-backend]$ mkdir notifications-backend
[theute@localhost notifications-backend]$ git mv pom.xml src/ notifications-backend/
```
then add your pom.xml and update `build_deploy.sh` to use the new location of Dockerfile

gwenneg:
Losing the git history would be really bad. You should probably undo all changes and start over with `git mv`.

PR #328, opened 2021-06-29
NOTIF-190 Add multi module project

bennetelli:


theute:
`.rhcicd` should remain at the root
`build_deploy.sh` should remain at the root (later it should move inside .rhcicd but will require app-interfaces changes) with a fix in DockerFile
`deploy`  should remain at the root (later it should move inside .rhcicd but will require app-interfaces changes)

(I don't know about checkstyles, I guess it could move later)

theute:
Don't forget to fix: 
https://github.com/RedHatInsights/notifications-backend/blob/61a5cf13a3ab7473c81a442cf32bf6c3e819713e/build_deploy.sh#L23

bennetelli:
@gwenneg please take a look at it again. I had to merge it manually.

theute:
Sounds good to me (beside the conflict of course)

Note: Please don't push to this repo directly (even with a new branch like here) but always push to your own clone/branch before you create a PR.

notifications-backend/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd"
+         xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>notifications-backend</artifactId>
+
+    <parent>
+        <groupId>com.redhat.cloud.notifications</groupId>
+        <artifactId>notifications</artifactId>
```
gwenneg, line 11: Shouldn't this be
```suggestion
        <artifactId>notifications-backend</artifactId>
```
?

pom.xml
```diff
@@ -3,31 +3,120 @@
          xmlns="http://maven.apache.org/POM/4.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
     <modelVersion>4.0.0</modelVersion>
+
     <groupId>com.redhat.cloud.notifications</groupId>
-    <artifactId>notifications-backend</artifactId>
+    <artifactId>notifications</artifactId>
```
gwenneg, line 8: Quarkus style:
```suggestion
    <artifactId>notifications-parent</artifactId>
```

notifications-backend/pom.xml
bennetelli: I would say no as we want the root project to be renamed from notifications-backend to just notifications. So the parent should be notifications instead of notifications-backend as well.

pom.xml
bennetelli: I am also fine with notifications-parent :)

notifications-backend/pom.xml
gwenneg: I edited my previous comment too late. Everything's fine, my comment was wrong.

notifications-backend/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+
+    <properties>
+        <mockserver-client-java.version>5.5.4</mockserver-client-java.version> <!-- not the newest, but matches what is in org.testcontainers:mockserver -->
+        <version.cloud-commons>0.1.1</version.cloud-commons>
+        <openapi-parser.version>4.0.4</openapi-parser.version>
+        <maven-exec-plugin.version>3.0.0</maven-exec-plugin.version>
+    </properties>
```
gwenneg, line 20: These could be in the parent pom.

gwenneg:
Is there any reason why the parent pom still contains dependencies?

notifications-backend/pom.xml
bennetelli: Does it make sense? We only used in the child project. I would keep it "as private" as possible and move it to the parent pom when we use it in the other project as well. Would this be okay for you?

gwenneg: Well, I had bad multi-modules migrations experiences in the past because of versions declared at module level so I'm more in favor of centralized versions management at parent level now 😄

notifications-backend/pom.xml
```diff
@@ -0,0 +1,110 @@
...
+        <groupId>com.redhat.cloud.notifications</groupId>
+        <artifactId>notifications-parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+
+    <properties>
+        <mockserver-client-java.version>5.5.4</mockserver-client-java.version> <!-- not the newest, but matches what is in org.testcontainers:mockserver -->
+        <openapi-parser.version>4.0.4</openapi-parser.version>
+    </properties>
+
```
gwenneg: Do the plugins declared in the parent pom really work here? Declaring everything at parent level and nothing at child level seems a bit unusual, but I'm not a Maven expert so maybe that works.

notifications-backend/pom.xml
bennetelli: I just moved the properties in the parent pom so you can sleep well today :D

notifications-backend/pom.xml
bennetelli: I took a look at policies-engine and it's pretty much the same there. At least I didn't find plugins that are in the parent as well as in the child pom. My expectation it, that the plugins and the dependencies gets inherited to the child projects

pom.xml
```diff
@@ -3,31 +3,123 @@
          xmlns="http://maven.apache.org/POM/4.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
     <modelVersion>4.0.0</modelVersion>
+
     <groupId>com.redhat.cloud.notifications</groupId>
-    <artifactId>notifications-backend</artifactId>
+    <artifactId>notifications-parent</artifactId>
```
gwenneg, line 8: Is there any reason why this pom still contains dependencies?

notifications-backend/pom.xml
bennetelli: WTF...

pom.xml
bennetelli: Same reason as with the properties. They won't be used in the aggregator project.

notifications-backend/pom.xml
bennetelli: I just moved all the plugins in the child pom. I would like to finish this PR and test it later when the second projects gets added to it.

bennetelli: By putting it in the child pom it recognizes checkstyle issues.

pom.xml
gwenneg: Ok but we're talking about the parent pom here. The dependencies should only be in the child pom, right?

bennetelli: Yes, I moved them in the child pom as well. Sorry.. I am getting confused of it.

notifications-backend/pom.xml
```diff
@@ -0,0 +1,187 @@
...
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>${compiler-plugin.version}</version>
+                <configuration>
+                    <parameters>${maven.compiler.parameters}</parameters>
+                </configuration>
```
gwenneg, line 156: Do we need this? It wasn't in the old pom.

notifications-backend/pom.xml
```diff
@@ -0,0 +1,187 @@
...
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>${surefire-plugin.version}</version>
+                <configuration>
+                    <systemPropertyVariables>
+                        <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+                        <maven.home>${maven.home}</maven.home>
```
gwenneg, line 118: Same question here. Since we're using the wrapper, the Maven home doesn't seem necessary.

notifications-backend/pom.xml
bennetelli: Maybe not. I remember this needs to be declared when using openApi

pom.xml
gwenneg: I can still see many dependencies at parent level on the branch. I don't think the parent pom should contain any of them.

gwenneg:
The order of the dependencies of a Maven POM has an impact on the transitive dependencies resolution. If two direct dependencies bring the same transitive dependency into the classpath but with different versions, the first direct dependencies declared in our POM will win the conflict.

We should stick with the existing order on `master` and always put Quarkus dependencies before anything else.

gwenneg (approved):
LGTM

PR #329, opened 2021-06-29
NOTIF-140 Connect RBAC s2s API

josejulio:
This sets some bases to allow Endpoint emails to admins, an specific rbac group and/or ignore user preferences. 

Replaces #129


backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/AuthRequestFilter.java
```diff
@@ -0,0 +1,38 @@
...
+    @ConfigProperty(name = "rbac.service-to-service.secret", defaultValue = "addme")
+    String secret;
+
+    @ConfigProperty(name = "rbac.service-to-service.application", defaultValue = "notifications")
+    String application;
+
+    AuthRequestFilter() {
+        String tmp = System.getProperty("rbac.service-to-service.exceptional.auth.info");
+        if (tmp != null && !tmp.isEmpty()) {
+            authInfo = Base64.getEncoder().encodeToString(tmp.getBytes());
```
gwenneg, line 23:
```suggestion
            authInfo = new String(Base64.getEncoder().encode(tmp.getBytes(UTF_8)), UTF_8);
```
Requires: `import static java.nio.charset.StandardCharsets.UTF_8;`

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -0,0 +1,81 @@
...
+                    }
+                });
+    }
+
+    private Uni<List<User>> getWithPagination(Function<Integer, CompletionStage<Page<RbacUser>>> fetcher) {
+        return Multi.createBy().repeating()
+                .completionStage(
+                        AtomicInteger::new,
+                        state -> fetcher.apply(state.getAndIncrement())
+                )
```
gwenneg, line 68: Can we avoid using `CompletionStage`? Something like this should work (unverified code) :
```suggestion
    private Uni<List<User>> getWithPagination(Function<Integer, Uni<Page<RbacUser>>> fetcher) {
        return Multi.createBy().repeating()
                .uni(
                        AtomicInteger::new,
                        state -> fetcher.apply(state.getAndIncrement())
                )
```

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -0,0 +1,81 @@
...
+                .until(page -> page.getData().isEmpty())
```
gwenneg, line 69: This would concatenate all `Multi` retrieved after we're done repeating:
```suggestion
                .until(page -> page.getData().isEmpty())
                .onItem().disjoint()
```
Then the following `transform` would only process one `RbacUser` at a time and at the end of the stream you could simply `.collect().asList()`.

gwenneg (changes requested):
I'll review the tests tomorrow but here's a few comments already.

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacGroup.java
```diff
@@ -0,0 +1,89 @@
+package com.redhat.cloud.notifications.recipients.rbac;
+
+import java.time.LocalDateTime;
+import java.util.UUID;
+
+public class RbacGroup {
```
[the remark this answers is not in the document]
gwenneg: Same question for `RbacUser`.

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.stream.Collectors;
+
+@ApplicationScoped
+public class RbacRecipientUsersProvider implements RecipientUsersProvider {
+
+    @Inject
+    @RestClient
+    RbacServiceToService rbacServiceToService;
+
+    @ConfigProperty(name = "recipient_provider.rbac.elements_per_page")
```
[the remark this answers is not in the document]
josejulio: I have this default on the properties file, but I can move it here.

backend/src/main/java/com/redhat/cloud/notifications/recipients/RecipientUsersProvider.java
```diff
@@ -0,0 +1,14 @@
+package com.redhat.cloud.notifications.recipients;
+
+import io.smallrye.mutiny.Uni;
+
+import java.util.List;
+import java.util.UUID;
+
+public interface RecipientUsersProvider {
```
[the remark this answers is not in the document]
josejulio: This (and the User class) just serve as a minimum usage of what we need to support. @pilhuhn wanted to allow integrations with LDAP and other things.
In some WIP I use the base objects when processing the email action.

That said, if you think is a good idea, I can remove it and we can add it later.

backend/src/main/java/com/redhat/cloud/notifications/recipients/User.java
```diff
@@ -0,0 +1,81 @@
...
+            return true;
+        }
+
+        if (!(o instanceof User)) {
+            return false;
+        }
+
+        User user = (User) o;
+        return username.equals(user.username) && email.equals(user.email) && firstName.equals(user.firstName) && lastName.equals(user.lastName)
+                && isActive.equals(user.isActive) && isAdmin.equals(user.isAdmin);
```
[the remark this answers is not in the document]
josejulio: > Also, shouldn't `equals` only compare the `username` field?

I'm not sure, but i think it makes sense to only compare the `username`

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacGroup.java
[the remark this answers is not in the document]
josejulio: So, this is curious - I recall testing it with the server, but I think I need to give other go.

Not all the properties are camel case. From their [openapi](https://cloud.redhat.com/api/rbac/v1/openapi.json) this is the group:

```
{
      "name": "GroupA",
      "description": "A description of GroupA",
      "uuid": "57e60f90-8c0c-4bd1-87a0-2143759aae1c",
      "created": "2019-01-21T17:32:28Z",
      "modified": "2019-03-04T07:25:58Z",
      "principalCount": 0,
      "roleCount": 0,
      "system": false,
      "platform_default": false
    }
```

It looks like `platform_default` is the only snake case.

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacUser.java
```diff
@@ -0,0 +1,60 @@
+package com.redhat.cloud.notifications.recipients.rbac;
+
+public class RbacUser {
```
josejulio, line 3: RbacUser: looks like this needs the snake_case.

```
{
      "username": "smithj",
      "email": "smithj@mytechco.com",
      "first_name": "John",
      "last_name": "Smith",
      "is_active": true,
      "is_org_admin": true
    }
```

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
josejulio: I'm a bit confused here, I though we wanted to avoid using Multi whenever possible when we could use stream operations.
What I understood here (and below) is that I'm getting multiple lists (as I'm not using the `disjoint`) and in the end I'm building a big list with all of them. 

I suppose the readability is improved when using `disjoint`. 

So is better to use the `disjoint` here and process as a multi?

gwenneg: Ignore my comment, I misread one line of the stream, sorry about that 😄

backend/src/main/java/com/redhat/cloud/notifications/recipients/RecipientUsersProvider.java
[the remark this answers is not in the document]
pilhuhn: For our current purpose, it is unlikely that we'll do anything else than RBAC, so we could omit it and change later if needed.

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
gwenneg: My comment was based on the asumption that we were transforming a `Multi<Multi<RbacUser>>` but I was wrong so the comment is irrelevant.

We should avoid creating `Multi` when `Uni` or simple Java stream are enough to do the job indeed.

backend/src/main/java/com/redhat/cloud/notifications/recipients/User.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        User user = (User) o;
+        return Objects.equals(username, user.username);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(username, email, firstName, lastName, isActive, isAdmin);
+    }
```
gwenneg, line 79:
```suggestion
    public int hashCode() {
        return Objects.hash(username);
    }
```

backend/src/main/java/com/redhat/cloud/notifications/recipients/RecipientUsersProvider.java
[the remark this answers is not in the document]
gwenneg: Let's delete it then. It will be easy to extract the interface if we need it later.

backend/src/test/java/com/redhat/cloud/notifications/recipients/rbac/AuthRequestFilterTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+        try {
+            rbacAuthRequestFilter.filter(context);
+            System.out.println(context.getHeaders());
+            MultivaluedMap<String, Object> map = context.getHeaders();
+            Assertions.assertEquals("this-is-a-secret-token", context.getHeaderString("x-rh-rbac-psk"));
+            Assertions.assertEquals("My nice app", context.getHeaderString("x-rh-rbac-client-id"));
+            Assertions.assertNull(context.getHeaderString("Authorization"));
+        } catch (IOException ex) {
+            Assertions.fail(ex);
```
gwenneg, line 38: The exception throw would already cause a test failure, do we really need that `catch` block?

backend/src/test/java/com/redhat/cloud/notifications/recipients/rbac/AuthRequestFilterTest.java
```diff
@@ -0,0 +1,87 @@
...
+            Assertions.assertNull(context.getHeaderString("x-rh-rbac-client-id"));
+
+            // Account is removed
+            Assertions.assertNull(context.getHeaderString("x-rh-rbac-account"));
+            Assertions.assertEquals(
+                    "Basic " + Base64.getEncoder().encodeToString("myuser:p4ssw0rd".getBytes()),
+                    context.getHeaderString("Authorization")
+            );
+        } catch (IOException ex) {
+            Assertions.fail(ex);
```
gwenneg, line 67: Same question.

backend/src/test/java/com/redhat/cloud/notifications/recipients/rbac/AuthRequestFilterTest.java
```diff
@@ -0,0 +1,87 @@
...
+    public void testServiceToServiceHeaders() {
+        ClientRequestContext context = configureContext();
+
+        AuthRequestFilter rbacAuthRequestFilter = new AuthRequestFilter();
+        rbacAuthRequestFilter.application = "My nice app";
+        rbacAuthRequestFilter.secret = "this-is-a-secret-token";
+
+        try {
+            rbacAuthRequestFilter.filter(context);
+            System.out.println(context.getHeaders());
```
gwenneg, line 32: Should this be removed?

backend/src/test/java/com/redhat/cloud/notifications/recipients/rbac/AuthRequestFilterTest.java
```diff
@@ -0,0 +1,87 @@
...
+        AuthRequestFilter rbacAuthRequestFilter = new AuthRequestFilter();
+        rbacAuthRequestFilter.application = "My nice app";
+        rbacAuthRequestFilter.secret = "this-is-a-secret-token";
+
+        // Setting x-rh-rbac-account
+        context.getHeaders().putSingle("x-rh-rbac-account", "the-account-id");
+
+        try {
+            rbacAuthRequestFilter.filter(context);
+            System.out.println(context.getHeaders());
```
gwenneg, line 56: Same question.

backend/src/test/java/com/redhat/cloud/notifications/recipients/rbac/AuthRequestFilterTest.java
```diff
@@ -0,0 +1,87 @@
...
+            // Account is removed
+            Assertions.assertNull(context.getHeaderString("x-rh-rbac-account"));
+            Assertions.assertEquals(
+                    "Basic " + Base64.getEncoder().encodeToString("myuser:p4ssw0rd".getBytes()),
+                    context.getHeaderString("Authorization")
+            );
+        } catch (IOException ex) {
+            Assertions.fail(ex);
+        } finally {
+            System.setProperty("rbac.service-to-service.exceptional.auth.info", "");
```
gwenneg, line 69: Do we need that here in addition to the `@BeforeEach` method?

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Inject
+    @RestClient
+    RbacServiceToService rbacServiceToService;
+
+    @ConfigProperty(name = "recipient-provider.rbac.elements-per-page", defaultValue = "40")
+    Integer rbacElementsPerPage;
+
+    @Override
+    public Uni<List<User>> getUsers(String accountId, boolean adminsOnly) {
```
gwenneg, line 31: When will this be called? Since we occasionally have response time issues while calling RBAC, I am wondering if there is a chance this new service could slow us down even more.

Can we cache the methods results in this class?

backend/src/test/java/com/redhat/cloud/notifications/recipients/rbac/AuthRequestFilterTest.java
josejulio: Yes, missed it, thanks!

backend/src/test/java/com/redhat/cloud/notifications/recipients/rbac/AuthRequestFilterTest.java
josejulio: Same answer.

backend/src/test/java/com/redhat/cloud/notifications/recipients/rbac/AuthRequestFilterTest.java
josejulio: I don't think so, the `@BeforeEach` is fine

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
josejulio: This will be called before sending an email, so we can have things like:
- `SendEmail: to all registered users` -> Should not be called
- `SendEmail: to all admins` -> Needs to call getUsers(accountId, true) and then filters by the registered users
- `SendEmail: to all admins (override)` -> Calls getUsers(accountId, true) but does not filter by the registered users
- `SendEmail: to all users (override)` -> Calls getUsers(accountId, false) - No filter
- `SendEmail: to all users of group Foo` -> getGroupUsers(....) - Filters
- etc

Not sure about caching, this won't take part of any http call

backend/src/test/java/com/redhat/cloud/notifications/recipients/rbac/AuthRequestFilterTest.java
josejulio: `IOException` is a checked exception, so either catch and fail or mark the function with `throws IOException`.

Which one is better in the tests?

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
josejulio: Maybe we could cache the result after all.

backend/src/test/java/com/redhat/cloud/notifications/recipients/rbac/AuthRequestFilterTest.java
gwenneg: I'd throw 😃

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @ConfigProperty(name = "recipient-provider.rbac.elements-per-page", defaultValue = "40")
+    Integer rbacElementsPerPage;
+
+    @CacheResult(cacheName = "rbac-recipient-users-provider-get-users")
+    public Uni<List<User>> getUsers(String accountId, boolean adminsOnly) {
+        return getWithPagination(
+                page -> rbacServiceToService
+                        .getUsers(accountId, adminsOnly, page * rbacElementsPerPage, rbacElementsPerPage)
+        ).memoize().indefinitely();
```
gwenneg, line 35: Can you add a TODO saying that `.memoize().indefinitely()` should be removed after the Quarkus 2.0 bump?

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -0,0 +1,79 @@
...
+                        // getGroupUsers doesn't have an adminOnly param.
+                        .onItem().transform(users -> {
+                            if (adminOnly) {
+                                return users.stream().filter(User::isAdmin).collect(Collectors.toList());
+                            }
+
+                            return users;
+                        });
+                    }
+                }).memoize().indefinitely();
```
gwenneg, line 58: Same here.

backend/src/test/java/com/redhat/cloud/notifications/recipients/rbac/AuthRequestFilterTest.java
```diff
@@ -0,0 +1,75 @@
...
+        context.getHeaders().putSingle("x-rh-rbac-account", "the-account-id");
+
+        rbacAuthRequestFilter.filter(context);
+        Assertions.assertNull(context.getHeaderString("x-rh-rbac-psk"));
+        Assertions.assertNull(context.getHeaderString("x-rh-rbac-client-id"));
+
+        // Account is removed
+        Assertions.assertNull(context.getHeaderString("x-rh-rbac-account"));
+        Assertions.assertEquals(
+                "Basic " + Base64.getEncoder().encodeToString("myuser:p4ssw0rd".getBytes()),
```
gwenneg, line 56: This is harmless here but it could be copied someday into runtime code and become a problem:
```suggestion
                "Basic " + new String(Base64.getEncoder().encode("myuser:p4ssw0rd".getBytes(UTF_8)), UTF_8),
```
Requires: `import static java.nio.charset.StandardCharsets.UTF_8;`

backend/src/test/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProviderTest.java
```diff
@@ -0,0 +1,225 @@
...
+    /*
+     * This would normally happen after a certain duration fixed in application.properties with the
+     * quarkus.cache.caffeine.rbac-recipient-users-provider-get-group-users.expire-after-write
+     * and
+     * quarkus.cache.caffeine.rbac-recipient-users-provider-get-users.expire-after-write key.
+     */
+    @CacheInvalidateAll.List(value = {
+        @CacheInvalidateAll(cacheName = "rbac-recipient-users-provider-get-users"),
+        @CacheInvalidateAll(cacheName = "rbac-recipient-users-provider-get-group-users")
+    })
```
gwenneg, line 183: It is repeatable:
```suggestion
    @CacheInvalidateAll(cacheName = "rbac-recipient-users-provider-get-users")
    @CacheInvalidateAll(cacheName = "rbac-recipient-users-provider-get-group-users")
```

backend/src/test/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProviderTest.java
```diff
@@ -0,0 +1,225 @@
...
+
+        users = rbacRecipientUsersProvider.getGroupUsers(accountId, false, group.getUuid()).await().indefinitely();
+        Assertions.assertEquals(updatedSize, users.size());
+    }
+
+    private void mockGetUsers(int elements, boolean adminsOnly) {
+        MockedUserAnswer answer = new MockedUserAnswer(elements, adminsOnly);
+        Mockito.when(rbacServiceToService.getUsers(
+                Mockito.eq(accountId),
+                Mockito.anyBoolean(),
```
gwenneg, line 141:
```suggestion
                Mockito.eq(adminsOnly),
```

backend/src/test/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProviderTest.java
```diff
@@ -0,0 +1,225 @@
...
+    @Test
+    public void getAllUsers() {
+        int elements = 95;
+        mockGetUsers(elements, false);
+
+        List<User> users = rbacRecipientUsersProvider.getUsers(accountId, false).await().indefinitely();
+        Assertions.assertEquals(elements, users.size());
+        for (int i = 0; i < elements; ++i) {
+            Assertions.assertEquals(String.format("username-%d", i), users.get(i).getUsername());
+        }
```
gwenneg, line 47: If you move this to `getAllUsersCache()`, the current test can be deleted.

backend/src/test/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProviderTest.java
```diff
@@ -0,0 +1,225 @@
...
+
+        List<User> users = rbacRecipientUsersProvider.getGroupUsers(accountId, false, defaultGroup.getUuid()).await().indefinitely();
+        Assertions.assertEquals(elements, users.size());
+        for (int i = 0; i < elements; ++i) {
+            Assertions.assertEquals(String.format("username-%d", i), users.get(i).getUsername());
+        }
+    }
+
+    @Test
+    public void getAllUsersFromGroup() {
```
gwenneg, line 69: `getAllGroupsCache()` already does everything that is in this test. This one can be removed.

gwenneg (approved):
LGTM

PR #331, opened 2021-06-30
NOTIF-190 Add aggregator

bennetelli:


theute:
See comment about branches here: https://github.com/RedHatInsights/notifications-backend/pull/328#issuecomment-870557376 ;-)
Also I though we were *not* adding submodules ?

bennetelli:
"Note: In the future please don't push to this repo directly (even with a new branch like here) but always push to your own clone/branch before you create a PR."

Does it mean I should fork the repository before?

bennetelli:
Locally there was no difference. I didn't add the submodule explicitly. This seems to be the default.

theute:
Yes :
1) otherwise you leave branches behind you: https://github.com/RedHatInsights/notifications-backend/branches https://github.com/RedHatInsights/policies-engine/branches
2) as a rule of thumb you should never push to this repo directly

theute:
Did you delete the .git in the aggregator directory before you pushed ?

theute:
It's definitely wrong, the icon also shows a submodule:
image
And then it can't be browsed:
https://github.com/RedHatInsights/notifications-backend/tree/add_aggregator

bennetelli:
> Did you delete the .git in the aggregator directory before you pushed ?

I would say yes.

PR #333, opened 2021-06-30
NOTIF-190 Add aggregator

bennetelli:


theute:
About the test failure @gwenneg is aware of this randomly failing issue, this is not related to this PR

aggregator/.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
+# To get started with Dependabot version updates, you'll need to specify which
```
gwenneg, line 1: The `aggregator` folder should not contain `.github` or `.mvn` subfolders.

aggregator/pom.xml
```diff
@@ -0,0 +1,170 @@
+<?xml version="1.0"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>aggregator</artifactId>
```
gwenneg, line 6:
```suggestion
    <artifactId>notifications-aggregator</artifactId>
```

gwenneg (changes requested):
A few comments about the project structure (I didn't review the code):

aggregator/src/main/docker/Dockerfile.jvm
```diff
@@ -0,0 +1,54 @@
+####
```
gwenneg, line 1: Do we need so many Docker files? Also, can we use the same Dockerfile pattern in all modules?

aggregator/.github/dependabot.yml
bennetelli: I know :-) it wasn't ready for review. Had to open a new PR to check if the submodule stuff works.

gwenneg: Oh sorry, I thought the PR was ready since it's not a draft 😃

bennetelli:
You are right. Now it's a draft :)

aggregator/src/main/java/com/redhat/cloud/notifications/db/EndpointEmailSubscriptionResources.java
```diff
@@ -0,0 +1,59 @@
...
+                .setParameter("userId", username)
+                .setParameter("bundleName", bundleName)
+                .setParameter("applicationName", applicationName)
+                .setParameter("subscriptionType", subscriptionType)
+                .executeUpdate();
+        session.flush();
+        return Boolean.TRUE;
+    }
+
+    @Transactional
```
gwenneg, line 48:
```suggestion
```

aggregator/src/main/java/com/redhat/cloud/notifications/db/EndpointEmailSubscriptionResources.java
```diff
@@ -0,0 +1,59 @@
...
+import javax.transaction.Transactional;
+
+@ApplicationScoped
+public class EndpointEmailSubscriptionResources {
+
+    @Inject
+    Session session;
+
+    @Transactional
+    public Boolean subscribe(String accountNumber, String username, String bundleName, String applicationName, EmailSubscriptionType subscriptionType) {
```
gwenneg, line 16: This method is only used by tests. It should not be part of the `main` sources.

aggregator/src/main/java/com/redhat/cloud/notifications/db/EndpointEmailSubscriptionResources.java
```diff
@@ -0,0 +1,59 @@
...
+                .setParameter("userId", username)
+                .setParameter("bundleName", bundleName)
+                .setParameter("applicationName", applicationName)
+                .setParameter("subscriptionType", subscriptionType.name())
+                .executeUpdate();
+        return Boolean.TRUE;
+    }
+
+    @Transactional
+    public Boolean unsubscribe(String accountNumber, String username, String bundleName, String applicationName, EmailSubscriptionType subscriptionType) {
```
gwenneg, line 33: This method is only used by tests. It should not be part of the `main` sources.

aggregator/src/main/java/com/redhat/cloud/notifications/db/EndpointEmailSubscriptionResources.java
```diff
@@ -0,0 +1,59 @@
...
+                .setParameter("subscriptionType", subscriptionType)
+                .executeUpdate();
+        session.flush();
+        return Boolean.TRUE;
+    }
+
+    @Transactional
+    public Long getEmailSubscribersCount(String accountNumber, String bundleName, String applicationName, EmailSubscriptionType subscriptionType) {
+        String query = "SELECT COUNT(id.userId) FROM EmailSubscription WHERE id.accountId = :accountId " +
+                "AND application.bundle.name = :bundleName AND application.name = :applicationName AND id.subscriptionType = :subscriptionType";
```
gwenneg, line 51: This query should be transformed into a native query. We wouldn't need the JPA model anymore to execute it.

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
```diff
@@ -0,0 +1,61 @@
...
+public class EmailAggregationResources {
+
+    @Inject
+    Session session;
+
+    @Transactional
+    public Boolean addEmailAggregation(EmailAggregation aggregation) {
+        try {
+            session.persist(aggregation);
+            session.flush();
```
gwenneg, line 22: This shouldn't be necessary, the enclosing transaction should flush automatically.

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
```diff
@@ -0,0 +1,61 @@
...
+    @Transactional
+    public Integer purgeOldAggregation(EmailAggregationKey key, LocalDateTime lastUsedTime) {
+        String query = "DELETE FROM EmailAggregation WHERE accountId = :accountId AND bundleName = :bundleName AND applicationName = :applicationName AND created <= :created";
+        final int result = session.createQuery(query)
+                .setParameter("accountId", key.getAccountId())
+                .setParameter("bundleName", key.getBundle())
+                .setParameter("applicationName", key.getApplication())
+                .setParameter("created", lastUsedTime)
+                .executeUpdate();
+        session.flush();
```
gwenneg, line 58: This shouldn't be necessary, the enclosing transaction should flush automatically.

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
```diff
@@ -0,0 +1,61 @@
...
+    @Inject
+    Session session;
+
+    @Transactional
+    public Boolean addEmailAggregation(EmailAggregation aggregation) {
+        try {
+            session.persist(aggregation);
+            session.flush();
+            return Boolean.TRUE;
+        } catch (Exception e) {
```
gwenneg, line 24: This exception should be logged.

gwenneg (changes requested):
This is a big PR so I'll split the review into several smaller reviews.

I added a comment about a native query (see below). If you apply it, I think most of the classes from the `com.redhat.cloud.notifications.models` package can be removed from this PR.

bennetelli:
> This is a big PR so I'll split the review into several smaller reviews.

Yea, it's a bit of a bad situation.. I didn't see a chance to have it smaller other than using lots of commits. Sorry :(

aggregator/.gitignore
```diff
@@ -0,0 +1,40 @@
+#Maven
```
gwenneg, line 1: Can we have only one `.gitignore` file at the repo root level for all Java modules?

aggregator/README.md
```diff
@@ -0,0 +1,54 @@
+# notifications-aggregator project
```
gwenneg, line 1: Do we want to keep that generated readme?

aggregator/src/main/resources/application.properties
```diff
@@ -0,0 +1,56 @@
...
+quarkus.log.sentry.in-app-packages=*
+quarkus.log.sentry.dsn=FILL_ME
+
+%test.quarkus.scheduler.enabled=false
+
+# Policies-notification migration
+policynotifications/mp-rest/url=http://localhost:8090
+policynotifications/mp-rest/connectTimeout=2000
+policynotifications/mp-rest/readTimeout=2000
+
```
gwenneg, line 48:
```suggestion
```

aggregator/src/main/resources/application.properties
```diff
@@ -0,0 +1,56 @@
...
+%test.quarkus.http.access-log.category=info
+
+# RBAC server
+rbac.enabled=true
+#rbac/mp-rest/url=http://ci.foo.redhat.com:1337
+rbac/mp-rest/url=https://ci.cloud.redhat.com
+rbac/mp-rest/connectTimeout=2000
+rbac/mp-rest/readTimeout=2000
+# Duration rbac entries are kept in cache
+quarkus.cache.caffeine.rbac-cache.expire-after-write=PT120s
```
gwenneg, line 30:
```suggestion
```

aggregator/src/main/resources/application.properties
```diff
@@ -0,0 +1,56 @@
...
+
+# Change port for tests to avoid messing with local Kafka instance
+%test.quarkus.http.port=9085
+%test.quarkus.http.test-port=9085
+
+# configure your datasource
+quarkus.datasource.db-kind=postgresql
+quarkus.datasource.username=postgres
+quarkus.datasource.password=postgres
+#quarkus.datasource.jdbc=false
```
gwenneg, line 11:
```suggestion
```

aggregator/src/main/resources/application.properties
```diff
@@ -0,0 +1,56 @@
+quarkus.http.port=8085
+
+# Change port for tests to avoid messing with local Kafka instance
+%test.quarkus.http.port=9085
+%test.quarkus.http.test-port=9085
```
gwenneg, line 5: The ports should probably be different from the backend.

aggregator/src/main/resources/application.properties
```diff
@@ -0,0 +1,56 @@
...
+# Quarkus since 1.11 redirects non-apps to /q/. We need to prevent this
+quarkus.http.redirect-to-non-application-root-path=false
+quarkus.http.non-application-root-path=/
+
+# Sentry logging. Off by default, enabled on OpenShift
+# See https://quarkus.io/guides/logging-sentry#in-app-packages
+quarkus.log.sentry=false
+quarkus.log.sentry.in-app-packages=*
+quarkus.log.sentry.dsn=FILL_ME
+
```
gwenneg, line 41: I don't see Sentry in the dependencies but I think we will need it.

aggregator/src/main/java/com/redhat/cloud/notifications/db/EndpointEmailSubscriptionResources.java
bennetelli: Good catch. I found this as well but didn't want to touch it as it's copied from notifications-backend and I thought we would get rid of it later when removing the duplicates.

I will move it for now.

aggregator/README.md
bennetelli: No :)

aggregator/src/main/resources/application.properties
bennetelli: Do you think this is important? I would say it's not. I mean if it would be important, then we would need to know every port of every project to be able to deploy the applications on OpenShift.

gwenneg: It's not important for OpenShift deployment, but it is for local testing. We need to be able to run all apps (backend, aggregator...) at once on the same machine.

aggregator/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jdbc-postgresql</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>io.quarkiverse.logging.cloudwatch</groupId>
+            <artifactId>quarkus-logging-cloudwatch</artifactId>
+            <version>3.0.0</version>
```
josejulio, line 62: Could we share this version (in a `notification-parent` variable) with the `notifcations-backend` ?

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private final Logger log = Logger.getLogger(this.getClass().getName());
+
+    @Inject
+    Session session;
+
+    @Transactional
+    public Boolean addEmailAggregation(EmailAggregation aggregation) {
+        try {
+            session.persist(aggregation);
```
josejulio, line 24: Just to clarify, this is the aggregated data?

bennetelli: Yes. 

Not sure if important, but this is exactly the same as in notfications-backend.

josejulio:
I'm concerned that we have lots of duplicated code backend / aggregator.
Could we split into submodules or if not possible at least use symlinks?

bennetelli:
> I'm concerned that we have lots of duplicated code backend / aggregator.
> Could we split into submodules or if not possible at least use symlinks?

Yes of course we have duplicated code at the moment. This PR is not the final solution. I don't want to commit 10k lines of code in a single PR :-)
notifications-backend still contains all the email aggregation features it had before. This code will be deleted.

That's the plan: put it in a multi module, check if it's working (we are currently here), if it does -> remove what's not necessary and build an additional module which contains the code needed in both projects.

bennetelli:
@gwenneg the current build will fail. There is a sequence table missing. This is probably related to the commit before which deleted flyway. Do you have an idea how to fix it without adding flyway again? I've never worked with it before.

gwenneg:
> I'm concerned that we have lots of duplicated code backend / aggregator.
> Could we split into submodules or if not possible at least use symlinks?

We won't have that much duplicated code after [the model is gone](https://github.com/RedHatInsights/notifications-backend/pull/333#pullrequestreview-697208983) from aggregator.

gwenneg:
> @gwenneg the current build will fail. There is a sequence table missing. This is probably related to the commit before which deleted flyway. Do you have an idea how to fix it without adding flyway again? I've never worked with it before.

I think Flyway will be required in this project, but only during the tests (unlike backend where it's part of `main`). The schema needed for the tests should contain as few tables as possible.

bennetelli:
> > @gwenneg the current build will fail. There is a sequence table missing. This is probably related to the commit before which deleted flyway. Do you have an idea how to fix it without adding flyway again? I've never worked with it before.
> 
> I think Flyway will be required in this project, but only during the tests (unlike backend where it's part of `main`). The schema needed for the tests should contain as few tables as possible.

That's why I had it in the aggregator before. It didn't work without it. But.. just for my understanding. Isn't flyway for relational database migrations? So I would expect it should even work without it. But I don't know what to do to achieve this.

bennetelli:
Interesting.. methods annotated with transactional which have a private modifier result in failing tests and strange error messages.

gwenneg:
I've seen issues in Quarkus with interceptors on private methods in the past. Caching annotations from `quarkus-cache` are forbidden on private methods for that reason and that's mentioned in the extension doc. Maybe `@Transactional` is affected as well but I didn't see anything about this in the Quarkus doc.

gwenneg:
> That's why I had it in the aggregator before. It didn't work without it. But.. just for my understanding. Isn't flyway for relational database migrations? So I would expect it should even work without it. But I don't know what to do to achieve this.

Flyway is a migration tool but it can be used to simply create and populate a schema for tests. You could add in the `test` sources folder something similar to [FlywayWorkaround](https://github.com/gwenneg/notifications-backend/blob/master/backend/src/main/java/com/redhat/cloud/notifications/db/FlywayWorkaround.java) alongside with a SQL script that would only create the tables this module needs (two tables).

bennetelli:
> > I'm concerned that we have lots of duplicated code backend / aggregator.
> > Could we split into submodules or if not possible at least use symlinks?
> 
> We won't have that much duplicated code after [the model is gone](https://github.com/RedHatInsights/notifications-backend/pull/333#pullrequestreview-697208983) from aggregator.
> 
> And I'm now pretty sure we don't have to share anything between backend and aggregator in terms of model.

Yes, probably. That's what I said multiple times already.. :-D "let's keep the duplicates, remove the duplicated code one by one and see what will be left in the end"

aggregator/src/main/java/com/redhat/cloud/notifications/db/EndpointEmailSubscriptionResources.java
bennetelli: I am too dumb to create this query.

aggregator/src/main/docker/Dockerfile.jvm
bennetelli: I will improve it in another PR :)

josejulio (approved):
All changes are isolated in a separate module - Lets continue iterating in a separate PR.

PR #355, opened 2021-07-01
NOTIF-153 Adds options to email endpoints

josejulio:
~Based on #329~
Based on #362


backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -130,143 +137,174 @@ public void addEmail(Email email) {
...
-                });
+                // So we are about to send the email
+                // but since we do de-duplication we have multiple email endpoints but only one request.
+                // Should we
+                // a) Save the NotificationHistory on the related endpoints (refactoring doHttpRequest method)
+                // b) Skip saving the NotificationHistory (refactoring doHttpRequest)
+                // c) Have a "default" email endpoint whose only purpose is to save all the NotificationHistory
+                // d) Pick any?
+
+                return webhookSender.doHttpRequest(item, bopRequest, payload);
```
josejulio, line 306: About to send the email, at this point all the email endpoints have been processed and fetched all the users according to each endpoint. De-duplication of users is also done.

Question is, where do we store the `NotificationHistory` ?  It's linked to an `Endpoint`, but we could have multiple Email endpoints here (because of behavior groups).

Some options that I have considered are:

a. Save the NotificationHistory on each of the related endpoints (refactoring doHttpRequest method)
b. Skip saving the NotificationHistory (refactoring doHttpRequest)
c. Have a "default" email endpoint whose only purpose is to save all the NotificationHistory
d. Pick any/first endpoint?

theute: I'm not familiar enough with all the concepts here but I'll try to comment...
I assume NotificationHistory is endpoint history which is what we use to show the last 5 attempts for webhooks ?

I saw a) or c) is your preference, but I guess b) would have no impact on current features (AFAIU we don't use those ?)

I don't think I understand c)

Thinking further, I think the [notifications log](https://issues.redhat.com/browse/NOTIF-209) will be more important API, in this API to query all the events an account received, it could be useful to list all triggered endpoints (and their success or failure), in that case I guess we will just store the unique email endpoint that was triggered. I'm not sure if we'll to duplicate the info or not with notification history

gwenneg: The history details for an email endpoint contain the list of recipients. Knowing that, it would be a bit weird to do `a` if that includes saving the history details.

Let's consider an event type is triggered and linked with two email endpoints. `endpoint1` has `[recipient1, recipient2]` and `endpoint2` has `[recipient2, recipient3]`. Does it make sense to put in the `endpoint1` _and_ `endpoint2` history details that `[recipient1, recipient2, recipient3]` were notified?

I'm not sure I understand `c` as well because now that email endpoints can have different properties, they can no longer be considered as a single endpoint.

I didn't see anything in the notifications log epic about the history details being used for email endpoints. That would mean we don't really need to know the list of notified recipients for an email notification. If that is correct, then maybe a modified version of `a` could be done: save the history **without the details** on all email endpoints involved during the notification process.

josejulio: > I saw a) or c) is your preference, but I guess b) would have no impact on current features (AFAIU we don't use those ?)

Yes, b would have no impact the UI and would be the easier to implement. We could update (if we require to) later to store history of emails.

> I don't think I understand c)

`c` is like having a default place to put all the email history (In practice it will be an Endpoint that we might implement anyway for [NOTIF-242](https://issues.redhat.com/browse/NOTIF-242)). This could serve as a log that `policies-triggrered` caused a mail to be sent to user `username-123`.


> Thinking further, I think the [notifications log](https://issues.redhat.com/browse/NOTIF-209) will be more important API, in this API to query all the events an account received, it could be useful to list all triggered endpoints (and their success or failure), in that case I guess we will just store the unique email endpoint that was triggered. I'm not sure if we'll to duplicate the info or not with notification history

Yes, that's more general (an event triggered) and this is more what actions happened and their outcome - I could probably choose one that is easy to implement and is something we could update later.

josejulio: > Let's consider an event type is triggered and linked with two email endpoints. `endpoint1` has `[recipient1, recipient2]` and `endpoint2` has `[recipient2, recipient3]`. Does it make sense to put in the `endpoint1` _and_ `endpoint2` history details that `[recipient1, recipient2, recipient3]` were notified?

As part of `a` i was thinking to go into the details and store only the triggered recipients for each endpoint, right now we only send one email, but as part of [NOTIF-200](https://issues.redhat.com/browse/NOTIF-200) we want to send one email per recipient, so it would make sense.


> I'm not sure I understand `c` as well because now that email endpoints can have different properties, they can no longer be considered as a single endpoint.

This was bothering me too, but, the question is if we really care what specific configuration triggered the email or if we only care that an email was sent. 

This is different than a `WebhookEndpoint` because different webhooks hit different urls. 
In the case o an `EmailEndpoint`, all emails are about an sending mail to someone.

> I didn't see anything in the notifications log epic about the history details being used for email endpoints. That would mean we don't really need to know the list of notified recipients for an email notification. If that is correct, then maybe a modified version of a could be done: save the history without the details on all email endpoints involved during the notification process.

That actually seems like `c` :)

josejulio:
Replaced by #396

PR #362, opened 2021-07-02
NOTIF-242 Prevent from creating/updating Email_Subscription endpoints

josejulio:
This prevents the creation of multiple Endpoint with type `EmailSubscription` and adds an API to get the `Endpoint` (with `EmailSubscription` type) represented by the `EmailSubscriptionProperties`. 
Currently the properties is an empty object, but going soon this will change to be able to configure it. 

This PR also has a migration function that is in charge of collapsing the multiple `EmailSubscription Endpoint's` into a single one (per account) and update the `BehaviorGroupAction` (The link between a `BehaviorGroup` and an `Endpoint`) to use the account's default - Removing any duplicate in the process so that each behavior group can at max one `EmailSubscription Endpoint`.

To merge with https://github.com/RedHatInsights/notifications-frontend/pull/141

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
@@ -103,13 +103,13 @@ public class EndpointResources {
     }
 
     public Uni<Endpoint> getOrCreateEmailSubscriptionEndpoint(String accountId, EmailSubscriptionProperties properties) {
-        return getEndpointsPerType(accountId, EndpointType.EMAIL_SUBSCRIPTION, true, null)
+        return getEndpointsPerType(accountId, EndpointType.EMAIL_SUBSCRIPTION, null, null)
                 .onItem().call(this::loadProperties)
                 .onItem().transformToUni(emailEndpoints -> {
                     Optional<Endpoint> endpointOptional = emailEndpoints
                             .stream()
-                            // Todo: This should be changed once we store the properties - (properties = null right now)
-                            // .filter(endpoint -> endpoint.getProperties().equals(properties))
+                            // Todo: This should be changed once we store the properties - (properties are null right now)
+                            // .filter(endpoint -> properties.hasSameProperties((EmailSubscriptionProperties) endpoint.getProperties()))
                             .findFirst();
                     if (endpointOptional.isPresent()) {
                         return Uni.createFrom().item(endpointOptional.get());
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
@@ -94,6 +94,14 @@ public class EndpointResources {
                 .onItem().call(this::loadProperties);
     }
 
+    public Uni<EndpointType> getEndpointTypeById(String accountId, UUID endpointId) {
+        String query = "Select e.type from Endpoint e WHERE e.accountId = :accountId AND e.id = :endpointId";
+        return session.createQuery(query, EndpointType.class)
+                .setParameter("accountId", accountId)
+                .setParameter("endpointId", endpointId)
+                .getSingleResultOrNull();
+    }
+
     public Uni<Endpoint> getOrCreateEmailSubscriptionEndpoint(String accountId, EmailSubscriptionProperties properties) {
         return getEndpointsPerType(accountId, EndpointType.EMAIL_SUBSCRIPTION, true, null)
                 .onItem().call(this::loadProperties)
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
@@ -1,6 +1,7 @@
 package com.redhat.cloud.notifications.db;
 
 import com.redhat.cloud.notifications.models.CamelProperties;
+import com.redhat.cloud.notifications.models.EmailSubscriptionProperties;
 import com.redhat.cloud.notifications.models.Endpoint;
 import com.redhat.cloud.notifications.models.EndpointProperties;
 import com.redhat.cloud.notifications.models.EndpointType;
@@ -14,6 +15,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.function.Function;
@@ -90,6 +92,31 @@ public class EndpointResources {
 
         return mutinyQuery.getResultList()
                 .onItem().call(this::loadProperties);
+    }
+
+    public Uni<Endpoint> getOrCreateEmailSubscriptionEndpoint(String accountId, EmailSubscriptionProperties properties) {
+        return getEndpointsPerType(accountId, EndpointType.EMAIL_SUBSCRIPTION, true, null)
+                .onItem().call(this::loadProperties)
+                .onItem().transformToUni(emailEndpoints -> {
+                    Optional<Endpoint> endpointOptional = emailEndpoints
+                            .stream()
+                            // Todo: This should be changed once we store the properties - (properties = null right now)
+                            // .filter(endpoint -> endpoint.getProperties().equals(properties))
+                            .findFirst();
+                    if (endpointOptional.isPresent()) {
+                        return Uni.createFrom().item(endpointOptional.get());
+                    }
+
+                    Endpoint endpoint = new Endpoint();
+                    endpoint.setProperties(properties);
+                    endpoint.setAccountId(accountId);
+                    endpoint.setEnabled(true);
+                    endpoint.setDescription("System email endpoint");
+                    endpoint.setName("Email endpoint");
+                    endpoint.setType(EndpointType.EMAIL_SUBSCRIPTION);
+
+                    return createEndpoint(endpoint);
+                });
     }
 
     public Uni<Long> getEndpointsCountPerType(String tenant, EndpointType type, Boolean activeOnly) {
--- a/backend/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
@@ -63,6 +63,7 @@ public class BehaviorGroup extends CreationUpdateTimestamped {
     @JsonIgnore
     private String accountId;
 
+    @NotNull
     @NotBlank
     @Schema(name = "display_name")
     private String displayName;
--- a/backend/src/main/java/com/redhat/cloud/notifications/models/EmailSubscriptionProperties.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/models/EmailSubscriptionProperties.java
@@ -1,4 +1,22 @@
 package com.redhat.cloud.notifications.models;
 
 public class EmailSubscriptionProperties extends EndpointProperties {
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+
+        if (!(o instanceof EndpointProperties)) {
+            return false;
+        }
+
+        return true;
+    }
+
+    @Override
+    public int hashCode() {
+        return 0;
+    }
 }
--- a/backend/src/main/java/com/redhat/cloud/notifications/models/EmailSubscriptionProperties.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/models/EmailSubscriptionProperties.java
@@ -8,7 +8,7 @@ public class EmailSubscriptionProperties extends EndpointProperties {
             return true;
         }
 
-        if (!(o instanceof EndpointProperties)) {
+        if (!(o instanceof EmailSubscriptionProperties)) {
             return false;
         }
 
--- a/backend/src/main/java/com/redhat/cloud/notifications/models/EmailSubscriptionProperties.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/models/EmailSubscriptionProperties.java
@@ -2,21 +2,7 @@ package com.redhat.cloud.notifications.models;
 
 public class EmailSubscriptionProperties extends EndpointProperties {
 
-    @Override
-    public boolean equals(Object o) {
-        if (this == o) {
-            return true;
-        }
-
-        if (!(o instanceof EmailSubscriptionProperties)) {
-            return false;
-        }
-
+    public boolean hasSameProperties(EmailSubscriptionProperties otherProps) {
         return true;
-    }
-
-    @Override
-    public int hashCode() {
-        return 0;
     }
 }
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/EmailEndpointMigrationService.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/EmailEndpointMigrationService.java
@@ -46,11 +46,11 @@ public class EmailEndpointMigrationService {
         MigrationReport report = new MigrationReport();
 
         return session.withTransaction(transaction -> {
-            LOGGER.debugf("[NOTIF-240] Starting");
+            LOGGER.debugf("[NOTIF-242] Starting");
             return session.createQuery("SELECT DISTINCT accountId from Endpoint WHERE type = :endpointType", String.class)
                     .setParameter("endpointType", EndpointType.EMAIL_SUBSCRIPTION)
                     .getResultList()
-                    .onItem().invoke(accounts -> LOGGER.debugf("[NOTIF-240] Collapsing Email endpoints"))
+                    .onItem().invoke(accounts -> LOGGER.debugf("[NOTIF-242] Collapsing Email endpoints"))
                     .onItem().transformToMulti(Multi.createFrom()::iterable)
                     .invoke(s -> report.updatedAccounts.incrementAndGet())
                     .onItem().transformToUniAndConcatenate(accountId -> session.createQuery("FROM Endpoint WHERE type = :endpointType AND accountId = :accountId", Endpoint.class)
@@ -60,7 +60,7 @@ public class EmailEndpointMigrationService {
                             .onItem().transformToUni(endpoints -> {
                                 if (endpoints.size() > 1) {
                                     Endpoint defaultEmailEndpoint = endpoints.remove(0);
-                                    LOGGER.debugf("[NOTIF-240] Account [%s] has [%d] EmailEndpoints. Using [%s] as the default", accountId, endpoints.size() + 1, defaultEmailEndpoint.getId());
+                                    LOGGER.debugf("[NOTIF-242] Account [%s] has [%d] EmailEndpoints. Using [%s] as the default", accountId, endpoints.size() + 1, defaultEmailEndpoint.getId());
                                     Set<UUID> endpointIds = endpoints.stream().map(Endpoint::getId).collect(Collectors.toSet());
 
                                     return session.createQuery("FROM BehaviorGroupAction bga WHERE bga.id.endpointId IN (:endpointIds)", BehaviorGroupAction.class)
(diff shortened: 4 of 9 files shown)


josejulio:
@gwenneg could you review? I would need to update the UI to merge at the same time of this PR.

backend/src/main/java/com/redhat/cloud/notifications/models/EmailSubscriptionProperties.java
```diff
@@ -1,4 +1,22 @@
...
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+
+        if (!(o instanceof EndpointProperties)) {
+            return false;
+        }
+
+        return true;
```
josejulio, line 15: This is a placeholder when we add more properties.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -54,6 +56,10 @@
 // TODO Needs documentation annotations
 public class EndpointService {
 
+    private static final List<EndpointType> systemEndpointType = List.of(
+            EndpointType.EMAIL_SUBSCRIPTION
```
josejulio, line 60: The same will happen with other Endpoints that are not meant to be created by the user, like the `Send to drawer`

backend/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
```diff
@@ -63,6 +63,7 @@
     @JsonIgnore
     private String accountId;
 
+    @NotNull
```
josejulio, line 66: Not really in the scope of this PR, but is a minor change. This makes the `field` to be required (in the openapi)

gwenneg: OpenAPI does not understand an annotation from `javax.validation.constraints`? 😢

josejulio: Might be a bug in Smallrye OpenAPI extension

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -119,6 +127,38 @@
...
+                    Endpoint endpoint = new Endpoint();
+                    endpoint.setProperties(properties);
+                    endpoint.setAccountId(principal.getAccount());
+                    endpoint.setEnabled(true);
+                    endpoint.setDescription("System email endpoint");
+                    endpoint.setName("Email endpoint");
+                    endpoint.setType(EndpointType.EMAIL_SUBSCRIPTION);
+
+                    return resources.createEndpoint(endpoint);
+                });
```
josejulio, line 159: I'm going to move this code to `EndpointResources` - To make it available to other parts of the application.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -168,10 +197,18 @@
...
     @APIResponse(responseCode = "200", content = @Content(schema = @Schema(type = SchemaType.STRING)))
     public Uni<Response> updateEndpoint(@Context SecurityContext sec, @PathParam("id") UUID id, @NotNull @Valid Endpoint endpoint) {
+        // This prevents from updating an endpoint from whatever EndpointType to a system EndpointType
+        checkSystemEndpoint(endpoint.getType());
         RhIdPrincipal principal = (RhIdPrincipal) sec.getUserPrincipal();
         endpoint.setAccountId(principal.getAccount());
         endpoint.setId(id);
-        return resources.updateEndpoint(endpoint)
+
+        return resources.getEndpoint(principal.getAccount(), id)
```
gwenneg, line 206: This has a minor impact on performances, but we could run a query that would only return the `Endpoint` type from its `id` instead of fetching the entire `Endpoint`.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -168,10 +197,18 @@
     @RolesAllowed(RbacIdentityProvider.RBAC_WRITE_INTEGRATIONS_ENDPOINTS)
     @APIResponse(responseCode = "200", content = @Content(schema = @Schema(type = SchemaType.STRING)))
     public Uni<Response> updateEndpoint(@Context SecurityContext sec, @PathParam("id") UUID id, @NotNull @Valid Endpoint endpoint) {
+        // This prevents from updating an endpoint from whatever EndpointType to a system EndpointType
+        checkSystemEndpoint(endpoint.getType());
```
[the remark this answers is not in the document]
gwenneg: Will it be possible to update `EmailSubscriptionProperties` fields in the future?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -119,6 +126,16 @@
         return resources.createEndpoint(endpoint);
     }
 
+    @POST
+    @Path("/system/email_subscription")
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    @RolesAllowed(RbacIdentityProvider.RBAC_READ_INTEGRATIONS_ENDPOINTS)
```
gwenneg, line 133: `READ` or `WRITE`?

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -92,6 +94,31 @@
...
     }
 
+    public Uni<Endpoint> getOrCreateEmailSubscriptionEndpoint(String accountId, EmailSubscriptionProperties properties) {
+        return getEndpointsPerType(accountId, EndpointType.EMAIL_SUBSCRIPTION, true, null)
+                .onItem().call(this::loadProperties)
+                .onItem().transformToUni(emailEndpoints -> {
+                    Optional<Endpoint> endpointOptional = emailEndpoints
+                            .stream()
+                            // Todo: This should be changed once we store the properties - (properties = null right now)
+                            // .filter(endpoint -> endpoint.getProperties().equals(properties))
```
gwenneg, line 104: Related with my comment on `EmailSubscriptionProperties`. Will that work knowing that properties equality is currently based on the DB primary key?

backend/src/main/java/com/redhat/cloud/notifications/routers/EmailEndpointMigrationService.java
```diff
@@ -0,0 +1,150 @@
...
+                                                        bgaId.behaviorGroupId = behaviorGroupAction.getId().behaviorGroupId;
+
+                                                        return session.find(BehaviorGroupAction.class, bgaId)
+                                                                .onItem().transformToUni(bgaFound -> {
+                                                                    if (bgaFound == null) {
+                                                                        BehaviorGroupAction replacement = new BehaviorGroupAction(
+                                                                                behaviorGroupAction.getBehaviorGroup(),
+                                                                                defaultEmailEndpoint
+                                                                        );
+                                                                        replacement.setPosition(behaviorGroupAction.getPosition());
```
gwenneg, line 90: IIUC, some positions may be empty after the migration, but I don't think that's a problem.

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
josejulio: Currently we never load/save `EmailSubscriptionProperties`, it's always null.

backend/src/main/java/com/redhat/cloud/notifications/models/EmailSubscriptionProperties.java
josejulio: Not really, this could be part of this PR or the next PR that actually adds attributes to this class.

backend/src/main/java/com/redhat/cloud/notifications/routers/EmailEndpointMigrationService.java
josejulio: Yes, some positions will be like 1, 2, 5, 8. But that shouldn't be a problem.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -119,6 +126,16 @@
         return resources.createEndpoint(endpoint);
     }
 
+    @POST
+    @Path("/system/email_subscription")
```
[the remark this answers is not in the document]
josejulio: Not hidden. If you have a better name it would be good :-)

The purpose of this path is to group `Endpoint`s that are meant to me managed by us (EmailSubscription and NotificationDrawer maybe)

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
josejulio: I think it's debatable.
Only `READ` permissions are fine, we want to get the default email subscription endpoint - If in the process it has to create one, its fine IMO.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
josejulio: That would work.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -278,4 +315,13 @@
                 ));
     }
 
+    private static void checkSystemEndpoint(EndpointType endpointType) {
+        if (systemEndpointType.contains(endpointType)) {
+            throw new BadRequestException(String.format(
+                    "Is not possible to create or alter endpoint with type %s, check API for alternatives",
```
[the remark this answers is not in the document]
josejulio: Well, it's not possible to `create or alter`, there are no alternatives for these APIs. The only valid alternative here would be the `system/email_subscription`

backend/src/main/java/com/redhat/cloud/notifications/models/EmailSubscriptionProperties.java
gwenneg: I'm OK with having a placeholder here, that's not why I commented this. My concern is about the compatibility between the existing `EndpointProperties` `equals/hashCode` implementation and the future implementatations based on `EmailSubscriptionProperties` attributes.

josejulio: The issue here is that for the `getOrCreate` we won't have any `id` on the received properties, so we will need to compare the attributes instead.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
gwenneg: I agree and in the meantime it's weird to have a `POST` operation with `READ` requirement 😄

gwenneg: Let's stick with READ, it makes sense.

backend/src/main/java/com/redhat/cloud/notifications/routers/EmailEndpointMigrationService.java
```diff
@@ -0,0 +1,150 @@
...
+        // Fetch every account and ensure this
+        long start = System.currentTimeMillis();
+        if (!CONFIRMATION_TOKEN.equals(confirmationToken)) {
+            throw new BadRequestException("You forgot to confirm the migration!");
+        }
+        MigrationReport report = new MigrationReport();
+
+        return session.withTransaction(transaction -> {
+            LOGGER.debugf("[NOTIF-242] Starting");
+            return session.createQuery("SELECT DISTINCT accountId from Endpoint WHERE type = :endpointType", String.class)
```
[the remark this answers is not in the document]
josejulio: That would be a way, but I think current is OK'sh enough for the migration code :)

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
josejulio: but when we load/save them, as the received `properties` is not in the database, it doesn't have any `id` and we need to check on the contents.

backend/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
josejulio: I created this issue there: https://github.com/smallrye/smallrye-open-api/issues/851 and will follow up

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
josejulio: Think of an API user with `READ` permissions that wants to get the `EmailSubscription` to see the history.

backend/src/main/java/com/redhat/cloud/notifications/routers/EmailEndpointMigrationService.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
+
+import static com.redhat.cloud.notifications.Constants.INTERNAL;
+
+// Todo: Delete after migration
+@Path(INTERNAL + "/email_endpoint/migrate")
+public class EmailEndpointMigrationService {
```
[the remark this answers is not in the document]
josejulio: Updated

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
josejulio: done

backend/src/main/java/com/redhat/cloud/notifications/models/EmailSubscriptionProperties.java
gwenneg: I'm resolving this since there's another discussion about the same subject. To be continued 😄

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
gwenneg: Can we simply use a new method for the comparison and keep the existing `equals`/`hashCode` methods unchanged? It could be something like `hasSameProperties` or `equalsProperties` that would only be used for that use case.

I am really concerned about overriding `equals`/`hashCode` for that purpose. It could have bad consequence in the near future if we put properties in a `Set` or even with Hibernate. I know the non-reactive version of Hibernate requires proper `equals`/`hashCode` implementations in some cases (that we are not using in notifications) but I am really unsure about Hibernate Reactive.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -135,7 +152,11 @@
     @APIResponse(responseCode = "204", description = "The integration has been deleted", content = @Content(schema = @Schema(type = SchemaType.STRING)))
     public Uni<Response> deleteEndpoint(@Context SecurityContext sec, @PathParam("id") UUID id) {
         RhIdPrincipal principal = (RhIdPrincipal) sec.getUserPrincipal();
-        return resources.deleteEndpoint(principal.getAccount(), id)
+        return resources.getEndpoint(principal.getAccount(), id)
```
gwenneg, line 155: It looks like this could be replaced with the new `resources.getEndpointTypeById` method.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -147,7 +168,11 @@
     @APIResponse(responseCode = "200", content = @Content(schema = @Schema(type = SchemaType.STRING)))
     public Uni<Response> enableEndpoint(@Context SecurityContext sec, @PathParam("id") UUID id) {
         RhIdPrincipal principal = (RhIdPrincipal) sec.getUserPrincipal();
-        return resources.enableEndpoint(principal.getAccount(), id)
+        return resources.getEndpoint(principal.getAccount(), id)
```
gwenneg, line 171: Same here.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -157,7 +182,11 @@
     @APIResponse(responseCode = "204", description = "The integration has been disabled", content = @Content(schema = @Schema(type = SchemaType.STRING)))
     public Uni<Response> disableEndpoint(@Context SecurityContext sec, @PathParam("id") UUID id) {
         RhIdPrincipal principal = (RhIdPrincipal) sec.getUserPrincipal();
-        return resources.disableEndpoint(principal.getAccount(), id)
+        return resources.getEndpoint(principal.getAccount(), id)
```
gwenneg, line 185: Same here.

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -92,6 +94,39 @@
...
+    public Uni<EndpointType> getEndpointTypeById(String accountId, UUID endpointId) {
+        String query = "Select e.type from Endpoint e WHERE e.accountId = :accountId AND e.id = :endpointId";
+        return session.createQuery(query, EndpointType.class)
+                .setParameter("accountId", accountId)
+                .setParameter("endpointId", endpointId)
+                .getSingleResultOrNull();
+    }
+
+    public Uni<Endpoint> getOrCreateEmailSubscriptionEndpoint(String accountId, EmailSubscriptionProperties properties) {
+        return getEndpointsPerType(accountId, EndpointType.EMAIL_SUBSCRIPTION, true, null)
```
gwenneg, line 106: I know email endpoints can't be disabled when this is merged, but that boolean add an extra condition to the HQL query for no reason 😃 

```suggestion
        return getEndpointsPerType(accountId, EndpointType.EMAIL_SUBSCRIPTION, false, null)
```

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
josejulio: Sure, I can create a method to compare `EmailSubscriptionProperties` and leave the equals / hasCode alone

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
josejulio: Ah, you are right, we don't need to check for that.

josejulio: Actually, it has to be `null`, right?
```suggestion
        return getEndpointsPerType(accountId, EndpointType.EMAIL_SUBSCRIPTION, null, null)
```

gwenneg: Good catch, I didn't pay enough attention to the method code 😃

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
josejulio: Updated

backend/src/test/java/com/redhat/cloud/notifications/routers/TestEmailEndpointMigrationService.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static io.restassured.RestAssured.given;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@QuarkusTest
+@QuarkusTestResource(TestLifecycleManager.class)
+public class TestEmailEndpointMigrationService {
```
gwenneg, line 25: It would be better if this extended `DbIsolatedTest` to prevent side-effects on other tests.

gwenneg (commented):
One last comment and I'm done!

backend/src/test/java/com/redhat/cloud/notifications/routers/TestEmailEndpointMigrationService.java
josejulio: done

gwenneg (approved):
LGTM

PR #367, opened 2021-07-06
NOTIF-190 Change to native query

bennetelli:


bennetelli:
need to fork the project first.

PR #368, opened 2021-07-06
NOTIF-190 Change to native query

bennetelli:
This commit changes the standard query to a native one so we can get rid of some models later.

### DIFF
--- a/aggregator/src/main/java/com/redhat/cloud/notifications/db/EndpointEmailSubscriptionResources.java
+++ b/aggregator/src/main/java/com/redhat/cloud/notifications/db/EndpointEmailSubscriptionResources.java
@@ -5,6 +5,7 @@ import org.hibernate.Session;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 import javax.transaction.Transactional;
+import java.math.BigInteger;
 
 @ApplicationScoped
 public class EndpointEmailSubscriptionResources {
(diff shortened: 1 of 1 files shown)


pilhuhn:
Are we sure we need this (premature(?)) optimisation? Shouldn't we get the aggregator up and running and deployed first and then see if that is really needed?

gwenneg:
This will allow `aggregator` to run without depending on the `Bundle`/`Application`/`EmailSubscription` model from `backend`. We're talking about hundreds of lines of code (probably around 500) that are currently required in `aggregator` while it's not necessary. That does not sound premature to me 😃

aggregator/src/main/java/com/redhat/cloud/notifications/db/EndpointEmailSubscriptionResources.java
```diff
@@ -14,13 +15,14 @@
...
+        String query = "SELECT COUNT(user_id) FROM endpoint_email_subscriptions es, applications a, bundles b WHERE es.account_id = :accountId " +
+                "AND es.application_id = a.id AND a.bundle_id = b.id AND b.name = :bundleName AND a.name = :applicationName AND es.subscription_type = :subscriptionType";
+        BigInteger result = (BigInteger) session.createNativeQuery(query)
                 .setParameter("accountId", accountNumber)
                 .setParameter("bundleName", bundleName)
                 .setParameter("applicationName", applicationName)
-                .setParameter("subscriptionType", subscriptionType)
+                .setParameter("subscriptionType", subscriptionType.name())
                 .getSingleResult();
+        return result.longValue();
```
gwenneg, line 26: Normally, converting a `BigInteger` to a long value would be an issue because of the potential data loss, but we're not actually using that value in the calling method. Sonar might be unhappy because of that.

gwenneg (approved):
LGTM

aggregator/src/main/java/com/redhat/cloud/notifications/db/EndpointEmailSubscriptionResources.java
bennetelli: We have a Sonar running?! I didn't know that! Please send me a link :)

gwenneg: It's well hidden behind the VPN. I didn't bookmark the link but I think you can find it somewhere in the Jenkins build log.

PR #370, opened 2021-07-06
NOTIF-190 add modules project

bennetelli:


bennetelli:
Hint: Because of the problem with the dependencies which doesn't allow us to use jackson and jackson-reactive and some other dependencies, the models project doesn't contain the common things anymore (for now, we will change this later) and backend as well as aggregator don't use the project.

This PR adds the project. It also adds configurations and the topic-name we will be using in stage, prod and so on. 
So with this PR, the aggregator part should be able to write things in a Kafka Queue. I tested it locally already but with other data than on stage.

aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
```diff
@@ -34,24 +37,27 @@
...
-        config.put("value.serializer", DEFAULT_STRING_SERIALIZER);
-        config.put("acks", DEFAULT_ACKS);
+    @ConfigProperty(name = "quarkus.aggregator.acks")
+    String acks;
+
+    @ConfigProperty(name = "quarkus.aggregator.topic.name")
+    String topicName;
+
+//    @Scheduled(identity = "dailyEmailProcessor", cron = "{email.subscription.daily.cron}")
+    @Scheduled(identity = "dailyEmailProcessor", every = "1m")
```
theute, line 53: Looks like this was used for testing ? I guess the line above would be the correct one ?

aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
```diff
@@ -63,7 +69,11 @@ public void processDailyEmail(ScheduledExecution se) {
         final LocalDateTime aggregateStarted = LocalDateTime.now();
 
         LOG.info(String.format("Running %s email aggregation for period (%s, %s)", EmailSubscriptionType.DAILY, startTime, endTime));
+
         final List<EmailAggregationKey> applicationsWithPendingAggregation = emailAggregationResources.getApplicationsWithPendingAggregation(startTime, endTime);
+
+        System.out.println("Aggregation Keys " + applicationsWithPendingAggregation);
```
theute, line 75: Local debugging remained here

aggregator/src/main/resources/application.properties
```diff
@@ -4,22 +4,34 @@ quarkus.http.port=8086
...
 %test.quarkus.http.test-port=9086
 
+## configure your datasource
+#quarkus.datasource.db-kind=postgresql
+#quarkus.datasource.username=postgres
+#quarkus.datasource.password=postgres
+#quarkus.datasource.jdbc.url=jdbc:postgresql://127.0.0.1:5432/notifications
 # configure your datasource
-quarkus.datasource.db-kind=postgresql
+quarkus.datasource.db-kind=postgres
```
theute, line 13: Why `postgres` here ? Doc says `postgresql` https://quarkus.io/guides/datasource

aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
bennetelli: Yes, I don't know how else I can test if the cron ob feature is working otherwise.
I would remove all of these things when the backend part is ready and it's working together as expected.

aggregator/src/main/resources/application.properties
bennetelli: Will change it back to postgresql. The other one worked with my local postgres database.

aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
theute: You can always override the EMAIL_SUBSCRIPTION_DAILY_CRON environment variable on your machine

aggregator/src/main/resources/application.properties
```diff
@@ -4,22 +4,34 @@ quarkus.http.port=8086
...
+quarkus.hibernate-orm.dialect=org.hibernate.dialect.PostgreSQL82Dialect
+#quarkus.hibernate-orm.database.generation=drop-and-create
 
 quarkus.hibernate-orm.physical-naming-strategy=com.redhat.cloud.notifications.db.naming.SnakeCasePhysicalNamingStrategy
 
+quarkus.aggregator.serializer=org.apache.kafka.common.serialization.StringSerializer
+quarkus.aggregator.broker-url=localhost:9092
+quarkus.aggregator.acks=1
+quarkus.aggregator.topic.name=platform.notifications.aggregation
+
```
theute, line 27: I assume the 4 variables above are ours and not quarkus ones ?
So I guess they shouldn't start with `quarkus.` but rather `notifications.`, no ?

aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
bennetelli: Fixed.

aggregator/src/main/resources/application.properties
bennetelli: Good catch. Thanks :) 
Fixed with the last PR.

aggregator/src/test/resources/application.properties
```diff
@@ -35,6 +49,8 @@ quarkus.log.sentry.dsn=FILL_ME
 
 %test.quarkus.scheduler.enabled=false
 
+email.subscription.daily.cron=1 * * * * ?
```
theute, line 52: 1. I would standardize and use the same prefix: `notifications.aggregator.`
2. Use `periodic` rather than `daily` in the case we add other frequency (I know more work would have to be done in the backend)
3. And use the real cron for daily (again, on your machine for test, you can always override with an environment variable and we should make sure the checked-in code is always correct)

So something like `notifications.aggregator.periodic.cron=0 0 0 * * ?` I guess

bennetelli: fixed.

aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
```diff
@@ -10,6 +12,7 @@
 import io.vertx.core.Vertx;
 import io.vertx.kafka.client.producer.KafkaProducer;
 import io.vertx.kafka.client.producer.KafkaProducerRecord;
```
gwenneg, line 14: Is there any reason why [quarkus-smallrye-reactive-messaging](https://quarkus.io/guides/kafka) isn't used here?

bennetelli: Yep. We tried to use it in cloudwatch-logging-kafka already and neither Heiko, me nor Peter Palaga got it up and running.
Feel free if you want to give it a try.

gwenneg: What was the problem exactly?

Having direct dependencies on `vertx-kafka-client` and `kafka-clients` isn't recommended by the Quarkus team. Depending on `quarkus-kafka-client` would be a bit better but still not the best way to go. Ideally we should depend on `quarkus-smallrye-reactive-messaging`.

bennetelli: I can't remember what exactly the problem was. I remember we spend some time on trying to get it up and running and used another dependency in the end.

Why isn't it recommended?

bennetelli: I can't remember what exactly the problem was. I remember we spend some time on trying to get it up and running and used another dependency in the end.

Why isn't it recommended?

Will try to use it later again, but won't spend much time with it. Timeboxed for 2 hours. If it's working then, it's fine. If not, I will put it on my list and do it later.

gwenneg: Because we won't benefit from any of the following things if we don't use the Quarkus extensions:
- Better integration between Kafka and Quarkus core and other extensions. I had a super quick look at `quarkus-kafka-client`, it contains integration code in many areas like startup/shutdown, serialization, health, dev mode, security... `quarkus-smallrye-reactive-messaging` certainly contains even more.
- Build time optimizations (faster boot, lower resources consumption) which mean better performances.
- Native mode compatibility (which is probably not important now). Kafka won't work in native mode without the Quarkus extensions.
- More importantly, support from the Quarkus team when we have a question or something goes wrong.

And I'm sure I am forgetting other important reasons 😄

aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
```diff
@@ -34,24 +37,26 @@
...
 
+    @Scheduled(identity = "dailyEmailProcessor", cron = "{notifications.aggregator.email.subscription.periodic.cron}")
+    public void processDailyEmail(ScheduledExecution se) throws JsonProcessingException {
+        List<EmailAggregation> aggregatedEmails = processAggregateEmails(se.getScheduledFireTime());
+
+        Map<String, String> config = createKafkaConfig();
         KafkaProducer<String, String> producer = KafkaProducer.create(Vertx.vertx(), config);
 
-        KafkaProducerRecord<String, String> records = KafkaProducerRecord.create("mytopic", aggregatedEmails.toString());
+        KafkaProducerRecord<String, String> records = KafkaProducerRecord.create(topicName, new ObjectMapper().writeValueAsString(aggregatedEmails));
```
gwenneg, line 59: The `ObjectMapper` instance provided by Quarkus should be used here (it can be injected).

aggregator/src/main/resources/application.properties
```diff
@@ -4,22 +4,34 @@ quarkus.http.port=8086
...
+#quarkus.datasource.jdbc.url=jdbc:postgresql://127.0.0.1:5432/notifications
 # configure your datasource
 quarkus.datasource.db-kind=postgresql
 quarkus.datasource.username=postgres
-quarkus.datasource.password=postgres
-quarkus.datasource.jdbc.url=jdbc:postgresql://127.0.0.1:5432/notifications
+quarkus.datasource.password=mysecretpassword
+quarkus.datasource.jdbc.url=jdbc:postgresql://127.0.0.1:5432/postgres
+quarkus.datasource.jdbc.driver=org.postgresql.Driver
+quarkus.hibernate-orm.dialect=org.hibernate.dialect.PostgreSQL82Dialect
```
gwenneg, line 18: The 2 lines above shouldn't be necessary.

aggregator/src/main/resources/application.properties
```diff
@@ -4,22 +4,34 @@ quarkus.http.port=8086
...
+#quarkus.hibernate-orm.database.generation=drop-and-create
```
gwenneg, line 19: This is dangerous, let's not commit that even if it's commented:
```suggestion
```

aggregator/src/test/resources/application.properties
```diff
@@ -10,8 +10,22 @@ quarkus.datasource.username=postgres
 quarkus.datasource.password=postgres
 quarkus.datasource.jdbc.url=jdbc:postgresql://127.0.0.1:5432/notifications
 
+# configure your datasource
```
gwenneg, line 13: This entire file should be removed.

You can add test configuration values to the `aggregator/src/main/resources/application.properties` file by using the `%test.` prefix.

aggregator/src/main/resources/application.properties
```diff
@@ -4,22 +4,34 @@ quarkus.http.port=8086
 %test.quarkus.http.port=9086
 %test.quarkus.http.test-port=9086
 
+## configure your datasource
+#quarkus.datasource.db-kind=postgresql
+#quarkus.datasource.username=postgres
+#quarkus.datasource.password=postgres
+#quarkus.datasource.jdbc.url=jdbc:postgresql://127.0.0.1:5432/notifications
```
gwenneg, line 11: This must be reverted before the merge.

gwenneg (changes requested):

aggregator/src/test/resources/application.properties
bennetelli: Okay, then this is different to Spring Boot. Didn't know that.

aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
gwenneg: > Will try to use it later again, but won't spend much time with it. Timeboxed for 2 hours. If it's working then, it's fine. If not, I will put it on my list and do it later.

It can be fixed later indeed.

pilhuhn: Btw notifications-gw sends to Kafka like this https://github.com/RedHatInsights/notifications-gw/blob/main/src/main/java/com/redhat/cloud/notifications/GwResource.java#L52 and https://github.com/RedHatInsights/notifications-gw/blob/main/src/main/java/com/redhat/cloud/notifications/GwResource.java#L95

bennetelli: My expected timebox wasn't that bad.

aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
```diff
@@ -63,7 +54,9 @@ public void processDailyEmail(ScheduledExecution se) {
         final LocalDateTime aggregateStarted = LocalDateTime.now();
 
         LOG.info(String.format("Running %s email aggregation for period (%s, %s)", EmailSubscriptionType.DAILY, startTime, endTime));
```
pilhuhn, line 56: I am a bit worried about the start/end time. This assumes, that the job always exactly starts at e.g. the full hour and 0 seconds. Now what if for whatever reasons the job starts one day at e.g. midnight and 0s 0ms and the next day at midnight +3s. We would then not aggregate the emails of the first 3s of the previous day. 

I have no idea if that is really a problem or more like a thought experiment.

bennetelli: To be honest, I just took it from backend. It's the same there.

I think we should put it on our list. If it's a problem, it should still be a problem since we implemented it in notifications-backend.

pilhuhn: Yes, does not need to be changed now

pilhuhn (approved):
Lgtm, but ask @gwenneg :)

gwenneg (approved):
LGTM, thanks for the `smallrye-reactive-messaging` update!

It may be interesting to use [InMemoryConnector](https://smallrye.io/smallrye-reactive-messaging/smallrye-reactive-messaging/2/testing/testing.html) in `DailyEmailAggregationJobTest` but this can be done later.

bennetelli:
> LGTM, thanks for the `smallrye-reactive-messaging` update!
> 
> It may be interesting to use [InMemoryConnector](https://smallrye.io/smallrye-reactive-messaging/smallrye-reactive-messaging/2/testing/testing.html) in `DailyEmailAggregationJobTest` but this can be done later.

You're welcome :) thanks for the hint for using this dependency.

aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
theute: AFAIK it was a scheduled job of a running pod, but now it will be a new pod starting every day, so I think it's more subject to have delays as the startup time may change

josejulio: IIRC the start time and end time are taken from the "expected start date" and not the actual start date.

josejulio: So yes, the expected start date https://github.com/RedHatInsights/notifications-backend/pull/370/files#diff-ca42686170b3e81b0d58629ed3bea6644c4a638b80bbd7f8d85a43e57bdd5c6cR45

PR #379, opened 2021-07-09
NOTIF-190 consume kafka messages in backend

bennetelli:


backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -269,98 +235,29 @@ public void addEmail(Email email) {
...
+                action.setBundle(emailAggregationItem.getBundleName());
+
+                // We don't have a eventtype as this aggregates over multiple event types
+                action.setEventType(null);
+                action.setTimestamp(LocalDateTime.now(ZoneOffset.UTC));
+
+                // We don't have any endpoint (yet) as this aggregates multiple endpoints
+                Notification item = new Notification(action, null);
+
+                sendEmail(item, EmailSubscriptionType.DAILY);
```
gwenneg, line 258: This returns a `Uni` with no subscription afterwards which mean the `Uni` may never be resolved (executed). Does it actually send an email?

bennetelli: I don't know. I spend so much time in trying to test it but this still doesn't work.

theute:
Is it a draft or is it ready for review&merge ? If it's still a draft, please mark it as draft so we don't accidentally merge

bennetelli:
> Is it a draft or is it ready for review&merge ? If it's still a draft, please mark it as draft so we don't accidentally merge

I am not able to test the email sending thing. The unit test is always green, even if I remove all the mocks. I'm waiting for Gwennegs response here.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
gwenneg: [This guide](https://smallrye.io/smallrye-mutiny/guides/testing) contains some information about testing `Uni`/`Multi` but it doesn't really explain how to use `UniAssertSubscriber` so a [quick look](https://github.com/search?l=Java&q=UniAssertSubscriber&type=Code) at the GitHub code from all repos that use `UniAssertSubscriber` can also be helpful. That's one way to test reactive code, but if we strictly consider the question of the emails being sent or not, I don't think `UniAssertSubscriber` is the easiest way to test that.

In notifications-backend, a webhook is called each time an email is sent so mocking the webhook server side with `mockserver` could be a solution to check that the emails are sent. We do that in [WebhookTest](https://github.com/RedHatInsights/notifications-backend/blob/master/backend/src/test/java/com/redhat/cloud/notifications/processors/webhook/WebhookTest.java) for example. But that's a bit overkill because I don't think the entire webhooks code should be tested here.

Mocking the `WebhookTypeProcessor webhookSender` attribute from `EmailSubscriptionTypeProcessor` with Mockito and verifying that the `doHttpRequest` method is called is probably enough to make sure emails are sent during the `EmailSubscriptionTypeProcessor#consumeEmailAggregations` execution.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -269,98 +192,79 @@ public void addEmail(Email email) {
...
+            });
+            return null;
+        });
+    }
```
gwenneg, line 269: This (untested code written from GH) should fix the issue.

I added a few comments for the review process but don't commit them 😃 
```suggestion
    @Incoming("aggregation")
    // If the method returns Uni<Void> instead of void, Quarkus will automatically subscribe to the Uni.
    public Uni<Void> consumeEmailAggregations(String aggregationsJson) throws JsonProcessingException {
        List<EmailAggregation> aggregations = objectMapper.readValue(aggregationsJson, new TypeReference<>() { });
        // Each element of the "aggregations" list must be processed as a reactive (Uni) operation because the email won't be sent otherwise, so we need to transform the List into a Multi.
        return Multi.createFrom().iterable(aggregations)
                // Each element of the Multi will be treated as a Uni.
                // Concatenate prevents concurrent access to the Hibernate Reactive session (that's forbidden).
                .onItem().transformToUniAndConcatenate(emailAggregationItem -> {
                    Action action = new Action();
                    action.setContext(emailAggregationItem.getPayload().getMap());
                    action.setEvents(List.of());
                    action.setAccountId(emailAggregationItem.getAccountId());
                    action.setApplication(emailAggregationItem.getApplicationName());
                    action.setBundle(emailAggregationItem.getBundleName());

                    // We don't have a eventtype as this aggregates over multiple event types
                    action.setEventType(null);
                    action.setTimestamp(LocalDateTime.now(ZoneOffset.UTC));

                    // We don't have any endpoint (yet) as this aggregates multiple endpoints
                    Notification item = new Notification(action, null);

                    return sendEmail(item, EmailSubscriptionType.DAILY);
                })
                // The following line will wait for all the Unis above to be resolved and then replace everything with a Uni<Void> value because that's all we need in the end.
                .onItem().ignoreAsUni();
    }
```

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessorTest.java
```diff
@@ -0,0 +1,82 @@
...
+        endpoint.setName("someName");
+        endpoint.setEnabled(true);
+        endpoint.setDescription("someDescription");
+        endpoint.setAccountId("someAccountId");
+
+        testee.process(action, List.of(endpoint));
+    }
+
+    @Test
+    void shouldSendEmail() {
```
gwenneg, line 66: `EmailSubscriptionTypeProcessor#consumeEmailAggregations` should be tested using [SmallRye Reactive Messaging testing features](https://smallrye.io/smallrye-reactive-messaging/smallrye-reactive-messaging/2/testing/testing.html). There is an example available in [EventConsumerTest](https://github.com/RedHatInsights/notifications-backend/blob/master/backend/src/test/java/com/redhat/cloud/notifications/events/EventConsumerTest.java#L42).

The idea is to replace the incoming Kafka `aggregation` channel with a mocked channel `InMemoryConnector.source("aggregation")` and then send a payload throught that channel. It will trigger a reactive messaging processing in `EmailSubscriptionTypeProcessor#consumeEmailAggregations` exactly like what would happen on prod. Then you should be able to run the Mockito `verify`.

bennetelli: I am not able to test it. Debugging doesn't work, changing the method to:     

@Incoming("aggregation")
    public Uni<Void> consumeEmailAggregations(String aggregationsJson) throws JsonProcessingException {throw new IllegalArgumentException("");}

Also doesn't result in exceptions in tests. I lost so many hours and days with this stuff and don't want to lose more. Testing single methods which deal with Unis and Multis doesn't work and testing this is also not fun. I will stop it here and mark it as resolved.

gwenneg: I can try fixing this if you want.

gwenneg: I pushed a test for the Kafka messages processing. This only works for successfull test cases.

Testing a failure because of an invalid payload will require another kind of test based on [this](https://smallrye.io/smallrye-mutiny/guides/testing).

bennetelli: Thanks :) I actually wanted to test more specific things, but maybe this is enough.

pilhuhn:
Thank you @gwenneg for helping here.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -269,98 +194,78 @@ public void addEmail(Email email) {
...
-                            )
-                    );
-                });
-    }
-
-    @Scheduled(identity = "dailyEmailProcessor", cron = "{email.subscription.daily.cron}")
-    public void processDailyEmail(ScheduledExecution se) {
-        // Only delete on the largest aggregate time frame. Currently daily.
-        processAggregateEmails(se.getScheduledFireTime(), EmailSubscriptionType.DAILY, true).await().indefinitely();
-    }
```
gwenneg, line 364: If we merge this, the old way of sending the daily email will be gone. That means the new way _has to_ work and be reliable immediately. Otherwise, we won't be able to fix any production issue that would require to change and redeploy notifications-backend.

Shouldn't we keep the old way and migrate to the new way using a configuration property?

cc @josejulio @pilhuhn

backend/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/PoliciesEmailPayloadAggregator.java
```diff
@@ -27,14 +30,22 @@
...
+    private final Map<String, HashSet<String>> uniqueHostPerPolicy = new HashMap<>();
 
-    public PoliciesEmailPayloadAggregator() {
+    static AbstractEmailPayloadAggregator by(EmailAggregationKey aggregationKey) {
+        if (aggregationKey.getBundle().equals("rhel") && aggregationKey.getApplication().equals("policies")) {
+            return new PoliciesEmailPayloadAggregator();
+        }
+
+        return null;
+    }
```
gwenneg, line 42: Unless I missed something, this is never used.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/Email.java
```diff
@@ -1,70 +1,70 @@
...
+package com.redhat.cloud.notifications.processors.email;
 
 import java.util.HashSet;
 import java.util.Set;
 
 /**
  * Input JSON format accepted by the BOP
  */
-public class Email {
+class Email {
```
gwenneg, line 9: There are lots of access modifier changes and minor refactors in this PR which make the review longer/harder, especially in `EmailSubscriptionTypeProcessor`. Please don't put too many of these in a PR that contains a major change such as the daily email job. 😃

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
bennetelli: I would like to put it on stage first and check if it's working there. 
There is no plan to put it in prod immediately. But this approach could be a problem when someone else also wants to put something in prod in the meantime.

By using a feature toggle, this would be a no brainer. 

I think we should keep the old way, indeed.

gwenneg (commented):
Intermediate comments (I'm not done with the review):

backend/src/main/java/com/redhat/cloud/notifications/processors/email/Email.java
bennetelli: Argh.. sorry. Got used to the pathfinder rule and clean up things the moment I see them.

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailTest.java
```diff
@@ -1,461 +0,0 @@
...
-import static org.junit.jupiter.api.Assertions.assertIterableEquals;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.fail;
-import static org.mockserver.model.HttpResponse.response;
-
-@QuarkusTest
-@TestInstance(TestInstance.Lifecycle.PER_CLASS)
-@QuarkusTestResource(TestLifecycleManager.class)
-public class EmailTest extends DbIsolatedTest {
```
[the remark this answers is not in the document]
gwenneg: It should also stay there for a while if we keep both the old and the new ways to aggregate emails in the code.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -269,98 +194,78 @@ public void addEmail(Email email) {
...
-                        return Multi.createFrom().empty();
-                    }
+                    return objects;
+                });
+    }
 
-                    aggregator.setStartTime(startTime);
-                    aggregator.setEndTimeKey(endTime);
+    @Incoming(AGGREGATION_CHANNEL)
+    public Uni<Void> consumeEmailAggregations(String aggregationsJson) throws JsonProcessingException {
```
gwenneg, line 248: Note for later: we'll need metrics in this method.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -269,98 +194,78 @@ public void addEmail(Email email) {
...
+        List<EmailAggregation> aggregations = objectMapper.readValue(aggregationsJson, new TypeReference<>() {
+        });
```
gwenneg, line 250: If that line throws a `JsonProcessingException`, the Kafka message won't be acknowledged and there will be another attempt at processing the Kafka message which will likely be invalid again (there's no reason an invalid JSON payload would magically become valid). That sounds like an infinite loop of failed attempts.

If that happens, should we ack the Kafka message and add an ERROR log entry so that we get notified through Sentry that something went wrong and we have to manually fix it? If so, we may need to save the data that was unsuccessfully sent.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -269,98 +194,78 @@ public void addEmail(Email email) {
...
+        return Multi.createFrom().iterable(aggregations)
+                .onItem().transformToUniAndConcatenate(emailAggregationItem -> {
                     Action action = new Action();
-                    action.setContext(aggregator.getContext());
+                    action.setContext(emailAggregationItem.getPayload().getMap());
```
gwenneg, line 254: Is there an NPE risk here because `emailAggregationItem.getPayload()` would be null?

gwenneg (changes requested):
More comments, still not done 😃

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
josejulio: Yes, lets keep the old way meanwhile and use a flag (maybe a variable in properties?) to turn on/off (default on :D).

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessorTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @InjectMock
+    EndpointEmailSubscriptionResources subscriptionResources;
+
+    @Test
+    void shouldNotProcessWhenEndpointsAreNull() {
+        final Action action = new Action("someBundle", "someApplication", "someEventType", LocalDateTime.now(), "someAccountId", Map.of(), List.of());
+        final Multi<NotificationHistory> process = testee.process(action, null);
+
+        assertEquals(0, process.collect().asList().await().indefinitely().size());
```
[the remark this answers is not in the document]
josejulio: :cry:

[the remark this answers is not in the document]
bennetelli: Hey :-D the suggestion to use .await().indefinitely() comes from you :-D

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
bennetelli: I would say no. There is a NotNull annotation on the payload attribute.

    @NotNull
    @Convert(converter = JsonObjectConverter.class)
    private JsonObject payload;

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
bennetelli: I will create a Jira ticket for it so we don't lose this note

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessorTest.java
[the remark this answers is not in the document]
gwenneg: You're right, but it changed from OK to bad practice recently 😄 It's still OK with Quarkus 1 but I'm sure we'll migrate to Quarkus 2 soon so we'd better stop awaiting indefinitely ASAP.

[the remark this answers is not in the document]
bennetelli: a lot happening under the hood of Quarkus :-D where did you get this information from? :)

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
bennetelli: Is "having it on status ON" a joke? Or do you want it to be on ON by default? :D

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
bennetelli: That's what I wanted to test. Doesn't work. Maybe someone else wants to give it a try?

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
josejulio: I mean to have the "old way"  as default :D

bennetelli: Okay. I disabled the "new way" with my last commit. Your desired behavior should be there now

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
bennetelli: As far as I understand this here correctly: https://smallrye.io/smallrye-reactive-messaging/smallrye-reactive-messaging/2/acknowledgement/acknowledgement.html

They will get acknowledged with my latest change.

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessorTest.java
```diff
@@ -0,0 +1,130 @@
...
+        when(endpointEmailSubscriptionResources.getEmailSubscribers(anyString(), anyString(), anyString(), any())).thenReturn(listUni);
+
+        final WebhookTypeProcessor webhookTypeProcessor = spy(WebhookTypeProcessor.class);
+
+        testee.sendEmail(new Notification(new Action(), new Endpoint()), DAILY).await().indefinitely();
+
+        verify(webhookTypeProcessor, times(1)).doHttpRequest(any(), any(), any());
+    }
+
+    @Test
```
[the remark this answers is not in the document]
bennetelli: This would be a no brainer in "normal"  I am unable to test it in this reactive, kafka scenario.

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessorTest.java
```diff
@@ -0,0 +1,130 @@
...
+    void shouldSuccessfullySendEmail() {
```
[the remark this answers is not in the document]
bennetelli: I need help here. With Spring Boot this would be a no brainer. With this reactive and kafka thing it's not. At least not for me.

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessorTest.java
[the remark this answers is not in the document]
gwenneg: I've been involved in the Quarkus dev for more than 2 years so I naturally follow many subjects there. Being a liaison between our team and Quarkus is also part of my current responsibilites 😃

gwenneg:
Please let me resolve the comments I created, otherwise I'm never sure which ones need my attention and I have to check all of them 😃

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -269,6 +204,121 @@ public void addEmail(Email email) {
...
+    }
+
+    @Incoming(AGGREGATION_CHANNEL)
+    public Uni<Void> consumeEmailAggregations(String aggregationsJson) {
+        List<EmailAggregation> aggregations;
+        try {
+            aggregations = objectMapper.readValue(aggregationsJson, new TypeReference<>() {
+            });
+        } catch (JsonProcessingException e) {
+            log.info(e.getMessage());
```
gwenneg, line 264: We need the Sentry alert with a full stack trace if that happens:
```suggestion
            log.error("Kafka aggregation payload parsing failed", e);
```

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
gwenneg: The infinite loop is broken indeed. Now there's still the question of the data that may be lost after the scheduled aggregations purge following a failure here.

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailTest.java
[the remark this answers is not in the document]
gwenneg: This class should be left untouched until `aggregator` is enabled by default on prod. We need that test coverage to prevent regressions in backend.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
gwenneg: The [old scheduled job](https://github.com/bennetelli/notifications-backend/blob/7ec7cfc6126588c19a68c9459947c605df33db32/backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java#L288-L292) should be executed only if `notifications.aggregator.email.subscription.periodic.cron.enabled` is `true`.

bennetelli: I am not sure if this would work. This entry is in another project and they are in different jars and have different application.properties. I would say this doesn't work.
We can add another property in notifcations-backend, but this somehow doesnt feel the right way

gwenneg: Oh yes I forgot notifications was multi-module now, my bad 😄

I think adding the same config property in backend is fine as long as both properties (backend and aggregator) are commented with something explaining that they should always be modified in a synchronized way and the values should never be identical between both projects.

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailTest.java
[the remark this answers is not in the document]
bennetelli: added the missing tests again

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
bennetelli: Do you have an idea how to test it?

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
bennetelli: I just added the property and comments in both application properties.

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessorTest.java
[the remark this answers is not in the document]
gwenneg: I pushed a small test that ensures that we won't throw any exception (in other words the Kafka message will be acked) in case of invalid payload.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
gwenneg: See [this](https://github.com/RedHatInsights/notifications-backend/pull/379#discussion_r674636428) for the test.

I can only think of one case where we could have a payload parsing issue which could cause data loss: it could happen if `backend` and `aggregator` used different versions of the `EmailAggregation` class. As soon as both projects depend on the same version of that class from the `models` module, the problem will be gone.

The data loss isn't such a big deal, I don't think it's worth implementing anything to save that data in case of failure.

bennetelli:
> Please let me resolve the comments I created, otherwise I'm never sure which ones need my attention and I have to check all of them smiley

Sorry.. :(

aggregator/src/main/resources/application.properties
```diff
@@ -37,6 +37,10 @@ quarkus.log.sentry.dsn=FILL_ME
 
 %test.quarkus.scheduler.enabled=false
 
+# if you change this value, you should also change:
+# notifications.backend.email.subscription.periodic.cron.enabled in notifications-backend project
+# to avoid both cron jobs are running at the same time
+notifications.aggregator.email.subscription.periodic.cron.enabled=false
 notifications.aggregator.email.subscription.periodic.cron=0 0 0 * * ?
```
gwenneg, line 44: The scheduled job starts at 2 am in `backend` while it starts at 12 am here. Is there any reason for that change?

gwenneg:
> Sorry.. :(

No worries 😉

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessorTest.java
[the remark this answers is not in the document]
gwenneg: Let's not block this PR because of that test, it can be added later.

gwenneg:
I am [one comment](https://github.com/RedHatInsights/notifications-backend/pull/379#discussion_r674651589) away from approving the PR, but be careful with the merge as it will conflict pretty badly with https://github.com/RedHatInsights/notifications-backend/pull/388 (which I also reviewed). You should talk with @josejulio to determine who's the lucky guy who gets to merge first 😄

aggregator/src/main/resources/application.properties
bennetelli: The property: notifications.backend.email.subscription.daily.cron=0 0 2 * * ? is the same as the whole time. I just renamed it.

The new property is also the same as before. I mean Thomas suggested that time. But anyway, we will change it anyway as soon as it's on ci to something like "every 1-2 minutes" or so to be able to test it.

I would say the current time in the new cron job is not that important for now as it will be changed at least two times anyway.

gwenneg: I'm only concerned about prod:

- [Current prod cron value](https://github.com/RedHatInsights/notifications-backend/blob/56c53f81c8f24f0cf35f124298ab816879c359a5/backend/src/main/resources/application.properties#L93)  is `0 0 2 * * ?` (2 am)
- [Backend cron value after merge](https://github.com/bennetelli/notifications-backend/blob/22fc8a207ed4fe1b8a6d9bc62d528dc51ac44e70/backend/src/main/resources/application.properties#L102) will be `0 0 2 * * ?` (same as current prod)
- [Aggregator cron value after merge](https://github.com/bennetelli/notifications-backend/blob/22fc8a207ed4fe1b8a6d9bc62d528dc51ac44e70/aggregator/src/main/resources/application.properties#L44) will be `0 0 0 * * ?` (two hours earlier than current prod)

I would expect these to be all equal unless there's a good reason to change the execution hour.

gwenneg: I didn't check if this is overridden using env vars on prod though.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
bennetelli: > I am [one comment](https://github.com/RedHatInsights/notifications-backend/pull/379#discussion_r674651589) away from approving the PR, but be careful with the merge as it will conflict pretty badly with #388 (which I also reviewed). You should talk with @josejulio to determine who's the lucky guy who gets to merge first smile

Yea, I talked to him already :-D luckily he is still sleeping :-D so lets go :-D

bennetelli: > I can only think of one case where we could have a payload parsing issue which could cause data loss: it could happen if `backend` and `aggregator` used different versions of the `EmailAggregation` class. As soon as both projects depend on the same version of that class from the `models` module, the problem will be gone.

I just tried to move all the classes to the models project. This doesn't work at the moment. We would need to add reactive things to the aggregator project first.

aggregator/src/main/resources/application.properties
bennetelli: But aggregator won't get deployed to prod after merged this PR. There will be a second PR after we got it up and running on ci as I am sure there will be some things to clean up, maybe some things won't work, ... and so on. 
And we will also have to write the aggregator in reactive so we can get the benefits of the models project.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
bennetelli: > > I am [one comment](https://github.com/RedHatInsights/notifications-backend/pull/379#discussion_r674651589) away from approving the PR, but be careful with the merge as it will conflict pretty badly with #388 (which I also reviewed). You should talk with @josejulio to determine who's the lucky guy who gets to merge first smile
> 
> Yea, I talked to him already :-D luckily he is still sleeping :-D so lets go :-D

huhu, looking at you @josejulio :-D

aggregator/src/main/resources/application.properties
gwenneg: Ok let's take care of that later then.

gwenneg (approved):
LGTM

gwenneg:
There is one failing test! Will @josejulio win the merge race??

josejulio:
> There is one failing test! Will @josejulio win the merge race??

I'm sure I can merge faster ;D

gwenneg:
The test should be OK now.

josejulio:
After syncing with @bennetelli we are going to merge this one; I'm going to assess the impact on the other PR and we are  going to check what would it take for the aggregator to pass the AggregationKey to kafka instead of the aggregated payload.

PR #383, opened 2021-07-13
Notif 190 fix checkstyle location issue

bennetelli:


theute:
Please always use the right JIRA ID format: ie `NOTIF-190` (not `Notif 190`) and don't re-include prior commits (and commit messages).

theute:
And test locally as it still doesn't pass

theute:
You could replace all the COPY commands to:
COPY . /home/jboss

bennetelli:
tried it already. Still doesn't work. Will put checkstyle in every project again.

theute:
Don't put the files in each project again, it builds without, just need to adapt the dockerfile

PR #385, opened 2021-07-13
NOTIF-190 Fix Dockerfile

theute:
This builds on my machine

### DIFF
--- a/docker/Dockerfile-build.jvm
+++ b/docker/Dockerfile-build.jvm
@@ -5,13 +5,9 @@
 # Build the project
 FROM registry.access.redhat.com/ubi8/openjdk-11:latest AS build
 USER root
-COPY pom.xml /home/jboss
-COPY backend/src /home/jboss/backend/src
-COPY backend/pom.xml /home/jboss/backend
-COPY checkstyle/checkstyle/checkstyle.xml /home/jboss/backend
-COPY checkstyle/checkstyle/checkstyle-suppressions.xml /home/jboss/backend
+COPY . /home/jboss
 WORKDIR /home/jboss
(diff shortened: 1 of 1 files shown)


bennetelli:
Mom, want to take a look what causes my issue.

docker/Dockerfile-build.jvm
```diff
@@ -5,13 +5,9 @@
...
 USER root
-COPY pom.xml /home/jboss
-COPY backend/src /home/jboss/backend/src
-COPY backend/pom.xml /home/jboss/backend
-COPY checkstyle/checkstyle/checkstyle.xml /home/jboss/backend
-COPY checkstyle/checkstyle/checkstyle-suppressions.xml /home/jboss/backend
+COPY . /home/jboss
 WORKDIR /home/jboss
-RUN mvn clean package -DskipTests --projects :notifications-parent,:notifications-backend
+RUN mvn clean install -DskipTests
```
bennetelli, line 10: I think install isn't needed here. I would say we don't need the dependencies to be in the local .m2/repository folder and package is enough to work with it.

my issue was that I added checktyle at the --projects flag, but in the wrong order.

bennetelli (commented):
RUN mvn clean package -DskipTests -Dskip.yarn --projects :checkstyle,:notifications-parent,:notifications-backend

This works as well and should take less time and memory I would say.

bennetelli (approved):
You can take yours or mine. Whatever you like :)

theute:
Go for it, I'll close this one.
As long as you don't duplicate checkstyles ;-)

theute:
But don't skip yarn, we need it

PR #388, opened 2021-07-13
NOTIF-200 Adding support for personalized emails

josejulio:


gwenneg:
Did you cherry-pick commits from `master`. The branch contains some `master` commits in a weird order. 😃

josejulio:
Ah, I based the work on other PR - I'll clean off the commits.

josejulio:
> Ah, I based the work on other PR - I'll clean off the commits.

done

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailTest.java
```diff
@@ -157,47 +147,41 @@ void testEmailSubscriptionInstant() {
...
+        assertEquals(3, bodyRequests.size());
+        for (int i = 0; i < usernames.length; ++i) {
+            JsonObject body = new JsonObject(bodyRequests.get(i));
+            JsonArray emails = body.getJsonArray("emails");
+            assertNotNull(emails);
+            assertEquals(1, emails.size());
+            JsonObject firstEmail = emails.getJsonObject(0);
+            JsonArray recipients = firstEmail.getJsonArray("recipients");
+            assertEquals(1, recipients.size());
+            assertEquals(usernames[i], recipients.getString(0));
```
josejulio, line 159: Emails could arrive in any order. I need to ensure these are sorted (as done elsewhere) before comparing

backend/src/main/java/com/redhat/cloud/notifications/processors/email/RecipientResolver.java
```diff
@@ -0,0 +1,34 @@
...
+    public Uni<List<User>> recipientUsers(String accountId, List<Endpoint> endpoints, Set<String> subscribers) {
+        return Multi.createFrom().iterable(endpoints)
+                .onItem().transformToUni(e -> recipientUsers(accountId, e, subscribers))
+                .concatenate().collect().in(ArrayList<User>::new, List::addAll)
+                .onItem().transform(users -> users.stream().distinct().collect(Collectors.toList()));
+    }
+
+    private Uni<List<User>> recipientUsers(String accountId, Endpoint endpoint, Set<String> subscribers) {
+        Uni<List<User>> usersUni = rbacRecipientUsersProvider.getUsers(accountId, false);
+        return usersUni.onItem().transform(users -> users.stream().filter(user -> subscribers.contains(user.getUsername())).collect(Collectors.toList()));
```
gwenneg, line 31: Is there a chance `contains` would fail because of the case-sensitive comparison?

backend/src/main/java/com/redhat/cloud/notifications/processors/email/RecipientResolver.java
```diff
@@ -0,0 +1,34 @@
...
+
+@ApplicationScoped
+public class RecipientResolver {
+
+    @Inject
+    RbacRecipientUsersProvider rbacRecipientUsersProvider;
+
+    public Uni<List<User>> recipientUsers(String accountId, List<Endpoint> endpoints, Set<String> subscribers) {
+        return Multi.createFrom().iterable(endpoints)
+                .onItem().transformToUni(e -> recipientUsers(accountId, e, subscribers))
```
gwenneg, line 24: Unless I misunderstood the code, something's wrong here:
- the endpoint is not actually used by `recipientUsers`
- the whole stream invokes `recipientUsers` n times if there are n endpoints
- each `recipientUsers` invocation returns the same list of values
- the resulting lists are flattened into a single list with duplicates removal, but the result is the same as any `recipientUsers` invocation

backend/src/test/java/com/redhat/cloud/notifications/db/ResourceHelpers.java
```diff
@@ -131,9 +132,25 @@ public UUID createWebhookEndpoint(String tenant) {
         ep.setEnabled(true);
         ep.setAccountId(tenant);
         ep.setProperties(properties);
-        return resources.createEndpoint(ep).await().indefinitely().getId();
+        return endpointResources.createEndpoint(ep).await().indefinitely().getId();
     }
 
+    public UUID getBundleId(String accountId, String bundleName) {
```
gwenneg, line 138:
```suggestion
    public UUID getBundleId(String bundleName) {
```

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
```diff
@@ -0,0 +1,132 @@
...
+    String bopApiToken;
+
+    @ConfigProperty(name = "processor.email.bop_client_id")
+    String bopClientId;
+
+    @ConfigProperty(name = "processor.email.bop_env")
+    String bopEnv;
+
+    @ConfigProperty(name = "processor.email.no_reply")
+    String noReplyAddress;
```
gwenneg, line 54: Is that still needed?

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @Inject
+    WebhookTypeProcessor webhookSender;
+
+    @Inject
+    EndpointResources endpointResources;
+
+    public Uni<NotificationHistory> sendEmail(User user, Action action, TemplateInstance subject, TemplateInstance body) {
```
gwenneg, line 62: If we need to send the same email to multiple recipients, will this method be called once for each recipient?

gwenneg (commented):
Here's a few early comments, I'm not done with the review.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
gwenneg: Found the (obvious) answer after reading the Jira ticket: emails can be personalized so each recipient is processed separately.

Did we consider the potentially huge increase of emails we will send through the BOP? It would be interesting to know the average of bcc recipients per email currently sent on prod. We have to be extra careful when this goes to prod. The injected `WebClient` instance has a fixed capacity that may be exceeded which could lead to slow/delayed BOP queries.

gwenneg: The following instance is shared with unsecured webhook calls:
```
@Inject
    @SslVerificationDisabled
    WebClient unsecuredWebClient;
```
We should separate webhooks from emails in this PR and produce an additional `WebClient` instance that would be dedicated to emails only. That way, if we experience poor performances because of personalized emails, webhooks won't be impacted.

gwenneg: It would be great if we could monitor with metrics the average execution time between the moment we want to send an email and the moment when the BOP request is actually completed.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
josejulio: I don't think we need this anymore.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/RecipientResolver.java
josejulio: > * the endpoint is not actually used by `recipientUsers`
> * the whole stream invokes `recipientUsers` n times if there are n endpoints
> * each `recipientUsers` invocation returns the same list of values
> * the resulting lists are flattened into a single list with duplicates removal, but the result is the same as any `recipientUsers` invocation


This is preparation for a follow-up PR, this will be it's next form: https://github.com/RedHatInsights/notifications-backend/pull/396/files#diff-8df1f99650e663bc8ff48a385fec4f0f1511608d2bdddaf457b373db1df6521dR32-R48

image

josejulio: Also, endpoints are deduplicated at some point (or they should be - will double check). Right now we only have one EmailSubscription Endpoint, so `endpoints.size()` will always be at most 1.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/RecipientResolver.java
josejulio: We take the username from rbac data all the times, it should be the same case. I'm not sure if different cases means different users, might need to check that.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
josejulio: > Did we consider the potentially huge increase of emails we will send through the BOP? It would be interesting to know the average of bcc recipients per email currently sent on prod. We have to be extra careful when this goes to prod. The injected `WebClient` instance has a fixed capacity that may be exceeded, which could lead to slow/delayed BOP queries.

One *optimization* we could eventually apply here is to batch emails. BOP allows to make one call with multiple emails.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
```diff
@@ -0,0 +1,153 @@
...
+
+    static final String BOP_APITOKEN_HEADER = "x-rh-apitoken";
+    static final String BOP_CLIENT_ID_HEADER = "x-rh-clientid";
+    static final String BOP_ENV_HEADER = "x-rh-insights-env";
+    static final String BODY_TYPE_HTML = "html";
+    static final ZoneOffset UTC = ZoneOffset.UTC;
+
+    @Inject
+    @BopWebClient
+    WebClient bopWebClient;
```
josejulio, line 47: @gwenneg Using a `BopWebClient` now

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
```diff
@@ -0,0 +1,153 @@
...
+                    // TODO If the call fails - we should probably rollback Kafka topic (if BOP is down for example)
+                    //      also add metrics for these failures
+
+                    return webhookSender.doHttpRequest(
+                            notification,
+                            bopRequest,
+                            getPayload(user, action, subject, body)
+                    ).onItem().invoke(unused -> {
+                        processedCount.increment();
+                        processTime.record(Duration.between(start, LocalDateTime.now(UTC)));
```
josejulio, line 99: > It would be great if we could monitor with metrics the average execution time between the moment we want to send an email and the moment when the BOP request is actually completed.

@gwenneg 
Is this the right way of doing it?

backend/src/main/java/com/redhat/cloud/notifications/processors/email/RecipientResolver.java
josejulio: I tried login with my username in an alternate case and it worked.
Changed to use anyMatch -> equalsIgnoreCase to be on the safe side

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
gwenneg: That would be great. Let's do that in follow-up PR.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/RecipientResolver.java
gwenneg: Oh that makes more sense with the next form 😄

gwenneg: > edit: I could change the List to a Set to make it more obvious.

Which one, the list at line 25? It would be great I see that your follow-up PR already use an intermediate Set, I like that version better 😃

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
gwenneg: Yes that looks good. We don't really need that to run with `UTC` but I don't think using that ZoneOffset comes at a cost so you can leave it that way.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/RecipientResolver.java
josejulio: I already changed the `List<Endpoint>` to `Set<Endpoint>`. Can do the same for line 25

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
josejulio: https://issues.redhat.com/browse/NOTIF-271

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregator.java
```diff
@@ -0,0 +1,123 @@
...
+                    entry.getValue().setEndTimeKey(end);
+                })
+                .collect(
+                    Collectors.toMap(
+                            Map.Entry::getKey,
+                            entry -> entry.getValue().getContext()
+                    )
+                )
+        );
+    }
```
gwenneg, line 108: Would this work? Sorry, I didn't mean to rewrite your code 😃 

```suggestion
    private Uni<Set<String>> getEmailSubscribers(EmailAggregationKey aggregationKey, EmailSubscriptionType emailSubscriptionType) {
        return subscriptionResources
                .getEmailSubscribers(aggregationKey.getAccountId(), aggregationKey.getBundle(), aggregationKey.getApplication(), emailSubscriptionType)
                .onItem().transform(subscriptions ->
                        // TODO Return directly the userId from the database, we don't need the EmailSubscription.
                        subscriptions.stream().map(EmailSubscription::getUserId).collect(Collectors.toSet())
                )
                // This will prevent multiple database queries executions during the aggregation process.
                .memoize().indefinitely();
    }

    public Uni<Map<User, Map<String, Object>>> getAggregated(EmailAggregationKey aggregationKey, EmailSubscriptionType emailSubscriptionType, LocalDateTime start, LocalDateTime end) {
        Map<User, AbstractEmailPayloadAggregator> aggregated = new HashMap<>();
        Uni<Set<String>> subscribers = getEmailSubscribers(aggregationKey, emailSubscriptionType);

        // First, we retrieve all aggregations that match the given key.
        return emailAggregationResources.getEmailAggregation(aggregationKey, start, end)
                .onItem().transformToMulti(Multi.createFrom()::iterable)
                // For each aggregation...
                .onItem().transformToUniAndConcatenate(aggregation -> {
                    // We need its event type to determine the target endpoints.
                    String eventType = getEventType(aggregation);
                    // Let's retrieve these targets.
                    return endpointResources.getTargetEndpointsFromType(aggregationKey.getAccountId(), aggregationKey.getBundle(), aggregationKey.getApplication(), eventType, EndpointType.EMAIL_SUBSCRIPTION)
                            .onItem().transform(Set::copyOf)
                            // Now we want to determine who will actually receive the aggregation email.
                            .onItem().transformToUni(endpoints ->
                                    // All users who subscribed to the current application and subscription type combination are recipients candidates.
                                    subscribers.onItem().transformToUni(users ->
                                            /*
                                             * The actual recipients list is a subset of the candidates.
                                             * The target endpoints properties will determine whether or not each candidate will actually receive an email.
                                             */
                                            recipientResolver.recipientUsers(aggregationKey.getAccountId(), endpoints, users)
                                    )
                            )
                            /*
                             * We now have the final recipients list.
                             * Let's populate with synchronized side-effects (invoke) the Map that will be returned by the method.
                             */
                            .onItem().invoke(users -> {
                                users.forEach(user -> {
                                    // It's aggregation time!
                                    fillUsers(aggregationKey, user, aggregated, aggregation);
                                });
                            });
                })
                // We need to wait for everything to be done before proceeding to the final step.
                .onItem().ignoreAsUni()
                // We need to return the Map that was populated using synchronized side-effects.
                .replaceWith(aggregated)
                .onItem().transform(aggregatorMap ->
                        aggregatorMap
                                .entrySet()
                                .stream()
                                .peek(entry -> {
                                    // TODO These fields could be passed to EmailPayloadAggregatorFactory.by since we know them from the beginning.
                                    entry.getValue().setStartTime(start);
                                    entry.getValue().setEndTimeKey(end);
                                })
                                .collect(
                                        Collectors.toMap(
                                                Map.Entry::getKey,
                                                entry -> entry.getValue().getContext()
                                        )
                                )
                );
    }
```

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregator.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private void fillUsers(EmailAggregationKey aggregationKey, User user, Map<User, AbstractEmailPayloadAggregator> aggregated, EmailAggregation emailAggregation) {
+        if (!aggregated.containsKey(user)) {
+            aggregated.put(user, EmailPayloadAggregatorFactory.by(aggregationKey));
+        }
+
+        AbstractEmailPayloadAggregator aggregator = aggregated.get(user);
```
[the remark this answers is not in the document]
josejulio: I always forget about these APIs

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregator.java
josejulio: :hocho: :page_facing_up: :drop_of_blood: :drop_of_blood:

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -76,257 +54,107 @@
...
-                    emails.addEmail(email);
-                    Uni<JsonObject> payload = Uni.createFrom().item(JsonObject.mapFrom(emails));
+        Multi<Tuple2<NotificationHistory, EmailAggregationKey>> doDelete = delete ?
+                emailAggregationResources.purgeOldAggregation(aggregationKey, endTime)
+                        .onItem().transformToMulti(unused -> Multi.createFrom().empty()) :
+                Multi.createFrom().empty();
 
-                    // TODO Add recipients processing from policies-notifications processing (failed recipients)
-                    //      by checking the NotificationHistory's details section (if missing payload - fix in WebhookTypeProcessor)
+        if (!emailTemplate.isSupported(null, emailSubscriptionType)) {
```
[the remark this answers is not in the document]
josejulio: This is one of these things that will go away once we move the templates to a database.
The point here is that the Daily email subscriptions don't have any `eventType` attach, as they are supposed to aggregate over multiple evenTypes.

Maybe an API like: `isInstantSupported(eventType)` and `isAggregatedSupported(subscriptionType)` would work better.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -76,257 +54,107 @@
...
-
-            return baseTransformer.transform(item.getAction())
-                    .onItem().transform(transformedAction -> {
-                        aggregation.setPayload(transformedAction);
-                        return aggregation;
-                    })
-                    .onItem().transformToUni(emailAggregation -> this.emailAggregationResources.addEmailAggregation(emailAggregation))
-                    .onItem().transformToUni(aBoolean -> sendEmail(item, EmailSubscriptionType.INSTANT));
+    private Multi<NotificationHistory> sendEmail(Action action, Set<Endpoint> endpoints, EmailSubscriptionType emailSubscriptionType) {
+        final EmailTemplate emailTemplate = emailTemplateFactory.get(action.getBundle(), action.getApplication());
```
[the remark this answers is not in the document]
josejulio: I suppose it could.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -76,257 +54,107 @@
...
-                    // TODO If the call fails - we should probably rollback Kafka topic (if BOP is down for example)
-                    //      also add metrics for these failures
-                    return webhookSender.doHttpRequest(item, bopRequest, payload);
-                });
-    }
+        TemplateInstance subject = emailTemplate.getTitle(null, emailSubscriptionType);
+        TemplateInstance body = emailTemplate.getBody(null, emailSubscriptionType);
 
-    private Multi<Tuple2<NotificationHistory, EmailAggregationKey>> processAggregateEmailsByAggregationKey(EmailAggregationKey aggregationKey, LocalDateTime startTime, LocalDateTime endTime, EmailSubscriptionType emailSubscriptionType, boolean delete) {
-        return subscriptionResources.getEmailSubscribersCount(aggregationKey.getAccountId(), aggregationKey.getBundle(), aggregationKey.getApplication(), emailSubscriptionType)
```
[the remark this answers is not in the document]
josejulio: Checking...

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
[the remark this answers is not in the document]
josejulio: That said, it would be better to review that in a separate PR

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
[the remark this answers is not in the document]
josejulio: Not used anymore. Cleaning.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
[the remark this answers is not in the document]
josejulio: Done

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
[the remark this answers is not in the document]
josejulio: Done

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -76,257 +54,107 @@
...
+        TemplateInstance subject = emailTemplate.getTitle(action.getEventType(), emailSubscriptionType);
+        TemplateInstance body = emailTemplate.getBody(action.getEventType(), emailSubscriptionType);
+
+        if (subject == null || body == null) {
+            return Multi.createFrom().empty();
         }
 
-        return sendEmail(item, EmailSubscriptionType.INSTANT);
+        return subscriptionResources
+                .getEmailSubscribers(action.getAccountId(), action.getBundle(), action.getApplication(), emailSubscriptionType)
```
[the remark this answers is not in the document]
josejulio: Done

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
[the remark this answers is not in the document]
gwenneg: Yes let's deal with that later.

gwenneg (approved):
LGTM

gwenneg:
Only one conflicting file? That shouldn't be too hard to fix 😃

josejulio:
> Only one conflicting file? That shouldn't be too hard to fix

I suppose it's because the aggregator has it's own copies :D

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -103,158 +74,84 @@
...
+                    return emailSender.oldSendEmail(users, action, subject, body).toMulti();
                 });
     }
 
+    /*
+    * Warning: This method is not in sync with the current daily-aggregation as a cron job in the backend.
+    * It still sends a single email (as opposed to one email per user) with the users in the bcc
+    * This is going to be updated in a separated PR to match the personalized email.
+    * */
     @Incoming(AGGREGATION_CHANNEL)
```
josejulio, line 155: @gwenneg Adding `oldSendEmail et al` methods to support this flow temporally. I discussed with @bennetelli  and we are going to fix it on a following PR.

gwenneg (approved):
LGTM

PR #396, opened 2021-07-14
NOTIF-153 Add send options to emails

josejulio:
Depends on #388

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
@@ -250,6 +250,10 @@ public class EndpointResources {
         String camelQuery = "UPDATE CamelProperties SET url = :url, subType = :subType, " +
                 "disableSslVerification = :disableSslVerification, secretToken = :secretToken WHERE endpoint.id = :endpointId";
 
+        if (endpoint.getType() == EndpointType.EMAIL_SUBSCRIPTION) {
+            throw new RuntimeException("Unable to update an endpoint of type EMAIL_SUBSCRIPTION");
+        }
+
         return session.createQuery(endpointQuery)
                 .setParameter("name", endpoint.getName())
                 .setParameter("description", endpoint.getDescription())
@@ -287,9 +291,6 @@ public class EndpointResources {
                                         .executeUpdate()
                                         .call(session::flush)
                                         .onItem().transform(rowCount -> rowCount > 0);
-                            case EMAIL_SUBSCRIPTION:
-                                LOGGER.warning("Endpoint updating properties of EmailSubscription. It should never happen.");
-                                // Todo: Should throw? EmailSubscriptions are not updatable once they are created
                             default:
                                 return Uni.createFrom().item(Boolean.TRUE);
                         }
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
@@ -52,9 +52,9 @@ public class EndpointResources {
                         switch (endpoint.getType()) {
                             case CAMEL:
                             case WEBHOOK:
+                            case EMAIL_SUBSCRIPTION:
                                 return session.persist(endpoint.getProperties())
                                         .onItem().call(session::flush);
-                            case EMAIL_SUBSCRIPTION:
                             default:
                                 // Do nothing.
                                 break;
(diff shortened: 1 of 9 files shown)


backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -289,6 +288,8 @@
                                         .call(session::flush)
                                         .onItem().transform(rowCount -> rowCount > 0);
                             case EMAIL_SUBSCRIPTION:
+                                LOGGER.warning("Endpoint updating properties of EmailSubscription. It should never happen.");
+                                // Todo: Should throw? EmailSubscriptions are not updatable once they are created
```
gwenneg, line 292: Even if there's a check in `EndpointService` which should prevent this case, throwing would make sense, but the endpoint itself (not only its properties) can't be updated so I would throw at the beginning of this method.

backend/src/main/java/com/redhat/cloud/notifications/models/EmailSubscriptionProperties.java
```diff
@@ -1,8 +1,65 @@
...
+import javax.validation.constraints.NotNull;
+import java.util.Objects;
+import java.util.UUID;
+
+@Entity
+@Table(name = "email_properties")
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
 public class EmailSubscriptionProperties extends EndpointProperties {
 
+    @NotNull
```
gwenneg, line 18: A primitive attribute can't be `null`.

backend/src/main/java/com/redhat/cloud/notifications/models/EmailSubscriptionProperties.java
```diff
@@ -1,8 +1,65 @@
...
+    @Schema(name = "only_admins")
+    private boolean onlyAdmins;
+
+    @NotNull
```
gwenneg, line 22: Same.

backend/src/main/resources/db/migration/V1.22.0__NOTIF-153-email_properties.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE email_properties (
+    only_admins boolean NOT NULL,
+    ignore_preferences boolean NULL,
```
gwenneg, line 3: Why is it nullable here while the Java attribute is annotated with `@NotNull`?

backend/src/main/java/com/redhat/cloud/notifications/processors/email/RecipientResolver.java
```diff
@@ -26,16 +27,27 @@
...
-        return usersUni
-                .onItem().transform(users -> users
-                        .stream()
-                        .filter(user -> subscribers
-                                .stream()
-                                .anyMatch(subscriber -> subscriber.equalsIgnoreCase(user.getUsername()))
-                        )
-                        .collect(Collectors.toSet())
-                );
+        final EmailSubscriptionProperties props = (EmailSubscriptionProperties) endpoint.getProperties();
```
gwenneg, line 30:
```suggestion
        final EmailSubscriptionProperties props = endpoint.getProperties(EmailSubscriptionProperties.class);
```

backend/src/main/resources/db/migration/V1.22.0__NOTIF-153-email_properties.sql
josejulio: because of copy/paste

backend/src/main/java/com/redhat/cloud/notifications/models/EmailSubscriptionProperties.java
josejulio: Apparently the openapi builder doesn't know about this. If i omit this, the result is an optional type.

backend/src/main/java/com/redhat/cloud/notifications/models/EmailSubscriptionProperties.java
josejulio: Same.

josejulio:
@gwenneg This one is ready for review, just concerned a bit on the RBAC calls failing. I'll try to copy what you did for the other RBAC calls.

backend/src/test/java/com/redhat/cloud/notifications/db/BehaviorGroupResourcesTest.java
```diff
@@ -211,7 +210,7 @@ public void testAddMultipleEmailSubscriptionBehaviorGroupActions() {
         BehaviorGroup behaviorGroup = createBehaviorGroup("displayName", bundle.getId());
         Endpoint endpoint1 = createEndpoint(EMAIL_SUBSCRIPTION);
         Endpoint endpoint2 = createEndpoint(EMAIL_SUBSCRIPTION);
-        updateAndCheckBehaviorGroupActions(ACCOUNT_ID, bundle.getId(), behaviorGroup.getId(), BAD_REQUEST, endpoint1.getId(), endpoint2.getId());
+        updateAndCheckBehaviorGroupActions(ACCOUNT_ID, bundle.getId(), behaviorGroup.getId(), OK, endpoint1.getId(), endpoint2.getId());
```
gwenneg, line 213: This test doesn't really make sense now. Should it be deleted?

gwenneg (approved):
LGTM

backend/src/test/java/com/redhat/cloud/notifications/db/BehaviorGroupResourcesTest.java
josejulio: Not a strong opinion about it.
 
Before we didn't allow multiple Email endpoints, now we do, so I don't think it makes any harm in keeping it to ensure we still allow multiple Email endpoints.

PR #402, opened 2021-07-15
Use relative urls for internal UI

josejulio:
URL was fixed on `/internal`, moved so it can be relative to the base of the `index.html` file.


aferd:
lgtm

PR #404, opened 2021-07-16
Fix EMAIL_SUBSCRIPTION endpoints migration

gwenneg:
This should hopefully fix the email endpoints migration issue on stage. I am not 100% sure it will work because I couldn't reproduce the issue locally with the exact same error message (I got something really close). Anyway, this PR is harmless in case it doesn't fix the issue.

I suspect what happened on stage is a failure during an endpoint deletion because it had history entries in the database and the cascading rule between the `endpoints` and `notifications_history` tables wasn't applied:
```
TABLE "notification_history" CONSTRAINT "notification_history_endpoint_id_fkey" FOREIGN KEY (endpoint_id) REFERENCES endpoints(id) ON UPDATE CASCADE ON DELETE CASCADE
```

It wouldn't be the first time I see issues with `session.remove` and `session.removeAll`. I don't know exactly why but these methods do not behave exactly like an old-school `DELETE FROM` query so I tend to avoid them and write HQL queries which are always reliable. The Javadoc of `remove` and `removeAll` says it cascades delete operations and yet it didn't work when I tried to migrate data after adding history entries.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/EmailEndpointMigrationService.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/EmailEndpointMigrationService.java
@@ -74,7 +74,9 @@ public class EmailEndpointMigrationService {
                                                         defaultEmailEndpoint.getId()
                                                 );
                                             })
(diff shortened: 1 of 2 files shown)


josejulio (approved):
lgtm

PR #405, opened 2021-07-16
Other attempt to have the internal UI relative

josejulio:
Tested locally opening build assets in the browser.

Previous attempt didn't work, because react adds the scripts to `index.html` at build time using the `PUBLIC_URL` as a base.
Changing the PUBLIC_URL to `.` yields base as `./` instead.

```
<script src="./static/js/main.0af8c67b.chunk.js">
```


aferd:
lgtm

aferd (approved):

PR #414, opened 2021-07-20
NOTIF-270 Update template links

josejulio:


theute:
@josejulio so I guess we can undraft and merge ?

gwenneg:
Don't merge yet, I'm about to comment but I need to check something first :smiley:

theute:
If the comment is that we should use a parameter for the hostname and use ci | qa | stage depending on the env, I agree :) I almost created a JIRA for it ;-)

gwenneg (commented):
I don't know if that's out of the scope of this PR, but...
- `OApiService` contains a `https://cloud.redhat.com` url. Should it be updated?
- `RbacRestClientRequestFilter` also contains a (less important) check on `https://ci.cloud.redhat.com`. That same base path is present in `applications.properties` for the RBAC calls for CI.
- `clowdapp.yaml` contains a `cloud.redhat.com` API base path.

gwenneg:
> If the comment is that we should use a parameter for the hostname and use ci | qa | stage depending on the env, I agree :) I almost created a JIRA for it ;-)

No but I completely agree with that! :smiley:

josejulio:
> * `OApiService` contains a `https://cloud.redhat.com` url. Should it be updated?

IIUC, the API is still going to be on cloud.redhat.com - Only the UI is moving. console.redhat.com will be redirecting/proxying over to cloud.redhat.com.

Under that assumption it makes sense to leave it there.

> * `RbacRestClientRequestFilter` also contains a (less important) check on `https://ci.cloud.redhat.com`. That same base path is present in `applications.properties` for the RBAC calls for CI.

Same as above I think

> * `clowdapp.yaml` contains a `cloud.redhat.com` API base path.

I have no idea here :D

theute:
For clowdapp, it's just a reversed DNS name but it won't change.
Yes API won't change now, only the UI

josejulio:
> > If the comment is that we should use a parameter for the hostname and use ci | qa | stage depending on the env, I agree :) I almost created a JIRA for it ;-)
> 
> No but I completely agree with that!

https://issues.redhat.com/browse/POL-383

gwenneg (approved):

PR #419, opened 2021-07-22
Use AggregationCommand instead of aggregated payload

josejulio:
This is something I was discussing with @bennetelli after the call.

Instead of doing the aggregation in the aggregator, let the aggregator only fetch what needs to be aggregated and send instructions (AggregationCommand.java) over to kafka on what exactly needs aggregation.

This helps the `personalized email` and `send to admin` work, because the code won't be duplicated in both `backend` and `aggregator` (one uses mutiny and the other not).

So the aggregating part still resides in `notifications-backend`, but `notifications-aggregator` takes care of deciding what needs aggregating and sending it for processing (thus allowing to have multiple `notifications-backend` pods).

This PR only touches the aggregator and could be merged before #388 so that the changes to update the backend happen there (in PR #388).

Note: Probably more things could be deleted off the aggregator, but didn't go into much detail.


bennetelli (approved):

bennetelli:
LGTM

gwenneg:
@josejulio I'll have a look at this one after https://github.com/RedHatInsights/notifications-backend/pull/388 is merged and the conflicts are resolved.

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/AuthRequestFilter.java
```diff
@@ -11,14 +12,13 @@
...
-    @ConfigProperty(name = "rbac.service-to-service.secret", defaultValue = "addme")
     String secret;
-
-    @ConfigProperty(name = "rbac.service-to-service.application", defaultValue = "notifications")
     String application;
 
     AuthRequestFilter() {
+        Config config = ConfigProvider.getConfig();
+        secret = config.getOptionalValue("rbac.service-to-service.secret", String.class).orElse("addme");
+        application = config.getOptionalValue("rbac.service-to-service.application", String.class).orElse("notifications");
```
josejulio, line 21: The `@ConfigProperty` wasn't working, I was getting `null` on these two variables (same issue in CI). Probably related to the way this class is used (`@RegisterProvider(AuthRequestFilter.class)`)

josejulio:
@gwenneg Rebased and updated.

josejulio:
Note: service to service Rbac token is still needed on all environments, this PR only fixes an issue with loading `@ConfigProperty` values from the `AuthRequest` filter

gwenneg (approved):
As we discussed on Friday with @josejulio, I think we could push this one step further:

- The `notifications-aggregator` pod could send a Kafka payload (on a new topic) to the `notifications-backend` pods that would simply contain the start/end times and the `EmailSubscriptionType`.
- One of the `notifications-backend` pods would process that payload, build the list of `AggregationCommand` (which isn't an expensive process) and send one Kafka payload to the `notifications-backend` pods (in the current Kafka aggreation topic) for each element in the list.
- The `notifications-backend` pods would then process each `AggregationCommand` payload separately to distribute the emails sending load accross all available pods because that can be an expensive operation.
- ~The `notifications-aggregator` pod would still be responsible for the purge (which is a simple SQL query).~ (Edit: see comment below)
- We would need an additional Kafka topic for this to work. The existing one would still be used for the `AggregationCommand` payloads. The new one would be used from the `notifications-aggregator` pod to send the "start aggregating" payload to the `notifications-backend` pods.

Why do that?

- All of the aggregation process steps would be managed/tested at the same place, the maintenance would be easier.
- No code duplication or no need to share a single model class anymore between modules.
- We may even be able to transform the `notifications-aggregator` pod into something else than a Java/Quarkus app.

cc @theute @pilhuhn

gwenneg:
> * The `notifications-aggregator` pod would still be responsible for the purge (which is a simple SQL query).

On second thoughts, the `notifications-backend` pod that would create the `AggregationCommand` list would could perform the purge at the end of that step.

That means the `notifications-aggregator` pod would no longer need access to the notifications database.

josejulio:
> On second thoughts, the `notifications-backend` pod that would create the `AggregationCommand` list could perform the purge at the end of that step.

I think that we will eventually defer this purge, as we will need to persist the events

PR #421, opened 2021-07-23
Use PostgreSQL dump as aggregator initial Flyway script

gwenneg:


gwenneg:
I'm self-approving this as it only contains `test` SQL scripts cleaning with no actual schema or data change.

PR #432, opened 2021-07-28
NOTIF-190 Temporarily remove Kafka topic interactions in backend and …

bennetelli:
…aggregator

### DIFF
--- a/aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
+++ b/aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
@@ -1,15 +1,8 @@
 package com.redhat.cloud.notifications.processors.email.aggregators;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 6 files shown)


gwenneg (commented):
Why remove everything? The channel can simply be disabled like [this](https://github.com/RedHatInsights/notifications-backend/blob/master/backend/src/main/resources/application.properties#L37).

bennetelli:
> Why remove everything? The channel can simply be disabled like [this](https://github.com/RedHatInsights/notifications-backend/blob/master/backend/src/main/resources/application.properties#L37).

Changed. Thanks for the hint :)

gwenneg (approved):
You may have a health check issue with `aggregator` because of the disabled channel. Reactive messaging may be reported as DOWN because all channels are disabled. But since that's only a temporary state, you can apply a temporary workaround: `quarkus.health.extensions.enabled=false`.

`backend` will be fine, we already have two disabled channels on prod.

gwenneg (approved):
LGTM

bennetelli:
don't lgtm too early. The tests need to be green first ;)

gwenneg:
I trusted the GH tests result :smile:

gwenneg:
I shouldn't even LGTM a Draft :smiley:

bennetelli:
haha :-D

PR #434, opened 2021-07-29
NOTIF-175 Print git commit sha during startup

bennetelli:


pilhuhn (approved):

backend/pom.xml
```diff
@@ -279,6 +279,24 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>io.github.git-commit-id</groupId>
+                <artifactId>git-commit-id-maven-plugin</artifactId>
+                <version>4.9.9</version>
```
TheSnoozer, line 285: Perhaps just curious:
Is/was there a reason not to choose v4.9.9 over v5.0.0?

PR #437, opened 2021-07-30
NOTIF-184 Retry on webhook call failure

gwenneg:
This PR adds retry logic to webhook calls when one of the following happens:
- The remote server returned a 5xx HTTP status, which could be caused by a temporary remote server issue.
- We had a technical communication issue during the call (e.g. `IOException`).

The `WebhookTypeProcessor#shouldRetry` method is meant to contain all exceptions that should cause a retry.

The retry logic is entirely configurable:
| Key | Description | Default |
| - | - | - |
| processor.webhook.retry.max-attempts | Max number of retries for a single webhook call (initial call included) | 3 |
| processor.webhook.retry.back-off.initial-value | Initial delay before the first retry attempt | 1S |
| processor.webhook.retry.back-off.max-value | Max delay before the last retry attempt (delay grows exponentially) | 30S |

Default values are open for discussion. I think `back-off.max-value` could be much higher than that.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/processors/webhooks/ServerErrorException.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/processors/webhooks/ServerErrorException.java
@@ -1,0 +1,20 @@
+package com.redhat.cloud.notifications.processors.webhooks;
+
+import com.redhat.cloud.notifications.models.NotificationHistory;
+
+/**
+ * This exception is thrown when the response of a webhook call contains a 5xx HTTP status.
+ * Such status can be caused by a temporary remote issue so we should retry the call.
+ */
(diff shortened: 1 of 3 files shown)


gwenneg:
I don't want this to be merged before I'm back from PTO so the PR will remain a Draft for now.

It is ready for a review though :smiley:

josejulio:
> The retry logic

I'm wondering if we should document this logic for the client, so that they know that it's going to be retry in case of error 500s

josejulio (approved):

gwenneg:
> I'm wondering if we should document this logic for the client, so that they know that it's going to be retry in case of error 500s

Good point, I think this should be documented indeed.

PR #446, opened 2021-08-03
NOTIF-190 Use Kubernetes/OpenShift cron instead of Quarkus Scheduler

bennetelli:


theute:
Will it exit or is that still running as a server ?

aggregator/src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
```diff
@@ -0,0 +1,74 @@
...
+    @Inject
+    DailyEmailAggregationJob dailyEmailAggregationJob;
+
+    // we do need a event as parameter here, otherwise the init method won't get called.
+    void init(@Observes StartupEvent ev) {
+        initAccessLogFilter();
+
+        LOG.info(readGitProperties());
+
+        dailyEmailAggregationJob.processDailyEmail(Instant.now());
```
josejulio, line 36: I think we will need to work on this `Instant.now()` time (If there is something we can test, we can refine in a follow up PR). 
IIRC this value is used to determine the period to aggregate. With Quarkus we knew the "scheduled time", so we might need to think on an alternative to always have the same period (i.e. from 2am of today to 1:59:59 of tomorrow)
Could be hard coded or taken from an env.
It would be nice if the cron could send these values somehow.

bennetelli:
> Will it exit or is that still running as a server ?

With the latest commit it will stop after the methods inside public int run(String... args) have been called.

aggregator/src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
bennetelli: True, we are losing this information with the OpenShift/Kubernetes cronjob. First I would go for a hardcoded version. Later we can think about putting it in a openshift volume or so where we can just store the timestamp and change it every time we finished the job or so.

josejulio: Hardcoding works for me in this phase.

I would check if there is a way for the cron to pass the scheduled time (as quarkus does) and the "DAILY" string. If that's possible we could setup other periods without moving the aggregator code.

bennetelli: We could take a look if it's possible to pass the DAILY String as an argument in the 

public int run(String... args) 

method.

josejulio (approved):

PR #447, opened 2021-08-03
Expose internal api on /internal/openapi.json

josejulio:
Will be useful for the admin ui


backend/src/main/java/com/redhat/cloud/notifications/oapi/OApiFilter.java
```diff
@@ -0,0 +1,203 @@
...
+
+@ApplicationScoped
+public class OApiFilter {
+
+    public static final String NOTIFICATIONS = "notifications";
+    public static final String INTEGRATIONS = "integrations";
+    public static final String PRIVATE = "private";
+    public static final String INTERNAL = "internal";
+
+    static List<String> whats = List.of(INTEGRATIONS, NOTIFICATIONS, PRIVATE, INTERNAL);
```
bennetelli, line 27: What is "whats"? Is this something domain specific?

backend/src/main/java/com/redhat/cloud/notifications/oapi/OApiFilter.java
```diff
@@ -0,0 +1,203 @@
...
+
+    private WebClient client;
+
+    @PostConstruct
+    void initialize() {
+        this.client = WebClient.create(vertx,
+                new WebClientOptions().setDefaultHost("localhost").setDefaultPort(port));
+    }
+
+    public Uni<String> serveOpenApi(String what) {
```
bennetelli, line 43: same here :)

backend/src/main/java/com/redhat/cloud/notifications/oapi/OApiFilter.java
```diff
@@ -0,0 +1,203 @@
...
+        return client.get("/openapi.json")
+                .send()
+                .onItem()
+                .transform(response ->
+                        filterJson(response.bodyAsJsonObject(), what)
+                )
+                .onItem().transform(JsonObject::encode);
+    }
+
+    private JsonObject filterJson(JsonObject oapiModelJson, String what) {
```
bennetelli, line 57: and here

backend/src/main/java/com/redhat/cloud/notifications/oapi/OApiFilter.java
```diff
@@ -0,0 +1,203 @@
...
+                case "components":
+                case "openapi":
+                    // We just copy all of them even if they may only apply to one
+                    root.put(key, entry.getValue());
+                    break;
+                case "tags":
+                    JsonArray tags = (JsonArray) entry.getValue();
+                    JsonArray filteredTags = new JsonArray(tags
+                            .stream()
+                            .filter(o -> !((JsonObject) o).getString("name").equals(PRIVATE))
```
bennetelli, line 73: could getString() lead to an NPE here?

backend/src/main/java/com/redhat/cloud/notifications/oapi/OApiFilter.java
```diff
@@ -0,0 +1,203 @@
...
+            serversArray.add(createServer(false, INTEGRATIONS));
+        }
+
+        root.put("servers", serversArray);
+
+        return root;
+    }
+
+    private String uppify(String what) {
+        return what.substring(0, 1).toUpperCase() + what.substring(1);
```
bennetelli, line 140: Could it be that the variable "what" contains uppercase chars somewhere later? For example like: joseJulio?
If yes, the resultright now would be: JoseJulio and we probably need to change it to:

`what.substring(0, 1).toUpperCase() + what.substring(1).toLowerCase();`

backend/src/main/java/com/redhat/cloud/notifications/oapi/OApiFilter.java
```diff
@@ -0,0 +1,203 @@
...
+            serversArray.add(createServer(true, INTEGRATIONS));
+            serversArray.add(createServer(false, INTEGRATIONS));
+        }
+
+        root.put("servers", serversArray);
+
+        return root;
+    }
+
+    private String uppify(String what) {
```
bennetelli, line 139: We could rename `uppify` to `uppifyFirstLetter` which would be more clear IMO

backend/src/main/java/com/redhat/cloud/notifications/oapi/OApiFilter.java
```diff
@@ -0,0 +1,203 @@
...
+        }
+        if (in.startsWith(Constants.INTERNAL)) {
+            out = in.substring(Constants.INTERNAL.length());
+        }
+        if (out != null && out.isEmpty()) {
+            out = "/";
+        }
+
+        if (out == null) {
+            System.out.println("Out is null and in is:" + in);
```
bennetelli, line 198: Should we use a logger here?

backend/src/main/java/com/redhat/cloud/notifications/oapi/OApiFilter.java
```diff
@@ -0,0 +1,203 @@
...
+        });
+
+        if (newPathObject.size() == 0) {
+            return null;
+        }
+
+        return newPathObject;
+    }
+
+    private JsonObject createServer(boolean isProd, String what) {
```
bennetelli, line 161: IMO it would be clearer if we split it in two methods, one for prod one for development.

bennetelli:
Right now we don't have any tests for the OApiFilter.

bennetelli:
I just added a OApiFilterTest class and refactored the mangle method a bit afterwards.
Unfortunately I can't fork your fork so I will send you my changes via Slack. That's probably faster.

backend/src/main/java/com/redhat/cloud/notifications/oapi/OApiFilter.java
josejulio: Actually this was something quick to debug :D (Removing)

backend/src/main/java/com/redhat/cloud/notifications/oapi/OApiFilter.java
josejulio: Same as above, will update to a better name

backend/src/main/java/com/redhat/cloud/notifications/oapi/OApiFilter.java
josejulio: as above

backend/src/main/java/com/redhat/cloud/notifications/oapi/OApiFilter.java
josejulio: tags should always have `name` field. I think is OK to fail if that's not true.

https://swagger.io/specification/#tag-object

backend/src/main/java/com/redhat/cloud/notifications/oapi/OApiFilter.java
josejulio: probably `capitalize` would be better :)

backend/src/main/java/com/redhat/cloud/notifications/oapi/OApiFilter.java
josejulio: done

bennetelli (approved):

bennetelli (approved):

PR #448, opened 2021-08-03
Use cache for frontend dependencies

josejulio:
This should start using the cache for the frontend dependencies


bennetelli (approved):
As far as I can tell, it looks good.

PR #458, opened 2021-08-05
NOTIF-190 Enable cron by default

bennetelli:


theute:
Does it still make sense to have that parameter since the cronjob is managed by k8s ?

PR #459, opened 2021-08-05
Fixes issue with recipients query

josejulio:
Look like the `AuthRequestfilter` does not process the `@PostProcess`, maybe related to this one: https://github.com/quarkusio/quarkus/issues/5752

Maybe it works on quarkus 2.0+? Anyway, meanwhile I'm going to use the constructor to load the required params.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/AuthRequestFilter.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/AuthRequestFilter.java
@@ -7,9 +7,6 @@ import org.eclipse.microprofile.config.Config;
 import org.eclipse.microprofile.config.ConfigProvider;
 import org.jboss.logmanager.Level;
 
-import javax.annotation.PostConstruct;
-import javax.enterprise.context.ApplicationScoped;
-import javax.inject.Inject;
 import javax.ws.rs.client.ClientRequestContext;
 import javax.ws.rs.client.ClientRequestFilter;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


bennetelli (commented):
The documentation of javax.ws.rs.client.ClientRequestFilter says: "An extension interface implemented by client request filters. Filters implementing this interface MUST be annotated with @Provider to be discovered by the runtime. This type of filters is supported only as part of the Client API."

Do we need to do this, too? Does Quarkus do this or is this just not necessary anymore?

josejulio:
> The documentation of javax.ws.rs.client.ClientRequestFilter says: "An extension interface implemented by client request filters. Filters implementing this interface MUST be annotated with @Provider to be discovered by the runtime. This type of filters is supported only as part of the Client API."

Maybe that is what I was missing :D

bennetelli:
> > The documentation of javax.ws.rs.client.ClientRequestFilter says: "An extension interface implemented by client request filters. Filters implementing this interface MUST be annotated with @Provider to be discovered by the runtime. This type of filters is supported only as part of the Client API."
> 
> Maybe that is what I was missing :D

Could be :D

bennetelli:
I mean, if you tested it and it's working, than I can also approve it. I don't prefer that one way over the other.

josejulio:
> I mean, if you tested it and it's working, than I can also approve it. I don't prefer that one way over the other.

So I  was looking it at more detail and followed the PR i mention on the first comment. I couldn't see what's wrong, but it works (using the same quarkus version we use) on the sample. 

Anyhow regarding the `Provider` it appears that you have to add it if you want to have a global filter, if not, adding the `RegisterProvider` annotation on the specific client works.

So yes, this is working as we need.

bennetelli (approved):

josejulio:
+1
Lets merge it quickly before @gwenneg sees ;)

bennetelli:
Psssssscht! Not that loud! :P

PR #465, opened 2021-08-06
Add Docker compose file

bennetelli:
TIL: don't use comments as todo list for PRs. They can't be deleted afterwards.


bennetelli:
There is a sql file in src/test/resources already which adds test data and I would like to execute it after the initial flyway migration was successful.
Tried to trick docker with different versions of a healthcheck which returns true, but it doesn't want to accept it.

Does anyone of you have an idea what could work to create a successful healtcheck? @pilhuhn @theute

bennetelli:
I could add a init.sql file in src/test/resources in Quarkus, but I don't like that approach. Having something independent from Quarkus would be better.

helpers/docker-compose.yaml
```diff
@@ -0,0 +1,50 @@
...
+  postgres:
+    image: postgres:12
+    ports:
+      - "5432:5432"
+    environment:
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
+      - POSTGRES_DB=notifications
+      - PGUSER=postgres
+    healthcheck:
```
pilhuhn, line 13: You could just leave the healthcheck out ?

bennetelli: That doesn't work. I am getting "ERROR: for flyway2  Service "flyway" is missing a healthcheck configuration" 
when I remove it. Have to put one healtcheck on a service if I want to depends_on it.

The problem is, that I need "flyway" to be executed before "flyway2" because the first one executes the initial flyway stuff we also execute for every other build. And flyway2 should run after that and insert test data.

theute:
> There is a sql file in src/test/resources already which adds test data and I would like to execute it after the initial flyway migration was successful.
> Tried to trick docker with different versions of a healthcheck which returns true, but it doesn't want to accept it.
> 
> Does anyone of you have an idea what could work to create a successful healtcheck? @pilhuhn @theute

I never really played with docker-compose, I tried a few things but ultimately since v3 it seems that conditions to depend_on aren't really supported anymore:
https://docs.docker.com/compose/startup-order/

Apparently there used to be a convenient `service_completed_successfully` condition...

Problem could be that flyway is doing its thing then exit, so it's never "ready", the container is dead before it can be checked for readyness

bennetelli:
> > There is a sql file in src/test/resources already which adds test data and I would like to execute it after the initial flyway migration was successful.
> > Tried to trick docker with different versions of a healthcheck which returns true, but it doesn't want to accept it.
> > Does anyone of you have an idea what could work to create a successful healtcheck? @pilhuhn @theute
> 
> I never really played with docker-compose, I tried a few things but ultimately since v3 it seems that conditions to depend_on aren't really supported anymore:
> https://docs.docker.com/compose/startup-order/
> 
> Apparently there used to be a convenient `service_completed_successfully` condition...
> 
> Problem could be that flyway is doing its thing then exit, so it's never "ready", the container is dead before it can be checked for readyness

That's my experience as well. That's why I tried to add a bash script which sleeps for some time and exists with successful status then, but this didn't work.

I will leave it as it is right now. It's a bit sad. but I will give it a try later. Even without automated test data, this state is better than the state before.

bennetelli:
If it's okay for you, I would like to merge this PR.

gwenneg:
FYI, Quarkus 1.13 can automatically start a Postgres instance in `dev` and `test` mode, see [this](https://quarkus.io/blog/quarkus-1-13-0-final-released/#zero-config-setup-with-devservices) and [that](https://quarkus.io/guides/datasource#dev-services-configuration-free-databases). Same goes from Kafka and Zookeeper but this is only available starting Quarkus 2.0, see [this](https://quarkus.io/guides/kafka-dev-services).

bennetelli:
@gwenneg thanks for the hint. I will take a look at it :) ps: you should not be here :P

PR #466, opened 2021-08-06
NOTIF-190 Fix exception in cronjob

bennetelli:


aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
```diff
@@ -24,7 +25,7 @@
                 .getResultList();
     }
 
-    public List<EmailAggregation> getEmailAggregation(EmailAggregationKey key, LocalDateTime start, LocalDateTime end) {
+    List<EmailAggregation> getEmailAggregation(EmailAggregationKey key, LocalDateTime start, LocalDateTime end) {
```
bennetelli, line 28: This method is only used in an Integration Test. We should refactor it in a future PR.

aggregator/src/main/resources/application.properties
```diff
@@ -8,7 +8,7 @@ quarkus.http.port=8086
 quarkus.datasource.db-kind=postgresql
 quarkus.datasource.username=postgres
 quarkus.datasource.password=postgres
-quarkus.datasource.jdbc.url=jdbc:postgresql://127.0.0.1:5432/notifications_backend
+quarkus.datasource.jdbc.url=jdbc:postgresql://127.0.0.1:5432/notifications
```
theute, line 11: Shouldn't this remain as before ?

pilhuhn: It should match https://github.com/RedHatInsights/notifications-backend/blob/master/backend/src/main/resources/application.properties#L44 in my understanding, so I think that change is good.

bennetelli: I would say it should be the same as in backend, but with non-reactive: 
quarkus.datasource.reactive.url=postgresql://127.0.0.1:5432/notifications

theute: Yes, sorry, it just needs to match.
I got confused with the name [we use in OpenShift environments](https://github.com/RedHatInsights/e2e-deploy/blob/master/templates/notifications/notifications-aggregator.yml#L69)
But as long as they match in their respective environments, we're good

PR #468, opened 2021-08-09
NOTIF-190 add log for kafka messages

bennetelli:


backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -132,6 +132,7 @@
         AggregationCommand aggregationCommand;
         try {
             aggregationCommand = objectMapper.readValue(aggregationCommandJson, AggregationCommand.class);
+            log.info("Receiving payload from kafka topic: " + aggregationCommand);
```
josejulio, line 135: We are going to possibly get a lot of commands in here (is one event per account/app), maybe this should be a debug log that we can turn on/off?

bennetelli: I removed the log message.

We should discuss which logger we are using when Gwenneg is back. The one I used here doesn't have debug, but warning. In other services we are using either the JUL oder JBoss Logging. I think it's good to decide which to take and use this in every service.

josejulio (approved):

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
gwenneg: I'm in favor of `org.jboss.logging.Logger` because it is much more flexible than JUL. That's also the logger [used internally in Quarkus](https://quarkus.io/guides/logging#supported-logging-apis).

You can log at `FINE` level with JUL which is then transformed to `DEBUG` by Quarkus, see [this](https://quarkus.io/guides/logging#logging-levels) for more information about logging levels.

PR #474, opened 2021-08-11
NOTIF-190 Add last cronjob run

bennetelli:


backend/src/main/resources/db/migration/V1.22.0__NOTIF-190_cronjob_timestamp.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE cronjob_run (
+    id UUID NOT NULL,
+    last_run TIMESTAMP
+) WITH (OIDS=FALSE);
+
+INSERT INTO cronjob_run (id, last_run) VALUES
+('3db0955b-751b-48cd-b531-c1d81596d133', NOW());
```
bennetelli, line 7: NOW() is probably wrong. Does anyone have a preference what I should insert here?

bennetelli:
> I think we should probably use other name for the table, `cron_job` is too generic, we might have more `cron_jobs`.
> Some other thing, is that the real aggregation doesn't happen in the aggregator code. So we might want to move this code to the backend side and probably store the account / bundle / app where this happened?

Regarding the table name. Usually our table names are in singular, not in plural. A table "person" always stores multiple persons, but from a modeling and OR Mapping perspective we would also have a class like 

Car for example which stores a List<Person> instead of a List<Persons>. I would say cron_job is better than cron_jobs here.

josejulio:
> Regarding the table name. Usually our table names are in singular, not in plural. A table "person" always stores multiple persons, but from a modeling and OR Mapping perspective we would also have a class like Car for example which stores a List of entity Person instead of a List of entity PersonS. I would say cron_job is better than cron_jobs here.

Ah sorry, I wasn't suggestion `cron_jobs` as the table name, i was thinking that `cron_job` is too generic, and we might have something like `failed_aggregation` table or so, with the actual failed things that we need to repeat

bennetelli:
> I think we should probably use other name for the table, `cron_job` is too generic, we might have more `cron_jobs`.
> Some other thing, is that the real aggregation doesn't happen in the aggregator code. So we might want to move this code to the backend side and probably store the account / bundle / app where this happened?

Regarding "moving the code". I would say it's ok where it is. The aggregator is the cron job and from a responsibility point of view the cron job should know when its last run was. Not other services.

josejulio:
> Regarding "moving the code". I would say it's ok where it is. The aggregator is the cron job and from a responsibility point of view the cron job should know when its last run was. Not other services.

The point I want to convey here is that there isn't a lot of things that could fail in the aggregator code - It currently queries what needs to be aggregated and sends it to the backend (though kafka).

I'm suggesting that the backend should report what failed, so the aggregator can schedule a repeat of some aggregation.

bennetelli:
> > Regarding "moving the code". I would say it's ok where it is. The aggregator is the cron job and from a responsibility point of view the cron job should know when its last run was. Not other services.
> 
> Ah, I'm thinking that instead of writing when a cron was executed, check when an aggregation failed (and it's related params) to help retry in the future.

That's under control of Kubernetes/OpenShift. It automatically retries in case of a failure. 
With the previous solution with @Scheduled it would be necessary to do what you described. But with Kubernetes/OpenShift managing the cronjob, these things shouldn't be managed within the application when the platform is holding all the informations about this. 

Or am I wrong?

aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
```diff
@@ -59,11 +65,13 @@ public void processDailyEmail(Instant scheduledFireTime) {
...
     }
 
-    List<AggregationCommand> processAggregateEmails(Instant scheduledFireTime) {
-        Instant yesterdayScheduledFireTime = scheduledFireTime.minus(DAILY.getDuration());
+    List<AggregationCommand> processAggregateEmails() {
+        Instant scheduledFireTime = emailAggregationResources.getLastCronJobRun().getLastRun();
+        Instant yesterdayScheduledFireTime = scheduledFireTime.minus(EmailSubscriptionType.DAILY.getDuration());
 
         LocalDateTime endTime = LocalDateTime.ofInstant(scheduledFireTime, UTC);
         LocalDateTime startTime = LocalDateTime.ofInstant(yesterdayScheduledFireTime, UTC);
```
gwenneg, line 73: These conversions could be avoided if `CronJobRun#lastRun` was a `LocalDateTime` instead of an `Instant`.

aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
```diff
@@ -59,11 +65,13 @@ public void processDailyEmail(Instant scheduledFireTime) {
         }
     }
 
-    List<AggregationCommand> processAggregateEmails(Instant scheduledFireTime) {
-        Instant yesterdayScheduledFireTime = scheduledFireTime.minus(DAILY.getDuration());
+    List<AggregationCommand> processAggregateEmails() {
+        Instant scheduledFireTime = emailAggregationResources.getLastCronJobRun().getLastRun();
+        Instant yesterdayScheduledFireTime = scheduledFireTime.minus(EmailSubscriptionType.DAILY.getDuration());
```
gwenneg, line 70: Shouldn't `scheduledFireTime` be `Instant.now()` (the same object than the one persisted after this method is executed) and `yesterdayScheduledFireTime` be `emailAggregationResources.getLastCronJobRun().getLastRun()`?

gwenneg (changes requested):

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
```diff
@@ -35,4 +38,31 @@
                 .setParameter("end", end)
                 .getResultList();
     }
+
+    @Transactional
+    public Integer purgeOldAggregation(EmailAggregationKey key, LocalDateTime lastUsedTime) {
```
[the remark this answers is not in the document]
josejulio: I think it can be safely removed, we don't purge aggregations in the aggregator

gwenneg:
I can't comment that part because it's too far from this PR changes in the code, but why does `DailyEmailAggregationJob#processDailyEmail` contain two `emitter#send` calls in a row?

https://github.com/RedHatInsights/notifications-backend/blob/602c34d37ff44e0f9b8d838a2c6b5c2c3f711b98/aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java#L47-L48

josejulio:
> I can't comment that part because it's too far from this PR changes in the code, but why does `DailyEmailAggregationJob#processDailyEmail` contain two `emitter#send` calls in a row?

Yikes, that looks like a bug :bug:

gwenneg:
`CompletableFuture` is eager (unlike `Uni` which is lazy) so each aggregation command should be sent twice with the current version.

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
```diff
@@ -35,4 +38,31 @@
...
+    @Transactional
+    public CronJobRun getLastCronJobRun() {
+        String query = "SELECT id, last_run FROM cronjob_run";
+        return session.createNativeQuery(query, CronJobRun.class).getSingleResult();
+    }
+
+    public void updateLastCronJobRun(UUID id, Instant lastRun) {
+        String query = "UPDATE cronjob_run SET last_run = :lastRun WHERE id = :id";
+        session.createNativeQuery(query, CronJobRun.class)
+                .setParameter("last_run", lastRun)
```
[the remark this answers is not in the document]
bennetelli: isn't the native one more save as it's a prepared statement?

aggregator/src/test/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJobTest.java
```diff
@@ -4,17 +4,15 @@
...
-import com.redhat.cloud.notifications.models.EmailSubscriptionType;
 import io.quarkus.test.common.QuarkusTestResource;
 import io.quarkus.test.junit.QuarkusTest;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
 import javax.inject.Inject;
-import java.time.Duration;
-import java.time.Instant;
 import java.util.List;
```
gwenneg, line 14: This class needs to test the messages emitted through SmallRye Reactive Messaging using `InMemoryConnector`.

gwenneg (changes requested):

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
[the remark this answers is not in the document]
gwenneg: All Hibernate queries (native or not) use prepare statements under the hood.

aggregator/src/test/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJobTest.java
bennetelli: I changed every of your findings besides the missing InMemoryConnector test. I will add it tomorrow morning.

aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
```diff
@@ -38,19 +38,22 @@
     @Channel("aggregation")
     Emitter<String> emitter;
 
-    public void processDailyEmail(Instant scheduledFireTime) {
-        List<AggregationCommand> aggregationCommands = processAggregateEmails(scheduledFireTime);
+    public void processDailyEmail() {
+        List<AggregationCommand> aggregationCommands = processAggregateEmails();
```
gwenneg, line 42: This method and the one persisting the `LocalDateTime` into the database must use the same instance:
```suggestion
        LocalDateTime now = LocalDateTime.now(ZoneOffset.UTC);
        List<AggregationCommand> aggregationCommands = processAggregateEmails(now);
```

aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
```diff
@@ -38,19 +38,22 @@
...
                 final String payload = objectMapper.writeValueAsString(aggregationCommand);
-                emitter.send(payload);
                 futures.add(emitter.send(payload).toCompletableFuture());
             } catch (JsonProcessingException e) {
                 LOG.warning("Could not transform AggregationCommand to JSON object.");
             }
         }
 
+        final CronJobRun lastCronJobRun = emailAggregationResources.getLastCronJobRun();
+        emailAggregationResources.updateLastCronJobRun(lastCronJobRun.getId(), Instant.now());
```
gwenneg, line 55: Related to my previous comment:
```suggestion
        emailAggregationResources.updateLastCronJobRun(lastCronJobRun.getId(), now);
```

aggregator/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJob.java
```diff
@@ -59,11 +62,10 @@ public void processDailyEmail(Instant scheduledFireTime) {
         }
     }
 
-    List<AggregationCommand> processAggregateEmails(Instant scheduledFireTime) {
-        Instant yesterdayScheduledFireTime = scheduledFireTime.minus(DAILY.getDuration());
+    List<AggregationCommand> processAggregateEmails() {
+        LocalDateTime endTime = LocalDateTime.now();
```
gwenneg, line 66: Related to my previous comment:
```suggestion
    List<AggregationCommand> processAggregateEmails(LocalDateTime endTime) {
```

aggregator/src/test/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJobTest.java
gwenneg: There are a few `hidden conversations` above with comments/questions. 😃

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
```diff
@@ -35,4 +38,17 @@
...
+    public CronJobRun getLastCronJobRun() {
+        String query = "SELECT id, last_run FROM cronjob_run";
+        return session.createNativeQuery(query, CronJobRun.class).getSingleResult();
+    }
+
+    public void updateLastCronJobRun(UUID id, Instant lastRun) {
+        String query = "UPDATE CronJobRun SET lastRun = :lastRun WHERE id = :id";
+        session.createQuery(query)
+                .setParameter("id", id)
+                .setParameter("lastRun", lastRun);
```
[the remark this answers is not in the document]
bennetelli: f*** looks like I need vacation.

backend/src/main/resources/db/migration/V1.23.0__NOTIF-190_cronjob_timestamp.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE cronjob_run (
+    id UUID NOT NULL,
+    last_run TIMESTAMP
+) WITH (OIDS=FALSE);
+
+INSERT INTO cronjob_run (id, last_run) VALUES
+('3db0955b-751b-48cd-b531-c1d81596d133', '-infinity');
```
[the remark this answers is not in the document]
gwenneg: If the UUID purpose is to prevent multiple rows in the `cronjob_run` table, we already have something similar based on Postgres constraints with the application status. See [V1.20.0__NOTIF-226_maintenance_mode.sql](https://github.com/RedHatInsights/notifications-backend/blob/master/backend/src/main/resources/db/migration/V1.20.0__NOTIF-226_maintenance_mode.sql) and [CurrentStatus.java](https://github.com/RedHatInsights/notifications-backend/blob/master/backend/src/main/java/com/redhat/cloud/notifications/models/CurrentStatus.java#L34).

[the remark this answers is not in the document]
bennetelli: the UUID is the Generated PK of the CronJob class.
I would say it doesn't need to exist somewhere as we are initializing it in the flyway script,
then we always read the entry: 

        final CronJobRun lastCronJobRun = emailAggregationResources.getLastCronJobRun();
        emailAggregationResources.updateLastCronJobRun(lastCronJobRun.getId(), now);

and always update the same entry over and over again.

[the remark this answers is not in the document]
bennetelli: -Infinity is the oldest possible datetime entry in Postgres: https://stackoverflow.com/questions/19686635/earliest-timestamp-supported-in-postgresql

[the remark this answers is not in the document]
bennetelli: The latest change from Instant to LocalDate has broken the code completely.

I will make it a draft now and work on it when I am back from vacation.

aggregator/src/test/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJobTest.java
bennetelli: I didn't hide these. Thought someone of you did it and don't want to talk about it anymore.

bennetelli:
Could it be that GitHub is hiding comments when the respective lines have changed?

bennetelli:
@gwenneg I just tested it locally by starting the cron job multiple times and comparing the start/end intervals.
That's working as expected. It's 12:30 and I should be in vacation already.. will commit my latest changes and write you via Slack.

gwenneg:
> Could it be that GitHub is hiding comments when the respective lines have changed?

GH hides somes parts of PRs/issues when they get too long. There are 5 hidden comments above but depending on how you read this, maybe they're not shown.

bennetelli:
> > Regarding the table name. Usually our table names are in singular, not in plural. A table "person" always stores multiple persons, but from a modeling and OR Mapping perspective we would also have a class like Car for example which stores a List of entity Person instead of a List of entity PersonS. I would say cron_job is better than cron_jobs here.
> 
> Ah sorry, I wasn't suggestion `cron_jobs` as the table name, i was thinking that `cron_job` is too generic, and we might have something like `failed_aggregation` table or so, with the actual failed things that we need to repeat

@gwenneg right now the name of the table is "cronjob_run". What's your preference? Is it okay or do you want a different name?

aggregator/src/main/java/com/redhat/cloud/notifications/models/CronJobRun.java
```diff
@@ -0,0 +1,48 @@
...
+    @GeneratedValue
+    @JsonProperty(access = READ_ONLY)
+    private UUID id;
+
+    @NotNull
+    private Instant lastRun;
+
+    public CronJobRun() {
+    }
+
```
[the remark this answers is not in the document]
bennetelli: I thought JPA would need a default constructor? Or am I oldschool already? :D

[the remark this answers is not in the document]
gwenneg: It does need a default constructor, but there's no need to write when there's no other constructor in the class. Java will add the default one automatically during build.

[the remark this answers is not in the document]
bennetelli: I thought back in the JPA 2 days it was necessary to declare that constructor always. As well as making the class serializable. But could be that this changed.

I know that Java does it under the hood. My question was more related to the JPA framework which (at least in the past) needed this constructor and serializable for the annotation scanning magic and to be consistent with the JSR

aggregator/src/test/java/com/redhat/cloud/notifications/processors/email/aggregators/DailyEmailAggregationJobTest.java
```diff
@@ -28,19 +34,52 @@
...
+        when(cronJobRun.getLastRun()).thenReturn(LocalDateTime.now().minusYears(1337));
+
+        final List<EmailAggregationKey> aggregationCommands = new LinkedList<>();
+        aggregationCommands.add(new EmailAggregationKey("tenant", "rhel", "policies"));
+        when(emailAggregationResources.getApplicationsWithPendingAggregation(any(), any())).thenReturn(aggregationCommands);
+
+        final ObjectMapper objectMapper1 = mock(ObjectMapper.class);
+        when(objectMapper1.writeValueAsString(anyString())).thenReturn("");
+
+//        verify(emitter).send(anyString());
```
bennetelli, line 76: here.

aggregator/src/test/java/com/redhat/cloud/notifications/DailyEmailAggregationJobTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+        final String secondAggregation = results.received().get(1).getPayload();
+        assertTrue(secondAggregation.contains("accountId\":\"tenant"));
+        assertTrue(secondAggregation.contains("bundle\":\"rhel"));
+        assertTrue(secondAggregation.contains("application\":\"unknown-application"));
+        assertTrue(secondAggregation.contains("subscriptionType\":\"DAILY"));
+    }
+
+    @Test
+    void shouldNotChangeSomethingWhenCreatingSubscription() {
```
gwenneg, line 79: Why does the `src/test` folder still contain code related to `EmailSubscription` while there isn't a trace of `EmailSubscription` or `endpoint_email_subscriptions` in `src/main`? It looks like all code related to `EmailSubscription` should be entirely removed from this module, but maybe I missed some place in `src/main` where it is actually needed.

aggregator/pom.xml
```diff
@@ -80,6 +81,23 @@
...
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5-mockito</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit-pioneer</groupId>
+            <artifactId>junit-pioneer</artifactId>
+            <version>1.4.2</version>
+            <scope>test</scope>
+        </dependency>
```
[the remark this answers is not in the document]
bennetelli: Good catch. With Spring Boot this was a bit tricky and I somehow didn't think about trying it with Quarkus.
You are right. It's actually not necessary.

aggregator/pom.xml
```diff
@@ -80,6 +81,23 @@
...
+        <dependency>
+            <groupId>io.smallrye.reactive</groupId>
+            <artifactId>smallrye-reactive-messaging-in-memory</artifactId>
+            <version>2.0.3</version>
```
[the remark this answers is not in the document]
bennetelli: removed already.

aggregator/src/test/java/com/redhat/cloud/notifications/DailyEmailAggregationJobTest.java
```diff
@@ -0,0 +1,178 @@
...
+    void setUp() {
+        helpers.purgeAggregations();
+        helpers.purgeEmailSubscriptions();
+    }
+
+    @AfterEach
+    void tearDown() {
+        helpers.purgeAggregations();
+        helpers.purgeEmailSubscriptions();
+    }
```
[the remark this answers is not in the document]
bennetelli: oh nice! I didn't know this annotation exist. It seems to work :)

aggregator/src/test/java/com/redhat/cloud/notifications/db/ResourceHelpers.java
```diff
@@ -21,15 +20,12 @@
     @Inject
     Session session;
 
-    @Inject
-    EmailAggregationResources emailAggregationResources;
-
-    public void createSubscription(String tenant, String username, String bundle, String application, EmailSubscriptionType type) {
-        subscribe(tenant, username, bundle, application, type);
+    public void createDailySubscription(String tenant, String username, String bundle, String application) {
```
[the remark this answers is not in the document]
bennetelli: done

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
```diff
@@ -35,4 +37,19 @@
                 .setParameter("end", end)
                 .getResultList();
     }
+
+    @Transactional
```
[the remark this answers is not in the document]
bennetelli: done

aggregator/src/test/java/com/redhat/cloud/notifications/db/ResourceHelpers.java
```diff
@@ -21,15 +20,12 @@
...
+        subscribe(tenant, username, bundle, application);
     }
 
-    public void removeSubscription(String tenant, String username, String bundle, String application, EmailSubscriptionType type) {
-        unsubscribe(tenant, username, bundle, application, type);
+    public void removeDailySubscription(String tenant, String username, String bundle, String application) {
```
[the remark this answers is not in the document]
bennetelli: done

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -82,11 +89,10 @@ public void processDailyEmail(Instant scheduledFireTime) {
                         startTime,
                         endTime,
                         DAILY,
-                        SECONDS.between(aggregateStarted, LocalDateTime.now()),
+                        SECONDS.between(endTime, LocalDateTime.now()),
```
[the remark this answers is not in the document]
bennetelli: fixed

aggregator/src/test/java/com/redhat/cloud/notifications/DailyEmailAggregationJobTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+    @Test
+    @SetSystemProperty(key = "notifications.aggregator.email.subscription.periodic.cron.enabled", value = "true")
+    void shouldSentTwoAggregationsToKafkaTopic() {
+        helpers.addEmailAggregation("tenant", "rhel", "policies", "somePolicyId", "someHostId");
+        helpers.addEmailAggregation("tenant", "rhel", "unknown-application", "somePolicyId", "someHostId");
+
+        testee.processDailyEmail();
+
+        InMemorySink<String> results = connector.sink("aggregation");
```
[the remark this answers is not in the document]
bennetelli: done

aggregator/src/test/java/com/redhat/cloud/notifications/DailyEmailAggregationJobTest.java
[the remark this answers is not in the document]
gwenneg: I wish the same thing existed for Hibernate Reactive. I asked the Quarkus team for it but it's not ready yet 😢

aggregator/src/test/java/com/redhat/cloud/notifications/DailyEmailAggregationJobTest.java
bennetelli: Right now it's needed for the tests only. For creating test data and so on.

gwenneg: Yes I saw that, but why would we need that data if it's never used at run time?

gwenneg: Actually, the entire `com.redhat.cloud.notifications.models` package (as well as all related code) from `src/test` seems removable.

bennetelli: That's what I did with the latest 2-3 commits :)

bennetelli: the test I copied from backend was a bit misleading as it mixed subscription and aggregation features withint a large test and I got confused by it.

backend/src/main/resources/db/migration/V1.23.0__NOTIF-190_cronjob_timestamp.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE cronjob_run (
+    id UUID NOT NULL,
+    last_run TIMESTAMP,
+    CONSTRAINT pk_cronjob_run PRIMARY KEY (id)
+);
+
+INSERT INTO cronjob_run (id, last_run) VALUES
+('3db0955b-751b-48cd-b531-c1d81596d133', '08/20/1900 09:34:19.503344+0');
```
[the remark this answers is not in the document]
gwenneg: Here's how this script could look like with what I suggested:
```suggestion
CREATE TABLE cronjob_run (
    last_run TIMESTAMP,
    -- The following PK and CHECK constraints combination guarantees that the table will never contain more than one row.
    prevent_multiple_rows BOOLEAN PRIMARY KEY DEFAULT TRUE,
    CONSTRAINT status_table_should_never_contain_multiple_rows CHECK (prevent_multiple_rows)
);

INSERT INTO cronjob_run (value) VALUES ('08/20/1900 09:34:19.503344+0');
```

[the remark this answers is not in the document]
gwenneg: `CronJobRun` would need this attribute:

```
    /*
     * We don't need this field in the Java code but Hibernate needs each @Entity to have a field annotated with @Id.
     * This field has a real purpose in the database though: it is used to guarantee that the status table will never
     * contain more than one row. This is done with a combination of PK and CHECK SQL constraints.
     */
    @Id
    @JsonIgnore
    private boolean preventMultipleRows = true;
```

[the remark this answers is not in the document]
bennetelli: whereas 
INSERT INTO cronjob_run (value) VALUES ('08/20/1900 09:34:19.503344+0');

would need to be:

INSERT INTO cronjob_run (last_run) VALUES ('08/20/1900 09:34:19.503344+0');

I would say.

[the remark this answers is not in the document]
gwenneg: Yes, I copied that from `backend` and forgot a few details sorry 😄

backend/src/main/resources/db/migration/V1.28.0__NOTIF-190_cronjob_timestamp.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE cronjob_run (
+    last_run TIMESTAMP,
```
gwenneg, line 2: Sorry, I forgot this earlier 😄
```suggestion
    last_run TIMESTAMP NOT NULL,
```

aggregator/src/test/resources/db/migration/V1.0.0__aggregator.sql
```diff
@@ -39,17 +17,11 @@ CREATE TABLE email_aggregation (
...
 
-CREATE TABLE endpoint_email_subscriptions (
-    account_id VARCHAR(50) NOT NULL,
-    user_id VARCHAR(50) NOT NULL,
-    subscription_type VARCHAR(50) NOT NULL,
-    application_id UUID NOT NULL,
-    CONSTRAINT pk_endpoint_email_subscriptions PRIMARY KEY (account_id, user_id, subscription_type, application_id),
-    CONSTRAINT fk_endpoint_email_subscriptions_application_id FOREIGN KEY (application_id) REFERENCES applications(id)
+CREATE TABLE cronjob_run (
+    last_run TIMESTAMP,
```
gwenneg, line 21:
```suggestion
    last_run TIMESTAMP NOT NULL,
```

aggregator/src/main/java/com/redhat/cloud/notifications/models/CronJobRun.java
```diff
@@ -0,0 +1,33 @@
...
+@Table(name = "cronjob_run")
+public class CronJobRun {
+
+    /*
+     * We don't need this field in the Java code but Hibernate needs each @Entity to have a field annotated with @Id.
+     * This field has a real purpose in the database though: it is used to guarantee that the status table will never
+     * contain more than one row. This is done with a combination of PK and CHECK SQL constraints.
+     */
+    @Id
+    @JsonIgnore
```
gwenneg, line 20:
```suggestion
```

aggregator/src/test/java/com/redhat/cloud/notifications/helpers/ResourceHelpers.java
```diff
@@ -0,0 +1,63 @@
...
+    public void purgeEmailAggregations() {
+        session.createQuery("DELETE FROM EmailAggregation").executeUpdate();
+        session.flush();
+    }
+
+    @Transactional
+    public void purgeAggregations() {
+        session.createQuery("DELETE FROM EmailAggregation").executeUpdate();
+        session.flush();
+    }
```
gwenneg, line 49: Duplicate spotted!

aggregator/src/test/java/com/redhat/cloud/notifications/DailyEmailAggregationJobTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+    @BeforeEach
+    void setUp() {
+        helpers.purgeEmailAggregations();
+    }
+
+    @Test
+    @TestTransaction
+    void shouldSentTwoAggregationsToKafkaTopic() {
+        System.setProperty("notifications.aggregator.email.subscription.periodic.cron.enabled", "true");
```
gwenneg, line 50: There's no guarantee tests will always be run in the same order from one JUnit version to another. It would be safer to use this in **all** tests (not only the first one) that require the cron job enabled:

```
System.setProperty("notifications.aggregator.email.subscription.periodic.cron.enabled", "true");
// test content
System.clearProperty("notifications.aggregator.email.subscription.periodic.cron.enabled");
```

aggregator/src/test/java/com/redhat/cloud/notifications/DailyEmailAggregationJobTest.java
```diff
@@ -0,0 +1,157 @@
...
+        DailyEmailAggregationJob dailyEmailAggregationJob = new DailyEmailAggregationJob(emailAggregationResources, objectMapper);
+
+        dailyEmailAggregationJob.processDailyEmail();
+        verifyNoInteractions(emailAggregationResources);
+    }
+
+    @Test
+    @TestTransaction
+    void shouldNotProcessMailsWhenNewCronJobIsDisabledByEnvironmentVariable() {
+        System.setProperty("notifications.aggregator.email.subscription.periodic.cron.enabled", "false");
```
gwenneg, line 140:
```suggestion
```

gwenneg (commented):
[This old comment](https://github.com/RedHatInsights/notifications-backend/pull/474#discussion_r701271991) (which is hidden by GitHub on my machine) requires your attention 😃

aggregator/src/test/java/com/redhat/cloud/notifications/helpers/ResourceHelpers.java
bennetelli: removed

aggregator/src/test/java/com/redhat/cloud/notifications/DailyEmailAggregationJobTest.java
bennetelli: i am clearing system properties before and after each test execution now.

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -27,31 +27,41 @@
...
+        LocalDateTime now = LocalDateTime.now(UTC);
+        List<AggregationCommand> aggregationCommands = processAggregateEmails(now);
+
+        List<CompletableFuture<Void>> futures = new ArrayList<>();
+        for (AggregationCommand aggregationCommand : aggregationCommands) {
+            try {
+                final String payload = objectMapper.writeValueAsString(aggregationCommand);
+                futures.add(emitter.send(payload).toCompletableFuture());
+            } catch (JsonProcessingException e) {
+                LOG.warning("Could not transform AggregationCommand to JSON object.");
```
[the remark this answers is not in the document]
gwenneg: > [This old comment](https://github.com/RedHatInsights/notifications-backend/pull/474#discussion_r701271991) (which is hidden by GitHub on my machine) requires your attention 😃

This suggestion won't work out of the box because of JUL but we need to log the exception:
```suggestion
                LOG.warning("Could not transform AggregationCommand to JSON object.", e);
```

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -20,38 +20,48 @@
...
```
gwenneg, line 59: > [This old comment](https://github.com/RedHatInsights/notifications-backend/pull/474#discussion_r701271991) (which is hidden by GitHub on my machine) requires your attention 😃

It's been burried by GitHub above so here's a copy of it:

This suggestion won't work out of the box because of JUL but we need to log the exception:
```suggestion
                LOG.warning("Could not transform AggregationCommand to JSON object.", e);
```

bennetelli:
I am not able to find anything here :(

gwenneg (approved):
LGTM

PR #487, opened 2021-08-16
Replace BaseTransformer with default method

bennetelli:


josejulio:
Maybe i missed the discussion, but you could add information on why are you refactoring it? thanks! :smile:

bennetelli:
@josejulio I just found that piece of code and for me it didn't make much sense. The methods of BaseTransformer are only called in the Processor classes so they somehow "belong" to the processor and with default methods we have a good feature to put the code in the existing interface.

gwenneg:
This kind of looks like a composition vs inheritance discussion. Composition is usually easier to mock/test/maintain.

bennetelli:
then do whatever you want with this PR. It's just an offer. If you want it, merge it. If not, close this PR.

PR #493, opened 2021-08-17
Clean Docker files

gwenneg:
We don't need the deleted files. They were generated by Quarkus and can be very easily recreated later if needed.


docker/Dockerfile-build-aggregator.jvm
```diff
@@ -7,7 +7,7 @@ FROM registry.access.redhat.com/ubi8/openjdk-11:latest AS build
 USER root
 COPY . /home/jboss
 WORKDIR /home/jboss
-RUN mvn clean package -DskipTests -Dcheckstyle.skip=true --projects :checkstyle,:notifications-parent,:notifications-aggregator
+RUN mvn clean package --projects :checkstyle,:notifications-aggregator
```
gwenneg, line 10: It would be even better if we could use `./mvnw` but I'm not sure it would work from a Docker build.

gwenneg:
I'll test the wrapper after the current build is done on CI.

docker/Dockerfile-build-aggregator.jvm
gwenneg: It works \o/

PR #494, opened 2021-08-17
Improve build.yml

gwenneg:


.github/workflows/build.yml
```diff
@@ -1,33 +1,37 @@
...
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Prepare cache key
+        id: cache-key
+        shell: bash
+        run: echo "::set-output name=date::$(/bin/date -u "+%Y-%m")"
       - name: Cache Maven dependencies
-        uses: actions/cache@v2.1.6
```
gwenneg, line 13: This reduces the number of dependabot PRs because the latest version will automatically be used.

.github/workflows/build.yml
```diff
@@ -1,33 +1,37 @@
...
+        uses: actions/cache@v2
         with:
           path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
-          restore-keys: |
-            ${{ runner.os }}-maven-
-      - name: Cache frontend dependencies
-        uses: actions/cache@v2.1.6
+          # The cache key is changed every month to prevent unlimited growth.
+          key: maven-cache-${{ steps.cache-key.outputs.date }}
```
gwenneg, line 21: The cache we used before was never cleaned and kept growing build after build (there's a 5 GB limit for each cache though). Now it will be cleaned each month and the planet will be a bit happier 😄

.github/workflows/build.yml
```diff
@@ -1,33 +1,37 @@
 name: Build & Test
 
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
```
gwenneg, line 3: With `workflow_dispatch`, we can trigger the action from a button in the GitHub UI for any branch.

.github/workflows/build.yml
bennetelli: not only the planet :D

bennetelli (approved):

.github/workflows/build.yml
gwenneg: image

bennetelli: This is great!

PR #498, opened 2021-08-18
Fix admin console

josejulio:
- Generated service code uses . as its base path
 - Uses `@Schema` for snake_case properties

I was wrong about the snake_case thing :D.


gwenneg (approved):
LGTM

PR #508, opened 2021-08-23
NOTIF-291 Introduce Event - Phase 1

gwenneg:


backend/src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
```diff
@@ -83,13 +87,14 @@
                 .getSingleResultOrNull();
     }
 
+    // Note: This method uses a stateless session
     public Uni<EventType> getEventType(String bundleName, String applicationName, String eventTypeName) {
```
gwenneg, line 91: FYI this method was unused before this PR.

backend/src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java
```diff
@@ -412,8 +412,8 @@ private void pushMessage(int expectedWebhookCalls) throws IOException, Interrupt
             mockServerConfig.getMockServerClient().clear(expectedRequestPattern);
         }
 
+        counterAssertionHelper.awaitAndAssertIncrement(PROCESSED_MESSAGES_COUNTER_NAME, 1);
```
gwenneg, line 415: That change was necessary because of a race condition in tests caused by the new SQL queries executed by `EventConsumer` (retrieve event type and create event). Without the `await` logic, the check was run before the Kafka message had been processed and failed.

backend/src/test/java/com/redhat/cloud/notifications/routers/TestEmailEndpointMigrationService.java
```diff
@@ -38,6 +39,7 @@
     Mutiny.Session session;
 
     @Test
+    @Disabled
```
gwenneg, line 42: `EmailEndpointMigrationService` and `TestEmailEndpointMigrationService` must be removed before this PR is merged.

backend/src/main/resources/db/migration/V1.23.0__NOTIF-291_events.sql
```diff
@@ -0,0 +1,14 @@
...
+   account_id VARCHAR(50) NOT NULL,
+   event_type_id UUID,
+   payload TEXT,
+   created TIMESTAMP NOT NULL,
+   CONSTRAINT pk_event PRIMARY KEY (id),
+   CONSTRAINT fk_event_event_type_id FOREIGN KEY (event_type_id) REFERENCES event_type (id) ON DELETE CASCADE
+);
+
+ALTER TABLE notification_history
+    DROP COLUMN event_id,
```
gwenneg, line 12: This `DROP` is safe on prod. I verified the database content and the `event_id` field always contains an empty String.

gwenneg:
@josejulio Feel free to start reviewing this PR even if it's a draft. The code should be pretty stable now.

I only need to write a data migration API and the corresponding test before it is officially "ready for review".

backend/src/main/java/com/redhat/cloud/notifications/routers/EventLogMigrationService.java
```diff
@@ -0,0 +1,84 @@
...
+        // All DB changes will be rolled-back if anything goes wrong.
+        return session.withTransaction(tx -> {
+            LOGGER.debug("[NOTIF-291] Start");
+            return session.createQuery("FROM NotificationHistory WHERE event IS NULL", NotificationHistory.class)
+                    .getResultList()
+                    .onItem().transformToMulti(Multi.createFrom()::iterable)
+                    .onItem().transformToUniAndConcatenate(history -> {
+                        Event event = new Event();
+                        event.setAccountId(history.getAccountId());
+                        return session.persist(event)
```
josejulio, line 51: mmh, IIUC this will create one `Event` per each `NotificationHistory`? 

AFAIK one Event can cause multiple `NotificationHistory` (i.e. one event triggering an email and a 2 webhook should create 3 `NotificationHistory`) - but I don't think we had any way to identify `NotificationHistory` create from the same event.

If that's the case, I'm wondering if we should do this migration.

backend/src/test/java/com/redhat/cloud/notifications/routers/EventLogMigrationServiceTest.java
```diff
@@ -0,0 +1,138 @@
...
+        history.setInvocationResult(true);
+        session.persist(history)
+                .onItem().call(session::flush)
+                .await().indefinitely();
+        return history.getId();
+    }
+
+    private EventLogMigrationService.MigrationReport migrate() {
+        return given()
+                .queryParam("confirmation-token", EmailEndpointMigrationService.CONFIRMATION_TOKEN)
```
josejulio, line 108:
```suggestion
                .queryParam("confirmation-token", EventLogMigrationServiceTest.CONFIRMATION_TOKEN)
```

gwenneg: Oops 😄

backend/src/main/java/com/redhat/cloud/notifications/routers/EventLogMigrationService.java
gwenneg: You understood correctly, but here's the reason why we need that migration.

After this PR, `NotificationResources#getNotificationHistory` and `NotificationResources#getNotificationDetails` will need a `NotificationHistory <-> Event` relationship to return a non-empty/null result. If we don't migrate the current prod data, the _old_ history will disappear from the UI after we deploy this branch on prod. The migration will prevent that issue by adding fake `Events` but as you sensed it wouldn't make sense to display these `Events` in the UI because they won't reflect what really happened. That's why `Events` won't be displayed for 14 days (the retention delay) in the UI.

Here's the plan I had in mind:
- We deploy this branch to prod ASAP. Fake `Events` are added to the current history to keep everything running as usual with the old UI.
- For 14 days, the history is filled with real `Events` linked to the new `NotificationHistory` entries and the old history data with fake `Events` is slowly removed by a scheduled job that I will submit as a new PR tomorrow.
- After 14 days, the history only contains real `Events` and that's the moment when we can display the new event log UI.

This is very similar to what we did with policies-history.

WDYT?

gwenneg: TL;DR The migration result is not meant to be displayed with the new event log UI. Its only goal is to avoid breaking the old UI. 😄

josejulio: Makes sense!

josejulio (approved):

PR #509, opened 2021-08-23
NOTIF-245 When no bundle is specified, return all applications

josejulio:
This is something I need for event log ui filter.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
@@ -65,10 +65,18 @@ public class ApplicationResources {
     }
 
     public Uni<List<Application>> getApplications(String bundleName) {
-        String query = "FROM Application WHERE bundle.name = :bundleName";
-        return session.createQuery(query, Application.class)
-                .setParameter("bundleName", bundleName)
-                .getResultList();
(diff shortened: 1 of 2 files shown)


backend/src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
```diff
@@ -65,10 +65,18 @@
...
-        String query = "FROM Application WHERE bundle.name = :bundleName";
-        return session.createQuery(query, Application.class)
-                .setParameter("bundleName", bundleName)
-                .getResultList();
+        String sql = "FROM Application";
+        if (bundleName != null) {
+            sql += " WHERE bundle.name = :bundleName";
+        }
+
+        var query = session.createQuery(sql, Application.class);
```
gwenneg, line 73:
```suggestion
        let query = session.createQuery(sql, Application.class);
```

gwenneg: Just kidding

gwenneg (approved):

backend/src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java
josejulio: You got me there :grin:

PR #514, opened 2021-08-25
Minor fixes

theute:
- Used "Event Type ID" rather than "Event ID" as I found it confusing
- Added space around the table

(Sorry for the import reordering, I couldn't figure out how to disable this reordering on save feature in VSCode...)


theute:
Ok, eslint didn't care about my apologize, should be fine now.
(Now I need to learn how to make VSCode work properly)

josejulio:
yarn lint:ts --fix should fix the sorting (or -fix, don't remember)

theute:
I changed my VSCode settings now, it's using `.eslintrc.yml` now.

gwenneg:
> * Used "Event Type ID" rather than "Event ID" as I found it confusing

FYI Event ID exists in notifications-backend since yesterday, see [here](https://github.com/RedHatInsights/notifications-backend/blob/538e99c37df1853cca0ca3c373aaa84301f9d838/backend/src/main/java/com/redhat/cloud/notifications/models/Event.java#L32) so +1 for this change 😄

aferd (approved):

josejulio:
Could you remove the binary files? Adding `*.pyc ` to the `.gitignore` would be helpful too :-)

theute:
> Could you remove the binary files? Adding `*.pyc ` to the `.gitignore` would be helpful too :-)

Good catch... Thanks

PR #515, opened 2021-08-25
NOTIF-291 Fix OutOfMemoryError during notifications history migration

gwenneg:
CI database contains 265241 records in `notification_history`. The migration failed because it tried to load all these records into an `ArrayList` which caused an `OutOfMemoryError`.

The new code migrates 1000 records at a time and repeats itself until there's nothing more to migrate.


gwenneg:
⚠️ Check `Hide whitespace changes` for an easier review!

josejulio:
>  Check `Hide whitespace changes` for an easier review!

image

josejulio (approved):

gwenneg:
I tested the migration again on CI. The OutOfMemoryError was fixed as expected, but the migration is awfully slow. I'll come up with something better.

PR #516, opened 2021-08-25
NOTIF-291 Introduce EventLogCleaner

gwenneg:
Event log entries will be cleaned by a scheduled job after a configurable retention delay of 14 days.


gwenneg:
This will delete both `Event` and `NotificationHistory` DB records because of the [cascading foreign key](https://github.com/RedHatInsights/notifications-backend/blob/538e99c37df1853cca0ca3c373aaa84301f9d838/backend/src/main/resources/db/migration/V1.23.0__NOTIF-291_events.sql#L14).

backend/src/test/java/com/redhat/cloud/notifications/db/EventLogCleanerTest.java
```diff
@@ -0,0 +1,86 @@
...
+import static com.redhat.cloud.notifications.db.EventLogCleaner.EVENT_LOG_CLEANER_DELETE_AFTER_CONF_KEY;
+import static com.redhat.cloud.notifications.db.EventLogCleaner.now;
+
+@QuarkusTest
+@QuarkusTestResource(TestLifecycleManager.class)
+public class EventLogCleanerTest {
+
+    /*
+     * The Event#created field is automatically set because of the @PrePersist annotation in CreationTimestamped when
+     * an Event is persisted using a stateful session. @PrePersist does not work with a stateless session. We need to
```
josejulio, line 26: Isn't the Event created on a stateless session too?

josejulio: https://github.com/gwenneg/notifications-backend/blob/72339b14c99c1a2433c56bd9f48592e4c664fbb5/backend/src/main/java/com/redhat/cloud/notifications/db/ApplicationResources.java#L169

And i suppose that's the `@PrePersist`

gwenneg: I'm not sure I get the question. Yes the `Event` is persisted with the stateless session. The only difference with a stateful session is the invocation of `Event#prePersist` which sets the `created` field automatically with the stateful session. That invocation doesn't happen with the stateless session.

josejulio: I was asking because I didn't remember any call to `setCreated` in the previous PR, but then I recall about that `prePersist` call :-)

backend/src/main/java/com/redhat/cloud/notifications/db/EventLogCleaner.java
```diff
@@ -0,0 +1,52 @@
...
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import java.time.Duration;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+
+import static java.util.concurrent.TimeUnit.MINUTES;
+
+@Singleton
```
josejulio, line 16: Curious, why  `Singleton` and not `ApplicationScoped` ?

PR #517, opened 2021-08-25
NOTIF-291 Speed up Event data migration

gwenneg:


josejulio (approved):
lgtm, this is definitively simpler :D

PR #532, opened 2021-08-30
Refactor resources usage in tests

gwenneg:
This PR moves resources-related test code around to reduce code duplication and ease tests maintenance.

It will also help me with NOTIF-273 and NOTIF-291.

- 100% test code, no impact on prod
- 0% tests coverage change, it is exactly like it was before this PR
- 100% boring to review, I'll merge it by the end of the day (without approval) unless someone likes boring reviews 😄


bennetelli (approved):

bennetelli:
I still think we can improve these tests to make them more readable, but this is a good step :) thanks!

gwenneg:
Yes we can definitely improve the tests but now would be a bad time because the Quarkus 2 bump will impact the tests heavily. I have a local branch with tons of mandatory tests changes.

PR #533, opened 2021-08-30
NOTIF-291 Ease the first EventLogCleaner run

gwenneg:
Draft until we decide to push notifications-backend to prod.

### DIFF
--- a/backend/src/main/resources/db/migration/V1.25.0__NOTIF-291_event_cleaning_warm_up.sql
+++ b/backend/src/main/resources/db/migration/V1.25.0__NOTIF-291_event_cleaning_warm_up.sql
@@ -1,0 +1,5 @@
+-- This will ease the first run of EventLogCleaner by reducing the amount of data it will have to delete.
+-- The following date is based on the assumption that the event log retention delay is 31 days.
+
+DELETE FROM event
+WHERE created < '2021-08-02T00:00:00.000000';


backend/src/main/resources/db/migration/V1.25.0__NOTIF-291_event_cleaning_warm_up.sql
```diff
@@ -0,0 +1,5 @@
+-- This will ease the first run of EventLogCleaner by reducing the amount of data it will have to delete.
+-- The following date is based on the assumption that the event log retention delay is 31 days.
+
+DELETE FROM event
+WHERE created < '2021-07-31T00:00:00.000000';
```
gwenneg, line 5: This date has to be updated as late as possible, ideally right before we deploy on prod.

gwenneg: I changed the value to `now() minus 28 days` which means `EventLogCleaner` will start with the lowest possible number of records to delete on its first run, assuming we deploy notifications-backend on prod before thursday.

The retention delay should be 31 days and we will actually have 28 days of data but it doesn't matter because that data isn't used yet in the prod UI.

gwenneg:
This is no longer a draft, it can be merged safely now.

bennetelli:
Where does the event table come from? And is this query working? I can't check out the pr locally as the github cli is broken and I need to fix it first :(

gwenneg:
That's new table, it was added during your PTO for the [event log](https://issues.redhat.com/browse/NOTIF-291) feature.

We already had a very close query tested on CI:
https://github.com/RedHatInsights/notifications-backend/blob/c35f8f9559a52fe10477ec34c313be52f051fe16/backend/src/main/resources/db/migration/V1.24.0__NOTIF-291_data_migration.sql#L6-L7

I will confirm this one works on CI before the stage/prod deployment, the test will only take a few minutes.

bennetelli (approved):

bennetelli:
LGTM

PR #546, opened 2021-09-01
NOTIF-291 Remove Flyway repair

gwenneg:
`repair()` was required because the Flyway scripts `V1.22` to `V1.26` were deleted and replaced with `V1.27` while some of them had already been executed on some environments.

CI, stage and prod were already migrated to `V1.27`. I'm leaving this around a bit more to help fixing migration issues in the devs local database.


gwenneg:
No, don't worry 😄 The event log schema/data migration was a real pain and it took several migration attempts (and migration scripts rollbacks) on CI and stage before the final and stable version was ready.

PR #549, opened 2021-09-03
NOTIF-291 Add EventService

gwenneg:
This PR introduces the main REST API for the event log.

We will probably need two more (much smaller) REST APIs to retrieve:
- the properties of an endpoint displayed in the event log
- the details of an event

If this is confirmed, the missing APIs will be added in a separate PR.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
@@ -30,14 +30,14 @@ public class EventResources {
                 .replaceWith(event);
     }
 
-    public Uni<List<Event>> get(String accountId, Set<UUID> bundleIds, Set<UUID> appIds, String eventTypeName,
+    public Uni<List<Event>> get(String accountId, Set<UUID> bundleIds, Set<UUID> appIds, String eventTypeDisplayName,
                                 LocalDate startDate, LocalDate endDate, Integer limit, Integer offset, String sortBy) {
         String hql = "SELECT DISTINCT e FROM Event e " +
                 "JOIN FETCH e.eventType et JOIN FETCH et.application a JOIN FETCH a.bundle b " +
                 "LEFT JOIN FETCH e.historyEntries he LEFT JOIN FETCH he.endpoint " +
                 "WHERE e.accountId = :accountId";
 
-        hql = addHqlConditions(hql, bundleIds, appIds, eventTypeName, startDate, endDate);
+        hql = addHqlConditions(hql, bundleIds, appIds, eventTypeDisplayName, startDate, endDate);
 
         if (sortBy == null) {
             hql += " ORDER BY e.created DESC";
@@ -51,7 +51,7 @@ public class EventResources {
         }
 
         Mutiny.Query<Event> query = session.createQuery(hql, Event.class);
-        setQueryParams(query, accountId, bundleIds, appIds, eventTypeName, startDate, endDate);
+        setQueryParams(query, accountId, bundleIds, appIds, eventTypeDisplayName, startDate, endDate);
 
         if (limit != null) {
             query.setMaxResults(limit);
@@ -63,20 +63,20 @@ public class EventResources {
         return query.getResultList();
     }
 
-    public Uni<Long> count(String accountId, Set<UUID> bundleIds, Set<UUID> appIds, String eventTypeName,
+    public Uni<Long> count(String accountId, Set<UUID> bundleIds, Set<UUID> appIds, String eventTypeDisplayName,
                       LocalDate startDate, LocalDate endDate) {
         String hql = "SELECT COUNT(*) FROM Event e JOIN e.eventType et JOIN et.application a JOIN a.bundle b " +
                 "WHERE e.accountId = :accountId";
 
-        hql = addHqlConditions(hql, bundleIds, appIds, eventTypeName, startDate, endDate);
+        hql = addHqlConditions(hql, bundleIds, appIds, eventTypeDisplayName, startDate, endDate);
 
         Mutiny.Query<Long> query = session.createQuery(hql, Long.class);
-        setQueryParams(query, accountId, bundleIds, appIds, eventTypeName, startDate, endDate);
+        setQueryParams(query, accountId, bundleIds, appIds, eventTypeDisplayName, startDate, endDate);
 
         return query.getSingleResult();
     }
 
-    private static String addHqlConditions(String hql, Set<UUID> bundleIds, Set<UUID> appIds, String eventTypeName,
+    private static String addHqlConditions(String hql, Set<UUID> bundleIds, Set<UUID> appIds, String eventTypeDisplayName,
                                            LocalDate startDate, LocalDate endDate) {
         if (bundleIds != null && !bundleIds.isEmpty()) {
             hql += " AND b.id IN (:bundleIds)";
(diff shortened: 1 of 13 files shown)


gwenneg:
Here's an example of the API output:
```
[
   {
      "id":"970e557d-5872-47d8-8040-e9dadfc26510",
      "created":"2021-09-06T08:36:40.76311",
      "bundle":"Bundle 2",
      "application":"Application 2",
      "event_type":"Event type 2",
      "actions":[
         {
            "id":"046fbfaf-9900-4bfc-bb9a-a6a9a4abe5c3",
            "endpoint_type":"webhook",
            "invocation_result":true
         }
      ]
   },
   {
      "id":"cbc0bbf1-7935-4d02-b6b6-585b25ae483e",
      "created":"2021-09-04T08:36:40.76311",
      "bundle":"Bundle 2",
      "application":"Application 2",
      "event_type":"Event type 2",
      "actions":[
         {
            "id":"88effd41-3315-4605-85d2-c6d633b43dba",
            "endpoint_type":"email_subscription",
            "invocation_result":true
         }
      ]
   },
   {
      "id":"3caacdd3-cca4-47d0-aa77-35c49872e069",
      "created":"2021-09-01T08:36:40.76311",
      "bundle":"Bundle 1",
      "application":"Application 1",
      "event_type":"Event type 1",
      "actions":[
         {
            "id":"374276d6-4ae6-4df3-a589-de418b62d023",
            "endpoint_type":"webhook",
            "invocation_result":true
         },
         {
            "id":"58f02b09-0627-4e81-a91a-1001fb2718cc",
            "endpoint_type":"email_subscription",
            "invocation_result":true
         }
      ]
   }
]
```

backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
```diff
@@ -0,0 +1,119 @@
...
+        if (bundleIds != null && !bundleIds.isEmpty()) {
+            hql += " AND b.id IN (:bundleIds)";
+        }
+
+        if (appIds != null && !appIds.isEmpty()) {
+            hql += " AND a.id IN (:appIds)";
+        }
+
+        if (eventTypeName != null) {
+            hql += " AND et.name = :eventTypeName";
```
josejulio, line 49: Could we make this one a `LIKE` query?

backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
```diff
@@ -0,0 +1,119 @@
...
+            hql += " AND DATE(e.created) >= :startDate";
+        } else if (endDate != null) {
+            hql += " AND DATE(e.created) <= :endDate";
+        }
+
+        if (sortBy == null) {
+            hql += " ORDER BY e.created DESC";
+        } else {
+            Matcher sortByMatcher = SORT_BY_PATTERN.matcher(sortBy);
+            if (sortByMatcher.matches()) {
```
josejulio, line 64: I think this should fail the request if doesn't matches.

backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        if (startDate != null && endDate != null) {
+            query.setParameter("startDate", startDate)
+                    .setParameter("endDate", endDate);
+        } else if (startDate != null) {
+            query.setParameter("startDate", startDate);
+        } else if (endDate != null) {
+            query.setParameter("endDate", endDate);
+        }
```
josejulio, line 93: I think is easier to read, but no strong opinion.

```suggestion
        if (startDate != null) {
            query.setParameter("startDate", startDate);
        }
        if (endDate != null) {
            query.setParameter("endDate", endDate);
        }
```

backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
josejulio: Ah, this is done in the service.

josejulio: Perhaps this method should receive the column and direction as different params (maybe using an enum for the direction here)

backend/src/main/java/com/redhat/cloud/notifications/routers/EventService.java
```diff
@@ -0,0 +1,74 @@
...
+    public static final Pattern SORT_BY_PATTERN = Pattern.compile("^([a-z0-9_-]+):(asc|desc)$", CASE_INSENSITIVE);
+
+    @Inject
+    EventResources eventResources;
+
+    @GET
+    @Produces(APPLICATION_JSON)
+    @RolesAllowed(RBAC_READ_NOTIFICATIONS)
+    @Operation(summary = "Retrieve the event log entries.")
+    public Uni<List<EventLogEntry>> getEvents(@Context SecurityContext securityContext, @RestQuery Set<UUID> bundleIds, @RestQuery Set<UUID> appIds,
```
josejulio, line 43: Could you wrap this object into a [Page](https://github.com/RedHatInsights/notifications-backend/blob/master/backend/src/main/java/com/redhat/cloud/notifications/routers/models/Page.java)?

backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
gwenneg: Why did I write something that complicated? I don't know 😄

backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
gwenneg: I didn't put the `sortBy` parsing code in the service because it relies on the `getSortField` method which is tightly linked to the persistence layer. I can change that if you want.

Regarding the enum, I don't think that's necessary since the validating regex will only allow `asc` or `desc`.

josejulio: It's fine.

gwenneg:
@josejulio I'm done pushing. Sorry for the multiple commits, I had a few issues with the `lastOffset` calculation 😄

josejulio:
> @josejulio I'm done pushing. Sorry for the multiple commits, I had a few issues with the `lastOffset` calculation

Thanks! I'll review/test this later today :)

backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
```diff
@@ -0,0 +1,134 @@
...
+    private static String addHqlConditions(String hql, Set<UUID> bundleIds, Set<UUID> appIds, String eventTypeName,
+                                           LocalDate startDate, LocalDate endDate) {
+        if (bundleIds != null && !bundleIds.isEmpty()) {
+            hql += " AND b.id IN (:bundleIds)";
+        }
+        if (appIds != null && !appIds.isEmpty()) {
+            hql += " AND a.id IN (:appIds)";
+        }
+        if (eventTypeName != null) {
+            hql += " AND et.name LIKE :eventTypeName";
```
josejulio, line 88: I'm wondering if we should do the `LIKE` against `LOWER(et.displayName)` or both (in lower case) i.e. `" AND (LOWER(et.name) LIKE :eventTypeName OR LOWER(et.displayName) LIKE :eventTypeName)"`

backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
```diff
@@ -0,0 +1,134 @@
...
+                                       LocalDate startDate, LocalDate endDate) {
+        query.setParameter("accountId", accountId);
+        if (bundleIds != null && !bundleIds.isEmpty()) {
+            query.setParameter("bundleIds", bundleIds);
+        }
+        if (appIds != null && !appIds.isEmpty()) {
+            query.setParameter("appIds", appIds);
+        }
+        if (eventTypeName != null) {
+            query.setParameter("eventTypeName", "%" + eventTypeName + "%");
```
josejulio, line 110:
```suggestion
            query.setParameter("eventTypeName", "%" + eventTypeName.toLowerCase() + "%");
```

josejulio:
> We will probably need two more (much smaller) REST APIs to retrieve:
> 
> * the properties of an endpoint displayed in the event log
> * the details of an event

We don't need those yet

josejulio (approved):
lgtm

PR #555, opened 2021-09-06
NOTIF-273 Prepare tests for the Quarkus 2 bump

gwenneg:
This is phase 1 of the preparation work, there will be more. As a consequence, some parts of this PR may look unfinished but that's perfectly normal.

Quarkus 2 requires all tests involving Hibernate Reactive to be written in a reactive way. In other words, `.await().indefinitely()` has to be removed completely from tests (it won't happen before the bump).

### DIFF
--- a/backend/src/test/java/com/redhat/cloud/notifications/db/BehaviorGroupResourcesTest.java
+++ b/backend/src/test/java/com/redhat/cloud/notifications/db/BehaviorGroupResourcesTest.java
@@ -1,16 +1,16 @@
 package com.redhat.cloud.notifications.db;
 
 import com.redhat.cloud.notifications.TestLifecycleManager;
-import com.redhat.cloud.notifications.models.Application;
 import com.redhat.cloud.notifications.models.BehaviorGroup;
 import com.redhat.cloud.notifications.models.BehaviorGroupAction;
 import com.redhat.cloud.notifications.models.Bundle;
-import com.redhat.cloud.notifications.models.Endpoint;
 import com.redhat.cloud.notifications.models.EventType;
 import com.redhat.cloud.notifications.models.EventTypeBehavior;
 import io.quarkus.test.common.QuarkusTestResource;
 import io.quarkus.test.junit.QuarkusTest;
+import io.smallrye.mutiny.Uni;
 import org.hibernate.reactive.mutiny.Mutiny;
+import org.junit.jupiter.api.Assertions;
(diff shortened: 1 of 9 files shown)


gwenneg:
⚠️ Check `Hide whitespace changes` for an easier review! ⚠️

josejulio:
> `.await().indefinitely()` has to be removed completely from tests

:cry:

gwenneg:
@josejulio This is ready for a review. I'll split the preparation work into several PRs like this one to minimize conflicts with other ongoing PRs.

backend/src/test/java/com/redhat/cloud/notifications/db/BehaviorGroupResourcesTest.java
```diff
@@ -45,245 +45,321 @@
...
-        assertEquals(1, eventTypes.size());
-        assertEquals(eventType.getId(), eventTypes.get(0).getId());
+    void testFindEventTypesByBehaviorGroupId() {
+        resourceHelpers.createBundle()
+                .onItem().call(bundle -> resourceHelpers.createApplication(bundle.getId())
+                        .onItem().call(app -> createEventType(app.getId())
+                                .onItem().call(eventType -> resourceHelpers.createBehaviorGroup(DEFAULT_ACCOUNT_ID, "displayName", bundle.getId())
+                                        .onItem().call(behaviorGroup -> updateAndCheckEventTypeBehaviors(DEFAULT_ACCOUNT_ID, eventType.getId(), true, behaviorGroup.getId())
+                                                .chain(() -> resourceHelpers.findEventTypesByBehaviorGroupId(behaviorGroup.getId())
+                                                        .onItem().invoke(eventTypes -> {
```
josejulio, line 178: This starts to get really difficult to read (nesting too much). Is there any other way to write this up?

gwenneg: I already wrote as many nested instructions as possible, these are all mandatory unfortunately. It could be broken into methods but I don't think that would really serve readability.

gwenneg: I made a tiny improvement (shortcuts) but it doesn't affect the crazy nesting.

Can we merge this as is and maybe improve the readability later?

josejulio (approved):

PR #559, opened 2021-09-07
Update Grafana dashboard

bennetelli:


theute:
The file needs to remain as a k8s configmap descriptor Yaml (as mentioned on Slack ;-))

bennetelli:
too many new informations yesterday.. now I see what that means

theute:
We may have to tweak `uid` and `version` but we'll see...

PR #560, opened 2021-09-07
Update Grafana dashboard version

bennetelli:


bennetelli:
It looks like this isn't necessary.

PR #563, opened 2021-09-07
Improve RBAC metrics and increase elements per page

gwenneg:
See the [Micrometer doc](https://micrometer.io/docs/concepts#_storing_start_state_in_timer_sample) for details about `Timer.Sample`.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
@@ -17,7 +17,6 @@ import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 import java.io.IOException;
 import java.time.Duration;
-import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
@@ -32,7 +31,7 @@ public class RbacRecipientUsersProvider {
     @RestClient
     RbacServiceToService rbacServiceToService;
 
-    @ConfigProperty(name = "recipient-provider.rbac.elements-per-page", defaultValue = "40")
+    @ConfigProperty(name = "recipient-provider.rbac.elements-per-page", defaultValue = "1000")
     Integer rbacElementsPerPage;
 
     @ConfigProperty(name = "rbac.retry.max-attempts", defaultValue = "3")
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -47,48 +46,46 @@
...
+            page -> {
+                Timer.Sample getUsersPageTimer = Timer.start(meterRegistry);
+                return retryOnError(
+                        rbacServiceToService.getUsers(accountId, adminsOnly, page * rbacElementsPerPage, rbacElementsPerPage)
+                )
+                .onItem().invoke(() -> getUsersPageTimer.stop(meterRegistry.timer("rbac.get-users.page", "accountId", accountId)));
+            }
         )
-        .onItem().invoke(() -> getUsersProcessTime.record(Duration.between(start, LocalDateTime.now())))
+        .onItem().invoke(() -> getUsersTotalTimer.stop(meterRegistry.timer("rbac.get-users.total", "accountId", accountId)))
```
josejulio, line 68: I'm wondering if it would make sense to add the total elements that were fetched.

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -47,48 +46,46 @@
...
-            page -> retryOnError(
-                    rbacServiceToService
-                            .getUsers(accountId, adminsOnly, page * rbacElementsPerPage, rbacElementsPerPage)
-            )
+            page -> {
+                Timer.Sample getUsersPageTimer = Timer.start(meterRegistry);
+                return retryOnError(
+                        rbacServiceToService.getUsers(accountId, adminsOnly, page * rbacElementsPerPage, rbacElementsPerPage)
+                )
+                .onItem().invoke(() -> getUsersPageTimer.stop(meterRegistry.timer("rbac.get-users.page", "accountId", accountId)));
```
josejulio, line 65: I'm wondering if it would make sense to add the total elements that were fetched and/or the current page.

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -32,7 +31,7 @@
     @RestClient
     RbacServiceToService rbacServiceToService;
 
-    @ConfigProperty(name = "recipient-provider.rbac.elements-per-page", defaultValue = "40")
+    @ConfigProperty(name = "recipient-provider.rbac.elements-per-page", defaultValue = "1000")
```
josejulio, line 34: I though we were going to use a higher number based on Jay's comments.
Anyhow we can tweak this number based on the data that we collect.

josejulio (approved):
lgtm, some minor comments that could be addressed later if needed.

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
gwenneg: Yeah we could go higher from the start but I'd rather see the results with that one and then adjust with the configuration.

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
gwenneg: Done.

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
gwenneg: Done.

PR #568, opened 2021-09-08
Change the way we register the webhook gauge

bennetelli:


gwenneg:
Why this change? Dependencies injections are done with `@Inject` everywhere in notifications-backend. I'm not sure introducing constructor injections will help maintainability. Also `registry.counter("processor.webhook.processed");` is a shortcut for `Counter.builder("processor.webhook.processed").register(registry);`.

bennetelli:
constructor injection > field injection. At least when you want to unit test classes and methods.

yes, it's a shortcut. But it's not possible to add tags, descriptions and other things. It just contains the metric name. Therefore I would prefer the builder-way.

bennetelli:
the main purpose for this change is getting rid of the PostConstruct stuff.

gwenneg:
> constructor injection > field injection. At least when you want to unit test classes and methods.

There are pros and cons for each approach but I don't see how constructor injections would be better in tests. Ideally, we should _never_ have to call the constructor and always use mocks instead.

Anyway, I'm not taking side for one or the other but I think we should stick to one approach and avoid mixing them in the code for the sake of maintainability.

> yes, it's a shortcut. But it's not possible to add tags, descriptions and other things. It just contains the metric name. Therefore I would prefer the builder-way.

`counter` has a signature with tags.

bennetelli:
I wouldn't stick with one or the other approach. I would just take the best approach depending on the use case. Mixing isn't bad. 
It's only bad if you know the one way only. People also use the bike and the car depending on the way to drive instead of always taking the bike ;-)

bennetelli:
counter has a signature with tags, but it doesn't contain descriptions.

bennetelli:
the counter is already there.

PR #572, opened 2021-09-08
NOTIF-294 Create new event type 

aferd:


backend/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -16,6 +25,27 @@ export const ApplicationPage: React.FunctionComponent = () => {
     const applicationTypesQuery = useApplicationTypes(applicationId);
     const columns = [ 'Event Type', 'Event Type Id', 'Description' ];
 
+    const newEvent = useCreateEventType();
+    const [ id ] = React.useState<string | undefined>();
+    const [ displayName, setDisplayName ] = React.useState<string | undefined>();
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
josejulio, line 31: We should remove this. I think we are using the variables below it. (or remove the unused variables)

backend/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -32,14 +62,57 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
+                                        variant={ ModalVariant.medium }
+                                        title='Create Event Type'
+                                        isOpen={ isOpen }
+                                        onClose={ toggle }
+                                    ><Form isHorizontal><FormGroup label='Application Id' fieldId='application-id'
+                                            helperText='This is the id of the application'>
+                                            <TextInput
+                                                type='text'
+                                                value= { applicationId }
+                                                id='application-id' /></FormGroup>
```
josejulio, line 87: I don't think we really need to show the applicationId here, We can show in the title something like: `Create Event Type for ${the-bundle}` instead

backend/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -32,14 +62,57 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
+                                            <TextArea
+                                                type='text'
+                                                onChange={ setDescription }
+                                                id='description' /></FormGroup>
+                                        <ActionGroup>
+                                            <Button variant='primary' type='submit'
+                                                onClick= { onSubmit }>Submit</Button>
+                                            <Button variant='link' onClick={ toggle }>Cancel</Button>
+                                        </ActionGroup>
+                                        </Form><></>
```
josejulio, line 111: We could probably put this form in a separate file to reuse it when we want to edit.

backend/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -32,14 +62,57 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
+                                                onChange={ setDisplayName }
+                                                id='display-name' /></FormGroup>
+                                        <FormGroup label='Description' fieldId='description'
+                                            helperText='Optional short description that appears in the UI to help admin descide how to notify users.'>
+                                            <TextArea
+                                                type='text'
+                                                onChange={ setDescription }
+                                                id='description' /></FormGroup>
+                                        <ActionGroup>
+                                            <Button variant='primary' type='submit'
```
josejulio, line 107: This button should be disabled if any of the required fields is undefined

backend/src/main/webapp/src/pages/ApplicationPage.tsx
aferd: would it be for the bundle or for the application?

josejulio: The application, yes - my bad :)

backend/src/main/webapp/src/pages/ApplicationPage.tsx
aferd: in progress.

aferd:
https://user-images.githubusercontent.com/86322239/133104237-769bdb9b-3cd6-4eca-ac2d-f5d202b8d200.mov

josejulio (approved):
lgtm, we could eventually add validations with [formik](https://formik.org/docs/overview), but lets complete the edit/delete flow first.

aferd:
> lgtm, we could eventually add validations with [formik](https://formik.org/docs/overview), but lets complete the edit/delete flow first.

It does need some more validation, but that seems cool, I will check it out!

PR #576, opened 2021-09-09
feat(templates): integrate Managed Kafka notification templates

machi1990:
This is the first step of integrating Managed Kafka templates. We'll be working on adding some
content on each of the added templates but let's keep them empty for now to be able to perform all
the required Java bindings

/cc @gwenneg @akoserwal @theute

PR #578, opened 2021-09-09
Update method to check if subscription is supported

josejulio:
After reviewing #576 I noticed that we could use `isEmailSubscriptionSupported` in this check.


gwenneg (approved):
LGTM

PR #581, opened 2021-09-10
Get rid of the reactive tests crazy nesting

gwenneg:
@josejulio How do you like this new version? 😃

If you like it, I'll write everything like this.


josejulio (approved):
IMO it looks better!

PR #583, opened 2021-09-10
Add a gauge which displays when an application started

bennetelli:


bennetelli:
It's a try to use it as a deployment marker via Graphana annotations. Not sure if it will be displayed as expected, but I am not able to test it locally.

backend/src/main/java/com/redhat/cloud/notifications/NotificationsApp.java
```diff
@@ -22,11 +26,24 @@
...
     private static final Logger LOG = Logger.getLogger(NotificationsApp.class);
 
+    @Inject
+    MeterRegistry registry;
+
+    AtomicInteger started;
+
+    @PostConstruct
+    void postConstruct() {
+        started = registry.gauge("notifications.backend.started", new AtomicInteger(0));
```
gwenneg, line 36: Why do we need a gauge instead of a counter?

bennetelli: A counter can onle be incremented but not decremented.

gwenneg: I know what a gauge is, but I don't see why it's needed here.

gwenneg: I mean its value will be `1` for a few milliseconds at most and `0` during something like 99% of the startup time.

bennetelli: exactly. I want it to use to create the blue line here: https://www.inovex.de/wp-content/uploads/2020/10/example_simple.png

PR #584, opened 2021-09-10
Revert "Add a gauge which displays when an application started"

bennetelli:
Reverts RedHatInsights/notifications-backend#583


bennetelli:
@gwenneg unfortunately it didn't work :(

gwenneg:
Well the timeframe with the `1` value is so short that maybe there's not enough time for that value to be scrapped. That's what I was afraid of.

bennetelli:
yes, could be. Will probably need to add a ticket for app-sre and pitch the deployment markers there.

PR #585, opened 2021-09-10
NOTIF-190 Use system properties first. If not present, use application properties

bennetelli:


aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -96,7 +96,11 @@ public void processDailyEmail() {
     }
 
     private boolean isCronJobEnabled() {
-        // The scheduled job is disabled by default.
-        return ConfigProvider.getConfig().getOptionalValue("notifications.aggregator.email.subscription.periodic.cron.enabled", Boolean.class).orElse(false);
+        final String cronJobEnabled = System.getProperty("NOTIFICATIONS_AGGREGATOR_EMAIL_SUBSCRIPTION_PERIODIC_CRON_ENABLED");
```
josejulio, line 99: If I'm not mistaken, the ConfigProvider will override `notifications.aggregator.email.subscription.periodic.cron.enabled` with `NOTIFICATIONS_AGGREGATOR_EMAIL_SUBSCRIPTION_PERIODIC_CRON_ENABLED` :) So I don't think we need this change.

bennetelli: @josejulio but then I would expect that the new cron job should run here: https://console-openshift-console.apps.crcs02ue1.urby.p1.openshiftapps.com/k8s/ns/notifications-stage/pods/notifications-aggregator-27187200--1-8pk52/logs

Hope I am not overlooking something..

josejulio: How/Where are you defining `NOTIFICATIONS_AGGREGATOR_EMAIL_SUBSCRIPTION_PERIODIC_CRON_ENABLED` in there? I don't see it on the environment tab: https://console-openshift-console.apps.crcs02ue1.urby.p1.openshiftapps.com/k8s/ns/notifications-stage/pods/notifications-aggregator-27187200--1-8pk52/environment

bennetelli: Forget about it.. my PR is still open... https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/24963

gwenneg (changes requested):
I'm trying to figure out what's wrong, don't merge.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -175,8 +175,12 @@ public void processDailyEmail(ScheduledExecution se) {
     }
 
     private boolean isScheduleEnabled() {
-        // The scheduled job is enabled by default.
-        return ConfigProvider.getConfig().getOptionalValue("notifications.backend.email.subscription.periodic.cron.enabled", Boolean.class).orElse(true);
+        final String cronJobEnabled = System.getProperty("NOTIFICATIONS_BACKEND_EMAIL_SUBSCRIPTION_PERIODIC_CRON_ENABLED");
```
gwenneg, line 178: This shouldn't be necessary, but even if it was we would read the value from the environment variables and not from the system properties.

gwenneg: The old code works on my machine with the environment variable. I'm testing the other method now.

gwenneg: Same for the other method, everything works as intended for me without any code change.

bennetelli: Right now we don't have the system properties in prod. And we need to be able to add new versions on prod. 

I think from a responsibility point of view, having a logic like this is necessary. I think the application should contain everything it needs itself. Without it we would depend on configuration tools of other teams.

josejulio: > I think from a responsibility point of view, having a logic like this is necessary. I think the application should contain everything it needs itself. Without it we would depend on configuration tools of other teams.

I'm not sure I understand why we need this logic.
All our custom configuration are pulled from the environment variables (stored in our e2e and appsre tempĺates). Why do we need to add system properties into the mix?

gwenneg: As I wrote, we don't use system properties, we use environment variables.

As explained in the [Quarkus configuration guide](https://quarkus.io/guides/config-reference#configuration-sources), properties from `applications.properties` file can be overridden from multiple sources including environment variables and in that specific case (the two methods you changed), it works without any code change. If it didn't, Quarkus would have to fix a huge regression.

gwenneg: Maybe the reason why you had an issue is because you used `NOTIFICATIONS_AGGREGATOR_EMAIL_SUBSCRIPTION_PERIODIC_CRON_ENABLED` as the system property key instead of `notifications.backend.email.subscription.periodic.cron.enabled`. The "all caps + snake case" version is only used with env vars.

bennetelli: i am not able to get the variable in quarkus by exporting it before, but that's probably not the problem. 
The environment variables are missing in OpenShift on stage. What do I need to update? I probably missed something..

PR #591, opened 2021-09-13
NOTIF-308 Fix EventService path

gwenneg:
@josejulio I also changed `/event` to `/events` to be consistent with other APIs.


gwenneg:
Great label!

PR #595, opened 2021-09-14
NOTIF-310 Process recipients from action

josejulio:
Depends on https://github.com/RedHatInsights/insights-schemas-java/pull/17

This PR makes use of `Action v1.1.0` which allow applications to specify a recipient configuration. Currently it allows to specify if we want to send to `admins` and/or `ignore user preferences`.

Notifications-backend does take this configuration and extends whatever configuration we already had configured by an admin.
It tries to simplify the queries executed to RBAC by consolidating the same calls. 
e.g. if admin configured `Send to admin` and the application also used `Send to admin` it will make only one to RBAC looking for the admins.

This is achieved by transforming the Email configurations from the settings and the application into a [RecipientResolverRequest](https://github.com/RedHatInsights/notifications-backend/pull/595/files#diff-7911d0b48554d334448202730552f3bcdbc6b7c7e751a5b974249d25f9b5873b) and removing the repeated configurations.


backend/src/main/java/com/redhat/cloud/notifications/processors/email/RecipientResolverRequest.java
```diff
@@ -0,0 +1,98 @@
+package com.redhat.cloud.notifications.processors.email;
+
+import com.redhat.cloud.notifications.models.EmailSubscriptionProperties;
+import com.redhat.cloud.notifications.models.Endpoint;
+
+import java.util.Objects;
+import java.util.UUID;
+
+public class RecipientResolverRequest {
```
josejulio, line 9: I don't like this name :smile:

josejulio:
@gwenneg this is ready for review, but not ready for merge.
We would need to approve, merge and release the schema (IIRC the release is manually) before.

PR #599, opened 2021-09-15
Remove RBAC user query feature flag

gwenneg:
@josejulio We don't need that anymore, right?


josejulio (approved):
LGTM. The flag was only an interim step.

PR #600, opened 2021-09-15
NOTIF-309 Deduplicate incoming ingress Kafka messages

gwenneg:
By default, Kafka guarantees that each message will be consumed/processed _at least once_. This PR introduces Kafka messages deduplication in order to consume each message _exactly once_. The underlying goal is to protect us against Kafka failures like the one we had recently on prod when our consumer failed to commit its offset because it was kicked out of the consumer group. We never confirmed it but it most probably led to duplicate notifications being sent during the incident.

This PR is based on two things:
- A new optional (for now) `rh-message-id` Kafka header which can be sent by the onboarded apps with their message. It is then used as a unique identifier for the message.
- A new `kafka_message` DB table where all valid messages IDs are stored for a limited period of time (default: 1 day, configurable) and then purged.

ℹ️ This is a non-breaking PR that should allow a smooth migration for the onboarded apps. We may decide on a date when the new header will become mandatory, or not. I added a several new counters to allow us to monitor the migration progress. It should help us know when all apps are done migrating.

⚠️ The app behavior should be _exactly identical_ before and after this PR for an onboarded app that _doesn't_ send the new header. If you suspect a behavior change during the review, please add a comment about it.

When a message is received, if its ID is already known (found in the database), then the message is ignored. If the ID is missing (not sent by the onboarded app) or invalid, then the message is considered unknown and processed normally. Invalid IDs are reported at `WARN` level in the logs and will show up in Sentry, allowing us to inform the onboarded app of an integration issue.

⚠️ Version 4 of UUID is mandatory. Any other version received will be rejected.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
@@ -8,7 +8,6 @@ import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.smallrye.mutiny.Uni;
 import org.eclipse.microprofile.reactive.messaging.Acknowledgment;
-import org.eclipse.microprofile.reactive.messaging.Acknowledgment.Strategy;
 import org.eclipse.microprofile.reactive.messaging.Incoming;
 import org.eclipse.microprofile.reactive.messaging.Message;
 import org.jboss.logging.Logger;
@@ -17,12 +16,18 @@ import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
+import java.util.UUID;
+
+import static org.eclipse.microprofile.reactive.messaging.Acknowledgment.Strategy.PRE_PROCESSING;
+
 @ApplicationScoped
 public class EventConsumer {
 
     public static final String INGRESS_CHANNEL = "ingress";
     public static final String REJECTED_COUNTER_NAME = "input.rejected";
     public static final String PROCESSING_ERROR_COUNTER_NAME = "input.processing.error";
+    public static final String DUPLICATE_COUNTER_NAME = "input.duplicate";
+    public static final String CONSUMED_COUNTER_NAME = "input.consumed";
 
     private static final Logger LOGGER = Logger.getLogger(EventConsumer.class);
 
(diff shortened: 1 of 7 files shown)


backend/src/main/java/com/redhat/cloud/notifications/events/KafkaMessageDeduplicator.java
```diff
@@ -0,0 +1,128 @@
...
+        Optional<KafkaMessageMetadata> metadata = message.getMetadata(KafkaMessageMetadata.class);
+        if (metadata.isPresent()) {
+            Iterator<Header> headers = metadata.get().getHeaders().headers(MESSAGE_ID_HEADER).iterator();
+            if (headers.hasNext()) {
+                found = true;
+                Header header = headers.next();
+                if (header.value() == null) {
+                    invalidMessageIdCounter.increment();
+                    LOGGER.warnf("Application %s/%s sent an invalid Kafka header [%s=null]. They must change their " +
+                                    "integration and send a non-null value.", bundleName, applicationName, MESSAGE_ID_HEADER);
```
josejulio, line 66: Do we have this header for policies?

josejulio: and notifications-gw

josejulio: Ah. i see that this is something new.

gwenneg: Yes it's new.

backend/src/main/java/com/redhat/cloud/notifications/models/KafkaMessageId.java
```diff
@@ -0,0 +1,39 @@
...
+public class KafkaMessageId implements Serializable {
+
+    @NotNull
+    public String bundleName;
+
+    @NotNull
+    public String applicationName;
+
+    @NotNull
+    public UUID messageId;
```
gwenneg, line 19: I'm still not sure about the need to store the bundle/app in the database.

If we do that, then we can delegate the responsibility of the `messageId` unicity to the onboarded apps. Duplicates would be their responsibility, not ours.

On the other hand, the chance for a UUID V4 collision is so low that it's very unlikely. Furthermore, the `kafka_message` table is meant to be purged regularly because there's no point storing old messages IDs forever. This reduces even more the collision risk.

Version 4 of UUID is not enforced right now though. They could send us a V3 or lower and it would still be valid for `UUID.fromString`. Maybe we should enforce V4 UUID in the header and get rid of bundle/app in the DB because of the super low collision risk.

WDYT @josejulio?

gwenneg: cc @pilhuhn

gwenneg: You will probably need to have a look at `KafkaMessageDeduplicator` to understand the question since the PR description is still missing. 😄

josejulio: It would make sense to use only V4... but

This github issue scared me a bit: https://github.com/ramsey/uuid/issues/80
TLDR; There is/was a bug on openssl that increases the chance of collision (if the generator uses a particular function of openssl) when using forked process (collisions across the forked)

Anyhow, the collisions would be for a single application. 
I'm +1 on removing the app/bundle and enforce v4.

backend/src/main/java/com/redhat/cloud/notifications/events/KafkaMessageDeduplicator.java
pilhuhn: I think almost no one is using Kafka headers in all of platform (apps)

backend/src/main/java/com/redhat/cloud/notifications/events/KafkaMessageDeduplicator.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+
+    public Uni<Void> registerMessageId(String bundleName, String applicationName, UUID messageId) {
+        if (messageId == null) {
+            return Uni.createFrom().voidItem();
+        } else {
+            KafkaMessage kafkaMessage = new KafkaMessage(bundleName, applicationName, messageId);
+            kafkaMessage.prePersist(); // This method must be called manually while using a StatelessSession.
+            return statelessSession.insert(kafkaMessage);
```
pilhuhn, line 126: Are we cleaning those out?
I think we only need to keep them for a relatively short amount of time (1 day max)

backend/src/main/java/com/redhat/cloud/notifications/models/KafkaMessageId.java
pilhuhn: I wonder if we should store literally as done here, as opposed to use a FK to the apps table, which has a FK to the bundle.

Tbh. what is the collision risk even with that uuid/80 issue? a few hundreds on a million tries? That should be good enough (and the issue is 3-5y old, so most likely fixed)

I think just storing uuid + timestamp could be enough. TS so that we can purge entries older than e.g. a day (I guess the value could be lower)

backend/src/main/java/com/redhat/cloud/notifications/events/KafkaMessageDeduplicator.java
gwenneg: Yes, that's part of the plan (it's not committed yet though).

backend/src/main/java/com/redhat/cloud/notifications/models/KafkaMessageId.java
gwenneg: I don't think we need the relationship between `KafkaMessage` and the app in the database. I added the `bundle`/`application` columns only for unicity purposes but since we all agree that they can be removed, I'll get rid of them. Thanks for your opinion on this @josejulio @pilhuhn!

gwenneg: Done.

backend/src/main/java/com/redhat/cloud/notifications/events/KafkaMessageDeduplicator.java
gwenneg: Cleaner added.

gwenneg:
Description updated, this is ready for a review! 😃

backend/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
```diff
@@ -41,56 +46,128 @@
...
                 .onItemOrFailure().transformToUni((unused, throwable) -> {
-                    // If an exception was thrown during steps 1, 2 or 3, an INFO entry with the payload is added to the log.
                     if (throwable != null) {
+                        /*
+                         * An exception was thrown at some point during the Kafka message processing, it is logged.
+                         */
                         LOGGER.infof(throwable, "Could not process the payload: %s", payload);
                     }
+                    /*
+                     * This counter main purpose is to prevent a test race condition.
```
josejulio, line 167: image

josejulio: ah - nm. It increases on all payloads?  Should the success be counted?

gwenneg: They are counter in `EndpointProcessor` 😃

josejulio (approved):
lgtm

PR #604, opened 2021-09-16
NOTIF-273 Fix RestAssured body serialization with Quarkus 2

gwenneg:
Here's another prep work PR for Quarkus 2.

With Quarkus 1, we currently serialize the RestAssured request bodies with an `ObjectMapper` instance provided by Vert.x. That instance is customized in our test code because we use Jackson filters and `LocalDate(Time)`.

With Quarkus 2, the instance that was provided by Vert.x has been replaced by another one from Quarkus which is not customizable from tests. I created a Quarkus PR that made it customizable but it was refused, so we can no longer use the old way of serializing bodies.

This PR also contains a few unrelated changes that are all required for the Quarkus 2 bump.

### DIFF
--- a/backend/src/test/java/com/redhat/cloud/notifications/routers/AuthenticationTest.java
+++ b/backend/src/test/java/com/redhat/cloud/notifications/routers/AuthenticationTest.java
@@ -16,7 +16,6 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import static io.restassured.RestAssured.given;
-import static io.restassured.http.ContentType.JSON;
 
 @QuarkusTest
 @QuarkusTestResource(TestLifecycleManager.class)
@@ -67,8 +66,7 @@ public class AuthenticationTest {
                 .header(identityHeader)
                 .when().get("/endpoints")
                 .then()
-                .statusCode(403)
-                .contentType(JSON);
+                .statusCode(403);
 
         clearRbacCache();
 
--- a/backend/src/test/java/com/redhat/cloud/notifications/routers/InternalServiceTest.java
+++ b/backend/src/test/java/com/redhat/cloud/notifications/routers/InternalServiceTest.java
@@ -1,5 +1,6 @@
 package com.redhat.cloud.notifications.routers;
 
+import com.redhat.cloud.notifications.Json;
 import com.redhat.cloud.notifications.TestLifecycleManager;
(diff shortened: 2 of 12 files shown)


backend/src/test/java/com/redhat/cloud/notifications/routers/AuthenticationTest.java
```diff
@@ -67,8 +66,7 @@ void testEndpointRoles() {
                 .header(identityHeader)
                 .when().get("/endpoints")
                 .then()
-                .statusCode(403)
-                .contentType(JSON);
+                .statusCode(403);
```
bennetelli, line 69: would this test fail if the order would stay the same?

backend/src/test/java/com/redhat/cloud/notifications/routers/InternalServiceTest.java
```diff
@@ -320,7 +321,7 @@ private static void updateBundle(String bundleId, Bundle bundle, int expectedSta
                 .put("/internal/bundles/{bundleId}")
                 .then()
                 .statusCode(expectedStatusCode)
-                .contentType(familyOf(expectedStatusCode) == Family.SUCCESSFUL ? is(TEXT.toString()) : any(String.class));
+                .contentType(familyOf(expectedStatusCode) == Family.SUCCESSFUL ? containsString(TEXT.toString()) : any(String.class));
```
bennetelli, line 324: did the text change? I mean is() is more precise than contains()

gwenneg: Yes, Quarkus 1 returned the media type only, but Quarkus 2 also returns the UTF-8 encoding.

backend/src/test/java/com/redhat/cloud/notifications/routers/AuthenticationTest.java
gwenneg: Quarkus 2 no longer returns a content type when the returned status is 403, that's why the content type check has to be removed.

bennetelli: ah.. now i get it. you removed the contenttype.. i thought you moved the lines around.. ok, makes sense to me.

bennetelli (approved):

bennetelli:
lgtm

PR #610, opened 2021-09-17
NOTIF-190 Add activate request context annotation

bennetelli:


gwenneg:
Did you reproduce the issue locally (by calling `NotificationsApp#run`) and confirm this is the right fix?

bennetelli:
> Did you reproduce the issue locally (by calling `NotificationsApp#run`) and confirm this is the right fix?

yes I did. Both work. @Transactional on getLastCronJobRun() as well as @ActivateRequestContext on processDailyEmail()

gwenneg (approved):
LGTM

PR #615, opened 2021-09-20
Fix daily aggregation date on new aggregator first execution

gwenneg:
With this PR, the `cronjob_run.last_run` DB field will be updated each time the old aggregator is done processing aggregations. As a consequence, when the new aggregator will run for the first time on prod, it won't use the following date:

https://github.com/RedHatInsights/notifications-backend/blob/ea5d165851c8a736d58423e02741bb67c90cf52d/backend/src/main/resources/db/migration/V1.28.0__NOTIF-190_cronjob_timestamp.sql#L8

Otherwise, this would be sent on prod (it happened on stage today):

image


josejulio (approved):
lgtm

bennetelli (approved):

bennetelli (approved):

bennetelli:
lgtm

PR #616, opened 2021-09-20
NOTIF-273 Add missing configuration value

gwenneg:
`backend` fails to start because of that missing conf value.


gwenneg:
Self-approved to fix a customer issue ASAP.

josejulio:
@gwenneg  Is there any way we could catch these on tests? (missing conf values)

gwenneg:
I'm surprised the app started during the tests. I'll have a look at this later today.

gwenneg:
This is actually related to NOTIF-309 and not NOTIF-273 😄

PR #625, opened 2021-09-21
NOTIF-190 Add gauge to visualize aggregated mails in grafana

bennetelli:


aggregator/pom.xml
```diff
@@ -42,6 +42,14 @@
...
             <artifactId>quarkus-smallrye-reactive-messaging-kafka</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-registry-prometheus</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-micrometer</artifactId>
+        </dependency>
```
gwenneg, line 52:
```suggestion
        <dependency>
            <groupId>io.quarkus</groupId>
            <artifactId>quarkus-micrometer-registry-prometheus</artifactId>
        </dependency>
```

aggregator/src/test/java/com/redhat/cloud/notifications/DailyEmailAggregationJobTest.java
```diff
@@ -136,7 +138,7 @@ void shouldNotIncreaseAggregationsWhenHostIdIsDifferent() {
     void shouldNotProcessMailsWhenNewCronJobIsDisabledByDefault() {
         final EmailAggregationResources emailAggregationResources = mock(EmailAggregationResources.class);
         final ObjectMapper objectMapper = mock(ObjectMapper.class);
-        DailyEmailAggregationJob dailyEmailAggregationJob = new DailyEmailAggregationJob(emailAggregationResources, objectMapper);
+        DailyEmailAggregationJob dailyEmailAggregationJob = new DailyEmailAggregationJob(emailAggregationResources, objectMapper, new SimpleMeterRegistry());
```
gwenneg, line 141: Shouldn't the counter value be checked in all tests?

aggregator/src/test/java/com/redhat/cloud/notifications/DailyEmailAggregationJobTest.java
```diff
@@ -44,6 +45,7 @@
     void setUp() {
         helpers.purgeEmailAggregations();
         System.clearProperty("notifications.aggregator.email.subscription.periodic.cron.enabled");
+        testee.registry = new SimpleMeterRegistry();
```
gwenneg, line 48: Note for later:
Injected fields should not be instantiated with `new` in a `QuarkusTest`, they should be mocked instead.

gwenneg (changes requested):

aggregator/src/test/java/com/redhat/cloud/notifications/DailyEmailAggregationJobTest.java
bennetelli: I actually don't want to test the gauge in that state. Will probably have to add another gauge to visualize more things and will then aggregate all gauges containing to the cron job in the same custom gauge. This is just a first step.

aggregator/src/test/java/com/redhat/cloud/notifications/DailyEmailAggregationJobTest.java
bennetelli: Yes, this test is also a bit misleading as it's testing with mocks, as asll as integration testing without mocks.
I think in the long term it's better to split the test.

gwenneg (approved):
LGTM

PR #630, opened 2021-09-21
NOTIF-314 Prevent in memory events pagination

gwenneg:
This fixes the following warning:
```
HHH000104: firstResult/maxResults specified with collection fetch; applying in memory!
```
The warning cause as well as the solution implemented in this PR are explained here: https://vladmihalcea.com/fix-hibernate-hhh000104-entity-fetch-pagination-warning-message/

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
@@ -8,6 +8,7 @@ import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 import java.time.LocalDate;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
```diff
@@ -30,27 +31,67 @@
...
+    private Uni<List<UUID>> getEventIds(String accountId, Set<UUID> bundleIds, Set<UUID> appIds, String eventTypeDisplayName,
+                                        LocalDate startDate, LocalDate endDate, Integer limit, Integer offset, Optional<String> orderByCondition) {
+        String hql = "SELECT e.id FROM Event e JOIN e.eventType et JOIN et.application a JOIN a.bundle b " +
+                "WHERE e.accountId = :accountId";
+
+        hql = addHqlConditions(hql, bundleIds, appIds, eventTypeDisplayName, startDate, endDate);
 
-        Mutiny.Query<Event> query = session.createQuery(hql, Event.class);
+        if (orderByCondition.isPresent()) {
+            hql += orderByCondition.get();
```
josejulio, line 91: Do we really need the order by here? 
I think it's something that we don't really care about in the list of ids.

gwenneg: Yes, we have to order the records because of the limit.

gwenneg: From the [Postgres doc](https://www.postgresql.org/docs/13/queries-limit.html):

> When using LIMIT, it is important to use an ORDER BY clause that constrains the result rows into a unique order. Otherwise you will get an unpredictable subset of the query's rows. You might be asking for the tenth through twentieth rows, but tenth through twentieth in what ordering? The ordering is unknown, unless you specified ORDER BY.

josejulio: I missed the limit :facepalm: }
nothing to see here :grin:

josejulio (approved):
lgtm

PR #637, opened 2021-09-22
Prevent Kafka warning about a notifications config key

gwenneg:
This will prevent this warning from Kafka which showed up in Sentry:
```
The configuration 'messages.cleaner.period' was supplied but isn't a known config.
```


bennetelli (approved):

bennetelli:
lgtm

PR #640, opened 2021-09-23
NOTIF-316 Silence internal template exceptions

bennetelli:


bennetelli:
@josejulio this solution was the first thing which came to my mind.

gwenneg (commented):
I think the `logger.warnf` line from `EmailTemplateService` should be moved in this method, after the `.asTuple()` line :

https://github.com/RedHatInsights/notifications-backend/blob/854f75a825b8ba89bcd6c050e1669c5d4fa2ae42/backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java#L107-L122

gwenneg:
Previous comment edited.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
```diff
@@ -110,6 +111,16 @@ public void init() {
...
+                .onFailure().invoke(templateEx -> {
+                    if (!(templateEx instanceof TemplateException) && !isInternalUser(user)) {
+                        logger.warnf(templateEx,
+                                "Unable to render template for bundle: [%s] application: [%s], eventType: [%s].",
+                                action.getBundle(),
+                                action.getApplication(),
+                                action.getEventType()
+                        );
+                    }
+                })
```
gwenneg, line 123: We only need the old code here. There's no need for a an `instanceof TemplateException` check or the `isInternalUser` method. All exceptions will be logged at WARN level when an email is being sent (same behavior as before). Nothing will be logged when `EmailTemplateService` was called form the admin UI.

bennetelli:
hope this is what you expected.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
josejulio: Yes, I don't think the internal call reaches this point (no email is being sent).

bennetelli: sorry.. i was a bit confused by this comment.. but I forgot to push my changes.

josejulio (approved):
lgtm

PR #643, opened 2021-09-23
NOTIF-328 Split admin ui from backend code

josejulio:
This PR splits the admin-ui code into a separate project that is disabled by default. 
This will prevent building it when doing a `mvn install`

There is a new check that verifies if there was any change in this module(notifications-admin-console). If there is any the following will happen:
- It will build the admin UI
- If we are on master, the build files will be pushed to ~~[notifications-frontend-admin-build](https://github.com/RedHatInsights/notifications-frontend-admin-build)~~ to this repository, the files will be added to the `backend` module, in its `META-INF` folder.
 

~~Regardless of that the notifications-backend will do the following now:~~
- ~~Download latest master (zip file) from  [notifications-frontend-admin-build](https://github.com/RedHatInsights/notifications-frontend-admin-build)~~
- ~~Place contents in `target/classes/META-INF/resources/internal`~~
- No changes needed here, the compiled admin-console code is placed in META-INF folder.

~~Something else that can be done is to instead of having this separate repo, after building the admin UI, the build results are placed in the backend/META-INF and it's committed directly to this repo.~~


admin-console/pom.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>admin-console</artifactId>
```
gwenneg, line 6: The `artifactId` of other modules is `notifications-*`. Should this one be named `notifications-admin-console`? (without changing the sources folder name that would still be `admin-console`)

josejulio: Sounds good

gwenneg:
> Something else that can be done is to instead of having this separate repo, after building the admin UI, the build results are placed in the backend/META-INF and it's committed directly to this repo.

Is there any difference (advantage or drawback) between the two approaches, besides the files location? If not, I would probably prefer having everything in the same repo.

josejulio:
> Is there any difference (advantage or drawback) between the two approaches, besides the files location? If not, I would probably prefer having everything in the same repo.

I don't know if it would be a drawback, but on merging, a new commit would appear (leading to a new build maybe?) . The previous commit/build would not have the changes applied. (i.e. commit `n` changing admin-console won't have the admin-console, but commit `n+1` would)

josejulio:
As a side note, i though of it after finishing this :-). 
I think i like more the single repo strategy, just wondering if having the extra commit would be an issue.

gwenneg:
With the current PR, will Jenkins automatically build notifications-backend and push a new image to Quay when new files are pushed to [notifications-frontend-admin-build](https://github.com/RedHatInsights/notifications-frontend-admin-build)?

josejulio:
> With the current PR, will Jenkins automatically build notifications-backend and push a new image to Quay when new files are pushed to [notifications-frontend-admin-build](https://github.com/RedHatInsights/notifications-frontend-admin-build)?

Current PR will (from github actions) buiild the admin-console when there are changes and push it to the frontend-admin-build repo.

The regular building for the notifications-backend will pull the code from the frontend-admin-build repo. 

IIUC, the jenkins build will do a build of the notifications-backend (thus pulling the code from the frontend-admin-build repo.

gwenneg:
With the current PR, isn't there a risk that Jenkins will build the Docker image from the old admin-console because the files were not pushed to the other repo before Jenkins started? I mean both GH actions and Jenkins build start at the same time so there could be some kind of race condition here, unless we can make sure Jenkins doesn't start before the GH action is complete.

josejulio:
@gwenneg updated the code, we no longer rely on the extra repository, a new commit is pushed to this repo with the result build in the `META-INF`.

.github/workflows/build.yml
```diff
@@ -30,6 +32,17 @@ jobs:
...
+        id: admin-ui-changed
+        uses: tj-actions/changed-files@v1.1.2
+        with:
+          files: admin-console
+      - name: Build admin UI
+        if: steps.admin-ui-changed.outputs.any_changed == 'true'
+        run: ./mvnw clean verify -Padmin-console -pl :checkstyle -pl :notifications-admin-console
+      - name: Push admin console
+        if: steps.admin-ui-changed.outputs.any_changed == 'true' && github.event_name == 'push' && github.ref == 'refs/heads/master'
+        run: bash .github/scripts/push-admin-console.sh
```
gwenneg, line 45:
```suggestion
      - name: Verify if admin-console changed
        id: admin-console-changed
        uses: tj-actions/changed-files@v1.1.2
        with:
          files: admin-console
      - name: Build admin-console
        if: steps.admin-console-changed.outputs.any_changed == 'true'
        run: ./mvnw clean verify -Padmin-console -pl :checkstyle,:notifications-admin-console
      - name: Push admin-console
        if: steps.admin-console-changed.outputs.any_changed == 'true' && github.event_name == 'push' && github.ref == 'refs/heads/master'
        run: bash .github/scripts/push-admin-console.sh
```

.github/scripts/push-admin-console.sh
```diff
@@ -0,0 +1,15 @@
...
+TARGET="backend/src/main/resources/META-INF/resources/internal"
+
+rm -rf "${TARGET}"
+mkdir -p "${TARGET}"
+cp -r admin-console/src/main/webapp/build/* "${TARGET}"
+
+git config --global user.name "${GITHUB_ACTOR}[automatic action]"
+git config --global user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+git add -f "${TARGET}"
+git commit -m "Updating admin-console build"
```
gwenneg, line 14:
```suggestion
git commit -m "Update admin-console build artifacts"
```

.github/scripts/push-admin-console.sh
```diff
@@ -0,0 +1,15 @@
...
+# When the builds succeeds, we need to place the results in the backend META-INF code and commit it.
+# Delete whatever is there and add the new build.
+
+TARGET="backend/src/main/resources/META-INF/resources/internal"
+
+rm -rf "${TARGET}"
+mkdir -p "${TARGET}"
+cp -r admin-console/src/main/webapp/build/* "${TARGET}"
+
+git config --global user.name "${GITHUB_ACTOR}[automatic action]"
```
gwenneg, line 11: I'm not sure where this will be shown, but maybe the username should be a bit more specific:
```suggestion
git config --global user.name "${GITHUB_ACTOR}[admin-console push script]"
```

josejulio: This will show in the commit

.github/workflows/build.yml
josejulio: thanks!

gwenneg (approved):
LGTM

gwenneg:
The Jenkins build is sooooooo much faster now, I love it!

PR #645, opened 2021-09-24
NOTIF-329 Replace run-java.sh with simple starter

bennetelli:


docker/Dockerfile-build-aggregator.jvm
```diff
@@ -1,42 +1,33 @@
...
     && echo "securerandom.source=file:/dev/urandom" >> /etc/alternatives/jre/lib/security/java.security
 
-# Configure the JAVA_OPTIONS, you can add -XshowSettings:vm to also display the heap size.
 ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
 
 COPY --from=build --chown=1001 /home/jboss/aggregator/target/quarkus-app/ /deployments/
+#COPY --from=build --chown=1001 /home/jboss/aggregator/target/quarkus-app/lib/ /deployments/lib/
+#COPY --from=build --chown=1001 /home/jboss/aggregator/target/quarkus-app/*.jar /deployments/
+#COPY --from=build --chown=1001 /home/jboss/aggregator/target/quarkus-app/app/ /deployments/app/
+#COPY --from=build --chown=1001 /home/jboss/aggregator/target/quarkus-app/quarkus/ /deployments/quarkus/
```
gwenneg, line 28: This is commented. It should probably be removed.

docker/Dockerfile-build-backend.jvm
```diff
@@ -1,42 +1,33 @@
...
     && echo "securerandom.source=file:/dev/urandom" >> /etc/alternatives/jre/lib/security/java.security
 
 # Configure the JAVA_OPTIONS, you can add -XshowSettings:vm to also display the heap size.
 ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
 
-COPY --from=build --chown=1001 /home/jboss/backend/target/quarkus-app/ /deployments/
+COPY --from=build --chown=1001 /home/jboss/backend/target/quarkus-app/lib/ /deployments/lib/
+COPY --from=build --chown=1001 /home/jboss/backend/target/quarkus-app/*.jar /deployments/
+COPY --from=build --chown=1001 /home/jboss/backend/target/quarkus-app/app/ /deployments/app/
+COPY --from=build --chown=1001 /home/jboss/backend/target/quarkus-app/quarkus/ /deployments/quarkus/
```
gwenneg, line 28: Wasn't the old version enough?

docker/Dockerfile-build-aggregator.jvm
```diff
@@ -1,42 +1,33 @@
-####
-# This Dockerfile is used in order to build a container that runs the Quarkus application in JVM mode
-###
-
-# Build the project
 FROM registry.access.redhat.com/ubi8/openjdk-11:latest AS build
 USER root
-COPY . /home/jboss
+COPY . /home/jboss/
```
gwenneg, line 3: Does this trailing slash change something?

docker/Dockerfile-build-backend.jvm
```diff
@@ -1,42 +1,33 @@
-####
-# This Dockerfile is used in order to build a container that runs the Quarkus application in JVM mode
-###
-
-# Build the project
 FROM registry.access.redhat.com/ubi8/openjdk-11:latest AS build
 USER root
-COPY . /home/jboss
+COPY . /home/jboss/
```
gwenneg, line 3: Same question.

docker/Dockerfile-build-backend.jvm
```diff
@@ -1,42 +1,33 @@
-####
```
gwenneg, line 1: Why remove the comments? They can be useful for someone who would be unfamiliar with that kind of file.

docker/Dockerfile-build-backend.jvm
```diff
@@ -1,42 +1,33 @@
...
+COPY --from=build --chown=1001 /home/jboss/backend/target/quarkus-app/lib/ /deployments/lib/
+COPY --from=build --chown=1001 /home/jboss/backend/target/quarkus-app/*.jar /deployments/
+COPY --from=build --chown=1001 /home/jboss/backend/target/quarkus-app/app/ /deployments/app/
+COPY --from=build --chown=1001 /home/jboss/backend/target/quarkus-app/quarkus/ /deployments/quarkus/
 
 EXPOSE 8080
 USER 1001
 
-ENTRYPOINT [ "/deployments/run-java.sh" ]
+ENTRYPOINT ["sh", "-c", "java $JAVA_OPTIONS -jar /deployments/quarkus-run.jar"]
```
gwenneg, line 33: The log from prod show this at startup:
```
exec java -Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager -XX:+ExitOnOutOfMemoryError -cp . -jar /deployments/quarkus-run.jar
```
Shouldn't this file contain the `-XX:+ExitOnOutOfMemoryError` argument?

docker/Dockerfile-build-backend.jvm
bennetelli: Both are doing the same, but splits it in four distinct layers so if there are application changes the library layers can be re-used. This is also the Quarkus way which everyone will get when generating a project via https://code.quarkus.io/

docker/Dockerfile-build-backend.jvm
bennetelli: They were auto generated and the same in every quarkus project. Also the dockerfile is 99% the same as in every Quarkus project generated via https://code.quarkus.io/

I would say they are not useful here. I mean, when "could be useful for someone", we should also comment our java code very often as this code is less intuitive than this dockerfile :)

docker/Dockerfile-build-backend.jvm
bennetelli: I would say no. It didn't contain it in the version before. Not sure where -XX:+ExitOnOutOfMemoryError comes from, but I did not remove it

docker/Dockerfile-build-aggregator.jvm
bennetelli: doesn't make a difference.

docker/Dockerfile-build-backend.jvm
bennetelli: Also doesn't make a difference

docker/Dockerfile-build-backend.jvm
pilhuhn: This is originally buried deep in the run-java.sh script ( https://repo1.maven.org/maven2/io/fabric8/run-java-sh/1.3.8/ ) 

```
  if [ -z "${JAVA_MAJOR_VERSION:-}" ] || [ "${JAVA_MAJOR_VERSION:-}" != "7" ]; then
    opts="${opts} -XX:+ExitOnOutOfMemoryError"
  fi
```

docker/Dockerfile-build-aggregator.jvm
```diff
@@ -1,42 +1,32 @@
...
-###
-
-# Build the project
 FROM registry.access.redhat.com/ubi8/openjdk-11:latest AS build
 USER root
 COPY . /home/jboss
 WORKDIR /home/jboss
 RUN ./mvnw clean package -DskipTests --projects :checkstyle,:notifications-aggregator
 
-# Build the container
```
pilhuhn, line 12: This comment I would keep as not everybody is yet familiar with multi-stage container builds

docker/Dockerfile-build-aggregator.jvm
```diff
@@ -1,42 +1,32 @@
...
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
 
 ARG JAVA_PACKAGE=java-11-openjdk-headless
 ARG RUN_JAVA_VERSION=1.3.8
```
pilhuhn, line 10: No longer needed

docker/Dockerfile-build-backend.jvm
```diff
@@ -1,42 +1,32 @@
...
 USER root
 COPY . /home/jboss
 WORKDIR /home/jboss
 RUN ./mvnw clean package -DskipTests --projects :checkstyle,:notifications-backend
 
-# Build the container
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
 
 ARG JAVA_PACKAGE=java-11-openjdk-headless
 ARG RUN_JAVA_VERSION=1.3.8
```
pilhuhn, line 10: No longer needed

docker/Dockerfile-build-backend.jvm
pilhuhn: I am +1 on splitting. I am not sure though if the CI pipeline really caches the old image or if we always start from scratch. For local builds it definitively helps. And I think when pushing to a repo, it is also quicker, as the lower layers will already be present and don't need to be re-pushed.

pilhuhn (commented):
Some smaller cosmetic comments.

docker/Dockerfile-build-backend.jvm
pilhuhn: the `ARG RUN_JAVA_VERSION=1.3.8` I meant. The comment could/should still be there as in aggregator.

bennetelli: > the `ARG RUN_JAVA_VERSION=1.3.8` I meant. The comment could/should still be there as in aggregator.

what's wrong with ARG RUN_JAVA_VERSION?

docker/Dockerfile-build-backend.jvm
bennetelli: added

docker/Dockerfile-build-backend.jvm
pilhuhn: A few lines down you remove its usage.

docker/Dockerfile-build-backend.jvm
gwenneg: They are generated for a good reason: because they're considered helpful for people unfamiliar with Dockerfiles. I still think we should keep them, not for us because we're all familiar with these files, but for someone who would join the team in the future.

> we should also comment our java code very often

Yes we should and it is actually already done in many classes.

docker/Dockerfile-build-backend.jvm
```diff
@@ -1,42 +1,32 @@
...
-    && chown 1001 /deployments/run-java.sh \
-    && chmod 540 /deployments/run-java.sh \
     && echo "securerandom.source=file:/dev/urandom" >> /etc/alternatives/jre/lib/security/java.security
 
-# Configure the JAVA_OPTIONS, you can add -XshowSettings:vm to also display the heap size.
-ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
+ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager -XX:+ExitOnOutOfMemoryError"
 
-COPY --from=build --chown=1001 /home/jboss/backend/target/quarkus-app/ /deployments/
+COPY --from=build --chown=1001 /home/jboss/backend/target/quarkus-app/lib/ /deployments/lib/
```
gwenneg, line 24:
```suggestion
# We make four distinct layers so if there are application changes the library layers can be re-used
COPY --from=build --chown=1001 /home/jboss/backend/target/quarkus-app/lib/ /deployments/lib/
```

docker/Dockerfile-build-aggregator.jvm
```diff
@@ -1,42 +1,32 @@
...
-    && chown 1001 /deployments/run-java.sh \
-    && chmod 540 /deployments/run-java.sh \
     && echo "securerandom.source=file:/dev/urandom" >> /etc/alternatives/jre/lib/security/java.security
 
-# Configure the JAVA_OPTIONS, you can add -XshowSettings:vm to also display the heap size.
-ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
+ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager -XX:+ExitOnOutOfMemoryError"
 
-COPY --from=build --chown=1001 /home/jboss/aggregator/target/quarkus-app/ /deployments/
+COPY --from=build --chown=1001 /home/jboss/aggregator/target/quarkus-app/lib/ /deployments/lib/
```
gwenneg, line 24:
```suggestion
# We make four distinct layers so if there are application changes the library layers can be re-used
COPY --from=build --chown=1001 /home/jboss/aggregator/target/quarkus-app/lib/ /deployments/lib/
```

docker/Dockerfile-build-backend.jvm
bennetelli: added again

bennetelli: ps: I personally prefer to write the cleanest possible software so its self explaining and we don't need coments :-D

gwenneg: > ps: I personally prefer to write the cleanest possible software so its self explaining and we don't need coments :-D

I completely agree but it's not always possible unfortunately 😃

docker/Dockerfile-build-backend.jvm
bennetelli: removed the argument

gwenneg (approved):
LGTM

bennetelli:
@gwenneg not sure what I can do here.. this is a yarn error and I didn't change something here.

gwenneg:
It looks like a network issue. Retrying the Jenkins job may help fix the error.

pilhuhn (approved):

PR #647, opened 2021-09-24
feat(rhosak): add initial scheduled upgrade template

machi1990:
Also allow all events to be Instant or Daily. The user will have to create a behaviour group and decide how they want to receive them.

Closes MGDSTRM-5089

/cc @davidffrench @akoserwal for internal review 

/cc @gwenneg @josejulio for anything notifications backend related.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/templates/Rhosak.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/templates/Rhosak.java
@@ -4,6 +4,9 @@ import com.redhat.cloud.notifications.models.EmailSubscriptionType;
 import io.quarkus.qute.CheckedTemplate;
 import io.quarkus.qute.TemplateInstance;
 
+import java.util.Arrays;
+import java.util.List;
+
 // Name needs to be "Rhosak" to read templates from resources/templates/Rhosak
 public class Rhosak implements EmailTemplate {
     private static final String SCHEDULED_UPGRADE = "scheduled-upgrade";
@@ -11,6 +14,7 @@ public class Rhosak implements EmailTemplate {
     private static final String INSTANCE_CREATED = "instance-created";
     private static final String INSTANCE_DELETED = "instance-deleted";
     private static final String ACTION_REQUIRED = "action-required";
+    private static final List<String> ALLOWED_EVENT_TYPES = Arrays.asList(SCHEDULED_UPGRADE, DISRUPTION, INSTANCE_DELETED, INSTANCE_CREATED, ACTION_REQUIRED);
 
     @Override
     public TemplateInstance getTitle(String eventType, EmailSubscriptionType type) {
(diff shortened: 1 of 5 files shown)


backend/src/main/java/com/redhat/cloud/notifications/templates/Rhosak.java
```diff
@@ -29,7 +33,9 @@ public TemplateInstance getTitle(String eventType, EmailSubscriptionType type) {
         if (eventType.equals(ACTION_REQUIRED)) {
             return Templates.actionRequiredTitle();
         }
-
+        if (eventType.equals(ACTION_REQUIRED)) {
```
josejulio, line 36:
```suggestion
        if (eventType.equals(SCHEDULED_UPGRADE)) {
```
I think you mean this here

josejulio:
I noticed that there is no change in the aggregator, but several event types were added for daily processing. Is that expected?

backend/src/main/resources/templates/Rhosak/scheduledUpgradeBody.html
```diff
@@ -0,0 +1,82 @@
...
+        <p>This notification is for your {action.events.size()} OpenShift Streams {#if action.events.size() is 1} instance {#else} instances{/if} below.</p>
+        <ul>
+            {#each action.events}
+            <li style="font-size: 14px;">
+                {it.payload.name}
+            </li>
+            {/each}
+        </ul>
+
+        Your {#if action.events.size() is 1} instance is {#else} instances are{/if} scheduled for upgrade maintenance to {action.context.kafka_version} on {action.context.upgrade_time.toUtcFormat()}
```
davidffrench, line 16:
```suggestion
        Your {#if action.events.size() is 1} instance is {#else} instances are{/if} scheduled for upgrade maintenance to {action.context.kafka_version} Kafka version on {action.context.upgrade_time.toUtcFormat()}
```

davidffrench: Might be good to call out that it is a Kafka version upgrade.

davidffrench (commented):
lgtm, just one comment

backend/src/main/java/com/redhat/cloud/notifications/templates/Rhosak.java
machi1990: Good catch. Thanks

backend/src/main/resources/templates/Rhosak/scheduledUpgradeBody.html
machi1990: Makes total sense, thanks.

machi1990:
> I noticed that there is no change in the aggregator, but several event types were added for daily processing. Is that expected?

Great spot. This is indeed true. For the moment, the daily upgrade aggregator is still empty, and so is the template. When we update the template, we'll work on the aggregator at the same time. This will likely be next week. Let me know if you see any issues with this workflow/division of task.

josejulio:
> Let me know if you see any issues with this workflow/division of task.

No issues :-)

josejulio (approved):
lgtm

josejulio:
Let us know when you are ready to merge

machi1990:
> Let us know when you are ready to merge

Let's get this in today. Thanks @josejulio @gwenneg

gwenneg (approved):
LGTM

machi1990:
Thanks

PR #650, opened 2021-09-27
NOTIF-333 Improve metrics to detect service interruptions

gwenneg:
With this PR, the `input.consumed` metric will now have `bundle` and `application` tags.

We should monitor in Grafana the incoming Kafka traffic for each app with that metric and, if possible, configure automatic alerts in case of traffic drop.


gwenneg:
Tests are unstable, fix in progress.

gwenneg:
Tests look stable now, this is ready for a review.

gwenneg:
@josejulio The "test" counter is a real metric now 😄

backend/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
```diff
@@ -52,20 +53,23 @@
...
 
     @Incoming(INGRESS_CHANNEL)
     @Acknowledgment(PRE_PROCESSING)
     public Uni<Void> processAsync(Message<String> message) {
+        // This timer will have dynamic tag values based on the action parsed from the received message.
+        Timer.Sample consumedTimer = Timer.start(registry);
         String payload = message.getPayload();
+        // The two following variables have to be final or effectively final. That why their type is String[] instead of String.
+        String[] bundleName = new String[1];
+        String[] appName = new String[1];
```
josejulio, line 72: Looks like these are pointers to a String

```suggestion
        String *bundleName;
        String *appName;
```

only kidding :grin:

backend/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
```diff
@@ -163,10 +167,10 @@ public void init() {
...
                         LOGGER.infof(throwable, "Could not process the payload: %s", payload);
                     }
-                    /*
-                     * This counter main purpose is to prevent a test race condition.
-                     */
-                    consumedCounter.increment();
+                    // bundleName[0] and appName[0] are null when the action parsing failed.
+                    String bundle = bundleName[0] == null ? "" : bundleName[0];
+                    String application = appName[0] == null ? "" : appName[0];
+                    consumedTimer.stop(registry.timer(CONSUMED_TIMER_NAME, "bundle", bundle, "application", application));
```
josejulio, line 173: is this tested anywhere?

backend/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
gwenneg: That's the idea 😄

backend/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
gwenneg: Good question, the answer is no but I'll add something.

gwenneg: Done.

josejulio (approved):
lgtm

PR #660, opened 2021-09-28
Add RDS graph + fixed JVM Metrics

theute:
image

### DIFF
--- a/.rhcicd/grafana/grafana-dashboard-notifications-general.configmap.yaml
+++ b/.rhcicd/grafana/grafana-dashboard-notifications-general.configmap.yaml
@@ -18,9 +18,21 @@ data:
       "editable": true,
       "gnetId": null,
       "graphTooltip": 1,
-      "iteration": 1631861900028,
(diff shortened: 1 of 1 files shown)


.rhcicd/grafana/grafana-dashboard-notifications-general.configmap.yaml
```diff
@@ -2369,7 +2547,7 @@ data:
       "timezone": "",
       "title": "Notifications Dashboard",
       "uid": "KQIVyFuMk",
-      "version": 10
+      "version": 1
```
bennetelli, line 2550: updating the version to 11 would be great. Then we would be able to restore the older dashboards. Not sure if we would ever need it, but it could be useful.

theute: Ok I thought this was ignored and rolling back in this repo would be enough. Will send an update

bennetelli (approved):

PR #661, opened 2021-09-28
NOTIF-291 Finish the event log schema migration and clean code

gwenneg:
This PR will be a draft until I've verified the data migration won't cause any issue with the current production data (I'll write app-interface queries tomorrow) but you can already review it.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/models/Event.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/models/Event.java
@@ -30,8 +30,8 @@ public class Event extends CreationTimestamped {
     @Size(max = 50)
     private String accountId;
 
-    // TODO [Event log phase 2] Make this field @NotNull (in the SQL schema as well) and the @ManyToOne relationship not optional.
-    @ManyToOne
+    @NotNull
+    @ManyToOne(optional = false)
     @JoinColumn(name = "event_type_id")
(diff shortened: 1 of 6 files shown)


josejulio (approved):

gwenneg:
```
  # Check that the notification_history.event_id field is never null before adding a not null constraint.
  - "select count(id) from notification_history where event_id is null;"
  # Check that the event.event_type_id field is never null before adding a not null constraint.
  - "select count(id) from event where event_type_id is null;"
```

gwenneg:
The queries result is actually normal. We didn't wait for the 31 days retention delay so the `event` table contains records resulting from the `V.1.27.0` data migration. These records don't have a link between `event` and `event_type`. They will be purged soon by `EventLogCleaner`.

Let's put this PR on hold for now. I'll check the prod database again in 7 days.

gwenneg:
```
  # Check that the notification_history.event_id field is never null before adding a not null constraint.
  - "select count(id) from notification_history where event_id is null;"
  # Check that the event.event_type_id field is never null before adding a not null constraint.
  - "select count(id) from event where event_type_id is null;"
```
```
 count
-------
     0
(1 row)

 count
-------
     0
(1 row)
```

gwenneg:
Rebased to fix conflicts.

PR #663, opened 2021-09-29
Adding JVM dashboard

theute:
Adding a Grafana dashboard for JVM metrics.
This isn't specific to Notifications but that's the most convenient place I know right now...

This is a modified version of: https://grafana.com/grafana/dashboards/4701

This is to provision: 
https://grafana.stage.devshift.net/d/nID67wNnz/jvm-micrometer?orgId=1&from=now-1h&to=now&refresh=30s

### DIFF
--- a/.rhcicd/grafana/grafana-dashboard-jvm.configmap.yaml
+++ b/.rhcicd/grafana/grafana-dashboard-jvm.configmap.yaml
@@ -1,0 +1,4843 @@
+apiVersion: v1
+data:
+  Notifications.json: |-
+    {
+      "annotations": {
+        "list": [
+          {
+            "builtIn": 1,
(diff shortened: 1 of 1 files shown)


.rhcicd/grafana/grafana-dashboard-jvm.configmap.yaml
```diff
@@ -0,0 +1,4843 @@
...
+          "24h",
+          "2d",
+          "7d",
+          "30d"
+        ]
+      },
+      "timezone": "browser",
+      "title": "JVM (Micrometer)",
+      "uid": "nID67wNnz",
+      "version": 5
```
bennetelli, line 4835: i think this version could be 1

PR #699, opened 2021-10-10
fix(rhosak): change system_check_in time by start_time for daily digest report

machi1990:
The daily context object does not contain system_check_in time but start_time. Let's change the template to reflect that.

/cc @gwenneg


machi1990:
Favor request: Once merged, we would like to get it to Stage env for easy of testing.

gwenneg:
> Favor request: Once merged, we would like to get it to Stage env for easy of testing. 

Sure, we'll update stage tomorrow.

PR #705, opened 2021-10-12
Add a default limit to the endpoints /history API

gwenneg:
This should prevent an `OutOfMemoryError` when the `/history` API is called without a reasonnable limit for an endpoint that would have a huge number of history entries in the database.

The default limit will be enforced to 500 if:
- the request has no limit
- the request has a limit higher than 500 (in that case, the `offset` parameter from the query will be ignored as well)


gwenneg:
cc @josejulio @aferd
This PR may have an impact on the frontend. In your absence, I'll do my best to try and determine if we need to change something in the frontend.

josejulio:
> cc @josejulio @aferd This PR may have an impact on the frontend. In your absence, I'll do my best to try and determine if we need to change something in the frontend.

I can take a look if you want

PR #706, opened 2021-10-13
Increase RBAC S2S calls read timeout

gwenneg:
This PR is meant to fix:
```
io.vertx.core.http.impl.NoStackTraceTimeoutException: The timeout period of 2000ms has been exceeded while executing GET /api/rbac/v1/principals/?admin_only=false&offset=0&limit=1000 for server rbac******HIDDEN******
```
It was thrown on prod right after we deployed Quarkus 2.

The `rbac_get_users_page_seconds_max` metric shows durations over 30 seconds (max 45 seconds) multiple times on a daily basis.

We have to increase the rest-client timeout if we don't want lots of RBAC queries to be interrupted by the same exception.


gwenneg:
User actions from the frontend are affected by the RBAC timeout increase.

PR #708, opened 2021-10-13
[NOTIF-355] Enhancing the Camel support in notifications-backend

pilhuhn:
* Messages from/to Camel are now encoded as CloudEvents.
* An example of a camel sender is provided, which just logs the payload received and returns success

The format of the CloudEvents is not final.

This needs more work -- especially tests. 

cc @beav


gwenneg:
The test failure is certainly caused by the invalid reactive messaging topology because the channels are now enabled but there is no source in test mode. `InMemoryConnector` will fix that, I'll push something later.

PR #714, opened 2021-10-18
fix(rhosak): add goto kafka instance page link

machi1990:


machi1990:
Thanks @gwenneg this is good to get in once CI is happy

PR #735, opened 2021-11-01
Add auto formatter maven plugin

bennetelli:


bennetelli:
This ia a new try. I just found it in the quarkiverse extensions. Quarkus is already using this plugin.
Unfortunately it looks like it's not compatible with checkstyle..

josejulio:
> Unfortunately it looks like it's not compatible with checkstyle..

If this is automatic and still does a good job we could use it. WDYT?
I'm gonna check what it does with the code.

bennetelli:
> > Unfortunately it looks like it's not compatible with checkstyle..
> 
> If this is automatic and still does a good job we could use it. WDYT? I'm gonna check what it does with the code.

yea. most of the files changed in this PR are changed by the plugin. If we think it's a good idea, then we could give it a try.

gwenneg:
I would prefer if the plugin didn't enforce lines breaks, lines max length and code indentation.

backend/src/main/java/com/redhat/cloud/notifications/routers/NotificationService.java
```diff
@@ -259,20 +270,14 @@
...
-                                            .map(BehaviorGroup::getActions)
-                                            .filter(Objects::nonNull)
-                                            .flatMap(Collection::stream)
-                                            .map(BehaviorGroupAction::getEndpoint)
-                                            .collect(Collectors.toList())
-                            )
-                    );
+            return getAccountId(sec).onItem()
+                    .transformToUni(accountId -> behaviorGroupResources.findByBundleId(accountId, bundleId)).onItem()
+                    .call(behaviorGroups -> endpointResources.loadProperties(behaviorGroups.stream()
```
gwenneg, line 278: Here's an example where the reactive code is harder to read after the indentation in enforced.

`.onItem().transformToUni(...)` and `.onItem().call(...)` are kind of a single reactive instruction and should be kept together for the sake of readability.

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointEmailSubscriptionResources.java
```diff
@@ -15,77 +15,68 @@
...
-                "ON CONFLICT (account_id, user_id, application_id, subscription_type) DO NOTHING"; // The value is already on the database, this is OK
+    public Uni<Boolean> subscribe(String accountNumber, String username, String bundleName, String applicationName,
+            EmailSubscriptionType subscriptionType) {
+        String query = "INSERT INTO endpoint_email_subscriptions(account_id, user_id, application_id, subscription_type) "
+                + "SELECT :accountId, :userId, a.id, :subscriptionType "
+                + "FROM applications a, bundles b WHERE a.bundle_id = b.id AND a.name = :applicationName AND b.name = :bundleName "
+                + "ON CONFLICT (account_id, user_id, application_id, subscription_type) DO NOTHING"; // The value is
+                                                                                                     // already on the
+                                                                                                     // database, this
+                                                                                                     // is OK
```
gwenneg, line 26: That kind of thing is also bad. The comment only applies to line 23. We should be able to write both the line and the full comment in a single line.

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
```diff
@@ -18,23 +18,17 @@
...
-        return session.createQuery(query, EmailAggregation.class)
-                .setParameter("accountId", key.getAccountId())
-                .setParameter("bundleName", key.getBundle())
-                .setParameter("applicationName", key.getApplication())
-                .setParameter("start", start)
-                .setParameter("end", end)
-                .getResultList();
+        return session.createQuery(query, EmailAggregation.class).setParameter("accountId", key.getAccountId())
+                .setParameter("bundleName", key.getBundle()).setParameter("applicationName", key.getApplication())
+                .setParameter("start", start).setParameter("end", end).getResultList();
```
gwenneg, line 31: This code is harder to read after the formatting.

gwenneg:
I won't put more examples here but there are plenty of them.

josejulio:
> I would prefer if the plugin didn't enforce lines breaks / max length and code indentation.

Is this possible? It's a killer for these things that @gwenneg mentioned.

gwenneg:
> Is this possible? It's a killer for these things that @gwenneg mentioned.

I hope so 😃 That kind of plugin usually has a default configuration which can be changed to match the projects needs. I never tried to change the `formatter-maven-plugin` configuration though so I'm not 100% sure it's possible.

bennetelli:
> > Is this possible? It's a killer for these things that @gwenneg mentioned.
> 
> I hope so smiley That kind of plugin usually has a default configuration which can be changed to match the projects needs. I never tried to change the `formatter-maven-plugin` configuration though so I'm not 100% sure it's possible.

It actually is possible. But I will need to download and configure eclipse for it probably. I will do it when there is some time between two tasks or so. its not that urgent at the moment

PR #740, opened 2021-11-02
NOTIF-361 Protect Event log with new RBAC permission

bennetelli:


backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacIdentityProvider.java
```diff
@@ -108,6 +110,9 @@
                                                 })
                                                 // Otherwise, we can finish building the QuarkusSecurityIdentity and return the result
                                                 .onItem().transform(rbacRaw -> {
+                                                    if (rbacRaw.canRead("events", "events")) {
```
josejulio, line 113:
```suggestion
                                                    if (rbacRaw.canRead("notifications", "events")) {
```
Can't really remember the order, but one is the "application" and other is the "resource"

backend/src/test/resources/rbac-examples/rbac_example_full_access.json
```diff
@@ -15,6 +15,10 @@
       "permission": "notifications:*:*",
       "resourceDefinitions": []
     },
+    {
+      "permission": "events:*:*",
```
josejulio, line 19: We probably don't need this one, the first is the application (notifications). We are not creating a new `application`, only a new `resource`

bennetelli: i just tested it. When I remove this, the EventServiceTest.testAllQueryParams() fails.

josejulio: It goes with the other comment :)

backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacIdentityProvider.java
bennetelli: The first one is the application. The second one is an item.. maybe the parameter has got a misleading name.. probably this is the resource.

How should this be changed?

josejulio: Item should be the resource (or I confused the RBAC lingo). So my suggestion looks OK

josejulio:
I think this one should stay as draft, unless the permission already exists.

bennetelli:
> I think this one should stay as draft, unless the permission already exists.

see: https://github.com/RedHatInsights/rbac-config/pull/217

backend/src/test/resources/rbac-examples/rbac_example_full_access.json
bennetelli: ok, then it's resolved I guess :)

PR #742, opened 2021-11-03
Remove `@Schema` annotations

gwenneg:
`quarkus-smallrye-openapi` was updated and now correctly deals with `@JsonNaming(SnakeCaseStrategy.class)`.


gwenneg:
I made a before/after comparison for all openapi generated files. They are identical.

josejulio (approved):
lgtm

PR #745, opened 2021-11-03
Replace NotificationService NPE with a clean HTTP error

gwenneg:
This will fix a NPE that triggered Sentry alerts during QE tests.


josejulio (approved):
lgtm

PR #746, opened 2021-11-03
Cleanup OpenShift Advisor templates and tests

epapbak:
- Remove "weekly-digest" dead code
- Cleanup related UTs
- Can't format time in email title using local time format (java.time.LocalDateTime.format does not have such parameter), so simplified the corresponding UT to avoid it failing when local time is not in English.


backend/src/test/java/com/redhat/cloud/notifications/templates/TestOpenshiftAdvisorTemplate.java
```diff
@@ -19,7 +19,8 @@ public void testInstantEmailTitle() {
                 .data("action", action)
                 .render();
 
-        assertEquals("OpenShift - Advisor Instant Notification - 20 May 2021", result, "Title is the expected.");
+        assertTrue(result.contains("OpenShift - Advisor Instant Notification - "));
```
pilhuhn, line 22: You could use `startsWith()`. But it is good as is

josejulio (approved):

josejulio:
@epapbak Is OK to merge this?

gwenneg:
@epapbak Can we merge?

PR #751, opened 2021-11-03
Support multiple types in the endpoint listing API

josejulio:
- Add missing properties in camel update

~~edit: need to update tests~~

### DIFF
--- a/backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointServiceTest.java
+++ b/backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointServiceTest.java
@@ -464,6 +464,136 @@ public class EndpointServiceTest extends DbIsolatedTest {
     }
 
     @Test
+    void testEndpointTypeQuery() {
+        String tenant = "limiter";
+        String userName = "user";
+        String identityHeaderValue = TestHelpers.encodeIdentityInfo(tenant, userName);
+        Header identityHeader = TestHelpers.createIdentityHeader(identityHeaderValue);
+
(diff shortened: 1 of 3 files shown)


bennetelli:
somwhow I can't checkout the branch, but we should fix the failed tests first.

josejulio:
@bennetelli fixed the issue and added a new test

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointServiceTest.java
```diff
@@ -463,6 +463,136 @@ void testEndpointUpdates() {
...
+        camelProperties.setSubType("demo");
+
+        Endpoint camelEp = new Endpoint();
+        camelEp.setType(EndpointType.CAMEL);
+        camelEp.setName("endpoint to find");
+        camelEp.setDescription("needle in the haystack");
+        camelEp.setEnabled(true);
+        camelEp.setProperties(camelProperties);
+
+        response = given()
```
bennetelli, line 519: I would prefer two tests so its clear what the responsibility of the test is.
We can probably split it in a camel and webhook test don't we?

josejulio: The purpose of the test is to actually verify that the `type` param is used correctly and that it supports both 1 type or multiple types.

I could probably make the test more focused If you consider it is necesary

josejulio: Parametrized the test, let me know what you think.

bennetelli: That's much better :) thanks

bennetelli (approved):

PR #756, opened 2021-11-04
NOTIF-361 Protect Event Log with new RBAC permission

bennetelli:


bennetelli:
closes this https://github.com/RedHatInsights/notifications-backend/pull/740 as I did some crap while merging.

backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacIdentityProvider.java
```diff
@@ -30,6 +30,7 @@
 @ApplicationScoped
 public class RbacIdentityProvider implements IdentityProvider<RhIdentityAuthenticationRequest> {
 
+    public static final String RBAC_READ_EVENTS_NOTIFICATIONS = "read:events:notifications";
```
josejulio, line 33: This name is more concrete and mimics exactly what we should have in RBAC, but is not consistent with our other internal RBAC permissions. RBAC server permissions have a `verb`, a `resource` and an `application` while our internal ones only have a `verb` and a `resource`. 

No strong opinion, but to be consistent we should rename it to have the `verb` and `resource`.

```suggestion
    public static final String RBAC_READ_EVENTS_NOTIFICATIONS = "read:events";
```

bennetelli: Changed

josejulio (approved):
lgtm

backend/src/main/java/com/redhat/cloud/notifications/routers/EventService.java
```diff
@@ -50,7 +50,7 @@
 
     @GET
     @Produces(APPLICATION_JSON)
-    @RolesAllowed(RBAC_READ_NOTIFICATIONS)
+    @RolesAllowed(RBAC_READ_EVENTS_NOTIFICATIONS)
```
gwenneg, line 53: Could you add a test that tries to access this API with the `RBAC_READ_NOTIFICATIONS` permission?

bennetelli: added.

theute:
I think like Josejulio on the other PR :-) 
https://github.com/RedHatInsights/notifications-backend/pull/740#issuecomment-957504872

backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest.java
```diff
@@ -479,6 +479,26 @@ void testInvalidLimit() {
...
 
+    @Test
+    void shouldNotBeAllowedToGetEventLogs() {
+        Header identityHeader = mockRbac(DEFAULT_ACCOUNT_ID, "noaccess", NO_ACCESS);
+        given()
+                .header(identityHeader)
+                .when().get(PATH)
+                .then()
+                .statusCode(403);
+    }
```
gwenneg, line 490: This case is already tested here:
https://github.com/RedHatInsights/notifications-backend/blob/a103e4cd771007814295860501fc8d8fa0ed5d01/backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest.java#L441-L449

backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest.java
```diff
@@ -479,6 +479,26 @@ void testInvalidLimit() {
...
+
+    @Test
+    void shouldBeAllowedToGetEventLogs() {
+        Header noAccessIdentityHeader = mockRbac("tenant", "user", FULL_ACCESS);
+        given()
+                .header(noAccessIdentityHeader)
+                .when().get(PATH)
+                .then()
+                .statusCode(200);
+    }
```
gwenneg, line 500: The `FULL_ACCESS` case is already tested multiple times in `EventServiceTest`.

gwenneg (commented):
What I meant with my previous comment is that we should make sure that someone with the `RBAC_READ_NOTIFICATIONS` role cannot access the event log API.

backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest.java
bennetelli: oops :X

backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest.java
bennetelli: so what else do you want to be tested then?

bennetelli:
> I think like Josejulio on the other PR :-) [#740 (comment)](https://github.com/RedHatInsights/notifications-backend/pull/740#issuecomment-957504872)

afaik Keith ran all that stuff.

josejulio:
Converting to draft, the role does not seem to exist on prod or ci

bennetelli:
> Converting to draft, the role does not seem to exist on prod or ci

how do you check that? I always asked keith..

backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest.java
gwenneg: See my comment above:
> What I meant with my previous comment is that we should make sure that someone with the RBAC_READ_NOTIFICATIONS role can no longer access the event log API.

backend/src/test/resources/rbac-examples/rbac_example_events_notifications_notifications_read_access.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+  "links": {
+    "first": "/api/rbac/v1/access/?application=%2A&limit=10&offset=0",
+    "next": null,
+    "previous": null,
+    "last": "/api/rbac/v1/access/?application=%2A&limit=10&offset=0"
+  },
+  "data": [
+    {
+      "permission": "notifications:blabla:read",
```
josejulio, line 15:
```suggestion
      "permission": "notifications:notifications:read",
```

bennetelli: added

bennetelli:
need to add a new test here. The mock somehow returns 200 when I add the tests to the existing EventServiceTest

josejulio:
That's weird. Could you merge it back into a file so that we might be able to check helping what's wrong?

bennetelli:
the @QuarkusTestResource(TestLifecycleManager.class) should be triggered before any test is run. 
So before each test, there should be a new MockServerClientConfig object. The code in mockServerConfig.addMockRbacAccess() is the same in both tests. 

Maybe I overlooked something?

backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacIdentityProvider.java
```diff
@@ -30,6 +30,7 @@
 @ApplicationScoped
 public class RbacIdentityProvider implements IdentityProvider<RhIdentityAuthenticationRequest> {
 
+    public static final String RBAC_READ_EVENTS_NOTIFICATIONS = "read:events";
```
gwenneg, line 33: Shouldn't this be named `RBAC_READ_EVENTS` to be consistent with the other roles naming convention?

gwenneg (commented):
The tests should validate this:
| Role | Before this PR | After |
| - | - | - |
| RBAC_READ_NOTIFICATIONS | Could access the event log | Can _no longer_ access the event log |
| RBAC_READ_EVENTS_NOTIFICATIONS | Didn't exist | Can access the event log |

backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacIdentityProvider.java
bennetelli: To be really consistent it should be similar to RBAC_READ_INTEGRATIONS_ENDPOINTS. RBAC_READ_NOTICIATIONS_EVENTS i would say.

backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest2.java
```diff
@@ -0,0 +1,51 @@
...
+                .when().get(PATH)
+                .then()
+                .statusCode(403);
+    }
+
+    @Test
+    void shouldNotBeAllowedTogetEventLogsWhenUserHasNotificationsAccessRightsOnly() {
+        Header noAccessIdentityHeader = mockRbac(NOTIFICATIONS_ACCESS_ONLY);
+        given()
+                .header(noAccessIdentityHeader)
```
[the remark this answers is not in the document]
bennetelli: the RBAC_READ_NOTIFICATIONS test are already there. shouldNotBeAllowedTogetEventLogsWhenUserHasNotificationsAccessRightsOnly()

i have added the other one in the last commit.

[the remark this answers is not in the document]
bennetelli: I also change the header stuff

backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest2.java
```diff
@@ -0,0 +1,51 @@
...
+
+@QuarkusTest
+@QuarkusTestResource(TestLifecycleManager.class)
+public class EventServiceTest2 extends DbIsolatedTest {
+
+    @MockServerConfig
+    MockServerClientConfig mockServerConfig;
+
+    @Test
+    void shouldNotBeAllowedToGetEventLogsWhenUserHasWrongAccessRights() {
```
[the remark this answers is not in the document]
bennetelli: with point you mean test? There is a test which covers this case already.

backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest.java
```diff
@@ -75,7 +78,27 @@
     MockServerClientConfig mockServerConfig;
 
     // A new instance is automatically created by JUnit before each test is executed.
-    private ModelInstancesHolder model = new ModelInstancesHolder();
+    private final ModelInstancesHolder model = new ModelInstancesHolder();
+
+    @Test
+    void shouldNotBeAllowedToGetEventLogsWhenUserHasWrongAccessRights() {
+        Header noAccessIdentityHeader = mockRbac("tenant", "user", WRONG_ACCESS);
```
josejulio, line 85: I suspect the `mockRbac` call is not resetting the previous access. That would explain why you get `200` in the tests. Could you update it so that we have a `beforeEach` that resets the `MockServerClientConfig` instance on every call?

@gwenneg WDYT?

gwenneg: I think you're right. All tests share the same `mockserver` container.

Calling [MockServerClientConfig#clearRbac](https://github.com/RedHatInsights/notifications-backend/blob/1af499740a5478949819353b025851ab347102bf/backend/src/test/java/com/redhat/cloud/notifications/MockServerClientConfig.java#L55) before each test will probably fix the issue. Or you could pass a different username to `mockRbac` when the permissions change.

bennetelli: Adding a beforeEach and clearRbac() call didn't work, using a different username works. I think we can move forward with it, but it's still a workaround and we could lead to issues like these again.

backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacIdentityProvider.java
gwenneg: I never had a look at [rbac-config/tree/master/configs](https://github.com/RedHatInsights/rbac-config/tree/master/configs) before today. Now that I did it, I changed my mind and I think `RBAC_READ_NOTIFICATIONS_EVENTS` better reflects the actual RBAC permissions. :smiley:

bennetelli:
the JavaScript stuff still fails, but this isn't related to my changes in the Java classes I would say.

backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest2.java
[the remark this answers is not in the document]
gwenneg: What I meant is that this test doesn't make sense to me. We don't need to test an imaginary permission.

backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest.java
gwenneg: `FULL_ACCESS` is already tested by the existing code. We don't need an additional test for that.

backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest2.java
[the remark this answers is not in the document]
bennetelli: deleted.

backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest.java
bennetelli: deleted

backend/src/test/resources/rbac-examples/rbac_example_events_notifications_read_access_only.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+  "links": {
+    "first": "/api/rbac/v1/access/?application=%2A&limit=10&offset=0",
+    "next": null,
+    "previous": null,
+    "last": "/api/rbac/v1/access/?application=%2A&limit=10&offset=0"
+  },
+  "data": [
+    {
+      "permission": "notifications:*:read:",
```
gwenneg, line 15: There's an harmless typo (because we split on `:`) here:
```suggestion
      "permission": "notifications:*:read",
```
Otherwise, LGTM.

gwenneg (approved):
LGTM

gwenneg:
The PR check failed because of a timeout which is not related to this PR. I triggered a Jenkins job retry.

gwenneg:
Ephemeral is still not completely stable. Some projects (not us) sometimes have big variation in their deployment duration.

backend/src/test/resources/rbac-examples/rbac_example_events_notifications_read_access_only.json
bennetelli: I changed it myself instead of the commit suggestion. 

Thanks for the hint :)

PR #757, opened 2021-11-04
NOTIF-336 Work around an ephemeral RBAC timeout issue

gwenneg:
Tests from `pr_check.sh` are failing because RBAC isn't ready fast enough when the first test is run.


theute:
I'm glad ENV_NAME is useful for once :-D

PR #760, opened 2021-11-08
NOTIF-295 edit event types 

aferd:


admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -45,14 +58,24 @@ export const ApplicationPage: React.FunctionComponent = () => {
             name: name ?? '',
             description: description ?? ''
         });
```
josejulio, line 60: You could call the query here if you want to reload on create/update:
`mutate(...).then(eventTypesQuery.query);`

And probably you would have to reset the id/displayName/etc as well

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -75,42 +92,53 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
+                                        onClick={ createEventType }> Create Event Type </Button>
                                     <Modal
                                         variant={ ModalVariant.medium }
                                         title={ `Create Event Type for ${ (applicationTypesQuery.loading ||
                                             applicationTypesQuery.payload?.status !== 200) ?
                                             <Spinner /> : applicationTypesQuery.payload.value.displayName }` }
-                                        isOpen={ isOpen }
-                                        onClose={ toggle }
+                                        isOpen={ showModal }
+                                        onClose={ () => setShowModal(false) }
```
josejulio, line 103: In general, is better to use `React.useCallback` (and `useMemo`) whenever is possible. It helps optimize the rendering (prevent unesesary renders)

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -75,42 +92,53 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
                                     ><Form isHorizontal>
                                             <FormGroup label='Name' fieldId='name' isRequired
                                                 helperText='This is a short name, only composed of a-z 0-9 and - characters.'>
                                                 <TextInput
                                                     type='text'
-                                                    value={ name }
-                                                    onChange={ setName }
-                                                    id='name' /></FormGroup>
+                                                    value={ event.name }
+                                                    onChange={ () => setEvent }
```
josejulio, line 110: Is this call working?

aferd: no, I am working on this currently.

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -75,42 +97,56 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
-                                                <Button variant='primary' type='submit' isDisabled={ !name || !displayName }
-                                                    onClick= { onSubmit }>Submit</Button>
-                                                <Button variant='link' onClick={ toggle }>Cancel</Button>
+                                                <Button variant='primary' type='button'
+                                                    isDisabled={ isEdit }
+                                                    { ...(newEvent.loading || newEvent.payload?.status !== 200) ?
+                                                        <Spinner /> : eventTypesQuery.payload.value }
+                                                    onClick={ handleSubmit }>Submit</Button>
+                                                <Button variant='primary' type='button' isDisabled={ !isEdit }
+                                                    onClick={ handleSubmit }>Update</Button>
```
josejulio, line 143: We should hide instead of disable (for the edit vs create flow) the buttons depending the case

josejulio:
Could you add a screenshot/gif of the interaction? Thanks!

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -63,8 +84,9 @@ export const ApplicationPage: React.FunctionComponent = () => {
         <React.Fragment>
             <PageSection>
                 <Title headingLevel="h1"><Breadcrumb>
-                    <BreadcrumbItem to='#'> { getBundle.isLoading ?
-                        <Spinner /> : getBundle.bundles.map(bundle => bundle.displayName)} </BreadcrumbItem>
+                    <BreadcrumbItem to='#'>
+                        { getBundle.bundles.map(bundle => bundle.displayName) } </BreadcrumbItem>
```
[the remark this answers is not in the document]
aferd: that was not meant for this PR. I got switched around last Friday between branches.

aferd:
Hide button instead of disabled depending on case

https://user-images.githubusercontent.com/86322239/141865089-01b2e360-0cd8-4469-9ccc-c8726540f957.mov

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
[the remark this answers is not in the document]
josejulio: Could you rever this change then?

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -28,31 +32,52 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
-    const [ id ] = React.useState<string | undefined>();
-    const [ displayName, setDisplayName ] = React.useState<string | undefined>();
-    const [ name, setName ] = React.useState<string | undefined>();
-    const [ description, setDescription ] = React.useState<string | undefined>();
+    const [ eventType, setEventType ] = React.useState<Partial<EventType>>({});
+
+    const [ showModal, setShowModal ] = React.useState(false);
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [ isEdit, setIsEdit ] = React.useState(false);
+    const [ visableButton, setVisableButton ] = React.useState(false);
```
josejulio, line 40: It looks like `isEdit` is not used anywhere and on glance, it could replace or be-replaced-by the `visableButon`. 
Could you double check and remove the unused ones?

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -75,42 +101,54 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
+                                                    id='description'
+                                                    name="description" /></FormGroup>
                                             <ActionGroup>
-                                                <Button variant='primary' type='submit' isDisabled={ !name || !displayName }
-                                                    onClick= { onSubmit }>Submit</Button>
-                                                <Button variant='link' onClick={ toggle }>Cancel</Button>
+                                                <Button variant='primary' type='button'
+                                                    { ...(newEvent.loading || newEvent.payload?.status !== 200) ?
+                                                        <Spinner /> : eventTypesQuery.payload.value }
+                                                    onClick={ handleSubmit } onClickCapture={ () => setVisableButton(!visableButton) }>
```
josejulio, line 144: What's the purpose of the `onClickCapture` here? I feel is not needed here.

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -75,42 +101,54 @@ export const ApplicationPage: React.FunctionComponent = () => {
                         <Toolbar>
                             <ToolbarContent>
                                 <ToolbarItem>
-                                    <Button variant='primary' type='button' onClick={ toggle }> Create Event Type </Button>
+                                    <Button variant='primary' type='button'
+                                        onClick={ createEventType }> Create Event Type </Button>
                                     <Modal
                                         variant={ ModalVariant.medium }
                                         title={ `Create Event Type for ${ (applicationTypesQuery.loading ||
```
josejulio, line 107: The `isEdit` (or `visable`) could help here to find out if we are Creating an event type or Updating an event type.

josejulio (commented):
Looking good! Some comments/asks

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
aferd: no purpose after looking at it again, was thinking it needed a button prop to be executed initially.

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -75,42 +96,54 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
+                                                    id='description'
+                                                    name="description" /></FormGroup>
                                             <ActionGroup>
-                                                <Button variant='primary' type='submit' isDisabled={ !name || !displayName }
-                                                    onClick= { onSubmit }>Submit</Button>
-                                                <Button variant='link' onClick={ toggle }>Cancel</Button>
+                                                <Button variant='primary' type='button'
+                                                    { ...(newEvent.loading || newEvent.payload?.status !== 200) ?
+                                                        <Spinner /> : eventTypesQuery.payload.value }
+                                                    onClick={ handleSubmit } { ...() => setIsEdit(!isEdit) }>
```
josejulio, line 139: What does `{ ...() => setIsEdit(!isEdit) }` do ?

aferd: i am getting a "... expected " when trying to do this { setIsEdit(!idEdit) } and { ...setIsEdit(!isEdit) } crashes with too many re-renders - though I can just add it to the onClick `onClick={ handleSubmit || setIsEdit(!isEdit }`

josejulio: I actually don't think you need the `() => setIsEdit(!isEdit)` at all.

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -75,42 +96,54 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
+                                                    id='description'
+                                                    name="description" /></FormGroup>
                                             <ActionGroup>
-                                                <Button variant='primary' type='submit' isDisabled={ !name || !displayName }
-                                                    onClick= { onSubmit }>Submit</Button>
-                                                <Button variant='link' onClick={ toggle }>Cancel</Button>
+                                                <Button variant='primary' type='button'
+                                                    { ...(newEvent.loading || newEvent.payload?.status !== 200) ?
+                                                        <Spinner /> : eventTypesQuery.payload.value }
+                                                    onClick={ handleSubmit || setIsEdit(!isEdit) }>
```
josejulio, line 139: The handleSubmit is fine without the `setIsEdit`
```suggestion
                                                    onClick={ handleSubmit }>
```

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
aferd: i dont! I have the isEdit on the next line - overthinking what I was trying to do there .. felt I needed to set it more than once for some reason. removed!

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
josejulio: I think this is the only remaining ask I have here.
Could you conditionally change the title?
something like: `${isEdit ? 'Update' : 'Create'}` instead of the `Create`

aferd: of course!

josejulio (approved):
lgtm!

PR #762, opened 2021-11-09
Simplify Docker build scripts

gwenneg:
This also fixes a few camel-sender-demo issues.


bennetelli (approved):
lgtm. whereas I still think it's smarter to split the projects instead of building and deploying camel-sender and aggregator when we change notifications-backend.

gwenneg:
I agree, we should split later.

PR #763, opened 2021-11-09
NOTIF-373 Introduce EphemeralDataInitializer

gwenneg:
This PR introduces data initialization on ephemeral environments.

▶️ **Which data?**

The init will persist any provided `Bundle`, `Application` or `EventType`.

We can't initialize any other kind of data because it would require an `accountId` which can only be found when a REST API is called from the UI (or the python helpers) or when we receive a Kafka message.

▶️ **When?**

The init will only happen if the `ENV_NAME` environment variable contains a value that starts with `env-ephemeral`.

It is guaranteed to happen _after_ the Flyway DB migration is done. For now, we're migrating the DB with [FlywayWorkaround](https://github.com/RedHatInsights/notifications-backend/blob/master/backend/src/main/java/com/redhat/cloud/notifications/db/FlywayWorkaround.java) which observes `StartupEvent`, so I introduced `FlywayEndEvent` to enforce an order and make sure `FlywayWorkaround` runs first. `FlywayWorkaround` will be removed in the future and this won't break the enforced execution order because the Flyway execution from the `quarkus-flyway` extension happens before the beans observing `StartupEvent` are invoked.

▶️ **How?**

There are two ways to init data on ephemeral:
- Persistent: commit a change to the `src/main/resources/ephemeral/ephemeral_data.json` file.
- Temporary: put the data in the `NOTIFICATIONS_EPHEMERAL_DATA` environment variable.

Both can be used at the same time.

Since the `src/main/resources/ephemeral/ephemeral_data.json` file is currently empty, a data example is provided in  `src/main/resources/ephemeral/ephemeral_data.sample.json`.

Here's an example of data passed to the app with the environment variable:
```
NOTIFICATIONS_EPHEMERAL_DATA="{\"bundles\":[{\"name\":\"my-bundle\",\"display_name\":\"My Bundle\",\"applications\":[{\"name\":\"my-app\",\"display_name\":\"My Application\",\"event_types\":[{\"name\":\"my-event-type\",\"display_name\":\"My Event Type\",\"description\":\"This is my event type\"}]}]}]}"
```

▶️ **Why use JSON for the input data?**

Because it will be easier to use for developers who are not familiar with the programming languages we use in notifications.

### DIFF
--- a/.rhcicd/clowdapp-backend.yaml
+++ b/.rhcicd/clowdapp-backend.yaml
@@ -79,6 +79,12 @@ objects:
           value: ${MP_MESSAGING_INCOMING_INGRESS_MAX_POLL_INTERVAL_MS}
         - name: MP_MESSAGING_INCOMING_INGRESS_MAX_POLL_RECORDS
           value: ${MP_MESSAGING_INCOMING_INGRESS_MAX_POLL_RECORDS}
+        - name: NOTIFICATIONS_EPHEMERAL_DATA
+          valueFrom:
+            configMapKeyRef:
+              name: notifications-ephemeral-data
+              key: ephemeral_data.json
+              optional: true
         - name: PROCESSOR_EMAIL_BOP_APITOKEN
           valueFrom:
             secretKeyRef:
--- a/README.adoc
+++ b/README.adoc
@@ -9,3 +9,97 @@ which is invoked when running maven.
 
 This project uses the Clowder Config Source from https://github.com/RedHatInsights/clowder-quarkus-config-source.
 To configure this source to use a different file than `/cdappconfig/cdappconfig.json` you can use the property `clowder.file=/path/to/file.json`.
+
+## Deploying the backend with data on ephemeral
+
+If you deploy `notifications-backend` on ephemeral, the database may not contain all the data you need for your tests.
+It is possible to load data on ephemeral when the backend pod starts, but this is restricted to three kinds of database
+records: `Bundle`, `Application` and `EventType`.
+
+There are two ways of loading that data, as explained below. Both can be used at the same time as long as the `name`
+field is globally unique for each type of database record. If there is no `name` conflict, the data from both sources
+will be inserted into the database.
+
+### Loading data using a ConfigMap
+
+The `notifications-backend` ClowdApp template contains an environment variable definition that can be used to load data:
+
+```yaml
+env:
+- name: NOTIFICATIONS_EPHEMERAL_DATA
+  valueFrom:
+    configMapKeyRef:
+      name: notifications-ephemeral-data
+      key: ephemeral_data.json
+      optional: true
+```
+
+If a `ConfigMap` named `notifications-ephemeral-data` is created by any of the pods present in the ephemeral namespace,
+the backend pod will consume that `ConfigMap` as an environment variable and put the value of the `ephemeral_data.json`
+key into the `NOTIFICATIONS_EPHEMERAL_DATA` environment variable.
+
+[TIP]
+The `ConfigMap` is optional, it is not a requirement for the `notifications-backend` pod deployment.
+
+Here is an example of the `ConfigMap` you could add to your application ClowdApp template:
+
+```yaml
+- apiVersion: v1
+  kind: ConfigMap
+  metadata:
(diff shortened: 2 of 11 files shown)


gwenneg:
I have a test ready, but the JVM isn't very friendly with env vars changes in tests (unless we add a new dependency) 😢

gwenneg:
I tried a lot of things/workarounds to test this PR, including the [junit-pioneer](https://search.maven.org/artifact/org.junit-pioneer/junit-pioneer/1.4.2/jar) extension. Nothing worked. That's because of the nature of `EphemeralDataInitializer`: it reads the environment variables at application startup. Changes to these variables applied during tests happen too late.

pilhuhn:
This is only a feeling and not a hard fact (or request for change): I'd prefer that the data comes from a ConfigMap and not an env-variable.
The env-variable could though be populated from a ConfigMap.

gwenneg:
@pilhuhn Storing the data in a ConfigMap wouldn't change anything to the current code, would it? Our app would still read the data from the environment variable.

pilhuhn:
Yes, if we use the `valueFrom:` mechanism 

```
env:
        # Define the environment variable
        - name: SPECIAL_LEVEL_KEY
          valueFrom:
            configMapKeyRef:
              # The ConfigMap containing the value you want to assign to SPECIAL_LEVEL_KEY
              name: special-config
              # Specify the key associated with the value
              key: special.how

```
(from https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#define-container-environment-variables-using-configmap-data )

josejulio:
> Here's an example of data passed to the app with the environment variable:
> 
> ```
> NOTIFICATIONS_EPHEMERAL_DATA="{\"bundles\":[{\"name\":\"my-bundle\",\"display_name\":\"My Bundle\",\"applications\":[{\"name\":\"my-app\",\"display_name\":\"My Application\",\"event_types\":[{\"name\":\"my-event-type\",\"display_name\":\"My Event Type\",\"description\":\"This is my event type\"}]}]}]}"
> ```

Could you document that into the README. Would be great if you added a config-map example too

gwenneg:
Sure.

gwenneg:
Is this what you had in mind @pilhuhn?

```
[...]
- apiVersion: v1
  kind: ConfigMap
  metadata:
    name: notifications-ephemeral-data
  data:
    ephemeral-data: |
      {
        "bundles": [
          {
            "name": "my-bundle",
            "display_name": "My Bundle",
            "applications": [
              {
                "name": "my-app",
                "display_name": "My Application",
                "event_types": [
                  {
                    "name": "my-event-type",
                    "display_name": "My Event Type",
                    "description": "This is my event type"
                  }
                ]
              }
            ]
          }
        ]
      }
[...]
        env:
        - name: NOTIFICATIONS_EPHEMERAL_DATA
          valueFrom:
            configMapKeyRef:
              name: notifications-ephemeral-data
              key: ephemeral-data
[...]
```
OpenShift is happy with that and the env variable looks OK, but I have yet to test it with this PR.

gwenneg:
Doc updated: [deploying-the-backend-with-data-on-ephemeral](https://github.com/RedHatInsights/notifications-backend/blob/3f2a74ac877da6ee9bacbdb3bcaa2156567b04a2/README.adoc#deploying-the-backend-with-data-on-ephemeral)

backend/src/main/java/com/redhat/cloud/notifications/epehemeral/Application.java
```diff
@@ -0,0 +1,14 @@
+package com.redhat.cloud.notifications.epehemeral;
```
josejulio, line 1: small typo on the package name: should be `ephemeral` and not  ep**e**hemeral

josejulio (approved):
LGTM, small typo on the package name

gwenneg:
> LGTM, small typo on the package name

Good catch, that's fixed.

gwenneg:
I renamed the `ephemeral-data` ConfigMap key to `ephemeral_data.json`.

pilhuhn (approved):
Perhaps add one line (to the README) about what happens if env-var and internal file are present

PR #765, opened 2021-11-10
NOTIF-377 Persist Kafka messages ID before processing

gwenneg:
Check `Hide whitespace` for an easier review.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
@@ -122,50 +122,49 @@ public class EventConsumer {
                                     } else {
                                         /*
                                          * Step 4
-                                         * The message ID is new. We need to retrieve an EventType from the DB using the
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
```diff
@@ -122,50 +122,49 @@ public void init() {
...
-                                                )
-                                                .onFailure().invoke(() -> {
-                                                    /*
-                                                     * A NoResultException was thrown because no EventType was found. The
-                                                     * message is therefore considered rejected.
-                                                     */
-                                                    rejectedCounter.increment();
-                                                })
-                                                .onItem().transformToUni(eventType -> {
+                                        return kafkaMessageDeduplicator.registerMessageId(messageId)
```
josejulio, line 128: I'm wondering if in the future we should write the outcome of the processing, like if it was a failure and at what point to retry the message.

josejulio (approved):
lgtm

PR #770, opened 2021-11-11
NOTIF-360 Reject messages with invalid bundle app eventtype

bennetelli:


backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
```diff
@@ -0,0 +1,33 @@
...
+    @GET
+    public Uni<Response> validate(@QueryParam("bundle") String bundleName, @QueryParam("application") String applicationName, @QueryParam("eventtype") String eventType) {
+        return appResources.getEventType(bundleName, applicationName, eventType)
+                .onItem()
+                .transform(this::isValid);
+    }
+
+    private Response isValid(EventType e) {
+        if (e != null) {
+            return Response.ok().entity("bundle found").build();
```
pilhuhn, line 28: Bundle found is probably misleading, as the combo of bundle/app/eT needed to match.
Same below.

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
```diff
@@ -0,0 +1,33 @@
...
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.Response;
+
+@Path("/validation")
+public class ValidationEndpoint {
+
+    @Inject
+    ApplicationResources appResources;
+
+    @GET
```
pilhuhn, line 19: I wonder if we should add another path segment here (`@Path("/baet")`) in case we want to offer more validations in the future.

pilhuhn (approved):
Generally ok, smaller comments

pilhuhn:
I wonder how the setup of b/a/et for testing is done. Do we provide them somewhere already?

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
```diff
@@ -0,0 +1,33 @@
...
+import com.redhat.cloud.notifications.models.EventType;
+import io.smallrye.mutiny.Uni;
+
+import javax.inject.Inject;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.Response;
+
+@Path("/validation")
```
gwenneg, line 13: This path should either start with `com.redhat.cloud.notifications.Constants.INTERNAL` if this is an internal API or `com.redhat.cloud.notifications.Constants.API_NOTIFICATIONS_V_1_0` if it's supposed to be accessed over the Internet.

FYI all internal APIs are currently regrouped in [InternalService](https://github.com/RedHatInsights/notifications-backend/blob/master/backend/src/main/java/com/redhat/cloud/notifications/routers/InternalService.java).

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
gwenneg: I don't think an entity is needed in case of success.

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
```diff
@@ -0,0 +1,33 @@
...
+        return appResources.getEventType(bundleName, applicationName, eventType)
+                .onItem()
+                .transform(this::isValid);
+    }
+
+    private Response isValid(EventType e) {
+        if (e != null) {
+            return Response.ok().entity("bundle found").build();
+        } else {
+            return Response.status(404).entity("did not find triple of bundle").build();
```
gwenneg, line 30: [EventConsumer](https://github.com/RedHatInsights/notifications-backend/blob/8ba668f4dd6a10c075eb8424d32c679b06ad6e95/backend/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java#L36) logs this message: `No event type found for [bundleName=%s, applicationName=%s, eventTypeName=%s]`. Maybe something similar (params names are a bit different) could be used here as well?

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
```diff
@@ -0,0 +1,33 @@
...
+
+@Path("/validation")
+public class ValidationEndpoint {
+
+    @Inject
+    ApplicationResources appResources;
+
+    @GET
+    public Uni<Response> validate(@QueryParam("bundle") String bundleName, @QueryParam("application") String applicationName, @QueryParam("eventtype") String eventType) {
+        return appResources.getEventType(bundleName, applicationName, eventType)
```
gwenneg, line 21: FYI this method depends on an Hibernate _stateless_ session which is not something we usually use from REST APIs (that kind of session comes with many limitations). It's not causing any harm with the current code, but we may need to change the session to a stateful one in the future.

backend/src/test/java/com/redhat/cloud/notifications/routers/ValidationEndpointTest.java
```diff
@@ -0,0 +1,81 @@
...
+class ValidationEndpointTest {
+
+    @MockServerConfig
+    MockServerClientConfig mockServerConfig;
+
+    @InjectMock
+    ApplicationResources appResources;
+
+    @Test
+    void shouldReturnNotFoundWhenTripleIsInvalid() {
```
gwenneg, line 31: The mocked `ApplicationResources` should throw a `NoResultException` when the event type is not found to match the runtime code.

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
```diff
@@ -0,0 +1,33 @@
...
+                .onItem()
+                .transform(this::isValid);
+    }
+
+    private Response isValid(EventType e) {
+        if (e != null) {
+            return Response.ok().entity("bundle found").build();
+        } else {
```
gwenneg, line 29: This `else` block will be reached with the current tests, but not with the runtime code. When the event type is not found, `ApplicationResources#getEventType` will throw a `NoResultException`. See [EventConsumer](https://github.com/RedHatInsights/notifications-backend/blob/8ba668f4dd6a10c075eb8424d32c679b06ad6e95/backend/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java#L136).

backend/src/test/java/com/redhat/cloud/notifications/routers/ValidationEndpointTest.java
```diff
@@ -0,0 +1,81 @@
...
+        final Response response = given()
+                .header(identityHeader)
+                .when()
+                .param("bundle", "blabla")
+                .param("application", "Notifications")
+                .param("eventtype", "Any")
+                .get("/validation")
+                .then()
+                .statusCode(404)
+                .extract().response();
```
gwenneg, line 51:
```suggestion
                .extract().asString();
```
is a shortcut for `.extract().response().asString()`. It works with `pretty` as well.

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
```diff
@@ -0,0 +1,33 @@
...
+import javax.ws.rs.core.Response;
+
+@Path("/validation")
+public class ValidationEndpoint {
+
+    @Inject
+    ApplicationResources appResources;
+
+    @GET
+    public Uni<Response> validate(@QueryParam("bundle") String bundleName, @QueryParam("application") String applicationName, @QueryParam("eventtype") String eventType) {
```
gwenneg, line 20:
```suggestion
    public Uni<Response> validate(@QueryParam("bundle") String bundleName, @QueryParam("application") String applicationName, @QueryParam("eventType") String eventType) {
```
or event shorter with the new `quarkus-resteasy-reactive` annotations:
```suggestion
    public Uni<Response> validate(@RestQuery String bundleName, @RestQuery String applicationName, @RestQuery String eventType) {
```

gwenneg (changes requested):

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
bennetelli: It's not needed. I just wanted to be able to be able to separate between any possible success and the specific one in our case.

pilhuhn: While it is not needed, it is good to have it. But the text may be misleading.

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
pilhuhn: We should provide a `@Produces` annotation that tells about the returned media type. And depending on it also adjust messages (entities) we return.

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
bennetelli: I would say it's an internal one then. Do you agree @pilhuhn ?

backend/src/test/java/com/redhat/cloud/notifications/routers/ValidationEndpointTest.java
bennetelli: Thanks :)

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
bennetelli: done. thx

pilhuhn:
> I would say it's an internal one then. Do you agree @pilhuhn ?

Yes

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
```diff
@@ -20,7 +21,12 @@
     public Uni<Response> validate(@RestQuery String bundle, @RestQuery String application, @RestQuery String eventType) {
         return appResources.getEventType(bundle, application, eventType)
                 .onItem()
-                .transform(this::isValid);
+                .transform(this::isValid)
+                .onFailure(NoResultException.class).recoverWithItem(this::convertToResponse);
```
bennetelli, line 25: This still doesn't work as expected. It probably reacts on failures in the previous transform method instead of the getEventType() call.. I don't know how to fix it. Right now it returns a http 500 instead of a 404 in case an NoResultException appears.. can someone help me fixing it?

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
bennetelli: changed.

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
```diff
@@ -0,0 +1,46 @@
...
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+import static com.redhat.cloud.notifications.Constants.INTERNAL;
+
+@Path(INTERNAL + "/validation")
```
josejulio, line 19: I'm wondering if we should move this to the Internal file or create a new `internal` submodule (i.e. `routers/internal`)  and put all our internal API's in there

bennetelli: I was wondering the same. Whereas I don't like the grouping by internal. This probably leads to putting things that don't belong together from a domain perspective in the same class.
The same as using Util classes. Usually these classes belong pretty much everything and the methods inside utils classes have nothing in common.

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
josejulio: I think this is now unreachable code (even on tests) - We would probably could get rid of this function as well

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
bennetelli: added as well

josejulio (approved):

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
```diff
@@ -0,0 +1,45 @@
...
+public class ValidationEndpoint {
+
+    private static final String EVENT_TYPE_NOT_FOUND_MSG = "No event type found for [bundleName=%s, applicationName=%s, eventTypeName=%s]";
```
gwenneg, line 22:
```suggestion
    private static final String EVENT_TYPE_NOT_FOUND_MSG = "No event type found for [bundle=%s, application=%s, eventType=%s]";
```

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Inject
+    ApplicationResources appResources;
+
+    @GET
+    @Consumes(MediaType.APPLICATION_JSON)
```
gwenneg, line 28:
```suggestion
```

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
```diff
@@ -0,0 +1,45 @@
+package com.redhat.cloud.notifications.routers;
+
+import com.redhat.cloud.notifications.db.ApplicationResources;
+import com.redhat.cloud.notifications.models.EventType;
+import io.smallrye.mutiny.Uni;
+import org.jboss.resteasy.reactive.RestQuery;
+
+import javax.inject.Inject;
+import javax.persistence.NoResultException;
+import javax.ws.rs.Consumes;
```
gwenneg, line 10:
```suggestion
```

backend/src/main/java/com/redhat/cloud/notifications/routers/ValidationEndpoint.java
```diff
@@ -0,0 +1,45 @@
...
+    ApplicationResources appResources;
+
+    @GET
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/baet")
+    public Uni<Response> validate(@RestQuery String bundle, @RestQuery String application, @RestQuery String eventType) {
+        return appResources.getEventType(bundle, application, eventType)
+                .onItem()
+                .transform(this::convertToOkayResponse)
```
gwenneg, line 34: Optional change (the current version works):
```suggestion
                .replaceWith(() -> Response.ok().build())
```

pilhuhn: That would also require to remove the method below.

pilhuhn:
That is what I get from using the "commit suggestion" button

```
Error:    ValidationEndpointTest.shouldReturnNotFoundWhenTripleIsInvalid:52 expected: <No event type found for [bundleName=blabla, applicationName=Notifications, eventTypeName=Any]> but was: <No event type found for [bundle=blabla, application=Notifications, eventType=Any]>
```

Will not use that anymore ;-)

backend/src/test/java/com/redhat/cloud/notifications/routers/ValidationEndpointTest.java
```diff
@@ -0,0 +1,78 @@
...
+                .when()
+                .param("bundle", "blabla")
+                .param("application", "Notifications")
+                .param("eventType", "Any")
+                .get(INTERNAL + "/validation/baet")
+                .then()
+                .statusCode(404)
+                .extract().asString();
+
+        assertEquals("No event type found for [bundleName=blabla, applicationName=Notifications, eventTypeName=Any]", response);
```
pilhuhn, line 52:
```suggestion
        assertEquals("No event type found for [bundle=blabla, application=Notifications, eventType=Any]", response);
```

gwenneg (approved):

PR #772, opened 2021-11-12
NOTIF-336 Fix clowdapp.yaml deployment name

gwenneg:
This is the preferred deployment naming which is already used in most migrated Clowdapps.


gwenneg:
IQE tests are failing because of this PR, this is expected.

PR #775, opened 2021-11-12
NOTIF-373 Add ephemeral data and fix data init

gwenneg:
This PR:
- Adds to the ephemeral data the same `Bundle`/`Application`/`EventType` we currently have on CI.
- Fixes the insertion of data that would already exist in our Flyway scripts.


josejulio (approved):
lgtm

gwenneg:
The PR check is currently broken (fix in progress), it's not related to the changes in this PR.

PR #776, opened 2021-11-12
NOTIF-336 Add Postgres verification algorithm to clowdapp.yaml

gwenneg:
This is a temporary fix, the value should come from `clowder-quarkus-config-source` eventually.


gwenneg:
I'm merging without waiting for the checks, I need this ASAP on stage.

PR #785, opened 2021-11-15
Remove sort and limit from the notification details API

gwenneg:
The query will never return more than one DB record. There's no point sorting or limiting the data.


gwenneg:
@josejulio The frontend should stop sending the sort/limit parameters as well.

josejulio:
> @josejulio The frontend should stop sending the sort/limit parameters as well.

The UI is not using this API

gwenneg:
Yeah that's what I just understood from the Slack discussion 😄

josejulio (approved):

PR #788, opened 2021-11-16
Reverting changes on the aggregator code

josejulio:
- Do not parse using the action parser.
   Aggregation does not follow the same encoding rules

Reverts changes on `EmailAggregator` from https://github.com/RedHatInsights/notifications-backend/pull/595

`Hide whitespace` makes it easier to review

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregator.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregator.java
@@ -68,7 +68,7 @@ public class EmailAggregator {
                                     subscribers.onItem().transformToUni(users ->
                                             /*
                                              * The actual recipients list may differ from the candidates depending on the endpoint properties and the action settings.
-                                             * The target endpoints properties and the action settings will determine whether or not each candidate will actually receive an email.
+                                             * The target endpoints properties will determine whether or not each candidate will actually receive an email.
                                              */
                                             recipientResolver.recipientUsers(
                                                     aggregationKey.getAccountId(),
--- a/backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregator.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregator.java
@@ -11,9 +11,7 @@ import com.redhat.cloud.notifications.processors.email.aggregators.AbstractEmail
 import com.redhat.cloud.notifications.processors.email.aggregators.EmailPayloadAggregatorFactory;
 import com.redhat.cloud.notifications.recipients.RecipientResolver;
 import com.redhat.cloud.notifications.recipients.User;
-import com.redhat.cloud.notifications.recipients.request.ActionRecipientSettings;
 import com.redhat.cloud.notifications.recipients.request.EndpointRecipientSettings;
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregator.java
```diff
@@ -60,42 +57,38 @@
...
-                                         * Let's populate with synchronized side-effects (invoke) the Map that will be returned by the method.
-                                         */
-                                        .onItem().invoke(users -> {
-                                            users.forEach(user -> {
-                                                // It's aggregation time!
-                                                fillUsers(aggregationKey, user, aggregated, aggregation);
-                                            });
-                                        });
+                    // We need its event type to determine the target endpoints.
+                    String eventType = getEventType(aggregation);
```
josejulio, line 61: This is still not quite right, but it's what we had before (and similar to what we had before this PR).
We might need to collect all event types so that it's better.

gwenneg (changes requested):
I haven't reviewed this yet, but don't merge it.

I need to test another PR (pod shutdown fix) before this one is merged.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregator.java
```diff
@@ -60,42 +57,38 @@
                 .onItem().transformToMulti(Multi.createFrom()::iterable)
                 // For each aggregation...
                 .onItem().transformToUniAndConcatenate(aggregation -> {
-                    // We need to parse the action to determine the target endpoints.
-                    return actionParser.fromJsonString(aggregation.getPayload().encode())
```
josejulio, line 64: It was failing here because our aggregation payload is not a valid `Action` nor do we need it to be a valid (at least not for now). 
It's build by the aggregation code and is not encoded to `Action`, so the payload and context are actually json objects and not json strings.

backend/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregator.java
```diff
@@ -60,42 +57,38 @@
...
+                    return endpointResources.getTargetEndpointsFromType(aggregationKey.getAccountId(), aggregationKey.getBundle(), aggregationKey.getApplication(), eventType, EndpointType.EMAIL_SUBSCRIPTION)
+                            .onItem().transform(Set::copyOf)
+                            // Now we want to determine who will actually receive the aggregation email.
+                            .onItem().transformToUni(endpoints ->
+                                    // All users who subscribed to the current application and subscription type combination are recipients candidates.
+                                    subscribers.onItem().transformToUni(users ->
+                                            /*
+                                             * The actual recipients list may differ from the candidates depending on the endpoint properties and the action settings.
+                                             * The target endpoints properties and the action settings will determine whether or not each candidate will actually receive an email.
+                                             */
```
gwenneg, line 72: Shouldn't this comment be reverted as well since the recipients no longer come from the action?

gwenneg (approved):
LGTM

gwenneg:
⚠️ We're still not ready for a merge ⚠️

josejulio:
> warning We're still not ready for a merge warning

I'm going to draft until we are ready - feel free to undraft and merge.

PR #803, opened 2021-11-24
Test GH build with JDK 17

gwenneg:
Quick and dirty test to see if the app builds with JDK 17.


gwenneg:
Some checks failed for unrelated reasons. The notifications-backend build was successful with JDK 17.

PR #804, opened 2021-11-24
Increase backend PR check timeout

gwenneg:
The current 360 seconds timeout can lead to PR check failures if some dependencies are slower than usual.

Let's see how it goes with 420 seconds.


gwenneg:
The CodeQL failure is unrelated.

PR #805, opened 2021-11-24
Fix aggregator minor issues

gwenneg:


aggregator/pom.xml
```diff
@@ -80,11 +72,6 @@
             <artifactId>quarkus-flyway</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-junit5</artifactId>
```
bennetelli, line 85: did you remove it because it's included in the quarkus-junit5-mockito dependency already?

gwenneg: Yes

bennetelli: ok thx

bennetelli (approved):

PR #806, opened 2021-11-24
NOTIF-296 delete event type

aferd:


admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -139,7 +166,7 @@ export const ApplicationPage: React.FunctionComponent = () => {
                                                 helperText='This is a short name, only composed of a-z 0-9 and - characters.'>
                                                 <TextInput
                                                     type='text'
-                                                    value={ eventType.name }
+                                                    defaultValue={ eventType.name }
```
josejulio, line 169: I don't think we need to change this to `defaultValue`. `value` seems fine.

```suggestion
                                                    value={ eventType.name }
```

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -148,16 +175,22 @@ export const ApplicationPage: React.FunctionComponent = () => {
                                                 helperText='This is the name you want to display on the UI'>
                                                 <TextInput
                                                     type='text'
-                                                    value={ eventType.displayName }
+                                                    defaultValue={ eventType.displayName }
```
josejulio, line 178: I don't think we need to change this to defaultValue. value seems fine.

```suggestion
                                                    value={ eventType.displayName }
```

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -148,16 +175,22 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
+                                                    onChange={ handleChange }
+                                                    id='display-name' /></FormGroup>
```
josejulio, line 180: This is duplicated (see line below)
```suggestion
```

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -148,16 +175,22 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
                                                     onChange={ handleChange }
                                                     id='display-name'
                                                     name="displayName" /></FormGroup>
+
```
josejulio, line 184: nitpick, we can live without this space :D 
```suggestion
```

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -148,16 +175,22 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
                                             <FormGroup label='Description' fieldId='description'
                                                 helperText='Optional short description that appears in the UI
                                                 to help admin descide how to notify users.'>
                                                 <TextArea
                                                     type='text'
-                                                    value={ eventType.description }
+
+                                                    defaultValue={ eventType.description }
```
josejulio, line 191:
```suggestion
                                                    value={ eventType.description }
```

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -148,16 +175,22 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
+                                                    onChange={ handleChange }
+                                                    id='description' /></FormGroup>
```
josejulio, line 193: Duplicated?
```suggestion
```

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -173,6 +206,30 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
+                                                <Spinner /> }/${ (applicationTypesQuery.loading
+                                             || applicationTypesQuery.payload?.status !== 200) ?
+                                                <Spinner /> : applicationTypesQuery.payload.value.displayName } will be deleted. 
+                                                If an application is currently sending this event, it will no longer be processed.`}
+                                            <br />
+                                            <br />
+                                            Type <b>{ eventType.name }</b> to confirm:
+                                            <br />
+                                            <TextInput type='text' onChange={ handleDeleteChange }
+                                                defaultValue={ eventType.name } name="name" id='name' isRequired />
```
josejulio, line 223: Lets use value:

```suggestion
                                                value={ eventType.name } name="name" id='name' isRequired />
```

josejulio: https://en.reactjs.org/docs/uncontrolled-components.html#default-values

josejulio (changes requested):
Some changes - As a next step it would be great to split each modals into its own file.

aferd:
working on splitting the modals from Application.

josejulio:
> working on splitting the modals from Application, should I do a seperate PR or attach to this one?

Separate one is fine

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -150,17 +177,15 @@ export const ApplicationPage: React.FunctionComponent = () => {
                                                     type='text'
                                                     value={ eventType.displayName }
                                                     onChange={ handleChange }
-                                                    id='display-name'
-                                                    name="displayName" /></FormGroup>
```
josejulio, line 154: Isn't the name needed for the edit ?

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -150,17 +177,15 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
+                                                    id='display-name' /></FormGroup>
                                             <FormGroup label='Description' fieldId='description'
                                                 helperText='Optional short description that appears in the UI
                                                 to help admin descide how to notify users.'>
                                                 <TextArea
                                                     type='text'
                                                     value={ eventType.description }
                                                     onChange={ handleChange }
-                                                    id='description'
-                                                    name="description" /></FormGroup>
```
josejulio, line 163: same as above

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -173,6 +198,30 @@ export const ApplicationPage: React.FunctionComponent = () => {
                                         <>
                                         </>
                                     </Modal>
+                                    <React.Fragment>
+                                        <Modal variant={ ModalVariant.small } titleIconVariant="warning" isOpen={ showDeleteModal }
+                                            onClose={ () => setShowDeleteModal(false) }
+                                            title={ `Permanetly delete ${ eventType.name }` }>
```
josejulio, line 204:
```suggestion
                                            title={ `Permanently delete ${ eventType.name }` }>
```

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
aferd: yes, didn't delete that on purpose. adding back.

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -17,26 +17,32 @@ import { useParameterizedQuery } from 'react-fetching-library';
...
 import { useCreateEventType } from '../services/CreateEventTypes';
+import { useDeleteEventType } from '../services/DeleteEventType';
 import { useApplicationTypes } from '../services/GetApplication';
 import { getBundleAction  } from '../services/GetBundleAction';
 import { useEventTypes } from '../services/GetEventTypes';
 import { EventType } from '../types/Notifications';
 
 type ApplicationPageParams = {
     applicationId: string;
+    eventTypeId: string;
```
josejulio, line 28: I didn't notice this before. I think this one is not used.
The `ApplicationPageParams` is used to get the types of the route params.

```suggestion
```

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -173,6 +202,30 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
+                                                <Spinner /> }/${ (applicationTypesQuery.loading
+                                             || applicationTypesQuery.payload?.status !== 200) ?
+                                                <Spinner /> : applicationTypesQuery.payload.value.displayName } will be deleted. 
+                                                If an application is currently sending this event, it will no longer be processed.`}
+                                            <br />
+                                            <br />
+                                            Type <b>{ eventType.name }</b> to confirm:
+                                            <br />
+                                            <TextInput type='text' onChange={ handleDeleteChange }
+                                                value={ eventType.name } name="name" id='name' isRequired />
```
aferd, line 219: this is no longer working after being set to value vs. defaultValue... not sure why?

josejulio: Ah... 

What is happening here is that we don't update the `eventType.name` ever on the delete modal. And that's fine.
It used to work, because `defaultValue` only used the first value, ignores the changes and instead it keeps a record on its own.
When `value` is set, it uses that value always.

See: https://en.reactjs.org/docs/uncontrolled-components.html#default-values

I suggest to remove `value` too so that it uses its own internal state.

aferd: okay, I was just reading that doc snippet again from when you posted it prior - thanks for explaining!

josejulio (approved):
lgtm

PR #809, opened 2021-11-29
NOTIF-315 Adds support for default behavior

josejulio:
This PR adds a behavior group type that is not bound to an specifyc `account_id`. Instead it can be used globally across all accounts. 

## Default behavior group

The only considerated endpoint type for this kind of behavior groups is the `EmailSubscription`, and it needs to be global as well (`account_id` should be `null`). 
There are new endpoints in the Internal API for managing this configuration:

- POST `/defaultBehaviorGroups` (BehaviorGroup)
  -  Creates a default behavior group
- PUT `/defaultBehaviorGroups/{behaviorGroupId}/actions` (List of EmailSubscriptionProperties)
  -  Updates the set of action of a default behavior group
- PUT `/defaultBehaviorGroups/{behaviorGroupId}/eventType/{eventTypeId}`
  -  Links a default behavior group to an event type
- DELETE `/defaultBehaviorGroups/{behaviorGroupId}/eventType/{eventTypeId}`
  - Unlinks a default behavior group from an event type

Some remarks:
 - A new response-only attribute is dynamically set to the behavior group: `default_behavior`. It is true when the `account_id` is `null`, false otherwise.
 - This code also includes an utility (CommonStateSession) to easen the code when dealing with functions that could be used in both stateful and stateless. This could be potentially extracted as a separate PR to ease the review.

## CommonStateSession

We use stateful sessions when dealing with REST calls and stateless sessions with kakfa events. There are some functions that we need in both places, instead of duplicating the function to run in both scenarios, a wrapper was created for a subset of operations. 

The [CommonStateSessionFactory](https://github.com/RedHatInsights/notifications-backend/blob/bdf46be269cf51820702e087af296f05bceb25bb/backend/src/main/java/com/redhat/cloud/notifications/db/session/CommonStateSessionFactory.java) provides a method to request a `session` depending our needs - The developer has to specify the state type, it doesn't autodetect anything.

Stateless sessions using this tool will attempt to mimic what one could expect with the stateful sessions. Currently it does use introspection to find if there is any method annotated with `@PrePersit` that should be called. This process is cached per class to avoid recomputing this on every call.


backend/src/main/java/com/redhat/cloud/notifications/db/session/CommonStateSession.java
```diff
@@ -0,0 +1,99 @@
...
+        for (Method method: object.getClass().getDeclaredMethods()) {
+            PrePersist prePersist = method.getAnnotation(PrePersist.class);
+            if (prePersist != null) {
+                try {
+                    method.invoke(object);
+                } catch (InvocationTargetException | IllegalAccessException exception) {
+                    LOGGER.warnf(exception, "Unable to call PrePersist method [%s] found in class [%s]", method.getName(), object.getClass().getName());
+                }
+            }
+        }
```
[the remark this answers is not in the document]
gwenneg: FYI There are a few cases in tests where we don't want `prePersist()` to be called, but I don't think that will ever happen in runtime code.

backend/src/main/java/com/redhat/cloud/notifications/db/session/CommonStateSession.java
```diff
@@ -0,0 +1,99 @@
...
+
+public interface CommonStateSession {
+
+    public static CommonStateSession withSession(Mutiny.Session session) {
+        return new SessionAdapter(session);
+    }
+
+    public static CommonStateSession withStatelessSession(Mutiny.StatelessSession statelessSession) {
+        return new StatelessSessionAdapter(statelessSession);
+    }
```
gwenneg, line 20: I'm not sure we need these methods. `SessionAdapter` and `StatelessSessionAdapter` could be moved to the factory.

backend/src/main/java/com/redhat/cloud/notifications/db/session/CommonStateSession.java
```diff
@@ -0,0 +1,99 @@
...
+    <T> Uni<List<T>> find(Class<T> aClass, Object... ids);
+    <T> Mutiny.Query<T> createQuery(String query, Class<T> aClass);
+    Uni<Void> persist(Object object);
+    Uni<Void> flush();
+}
+
+class SessionAdapter implements CommonStateSession {
+    Mutiny.Session session;
+
+    public SessionAdapter(Mutiny.Session session) {
```
gwenneg, line 31: The adapters constructor could be private or package-protected.

backend/src/main/java/com/redhat/cloud/notifications/db/session/CommonStateSessionFactory.java
```diff
@@ -0,0 +1,24 @@
...
+import javax.inject.Inject;
+import java.util.function.Function;
+
+@ApplicationScoped
+public class CommonStateSessionFactory {
+
+    @Inject
+    Mutiny.SessionFactory sessionFactory;
+
+    public <T> Uni<T> withSession(boolean useStatelessSession, Function<CommonStateSession, Uni<T>> function) {
```
gwenneg, line 16:
```suggestion
    public <T> Uni<T> withSession(boolean stateless, Function<CommonStateSession, Uni<T>> function) {
```

gwenneg (commented):
I only had a look at the sessions changes, this looks interesting!

backend/src/main/java/com/redhat/cloud/notifications/db/session/CommonStateSession.java
```diff
@@ -0,0 +1,99 @@
...
+
+    public static CommonStateSession withSession(Mutiny.Session session) {
+        return new SessionAdapter(session);
+    }
+
+    public static CommonStateSession withStatelessSession(Mutiny.StatelessSession statelessSession) {
+        return new StatelessSessionAdapter(statelessSession);
+    }
+
+    <T> Uni<List<T>> find(Class<T> aClass, Object... ids);
```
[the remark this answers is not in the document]
josejulio: If we need one, yes!

backend/src/main/java/com/redhat/cloud/notifications/db/session/CommonStateSessionFactory.java
```diff
@@ -0,0 +1,24 @@
...
+        if (useStatelessSession) {
+            return sessionFactory.withStatelessSession(statelessSession -> function.apply(CommonStateSession.withStatelessSession(statelessSession)));
+        } else {
+            return sessionFactory.withSession(session -> function.apply(CommonStateSession.withSession(session)));
+        }
+    }
```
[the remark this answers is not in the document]
josejulio: It's more of an utility method. 

When we are executing a function than is compatible with both stateful and stateless sessions, we already have a `useStatelesssession` boolean around. This makes it easier to pass it around and reuse the same block for both.

see https://github.com/RedHatInsights/notifications-backend/pull/809/files#diff-567666f671274a21ea36eaf52aff1613c25b7302fc0c4146521a53c474d2edb8R40

backend/src/main/java/com/redhat/cloud/notifications/db/session/CommonStateSession.java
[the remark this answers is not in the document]
josejulio: Do you think it would take a significant amount of time? 

My only worry about doing that is that someone could forget to add the `PrePersisted` interface. 
But yes, current implementation is not smart at all (iterates over all the methods).

Would it be better if it stores a map of the results per class?

backend/src/main/java/com/redhat/cloud/notifications/db/session/Invoker.java
```diff
@@ -0,0 +1,42 @@
...
+                    cache.prePersist = method;
+                }
+            }
+
+            return cache;
+        });
+    }
+
+    public <T> void prePersist(T instance) {
+        InvokerCache cache = get(instance.getClass());
```
josejulio, line 35: @gwenneg Added a cache for the `prePersist` (per class). Most of the introspection methods should only be called once. Leaving the `invoke` to be called.
I'm not really sure if this is a valid approach :D.

backend/src/main/java/com/redhat/cloud/notifications/models/EndpointProperties.java
```diff
@@ -25,7 +25,7 @@
     private UUID id;
 
     @MapsId
-    @OneToOne(fetch = LAZY)
+    @OneToOne(fetch = EAGER) // Todo: This was causing issues on the stateless session createEndpoint - Revisit
```
josejulio, line 28: This was failing here: https://github.com/RedHatInsights/notifications-backend/blob/bf31e44617c4e84973c8ee3794beae218483a00a/backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java#L51

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -154,21 +166,33 @@
...
             return statelessSession.createQuery(query, Endpoint.class)
                     .setParameter("eventType", eventType)
                     .setParameter("accountId", tenant)
                     .getResultList()
-                    .onItem().call(endpoints -> loadProperties(endpoints, true));
+                    .onItem().call(endpoints -> loadProperties(endpoints, true))
+                    .onItem().transformToMulti(endpoints -> Multi.createFrom().iterable(endpoints))
+                    .onItem().transformToUniAndMerge(endpoint -> {
+                        if (endpoint.getAccountId() == null) {
+                            if (endpoint.getType().equals(EndpointType.EMAIL_SUBSCRIPTION)) {
```
gwenneg, line 180:
```suggestion
                            if (endpoint.getType() == EndpointType.EMAIL_SUBSCRIPTION) {
```

backend/src/main/java/com/redhat/cloud/notifications/routers/InternalService.java
```diff
@@ -282,6 +298,80 @@ public void httpRoot() {
         .onFailure().recoverWithItem(throwable -> Response.status(Response.Status.BAD_REQUEST).entity(new RenderEmailTemplateResponse.Error(throwable.getMessage())).build());
     }
 
+    @POST
+    @Path("/defaultBehaviorGroups")
```
gwenneg, line 302: `/defaultBehaviorGroups` or `/behaviorGroups/default`? Just asking, I don't have a strong preference.

gwenneg (commented):
FYI I will review the code again later, after most of the following comments have been resolved.

gwenneg:
Sorry for the delay before I reviewed this BTW 😃

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailTest.java
```diff
@@ -71,17 +72,22 @@
...
         String url = String.format("http://%s/v1/sendEmails", mockServerConfig.getRunningAddress());
-        System.setProperty("processor.email.bop_url", url);
-        System.setProperty("processor.email.bop_apitoken", BOP_TOKEN);
-        System.setProperty("processor.email.bop_env", BOP_ENV);
-        System.setProperty("processor.email.bop_client_id", BOP_CLIENT_ID);
+
+        TestHelpers.updateField(emailSender, "bopUrl", url, EmailSender.class);
+        TestHelpers.updateField(emailSender, "bopApiToken", BOP_TOKEN, EmailSender.class);
+        TestHelpers.updateField(emailSender, "bopEnv", BOP_ENV, EmailSender.class);
+        TestHelpers.updateField(emailSender, "bopClientId", BOP_CLIENT_ID, EmailSender.class);
```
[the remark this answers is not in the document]
josejulio: Yes, setting the property is a system-wide setting (it affects all the tests), so I added the reflection calls to keep them locally.

backend/src/main/resources/db/migration/V1.34.0__NOTIF-315_default_behavior_group.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE behavior_group
+    ALTER COLUMN account_id DROP NOT NULL;
+
+ALTER TABLE endpoints
+    ALTER COLUMN account_id DROP NOT NULL;
+
+ALTER TABLE endpoint_email_subscriptions
+    DROP CONSTRAINT fk_endpoint_email_subscriptions_application_id,
+    ADD CONSTRAINT fk_endpoint_email_subscriptions_application_id FOREIGN KEY (application_id) REFERENCES applications(id) ON DELETE CASCADE;
```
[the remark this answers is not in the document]
josejulio: IIRC a test  broke because I deleted the (test) application but the email_subscription was still in the db.

backend/src/main/java/com/redhat/cloud/notifications/routers/InternalService.java
josejulio: I didn't want to cause a conflict with the `/behaviorGroups/{something}` paths that's why I used that one - no strong preference either.

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -31,12 +33,23 @@
...
+    }
+
+    private Uni<BehaviorGroup> create(String accountId, BehaviorGroup behaviorGroup, boolean isDefaultBehaviorGroup) {
         return sessionFactory.withSession(session -> {
             return session.find(Bundle.class, behaviorGroup.getBundleId())
                     .onItem().ifNull().failWith(new NotFoundException("bundle_id not found"))
                     .onItem().invoke(bundle -> {
                         behaviorGroup.setBundle(bundle);
                         behaviorGroup.setAccountId(accountId);
+                        if (isDefaultBehaviorGroup != (accountId == null)) {
```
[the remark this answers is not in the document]
josejulio: What about `isDefaultBehaviorGroup != behaviorGroup.isDefaultBehavior()` ?

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -246,4 +333,19 @@
                     .onItem().invoke(behaviorGroups -> behaviorGroups.forEach(BehaviorGroup::filterOutActions));
         });
     }
+
+    private Uni<BehaviorGroup> getBehaviorGroup(Session session, UUID behaviorGroupId, boolean isDefaultBehaviorGroup) {
+        return session.find(BehaviorGroup.class, behaviorGroupId)
+                .onItem().invoke(behaviorGroup -> {
```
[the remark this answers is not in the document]
josejulio: :scream:

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -55,30 +68,83 @@
...
+            return this.getBehaviorGroup(session, behaviorGroup.getId(), isDefaultBehavior)
+                    .replaceWith(
+                            session.createQuery(query)
+                            .setParameter("displayName", behaviorGroup.getDisplayName())
+                            .setParameter("accountId", accountId)
+                            .setParameter("id", behaviorGroup.getId())
+                            .executeUpdate()
+                            .call(session::flush)
+                            .onItem().transform(rowCount -> rowCount > 0)
+                    );
```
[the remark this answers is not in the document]
josejulio: Not sure i understand. `this.getBehaviorGroup` will return the behavior group or throw an exception if there is a missmatch.
Although I get a warning: `Throw statement in Mutiny invoke operator ` .
After your other suggested change (throw if null) this shouldn't be an issue. or is it?

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -120,14 +132,14 @@
                         }
 
                         Endpoint endpoint = new Endpoint();
-                        endpoint.setProperties(properties);
+                        endpoint.setProperties(new EmailSubscriptionProperties(properties));
```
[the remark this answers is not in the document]
josejulio: I probably should not do it here, but one caller gets the `default email subscription`  and calls this method to create an `account email subscription`, I don't want to affect the default one. I'll move this new call to the caller.

backend/src/main/java/com/redhat/cloud/notifications/db/session/Invoker.java
```diff
@@ -0,0 +1,53 @@
...
+ * To lessen this burden, a cache is created per class once, so the methods are not inspected all the time.
+ */
+class Invoker {
+
+    private static final Logger LOGGER = Logger.getLogger(Invoker.class);
+    private final Map<Class<?>, InvokerCache> cacheMap = new HashMap<>();
+
+    private static class InvokerCache {
+        private Method prePersist;
+    }
```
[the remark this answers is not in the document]
josejulio: I think it would. That InvokerCache class would be needed if there is any extra method we need to call.

backend/src/main/java/com/redhat/cloud/notifications/db/session/Invoker.java
```diff
@@ -0,0 +1,53 @@
...
+        });
+    }
+
+    public <T> void prePersist(T instance) {
+        InvokerCache cache = get(instance.getClass());
+        if (cache.prePersist != null) {
+            try {
+                cache.prePersist.invoke(instance);
+            } catch (InvocationTargetException | IllegalAccessException exception) {
+                LOGGER.warnf(exception, "Unable to call PrePersist method [%s] found in class [%s]", cache.prePersist.getName(), instance.getClass().getName());
```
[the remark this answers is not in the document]
josejulio: I wasn't sure. I can make it fail instead.

backend/src/main/java/com/redhat/cloud/notifications/db/session/CommonStateSession.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public Uni<Void> flush() {
+        return session.flush();
+    }
+}
+
+class StatelessSessionAdapter implements CommonStateSession {
+    private final Mutiny.StatelessSession statelessSession;
+    private final Invoker invoker = new Invoker();
```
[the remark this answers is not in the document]
josejulio: wow. that's right.

josejulio:
@gwenneg I addressed most of the suggested changes

backend/src/main/java/com/redhat/cloud/notifications/db/session/Invoker.java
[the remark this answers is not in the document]
gwenneg: True, but we can easily add the class later if needed. I would remove it for now.

backend/src/main/java/com/redhat/cloud/notifications/routers/InternalService.java
gwenneg: `InternalService` does not contain any `/behaviorGroups/{something}`, they are all in `NotificationService` so the conflict risk is low. I have a slight preference for `/behaviorGroups/default` but I can live with `/defaultBehaviorGroups`. I'll let you choose 😄

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailTest.java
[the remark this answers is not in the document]
gwenneg: Clearing the properties with `System.clearProperty(...)` at the end of the tests didn't work?

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -55,30 +70,83 @@
...
     }
 
-    // TODO Should this be forbidden for default behavior groups?
     public Uni<Boolean> update(String accountId, BehaviorGroup behaviorGroup) {
+        return this.update(accountId, behaviorGroup, false);
+    }
+
+    public Uni<Boolean> updateDefault(BehaviorGroup behaviorGroup) {
+        return this.update(null, behaviorGroup, true);
+    }
```
gwenneg, line 79: This is never used. Should it be removed?

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -154,21 +166,38 @@
     // Note: This method uses a stateless session
```
gwenneg, line 166: I wonder if we should create and use an annotation like `@StatelessSession` (or a better name) with `RetentionPolicy.SOURCE`. It would be more visible than a comment in my IntelliJ configured with the dark theme 😄

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -154,21 +166,38 @@
...
                     .getResultList()
-                    .onItem().call(endpoints -> loadProperties(endpoints, true));
+                    .onItem().call(endpoints -> loadProperties(endpoints, true))
+                    .onItem().transformToMulti(endpoints -> Multi.createFrom().iterable(endpoints))
+                    .onItem().transformToUniAndConcatenate(endpoint -> {
+                        if (endpoint.getAccountId() == null) {
+                            if (endpoint.getType() == EndpointType.EMAIL_SUBSCRIPTION) {
+                                return this.getOrCreateEmailSubscriptionEndpoint(
+                                        tenant,
+                                        // We don't want to affect the original properties
```
gwenneg, line 183: How could the properties be affected? This method is called during a Kafka messages processing.

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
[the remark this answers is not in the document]
gwenneg: Yeah, I wrote this comment before the other one. This one is no longer relevant.

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -55,30 +70,83 @@
...
+            return this.getBehaviorGroup(session, behaviorGroup.getId(), isDefaultBehavior)
+                    .replaceWith(
+                            session.createQuery(query)
+                            .setParameter("displayName", behaviorGroup.getDisplayName())
+                            .setParameter("accountId", accountId)
+                            .setParameter("id", behaviorGroup.getId())
+                            .executeUpdate()
+                            .call(session::flush)
+                            .onItem().transform(rowCount -> rowCount > 0)
+                    );
```
gwenneg, line 93: We didn't throw an exception in case of unknown behavior group before.
```suggestion
                    )
                    .onFailure(NoResultException.class).recoverWithItem(Boolean.FALSE);
```

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -55,30 +70,83 @@
...
-                    .onItem().transform(rowCount -> rowCount > 0);
+            return this.getBehaviorGroup(session, behaviorGroupId, isDefaultBehavior)
+                    .replaceWith(
+                        session.createQuery(query)
+                                .setParameter("accountId", accountId)
+                                .setParameter("id", behaviorGroupId)
+                                .executeUpdate()
+                                .call(session::flush)
+                                .onItem().transform(rowCount -> rowCount > 0)
+                    );
```
gwenneg, line 116: We didn't throw an exception in case of unknown behavior group before.
```suggestion
                    )
                    .onFailure(NoResultException.class).recoverWithItem(Boolean.FALSE);
```

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -55,30 +70,83 @@
...
+                        String insertQuery = "INSERT INTO event_type_behavior (event_type_id, behavior_group_id, created) " +
+                                "VALUES (:eventTypeId, :behaviorGroupId, :created) " +
+                                "ON CONFLICT (event_type_id, behavior_group_id) DO NOTHING";
+                        return session.createNativeQuery(insertQuery)
+                                .setParameter("eventTypeId", eventTypeId)
+                                .setParameter("behaviorGroupId", behaviorGroupId)
+                                .setParameter("created", LocalDateTime.now(UTC))
+                                .executeUpdate();
+                    }).replaceWith(true)
+            .onFailure(BadRequestException.class).recoverWithItem(false);
```
gwenneg, line 133: What about the `NoResultException` case? Should we return `FALSE` as well?

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -55,30 +70,83 @@
...
+                    .onItem().transformToUni(_ignored -> {
+                        String deleteQuery = "DELETE FROM EventTypeBehavior b " +
+                                "WHERE eventType.id = :eventTypeId " +
+                                "AND id.behaviorGroupId = :behaviorGroupId";
+                        return session.createQuery(deleteQuery)
+                                .setParameter("eventTypeId", eventTypeId)
+                                .setParameter("behaviorGroupId", behaviorGroupId)
+                                .executeUpdate()
+                                .replaceWith(true);
+                    });
```
gwenneg, line 149: Should we "catch" the `NoResultException` and recover with `FALSE` here?

backend/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
```diff
@@ -91,6 +93,13 @@
     @JsonIgnore
     private Set<EventTypeBehavior> behaviors;
 
+    @JsonInclude
+    @JsonProperty(access = WRITE_ONLY)
```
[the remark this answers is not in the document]
josejulio: After reading the docs, yes, it should be. 

> (note that the direction of read and write is from perspective of the property, not from external data format: this may be confusing in some contexts).

backend/src/main/java/com/redhat/cloud/notifications/routers/InternalService.java
josejulio: You're right.
I don't have a preference in the name and can update it.
 Actually I think the "default" name is confusing for what it really does, but not really worried as is an internal api

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
josejulio: Being used now!

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
josejulio: That would be nice, would be even better if we could detect mixed usage (at least at compile or test time)

josejulio: That said, we could do that in a separate PR

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailTest.java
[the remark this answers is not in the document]
josejulio: Just tried again, they don't seem to work well together (I have other test that touches the bopUrl). 

IIUC, the reason is that quarkus starts only once, and once a config is used, the value won't change.
From quarkus test profiles [1]:

> The first time one of these tests is run the Quarkus test extension will start Quarkus. Quarkus will then remain running for the duration of the test run. This makes testing very fast, because Quarkus is only started once, however it does limit you to testing a single configuration per module, as you can’t restart Quarkus in a different configuration. 


Other way to do it would be to use test configuration - although I'm not sure it would help as we need some dynamic values and maybe it would make the tests slower.

[1] https://quarkus.io/blog/quarkus-test-profiles/

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
josejulio: Passing the properties directly breaks the tests that uses this part. I will do some tests later to check what it does in the database

josejulio: This is the error it throws:
```
2022-01-05 10:12:54,878 INFO  [com.red.clo.not.eve.EventConsumer] (vert.x-eventloop-thread-23) Could not process the payload: {"bundle":"my-bundle","application":"policies-lifecycle-test","event_type":"all","timestamp":"2022-01-05T10:12:54.837302","account_id":"tenant","context":"{}","events":[{"metadata":{},"payload":"{}"}],"version":"v1.0.0","recipients":[]}: org.hibernate.LazyInitializationException: Unable to perform requested lazy initialization [com.redhat.cloud.notifications.models.EmailSubscriptionProperties.endpoint] - no session and settings disallow loading outside the Session
        at org.hibernate.bytecode.enhance.spi.interceptor.EnhancementHelper.throwLazyInitializationException(EnhancementHelper.java:275)
        at org.hibernate.bytecode.enhance.spi.interceptor.EnhancementHelper.performWork(EnhancementHelper.java:165)
        at org.hibernate.bytecode.enhance.spi.interceptor.LazyAttributeLoadingInterceptor.loadAttribute(LazyAttributeLoadingInterceptor.java:77)
        at org.hibernate.bytecode.enhance.spi.interceptor.LazyAttributeLoadingInterceptor.fetchAttribute(LazyAttributeLoadingInterceptor.java:73)
        at org.hibernate.bytecode.enhance.spi.interceptor.LazyAttributeLoadingInterceptor.handleRead(LazyAttributeLoadingInterceptor.java:54)
        at org.hibernate.bytecode.enhance.spi.interceptor.AbstractInterceptor.readObject(AbstractInterceptor.java:153)
        at com.redhat.cloud.notifications.models.EmailSubscriptionProperties.$$_hibernate_read_endpoint(EmailSubscriptionProperties.java)
        at com.redhat.cloud.notifications.models.EmailSubscriptionProperties.$$_hibernate_write_endpoint(EmailSubscriptionProperties.java)
        at com.redhat.cloud.notifications.models.EndpointProperties.setEndpoint(EndpointProperties.java:46)
        at com.redhat.cloud.notifications.db.EndpointResources.lambda$createEndpoint$0(EndpointResources.java:51)
```

Luckily it's noisy because we are in a stateless session.

josejulio:
Updated again!

backend/src/test/java/com/redhat/cloud/notifications/processors/email/EmailTest.java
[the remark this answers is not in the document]
gwenneg: Thanks for checking! Reflection is fine then 😃

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
gwenneg: https://issues.redhat.com/browse/NOTIF-445

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
gwenneg: Wouldn't that work instead of calling `getOrCreateEmailSubscriptionEndpoint`?

```
                    .invoke(endpoint -> {
                        if (endpoint.getAccountId() == null) {
                            if (endpoint.getType() == EndpointType.EMAIL_SUBSCRIPTION) {
                                endpoint.setAccountId(tenant);
                            } else {
                                LOGGER.warnf("Invalid endpoint configured in default behavior group: %s", endpoint.getId());
                            }
                        }
                    })
```
Since the stateless session is used, calling `endpoint.setAccountId` should not affect the DB in any way.

gwenneg: If that works, we could even avoid the `Multi` transformation.

gwenneg:
```
            return statelessSession.createQuery(query, Endpoint.class)
                    .setParameter("eventType", eventType)
                    .setParameter("accountId", tenant)
                    .getResultList()
                    .onItem().call(endpoints -> loadProperties(endpoints, true))
                    .invoke(endpoints -> {
                        for (Endpoint endpoint : endpoints) {
                            if (endpoint.getAccountId() == null) {
                                if (endpoint.getType() == EndpointType.EMAIL_SUBSCRIPTION) {
                                    endpoint.setAccountId(tenant);
                                } else {
                                    LOGGER.warnf("Invalid endpoint configured in default behavior group: %s", endpoint.getId());
                                }
                            }
                        }
                    });
```

josejulio: Yes, I think that's fine - We don't really need to create it in the database just yet.

josejulio: Pushed the changes

backend/src/main/java/com/redhat/cloud/notifications/models/EmailSubscriptionProperties.java
```diff
@@ -22,6 +22,16 @@
...
+    public EmailSubscriptionProperties() {
+
+    }
+
+    public EmailSubscriptionProperties(EmailSubscriptionProperties properties) {
+        this.onlyAdmins = properties.onlyAdmins;
+        this.ignorePreferences = properties.ignorePreferences;
+        this.groupId = properties.groupId;
+    }
+
```
gwenneg, line 34: This is no longer needed, right?

josejulio: Correct - just removed it

josejulio: Correct - just removed it

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -55,30 +70,88 @@
...
     public Uni<Boolean> delete(String accountId, UUID behaviorGroupId) {
+        return this.delete(accountId, behaviorGroupId, false);
+    }
+
+    public Uni<Boolean> deleteDefault(UUID behaviorGroupId) {
+        return this.delete(null, behaviorGroupId, true);
+    }
+
+    public Uni<Boolean> delete(String accountId, UUID behaviorGroupId, boolean isDefaultBehavior) {
         String query = "DELETE FROM BehaviorGroup WHERE accountId = :accountId AND id = :id";
```
gwenneg, line 107: Same question about `Hibernate` and `accountId`.

gwenneg (commented):
I'm done reviewing the runtime code. I'll have a look at the test tomorrow.

backend/src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java
```diff
@@ -116,61 +155,101 @@ void test() throws IOException, InterruptedException {
...
         // Pushing a new message should trigger two webhook calls.
-        pushMessage(2);
+        pushMessage(2, 0, 0);
 
         // Let's check the notifications history.
         retry(() -> checkEndpointHistory(identityHeader, endpointId1, 1, true, 200));
         retry(() -> checkEndpointHistory(identityHeader, endpointId2, 1, true, 200));
+        retry(() -> checkEndpointHistory(identityHeader, emailEndpoint, 0, true, 200));
+
+        // We'll link with the default behavior group
```
gwenneg, line 191:
```suggestion
        // We'll link the event type with the default behavior group
```

backend/src/main/java/com/redhat/cloud/notifications/routers/InternalService.java
```diff
@@ -282,6 +298,92 @@ public void httpRoot() {
...
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    public Uni<BehaviorGroup> createDefaultBehaviorGroup(@NotNull @Valid BehaviorGroup behaviorGroup) {
+        return sessionFactory.withSession(session -> {
+            return behaviorGroupResources.createDefault(behaviorGroup);
+        });
+    }
+
+    @PUT
+    @Path("/behaviorGroups/default/{id}")
```
gwenneg, line 312: I didn't see a test for this, but maybe I missed it.

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -55,30 +70,88 @@
...
     }
 
-    // TODO Should this be forbidden for default behavior groups?
     public Uni<Boolean> delete(String accountId, UUID behaviorGroupId) {
+        return this.delete(accountId, behaviorGroupId, false);
+    }
+
+    public Uni<Boolean> deleteDefault(UUID behaviorGroupId) {
+        return this.delete(null, behaviorGroupId, true);
+    }
```
gwenneg, line 104: This method does not seem to be used anywhere.

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -55,30 +70,88 @@
...
     public Uni<Boolean> update(String accountId, BehaviorGroup behaviorGroup) {
+        return this.update(accountId, behaviorGroup, false);
+    }
+
+    public Uni<Boolean> updateDefault(BehaviorGroup behaviorGroup) {
+        return this.update(null, behaviorGroup, true);
+    }
+
+    private Uni<Boolean> update(String accountId, BehaviorGroup behaviorGroup, boolean isDefaultBehavior) {
         String query = "UPDATE BehaviorGroup SET displayName = :displayName WHERE accountId = :accountId AND id = :id";
```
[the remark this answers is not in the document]
josejulio: mmh, good catch. I don't think it complains, but I don't think it does what I want. I'll add a test and check.

backend/src/main/java/com/redhat/cloud/notifications/routers/InternalService.java
```diff
@@ -282,6 +298,92 @@ public void httpRoot() {
...
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    @Operation(summary = "Update a behavior group.")
+    public Uni<Boolean> createDefaultBehaviorGroup(@PathParam("id") UUID id, @NotNull @Valid BehaviorGroup behaviorGroup) {
```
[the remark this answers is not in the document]
josejulio: Whoops :D

backend/src/main/java/com/redhat/cloud/notifications/routers/InternalService.java
```diff
@@ -282,6 +298,92 @@ public void httpRoot() {
...
+    @Operation(summary = "Update the list of actions of a default behavior group.")
+    @APIResponse(responseCode = "200", content = @Content(schema = @Schema(type = SchemaType.STRING)))
+    public Uni<Response> updateDefaultBehaviorGroupActions(@PathParam("behaviorGroupId") UUID behaviorGroupId, List<EmailSubscriptionProperties> propertiesList) {
+        if (propertiesList == null) {
+            return Uni.createFrom().failure(new BadRequestException("The request body must contain a list of EmailSubscriptionProperties"));
+        }
+        // RESTEasy does not reject an invalid List<UUID> body (even when @Valid is used) so we have to do an additional check here.
+        if (propertiesList.contains(null)) {
+            return Uni.createFrom().failure(new BadRequestException("The list of EmailSubscriptionProperties should not contain empty values"));
+        }
```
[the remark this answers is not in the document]
josejulio: Not really

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
josejulio: Right, adding the corresponding endpoint to delete the default behavior group

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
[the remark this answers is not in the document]
josejulio: So yes, Hibernate is happy but it doesn't work. fixed in latest

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
josejulio: Same

backend/src/main/java/com/redhat/cloud/notifications/routers/InternalService.java
josejulio: There wasn't any, but the db code is now tested

gwenneg (approved):
LGTM

gwenneg:
The PR check failures seem unrelated.

PR #815, opened 2021-12-01
NOTIF-299 Run the backend DB cleaning jobs from a CronJob - Phase 1

gwenneg:
I tested this on ephemeral already.

### DIFF
--- a/.rhcicd/clowdapp-backend.yaml
+++ b/.rhcicd/clowdapp-backend.yaml
@@ -132,6 +132,69 @@ objects:
           value: https://3ff0dbd8017a4750a1d92055a1685263@o271843.ingest.sentry.io/5440905?environment=${ENV_NAME}
         - name: QUARKUS_LOG_SENTRY_ENVIRONMENT
           value: ${ENV_NAME}
+- apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: notifications-db-cleaner-config
+  data:
+    clean.sh: |
+      cat /notifications-db-cleaner/clean.sql | psql > /dev/null
+    clean.sql: |
+      CALL cleanEventLog();
+      CALL cleanKafkaMessagesIds();
+- apiVersion: batch/v1beta1
+  kind: CronJob
+  metadata:
+    name: notifications-db-cleaner-cronjob
+  spec:
+    schedule: ${DB_CLEANER_SCHEDULE}
+    suspend: ${{DISABLE_DB_CLEANER}}
+    concurrencyPolicy: Forbid
+    jobTemplate:
+      spec:
+        template:
+          spec:
+            restartPolicy: Never
+            volumes:
+            - name: notifications-db-cleaner-volume
+              configMap:
+                name: notifications-db-cleaner-config
+            containers:
+            - name: notifications-db-cleaner
+              image: quay.io/cloudservices/postgresql-rds:12-1
+              restartPolicy: Never
+              resources:
+                requests:
+                  cpu: 100m
+                  memory: 100Mi
+                limits:
+                  cpu: 200m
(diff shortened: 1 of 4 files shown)


gwenneg:
Let's wait for the end of the prod releasing process before this is merged.

backend/src/main/resources/db/migration/V1.32.0__NOTIF-299_notifications-db-cleaner.sql
```diff
@@ -0,0 +1,25 @@
+-- This stored procedure deletes the event log entries that are no longer needed.
+-- It is executed from an OpenShift CronJob.
+CREATE PROCEDURE cleanEventLog() AS $$
+DECLARE
+    deleted INTEGER;
+BEGIN
+    RAISE INFO '% Event log purge starting. Entries older than 31 days will be deleted.', NOW();
+    DELETE FROM notification_history WHERE created < now() - interval '31 days';
```
josejulio, line 8: nitpick: looks like interval is a reserved keyword, should it be in upper case? (same below and with `now`)
```suggestion
    DELETE FROM notification_history WHERE created < NOW() - INTERVAL '31 days';
```

backend/src/main/resources/db/migration/V1.32.0__NOTIF-299_notifications-db-cleaner.sql
```diff
@@ -0,0 +1,25 @@
...
+$$ LANGUAGE PLPGSQL;
+
+-- This stored procedure deletes the Kafka messages IDs that are no longer needed.
+-- It is executed from an OpenShift CronJob.
+CREATE PROCEDURE cleanKafkaMessagesIds() AS $$
+DECLARE
+    deleted INTEGER;
+BEGIN
+    RAISE INFO '% Kafka messages purge starting. Entries older than 1 day will be deleted.', NOW();
+    DELETE FROM kafka_message WHERE created < now() - interval '1 days';
```
josejulio, line 21: same

```suggestion
    DELETE FROM kafka_message WHERE created < NOW() - INTERVAL '1 days';
```

.rhcicd/clowdapp-backend.yaml
```diff
@@ -132,6 +132,69 @@ objects:
...
+- apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: notifications-db-cleaner-config
+  data:
+    clean.sh: |
+      cat /notifications-db-cleaner/clean.sql | psql > /dev/null
+    clean.sql: |
+      CALL cleanEventLog();
+      CALL cleanKafkaMessagesIds();
```
josejulio, line 144: Wondering, if it fails, do we get an alert?

josejulio (approved):
Looks great! some non blocking comments.

.rhcicd/clowdapp-backend.yaml
gwenneg: Good question!

The answer depends on the failure:
- It could fail because of an OpenShift template change, but we should see the failure after the new template has been deployed.
- It could fail because of a SQL schema change that would break the stored procedures and then unfortunately the pod cronjob pod will not be marked as failed. I'll try to add a unit test that will execute the stored procedures during the notifications build. That should be enough to prevent a schema issue.
- It could probably fail for other reasons...

gwenneg: The stored procedures are now tested with Java. This protects us against a breaking SQL schema change.

gwenneg: > Is there any way for the openshift jobs to alert us that it's failing?

I don't know yet but I will look into that.

PR #830, opened 2021-12-06
NOTIF-383 add warning text when editing name field

aferd:
next steps would be to add further confirmation(type 'name' to edit), similar to "delete" function in admin ui.


admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -158,7 +158,9 @@ export const ApplicationPage: React.FunctionComponent = () => {
                                         onClose={ () => setShowModal(false) }
                                     ><Form isHorizontal>
                                             <FormGroup label='Name' fieldId='name' isRequired
-                                                helperText='This is a short name, only composed of a-z 0-9 and - characters.'>
+                                                helperText={ isEdit ? <HelperText><HelperTextItem variant="warning"  hasIcon>
```
josejulio, line 161: Probably the linter is going to complain that there are 2 spaces here
```suggestion
                                                helperText={ isEdit ? <HelperText><HelperTextItem variant="warning" hasIcon>
```

josejulio (approved):
lgtm

PR #836, opened 2021-12-09
Add log entries and metrics to aggregation commands processing

gwenneg:
This PR will give us the possibility to monitor and analyze the aggregation commands processing in the backend.


josejulio (approved):
lgtm

PR #843, opened 2021-12-09
Fix issue with datasource labels in grafana dashboard

bennetelli:
Grafana is overwriting the label in the ui and we still have to add the label by hand.


theute:
Thanks @bennetelli  !

PR #844, opened 2021-12-09
Log webhook target host

gwenneg:
Our logs currently contain the HTTP status of each webhook call but when something is wrong we have no way to determine which host failed from the logs. This PR will change that.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
@@ -117,25 +117,25 @@ public class WebhookTypeProcessor implements EndpointTypeProcessor {
                         .onItem().transform(resp -> {
                             NotificationHistory history = buildNotificationHistory(item, startTime);
 
+                            HttpRequestImpl<Buffer> reqImpl = (HttpRequestImpl<Buffer>) req.getDelegate();
+
                             boolean serverError = false;
                             if (resp.statusCode() >= 200 && resp.statusCode() < 300) {
                                 // Accepted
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
```diff
@@ -117,26 +117,26 @@ private WebClient getWebClient(boolean disableSSLVerification) {
...
                             if (resp.statusCode() >= 200 && resp.statusCode() < 300) {
                                 // Accepted
-                                LOGGER.debugf("Target endpoint successful: %d", resp.statusCode());
+                                LOGGER.debugf("Webhook request to %s was successful: %d", reqImpl.host(), resp.statusCode());
                                 history.setInvocationResult(true);
                             } else if (resp.statusCode() >= 500) {
                                 // Temporary error, allow retry
                                 serverError = true;
-                                LOGGER.debugf("Target endpoint server error: %d %s", resp.statusCode(), resp.statusMessage());
+                                LOGGER.debugf("Webhook request to %s failed: %d %s", reqImpl.host(), resp.statusCode(), resp.statusMessage());
```
josejulio, line 130: Should we say that we will retry?

josejulio (approved):
lgtm

backend/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
gwenneg: Oops I saw this after clicking on merge, sorry 😄 That would have made sense, yeah, but now we'll have to wait for another PR.

PR #845, opened 2021-12-10
Updating Clowdapp

theute:
`web` is deprecated and [will go away](https://docs.google.com/document/d/1m_MgEhZjb-z-bE8RbTrii559e9Sb3yS7P3gYsqAwveI/edit#).

We'll need to do the same for:

- https://github.com/RedHatInsights/notifications-gw/blob/main/.rhcicd/clowdapp.yaml
- https://github.com/RedHatInsights/policies-ui-backend/blob/master/.rhcicd/clowdapp.yaml#L27
- https://github.com/RedHatInsights/policies-engine/blob/master/.rhcicd/clowdapp.yaml#L47

But testing here before...

### DIFF
--- a/.rhcicd/clowdapp-backend.yaml
+++ b/.rhcicd/clowdapp-backend.yaml
@@ -36,7 +36,9 @@ objects:
     deployments:
     - name: service
       minReplicas: ${{MIN_REPLICAS}}
-      web: true
+      webServices:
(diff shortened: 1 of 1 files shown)


gwenneg (approved):
I've seen that change before in other projects, LGTM

PR #847, opened 2021-12-10
NOTIF-425 Add Prometheus metrics for the aggregator cron job

bennetelli:


.rhcicd/clowdapp-aggregator.yaml
```diff
@@ -61,6 +61,8 @@ objects:
           value: https://39032857c8214cedaf226c4a26709b5b@o271843.ingest.sentry.io/5987783?environment=${ENV_NAME}
         - name: QUARKUS_LOG_SENTRY_ENVIRONMENT
           value: ${ENV_NAME}
+        - name: PROMETHEUS_PUSH_GATEWAY
+        - value: ${PROMETHEUS_PUSHGATEWAY}
```
bennetelli, line 65: @gwenneg is this correct? How do I set the PROMETHEUS_PUSHGATEWAY variable in clowder for each environment (stage, prod, ..) ?

theute: - There is a bogus `-` in front of `values`
- If you map the container parameter PROMETHEUS_PUSHGATEWAY to PROMETHEUS_PUSH_GATEWAY environment parameter you also need to declare the container parameter in the parameters section lower in that same file.

bennetelli: > * If you map the container parameter PROMETHEUS_PUSHGATEWAY to PROMETHEUS_PUSH_GATEWAY environment parameter you also need to declare the container parameter in the parameters section lower in that same file.

That's the problem here. I was able to find the the value of PROMETHEUS_PUSHGATEWAY somewhere deep in app-interface and wasn't sure if it's a good idea to copy and paste it in this file. I acutally thought I could get the value somehow with clowder-magic in the container/application.

theute: You just reference the name, not the value, the value will be picked from the environment file:
`data/products/insights/environments/production.yml` `PROMETHEUS_PUSHGATEWAY: blah`

It's exactly the same as for `ENV_NAME`

bennetelli: Thanks. I didn't see there is an entry without value.

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -34,48 +37,69 @@
...
+        } finally {
+            durationTimer.setDuration();
+            PushGateway pg = new PushGateway(prometheusPushGatewayUrl);
             try {
-                CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).get();
-            } catch (InterruptedException | ExecutionException ie) {
-                LOG.error("Writing AggregationCommands failed", ie);
+                pg.pushAdd(registry, "aggregator_job");
+            } catch (IOException e) {
+                LOG.warn("Could not push metrics to Prometheus Pushgateway.");
```
gwenneg, line 97:
```suggestion
                LOG.warn("Could not push metrics to Prometheus Pushgateway.", e);
```

aggregator/src/main/resources/application.properties
```diff
@@ -34,6 +34,9 @@ quarkus.log.sentry=false
 quarkus.log.sentry.in-app-packages=*
 quarkus.log.sentry.dsn=FILL_ME
 
+prometheus.pushgateway.url=${PROMETHEUS_PUSH_GATEWAY}
+%test.prometheus.pushgateway.url=localhost:8080
```
gwenneg, line 38: I would be simpler if `PROMETHEUS_PUSH_GATEWAY`, `PROMETHEUS_PUSHGATEWAY` and `prometheus.pushgateway.url` were a single thing.

The properties should only contain something like this (key to be confirmed):
```suggestion
prometheus.pushgateway.url=localhost:8080
```
And then the Clowder template could override the value with the `PROMETHEUS_PUSHGATEWAY_URL` env var.

aggregator/src/test/java/com/redhat/cloud/notifications/DailyEmailAggregationJobTest.java
```diff
@@ -79,7 +78,7 @@ void shouldProcessFourSubscriptions() {
         helpers.addEmailAggregation("tenant", "unknown-bundle", "policies", "somePolicyId", "someHostId");
         helpers.addEmailAggregation("tenant", "unknown-bundle", "unknown-application", "somePolicyId", "someHostId");
 
-        final List<AggregationCommand> emailAggregations = testee.processAggregateEmails(LocalDateTime.now(UTC));
+        final List<AggregationCommand> emailAggregations = testee.processAggregateEmails(LocalDateTime.now(UTC), new CollectorRegistry());
```
gwenneg, line 81: It would be great if the tests checked the metrics.

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -96,7 +120,12 @@ public void processDailyEmail() {
                 pendingAggregationCommands.size()
         );
 
-        processedGauge.set(pendingAggregationCommands.size());
+        Gauge pairsProcessed = Gauge
+                .build()
+                .name("aggregator_job_accountid_application_pairs_processed")
```
gwenneg, line 125: Metrics names are usually separated with `.`

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -34,48 +37,69 @@
...
+                    .register(registry);
+            lastSuccess.setToCurrentTime();
+        } finally {
+            durationTimer.setDuration();
+            PushGateway pg = new PushGateway(prometheusPushGatewayUrl);
             try {
-                CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).get();
-            } catch (InterruptedException | ExecutionException ie) {
-                LOG.error("Writing AggregationCommands failed", ie);
+                pg.pushAdd(registry, "aggregator_job");
```
gwenneg, line 95: Will `aggregator_job` be displayed somewhere?

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -34,48 +37,69 @@
...
+                    LOG.error("Writing AggregationCommands failed", ie);
+                }
             }
 
-            // resolve completable futures so the Quarkus main thread doesn't stop before everything has been sent
+            emailAggregationResources.updateLastCronJobRun(now);
+
+            Gauge lastSuccess = Gauge
+                    .build()
+                    .name("aggregator_job_last_success")
```
gwenneg, line 87: Metrics names are usually separated with `.`

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -34,48 +37,69 @@
...
+                    .help("Last time the aggregator job succeeded.")
+                    .register(registry);
+            lastSuccess.setToCurrentTime();
```
gwenneg, line 90: Does this automatically use the UTC timezone?

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -34,48 +37,69 @@
...
-        for (AggregationCommand aggregationCommand : aggregationCommands) {
-            try {
-                final String payload = objectMapper.writeValueAsString(aggregationCommand);
-                futures.add(emitter.send(payload).toCompletableFuture());
-            } catch (JsonProcessingException e) {
-                LOG.warn("Could not transform AggregationCommand to JSON object.", e);
+        CollectorRegistry registry = new CollectorRegistry();
+        Gauge duration = Gauge
+                .build()
+                .name("aggregator_job_duration_seconds")
```
gwenneg, line 57: Metrics names are usually separated with `.`

gwenneg: `notifications.aggregator.job.duration.seconds` could help avoiding collisions with another project that would publish metrics starting with `aggregator`. Same idea for other metrics names.

aggregator/pom.xml
```diff
@@ -31,6 +31,11 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-micrometer-registry-prometheus</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.prometheus</groupId>
+            <artifactId>simpleclient_pushgateway</artifactId>
+            <version>0.12.0</version>
+        </dependency>
```
gwenneg, line 38: Is the following dependency still required?
```
        <dependency>
            <groupId>io.quarkus</groupId>
            <artifactId>quarkus-micrometer-registry-prometheus</artifactId>
        </dependency>
```

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
bennetelli: Yes, in the Pushgateway of Prometheus.

bennetelli: that's the name of the job displayed in the pushgateay ui:
Screenshot from 2021-12-13 08-38-58

aggregator/pom.xml
bennetelli: I am able to create Gauges in the PushGateway without it, indeed. A bit unexpected as the simpleclient-pushgateway dependency doesn't contain micrometer-core dependency, but good to be able to remove it :)

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
bennetelli: Screenshot from 2021-12-13 08-51-39

yes, it does

aggregator/src/test/java/com/redhat/cloud/notifications/DailyEmailAggregationJobTest.java
bennetelli: Added-

.rhcicd/clowdapp-aggregator.yaml
```diff
@@ -76,6 +78,9 @@ parameters:
 - name: ENV_NAME
   description: ClowdEnvironment name (ephemeral, stage, prod)
   required: true
+- name: PROMETHEUS_PUSHGATEWAY_URL
```
theute, line 81: The parameter key for the provided value is `PROMETHEUS_PUSHGATEWAY` so the same key needs to be here, and then line 65 would need to refer to that key name as well.

theute (changes requested):

.rhcicd/clowdapp-aggregator.yaml
bennetelli: changed.

theute (approved):

PR #855, opened 2021-12-14
Add cache informations

bennetelli:


backend/pom.xml
```diff
@@ -71,6 +71,11 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-cache</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.github.ben-manes.caffeine</groupId>
```
pilhuhn, line 75: Is(n't) that pulled via the quarkus-cache dep in the entry above?

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -68,7 +84,11 @@ public void initCounters() {
                 .onItem().invoke(() -> getUsersPageTimer.stop(meterRegistry.timer("rbac.get-users.page", "accountId", accountId)));
             }
         )
-        .onItem().invoke(users -> getUsersTotalTimer.stop(meterRegistry.timer("rbac.get-users.total", "accountId", accountId, "users", String.valueOf(users.size()))));
+            .onItem().invoke(users -> getUsersTotalTimer.stop(meterRegistry.timer("rbac.get-users.total", "accountId", accountId, "users", String.valueOf(users.size()))));
+
+        rbacCache.put(new RbacCacheKey(accountId, adminsOnly), rbacCachedUsers.await().indefinitely());
```
gwenneg, line 89: `await().indefinitely()` must never be used in a reactive method.

You can add this between lines 70 and 71 of the old code:
```
        .onItem().invoke(users -> rbacCache.put(new RbacCacheKey(accountId, adminsOnly), users);
```

backend/pom.xml
gwenneg: It should and the version has to come from the Quarkus BOM anyway.

PR #858, opened 2021-12-15
Add logging for backoffice host and prometheus pushgateway

bennetelli:


gwenneg:
The backoffice URL is already visible in the logs:
```
2021-12-16 13:36:11,521 DEBUG [com.red.clo.not.pro.web.WebhookTypeProcessor] (vert.x-eventloop-thread-3) Webhook request to backoffice-proxy-insights-services.ext.us-east.dc.prod.paas.redhat.com was successful: 200
```
Is there a reason why we need to log it at application startup?

theute:
The context for this PR was more about debugging (it turned out that it was not need since we can see those in OpenShift directly for running pods)

But here we definitely don't need the backffice proxy URL since it's not used (and not set) in the aggregator pod (only in backend).

In general, I think it would be valuable to log as INFO a set of environment parameters like those 2 (and likely a few more, KAFKA...), sometimes they change and can explain some issues.
We won't want to show all of them though for security purposes.

gwenneg:
Oh that's in aggregator, I didn't see that 😄

gwenneg:
> In general, I think it would be valuable to log as INFO a set of environment parameters like those 2 (and likely a few more, KAFKA...), sometimes they change and can explain some issues. We won't want to show all of them though for security purposes.

👍

PR #860, opened 2021-12-16
NOTIF-427 Return the payload on demand in the /events response

gwenneg:
FYI we are loading the payload from the database even when it is not returned in the response, which is bad for performances.

I'll change that in another PR.

### DIFF
--- a/backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest.java
+++ b/backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest.java
@@ -52,6 +52,7 @@ import static java.lang.Boolean.TRUE;
 import static java.time.ZoneOffset.UTC;
 import static java.util.UUID.randomUUID;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @QuarkusTest
@@ -60,6 +61,7 @@ public class EventServiceTest extends DbIsolatedTest {
 
     private static final String OTHER_ACCOUNT_ID = "other-account-id";
     private static final LocalDateTime NOW = LocalDateTime.now(UTC);
+    private static final String PAYLOAD = "payload";
 
     @Inject
     Mutiny.SessionFactory sessionFactory;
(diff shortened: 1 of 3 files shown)


josejulio:
> FYI we are loading the payload from the database even when it is not returned in the response, which is bad for performances.

That could be the reason it's taking too long :D

backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest.java
```diff
@@ -438,13 +456,14 @@ void testAllQueryParams() {
...
                      */
-                    return getEventLogPage(defaultIdentityHeader, null, null, null, null, null, Set.of(EMAIL_SUBSCRIPTION), Set.of(FALSE), null, null, null);
+                    return getEventLogPage(defaultIdentityHeader, null, null, null, null, null, Set.of(EMAIL_SUBSCRIPTION), Set.of(FALSE), null, null, null, false);
                 }))
                 .emitOn(MutinyHelper.executor(vertx.getOrCreateContext()))
                 .invoke(page -> {
                     assertEquals(1, page.getMeta().getCount());
                     assertEquals(1, page.getData().size());
                     assertSameEvent(page.getData().get(0), model.events.get(0), model.notificationHistories.get(0), model.notificationHistories.get(1));
+                    assertNull(page.getData().get(0).getPayload());
```
josejulio, line 466: Is there any test when this is not null?

gwenneg: Yes, it's well hidden 😄

backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest.java
```diff
@@ -342,13 +354,14 @@ void testAllQueryParams() {
...
                      */
-                    return getEventLogPage(defaultIdentityHeader, Set.of(model.bundles.get(1).getId()), Set.of(model.applications.get(1).getId()), model.eventTypes.get(1).getDisplayName(), NOW.minusDays(3L), NOW.minusDays(1L), Set.of(EMAIL_SUBSCRIPTION), Set.of(TRUE), 10, 0, "created:desc");
+                    return getEventLogPage(defaultIdentityHeader, Set.of(model.bundles.get(1).getId()), Set.of(model.applications.get(1).getId()), model.eventTypes.get(1).getDisplayName(), NOW.minusDays(3L), NOW.minusDays(1L), Set.of(EMAIL_SUBSCRIPTION), Set.of(TRUE), 10, 0, "created:desc", true);
                 }))
                 .emitOn(MutinyHelper.executor(vertx.getOrCreateContext()))
                 .invoke(page -> {
                     assertEquals(1, page.getMeta().getCount());
                     assertEquals(1, page.getData().size());
                     assertSameEvent(page.getData().get(0), model.events.get(2), model.notificationHistories.get(3));
+                    assertEquals(PAYLOAD, page.getData().get(0).getPayload());
```
gwenneg, line 364: Here!

PR #861, opened 2021-12-16
Split DeleteModal from App

aferd:


backend/src/main/resources/META-INF/resources/internal/asset-manifest.json
```diff
@@ -1,21 +1,21 @@
 {
   "files": {
     "main.css": "./static/css/main.ebdcbaff.chunk.css",
-    "main.js": "./static/js/main.de9c4f91.chunk.js",
-    "main.js.map": "./static/js/main.de9c4f91.chunk.js.map",
+    "main.js": "./static/js/main.b0de9fd0.chunk.js",
```
josejulio, line 4: Files under `backend/src/main/resources/META-INF/resources/internal`  should not be modified by our commits.
I'm sure this one wasn't on purpose, but we need revert back these changes in this PR.

josejulio (approved):

PR #864, opened 2021-12-17
Include Kafka records consumed rate in health check

gwenneg:
This adds a new check (disabled by default) to the existing health checks.

After a 5 minutes initial delay (configurable), every 1 minute (configurable) the `kafka.consumer.fetch.manager.records.consumed.rate` Micrometer gauge value will be tested.

A gauge value equal to zero will be recorded as a minor failure. If the number of _consecutive_ minor failures exceeds the maximum number allowed (configurable), then the app will be marked as `DOWN` the next time OpenShift calls `/health`.

If the gauge value is greater than zero when it is tested, the minor failures counter will be reseted.

### DIFF
--- a/.rhcicd/clowdapp-backend.yaml
+++ b/.rhcicd/clowdapp-backend.yaml
@@ -77,6 +77,10 @@ objects:
         env:
         - name: ENV_NAME
           value: ${ENV_NAME}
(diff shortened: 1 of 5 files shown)


gwenneg:
It seems to work locally, I have to confirm that on stage now.

theute (approved):

PR #824, opened 2021-12-05
If the integration reports an error, re-inject for e.g. email notification

pilhuhn:
This is a first PoC to create a new event from the return information of a previous one.
Here if an integration reports a failure, we create a new event, which is then delivered via default-default-BG to admins of the account via Email.

B/a/et are

```
 .setBundle("platform")
 .setApplication("notifications")
 .setEventType("integration-failed")

```
There is a switch `reinject.enabled=true` to disable this reinjection. Default is off (also in code).

Bildschirmfoto 2021-12-05 um 11 58 29

Todos:
[ ] Decide on new b/a/et
[X] Add migration to add new b/a/et
[X] Decide if send_email_to_admin should be used for processing of the new event(type) by default
[ ] How to prevent loops if the admin sets processing to the failed integration?

PR #872, opened 2021-12-21
Reinject+default bg

pilhuhn:
This PR tries to apply the new default BGs to the reinject changes in order to have the email to admin as default when an integration fails.
It builds on top of #824 and #809 

TODO: 
[X] rewrite the 2 migrations into one
[ ] add tests


pilhuhn:
This is now incorporated in #824 (and master)

PR #875, opened 2021-12-22
Check the Kafka health once instead of periodically

gwenneg:
Checking periodically made sense at first because the check was initially based on `kafka.consumer.fetch.manager.records.consumed.rate` which is a gauge. But the latest version is based on a counter, so we can simply check once every 5 minutes and fail as soon as the number of processed messages didn't increase since the previous check.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/health/KafkaConsumedTotalChecker.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/health/KafkaConsumedTotalChecker.java
@@ -20,15 +20,12 @@ public class KafkaConsumedTotalChecker {
     @ConfigProperty(name = "notifications.kafka-consumed-total-checker.enabled", defaultValue = "false")
     boolean enabled;
 
-    @ConfigProperty(name = "notifications.kafka-consumed-total-checker.max-failures", defaultValue = "5")
-    int maxFailures;
-
     @Inject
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/redhat/cloud/notifications/health/KafkaConsumedTotalChecker.java
```diff
@@ -42,25 +39,19 @@ void postConstruct() {
...
-                    LOGGER.debugf("Kafka records consumed total check succeeded for topic '%s' after %d failures",
-                            INGRESS_TOPIC, currentFailures);
-                    currentFailures = 0;
-                }
-            } else {
-                currentFailures++;
+            if (currentTotal == previousTotal) {
                 LOGGER.debugf("Kafka records consumed total check failed for topic '%s'", INGRESS_TOPIC);
+                down = true;
             }
```
josejulio, line 49: I suppose the health checks would kill the pod before the next execution of this one? (i.e. having a `else { down = false }` would be superfluos?)

gwenneg: Yes that's the idea. If `down` is true even once, the pod should be killed.

josejulio (approved):

PR #880, opened 2021-12-23
Add Ansible bundle to ephemeral data

gwenneg:
Based on the information from https://issues.redhat.com/browse/NOTIF-398, this should work.


josejulio (approved):
LGTM, just notice that this won't be in prod soon

PR #881, opened 2021-12-23
NOTIF-410 Improve the /events response time

gwenneg:
I expect this PR to have a major impact on the `/events` response time issue we currently have on prod. It may not fix it entirely though, we'll need to confirm that with tests on prod.

I created a local database similar to prod, with millions of `notification_history` and `event` records, which allowed me to reproduce the prod issue on my machine. Then I analyzed (with Postgres' `EXPLAIN`) and timed each SQL query involved during an `/events` call. I tested several fix candidates and this one had the bigger impact by far.

⚠️ This PR will remain a draft until `app-interface!31022` is merged. I need to know the exact numbers of rows on prod before I can determine the final `autovacuum` settings changes.

For more information about autovacuum tuning, here's an interesting article: https://www.2ndquadrant.com/en/blog/postgresql-vacuum-and-analyze-best-practice-tips/

### DIFF
--- a/backend/src/main/resources/db/migration/V1.34.0__NOTIF-410_events_response_time.sql
+++ b/backend/src/main/resources/db/migration/V1.34.0__NOTIF-410_events_response_time.sql
@@ -1,0 +1,24 @@
+-- The response time improvement mostly comes from this.
+CREATE INDEX ix_notification_history_event_id ON notification_history (event_id);
+
+-- The following instructions change the Postgres autovacuum settings for several tables.
+
+-- Indexes on tables with lots of DELETE or UPDATE operations tend to become less effective over time because Postgres
+-- keeps the dead tuples that were deleted on the disk. Vacuuming the tables can fix that but the default autovacuum
+-- settings are not suitable for tables with a large amount of records. By default, Postgres will autovacuum a table
+-- when it contains [table total records] * autovacuum_vacuum_scale_factor (default: 0.2) + autovacuum_vacuum_threshold
+-- (default: 50). This means that a table with 1 million records will be vacuumed when the dead tuples count reaches
+-- 200050. Changing the autovacuum settings helps ensuring that autovacuum will run more often, will take less time and
+-- won't be interrupted by transactions that would need to lock the vacuumed table.
+
+-- The 'event' table contains ~3 million records on prod when this script is written. Old events are deleted at a rate
+-- of ~4000 per hour. Assuming this rate remains constant, the autovacuum will happen approximately every hour instead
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
```diff
@@ -116,7 +117,7 @@
     private static String addHqlConditions(String hql, Set<UUID> bundleIds, Set<UUID> appIds, String eventTypeDisplayName,
                                            LocalDate startDate, LocalDate endDate, Set<EndpointType> endpointTypes, Set<Boolean> invocationResults) {
         if (bundleIds != null && !bundleIds.isEmpty()) {
-            hql += " AND b.id IN (:bundleIds)";
+            hql += " AND a.bundle.id IN (:bundleIds)";
```
josejulio, line 120: I suppose this adds the JOIN on demand?

gwenneg: No the join with `bundles` is gone for good.

Old query:
```
[...]
 inner join
        bundles bundle3_
            on applicatio2_.bundle_id=bundle3_.id
[...]
```
New query:
```
[...]
 applicatio2_.bundle_id in (
                $2
            )
[...]
```

josejulio: nice!

gwenneg: Hmm not so nice, the tests showed me something wrong because of the ordering 😄 This needs more work anyway!

gwenneg: @josejulio Do you think this is needed?
https://github.com/RedHatInsights/notifications-backend/blob/bcb91e750c40b0128300ce89e41699797bc3c846/backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java#L179-L184

The frontend only sorts on the creation time which makes sense.

josejulio: I'm sure we will eventually be asked (if not already in the mocks) to sort by any column.

gwenneg: No longer relevant, my approach to fix the response time has changed.

gwenneg:

gwenneg:
This is ready for a review.

bennetelli (approved):

gwenneg:
PR check failures are still unrelated.

PR #882, opened 2021-12-23
PoC: Query builder

josejulio:


josejulio:
I liked this approach but it was too simple and had issues concatenating AND's

PR #888, opened 2022-01-05
Update grafana dashboard

bennetelli:


theute:
Should probably be stacked as well

bennetelli:
That is the plan. I would like to have it "running" first and remove the obsolete parts then.

PR #889, opened 2022-01-05
NOTIF-444 separate counters instead of one with different tags

bennetelli:


josejulio:
What's wrong with the tags ?

theute:
We can't use this PR, since we'll keep adding bundles/applications and we should limit code changes when we do so (I know we have to do to add email templates so we still need to to a PR when we add an app but we want to walk away from this)

theute:
> What's wrong with the tags ?

With the current code deployed to stage this morning (and "dynamic" tags), the tags don't show up in Prometheus:
https://prometheus.crcs02ue1.devshift.net/graph?g0.expr=processor_email_processed_total&g0.tab=0&g0.stacked=0&g0.show_exemplars=0&g0.range_input=2h

gwenneg:
We already have dynamic tags working on prod in [EventConsumer](https://github.com/RedHatInsights/notifications-backend/blob/master/backend/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java). We should change the code from https://github.com/RedHatInsights/notifications-backend/pull/886 to something similar.

theute:
> We already have dynamic tags working on prod in [EventConsumer](https://github.com/RedHatInsights/notifications-backend/blob/master/backend/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java). We should change the code from #886 to something similar.

But unless I read too fast the one we have is a timer, would be good to get the actual number of emails (same for events BTW)

gwenneg:
It is a timer indeed, but timers can be used to count things. See `rbac_get_users_total_seconds_count` in Prometheus.

theute:
> It is a timer indeed, but timers can be used to count things. See `input_consumed_seconds_count` in Prometheus.

Ah yes indeed.
Still wondering why the gauge + tags didn't work :-/

gwenneg:
BTW, the dynamic tags can be checked locally by simply calling `/metrics` from localhost. We don't need to check if that works from Prometheus.

bennetelli:
> 

Actually that's not a solution. Yes, the micrometer metrics are available in /metrics, but they will be exposed from there to Prometheus and in /metrics it's working, but not in Prometheus. So looking at /metrics only doesn't replace taking a look at Prometheus as well.

bennetelli:
> It is a timer indeed, but timers can be used to count things. See `input_consumed_seconds_count` in Prometheus.

Yes they can. But from a semantically point of view it's not an elegant solution. It's a bit like storing booleans in an integer instead of the boolean datatype. We should check why it doesn't work with counters instead of creating workarounds.

gwenneg:
> Actually that's not a solution. Yes, the micrometer metrics are available in /metrics, but they will be exposed from there to Prometheus and in /metrics it's working, but not in Prometheus. So looking at /metrics only doesn't replace taking a look at Prometheus as well.

You mean that with the current `master` implementation (not the latest one), you saw the tags in the `/metrics` output?

bennetelli:
Don't need this anymore.

PR #890, opened 2022-01-07
NOTIF-381 Speed up user fetching

bennetelli:
TODO:
* ~~add secrets to vault (PR is here: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/31795~~
* ~~add paging in case the query takes too long.~~
* add keystore password in clowder/as OpenShift secrets (for each environment)
* ~~retrieving "active: true" users only right now. Is that okay?~~
* renew prod certificate (Thomas needs to do it)


bennetelli:
There are two things I am fighting with at the moment:
* the integration test/the rest client doesn't like my request and returns unknown errors in combination with 401, 403 and so on.
* there are two methods: EmailTest.mockGetUsers() and RbacRecipientUsersProviderTest.mockGetUsers() that needs to be chanced, but I have no idea how it should look like so it works the same way as before.

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/ITUserServiceWrapper.java
```diff
@@ -0,0 +1,32 @@
...
+
+@ApplicationScoped
+public class ITUserServiceWrapper {
+
+    @Inject
+    @RestClient
+    ITUserService itUserService;
+
+    public Uni<Page<RbacUser>> getUsers(String accountId, boolean adminsOnly, int i, Integer rbacElementsPerPage) {
+        Path fileName = Path.of("src/main/resources/users.json");
```
josejulio, line 22: I don't see this file in this commit. What is it for?

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/AccountRelationship.java
```diff
@@ -0,0 +1,41 @@
...
+        "allOf",
+        "by"
+})
+public class AccountRelationship {
+
+    @JsonProperty("allOf")
+    private List<String> allOf = List.of("primary_email");
+
+    @JsonProperty("by")
+    private By__1 by;
```
josejulio, line 20: Was this name auto generated?

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/ITUserServiceWrapper.java
bennetelli: It's still a draft :-) this file is currently used to do a static http post. it will be replaced by dynamic content (accountId, adminsOnly and so on) later. I wanted to get it working first.

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/AccountRelationship.java
bennetelli: Yes

josejulio: Could you add some steps on how to generate it again in case we need to sync?

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -114,7 +110,32 @@ public void initCounters() {
...
     }
 
-    private Uni<List<User>> getWithPagination(Function<Integer, Uni<Page<RbacUser>>> fetcher) {
+    private Uni<List<User>> getWithPagination(Uni<List<ITUserResponse>> itUserResponses) {
+        return itUserResponses.onItem().transform(page -> page.stream().map(itUserResponse -> {
+                    User user = new User();
+//                    user.setUsername(itUserResponse.getPersonalInformation().getUsername());
+                    user.setEmail(itUserResponse.getAccountRelationships().get(0).getEmails().toString());
+//                    user.setAdmin(rbacUser.getOrgAdmin());
+//                    user.setActive(rbacUser.getActive());
```
josejulio, line 119: I understand that this is still a draft, but does the IT service allows to know the username / admin / active status?

bennetelli: admin, yes. username and active status I don't know yet

bennetelli: username is probably principal

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/AccountRelationship.java
```diff
@@ -0,0 +1,37 @@
+package com.redhat.cloud.notifications.recipients.itservice.pojo.request;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
+import java.util.List;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
gwenneg, line 9: General comment: since `@JsonInclude` only applies to serialization, it can be removed from all classes in this PR as we only do deserialization.

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/AccountRelationship.java
```diff
@@ -0,0 +1,37 @@
...
+@JsonPropertyOrder({
+        "allOf",
+        "by"
+})
+public class AccountRelationship {
+
+    @JsonProperty("allOf")
```
gwenneg, line 16: General comment: `@JsonProperty` is not useful if its `value` field is exactly the same than the Java field name. Most (maybe all) `@JsonProperty` annotations can be removed from this PR.

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/AccountRelationship.java
```diff
@@ -0,0 +1,37 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
+import java.util.List;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder({
+        "allOf",
+        "by"
+})
```
gwenneg, line 13: General comment: `@JsonPropertyOrder` is usually used to change the serialization output. In rare occasion it can affect the deserialization but we probably don't need it here.

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/AccountRelationship.java
```diff
@@ -0,0 +1,37 @@
...
+    @JsonProperty("by")
+    private By__1 by;
+
+    public List<String> getAllOf() {
+        return allOf;
+    }
+
+    public void setAllOf(List<String> allOf) {
+        this.allOf = allOf;
+    }
```
gwenneg, line 28: General comment: if we only use this class and the other similar ones for deserialization of the IT response, `public` fields  will do the same job with a lot less code:
```
public class AccountRelationship {
    public List<String> allOf; // no need for '= null'
    public By__1 by;
}
```

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/AccountRelationship.java
bennetelli: I actually like to have properties like this (even in JPA Entities) because the code gets independent from the table names/json attribute names and we can refactor the code without any side effects.

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/AccountRelationship.java
bennetelli: removed. thx

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/AccountRelationship.java
bennetelli: removed. thx

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/AccountRelationship.java
bennetelli: good catch. thx. The object structure is still on my list, but I wanted to make the tests green first.

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/AccountRelationship.java
bennetelli: sure. where do you want it to be? in the readme of the application?

josejulio: a new section on the README would be nice

bennetelli: done :)

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/AccountRelationship.java
gwenneg: > I actually like to have properties like this (even in JPA Entities) because the code gets independent from the table names/json attribute names and we can refactor the code without any side effects.

While this makes perfect sense, it is still debatable. 😃 

I never code for an hypotetical future situation because most of the time that situation will never happen. Coding for the present usually makes the code lighter and easier to maintain.

bennetelli: me neither normally. I am usually taking the pragmatic way. But in this case I like being "refactoring safe".

bennetelli: we can remove it as well.. i can live with both :)

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/ITUserServiceWrapper.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.List;
+
+@ApplicationScoped
+public class ITUserServiceWrapper {
+
+    @Inject
+    @RestClient
+    ITUserService itUserService;
+
+    public Uni<List<ITUserResponse>> getUsers(String accountId, boolean adminsOnly) {
```
josejulio, line 19: Not to apply right now - but we might want to optimize this query - Would it be possible in the future to also pass the list of users that we want to fetch? 

In most cases (but not all cases) we have the list of users (i.e. all the subscribed) that we would like to filter for.

bennetelli: Generally it's possible, yes. We can find users by a lot of attributes. If we need it, we can just add it.

bennetelli:
The Build & Test is failing because the password of the keystore isn't in the application.properties.

Maybe this is something I need to change somehow. Need to think about it again.

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/ITUserServiceWrapper.java
bennetelli: I can add a link to the gitlab IT User Service Readme if you want, @josejulio ?

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/ITUserService.java
```diff
@@ -0,0 +1,21 @@
...
+import javax.ws.rs.core.MediaType;
+import java.util.List;
+
+@RegisterRestClient(configKey = "it-s2s")
+public interface ITUserService {
+
+    @POST
+    @Path("/findUsers")
+    @Produces(MediaType.APPLICATION_JSON)
+    Uni<List<ITUserResponse>> getUsers(ITUserRequest string);
```
gwenneg, line 20: Shouldn't this be annotated with `@Consumes(APPLICATION_JSON)`?

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/ITUserServiceWrapper.java
```diff
@@ -0,0 +1,22 @@
...
+import com.redhat.cloud.notifications.recipients.itservice.pojo.response.ITUserResponse;
+import io.smallrye.mutiny.Uni;
+import org.eclipse.microprofile.rest.client.inject.RestClient;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+import java.util.List;
+
+@ApplicationScoped
+public class ITUserServiceWrapper {
```
gwenneg, line 13: Why is that wrapper needed?

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/ITUserRequest.java
```diff
@@ -0,0 +1,42 @@
...
+        include.allOf = List.of("authentications", "personal_information");
+        List<AccountRelationship> accountRelationships = new LinkedList<>();
+        AccountRelationship accountRelationship1 = new AccountRelationship();
+        accountRelationship1.allOf = List.of("primary_email");
+
+        final By__1 by1 = new By__1();
+        by1.active = true;
+        accountRelationship1.by = by1;
+        accountRelationships.add(accountRelationship1);
+        include.accountRelationships = accountRelationships;
```
gwenneg, line 38: `include` seems constant. Can we avoid rebuilding it each time we need to send a request?

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/ITUserRequest.java
```diff
@@ -0,0 +1,42 @@
...
+    public ITUserRequest(String accountId, boolean adminsOnly) {
+        final By by = new By();
+        AllOf allOf = new AllOf();
+        allOf.status = "enabled";
+        allOf.ebsAccountNumber = accountId;
+
+        if (adminsOnly) {
+            PermissionCode permissionCode = new PermissionCode();
+            permissionCode.value = "admin:org:all";
+            permissionCode.operand = "eq";
```
gwenneg, line 20: `permissionCode` seems constant. Can we avoid rebuilding it each time we need to send a request?

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/By__1.java
```diff
@@ -0,0 +1,6 @@
+package com.redhat.cloud.notifications.recipients.itservice.pojo.request;
+
+public class By__1 {
```
gwenneg, line 3:
```suggestion
public class AccountRelationshipBy {
```

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/By.java
```diff
@@ -0,0 +1,7 @@
+package com.redhat.cloud.notifications.recipients.itservice.pojo.request;
+
+public class By {
```
gwenneg, line 3:
```suggestion
public class ITUserRequestBy {
```

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/AccountRelationship.java
```diff
@@ -0,0 +1,9 @@
+package com.redhat.cloud.notifications.recipients.itservice.pojo.request;
+
+import java.util.List;
+
+public class AccountRelationship {
```
gwenneg, line 5: General comment: what should happen if we receive an unknown field? Right now, it will cause an exception. We could add `@JsonIgnoreProperties(ignoreUnknown = true)` to each pojo to prevent exceptions.

We could even drop all the IT response fields we don't use in notifications. That would make the response pojos simpler.

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/ITUserService.java
bennetelli: It's working with and without Procudes and Consumes, but I can add it f you want.

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/AccountRelationship.java
bennetelli: Not sure if this could ever happen. With the request we specify what attributes or kind of attributes we would like to get in the response. I tested several ids and it worked every time. Also the "huge" response of all users are working with the generated response classes.

We can add ignoreUnknown just to be 100% sure.

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/ITUserService.java
gwenneg: Probably because `APPLICATION_JSON` is the default in Quarkus right now. It would be safer to have an explicit value though, that kind of default can changed (happened in Quarkus already in the past).

bennetelli: let's just go with the annotation to be safe. It doesn't add much complexity so adding a more or less "obsolete" annotation doesn't change much.

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/ITUserServiceWrapper.java
josejulio: I think we need to store that somewhere for easier access, but not sure if on this README or somewhere else.

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/AccountRelationship.java
josejulio: +1 on adding it, we never know when the API will provide new fields

backend/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -114,7 +113,29 @@ public void initCounters() {
...
+            final List<Email> emails = itUserResponse.accountRelationships.get(0).emails;
+            for (Email email : emails) {
+                if (email != null && email.isPrimary != null && email.isPrimary) {
+                    String address = email.address;
+                    user.setEmail(address);
+                }
+            }
+
+            user.setAdmin(adminsOnly);
+            user.setActive(true);
```
josejulio, line 130: The API only provides active users?

josejulio: > retrieving "active: true" users only right now. Is that okay?

ah, I think this is OK.

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/ITUserServiceWrapper.java
bennetelli: It's already in the jira ticket. So in case we really need it and nobody knows the url, we are still able to find it via git blame

josejulio:
>  The Build & Test is failing because the password of the keystore isn't in the application.properties.
> 
> Maybe this is something I need to change somehow. Need to think about it again.

Could we use dummy keystore for testing?

backend/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/AccountRelationship.java
bennetelli: done already.

bennetelli:
> 

fixed. the keystore isn't necessary anymore.

josejulio:
What's missing here?

bennetelli:
> What's missing here?

we will take the other notif-381 PR. this one is just here for bookkeeping. I opened a new one as it was easier to merge.

theute:
Then shouldn't we close that one ?

bennetelli:
we could. initially i wanted to have it here to find it more easily, but I can also close it.

PR #899, opened 2022-01-11
NOTIF-152 Add recipient users

josejulio:
Allows notification message to specify a list of users.

This allows an applications to specify users they want to send a specific notification. This list of users is added to whatever the notification administrator already set and does not overwrite the setting.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/recipients/RecipientResolver.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/recipients/RecipientResolver.java
@@ -31,20 +31,24 @@ public class RecipientResolver {
             usersUni = rbacRecipientUsersProvider.getGroupUsers(accountId, request.isOnlyAdmins(), request.getGroupId());
         }
 
+        // The base list of recipients comes from RBAC.
         return usersUni
                 .onItem().transform(Set::copyOf)
                 .onItem().transform(users -> {
-                    // This step requires a specific set of users and all others needs to be ignored (for this step)
+                    // If the request contains a list of users, then the recipients from RBAC who are not included in
+                    // the request users list are filtered out.
                     if (request.getUsers().size() > 0) {
                         return filterUsers(users, request.getUsers());
                     }
-
+                    // Otherwise, the full list of recipients from RBAC will be processed by the next step.
                     return users;
                 }).onItem().transform(users -> {
+                    // If the user preferences should be ignored, the recipients from RBAC (possibly filtered by the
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -45,7 +45,8 @@
         <!-- Be careful with bumps, version 5.11.2 caused tests instability so we had to rollback to 5.5.4 -->
         <mockserver-client-java.version>5.5.4</mockserver-client-java.version>
 
-        <insights-notification-schemas-java.version>0.4</insights-notification-schemas-java.version>
+
+        <insights-notification-schemas-java.version>0.5</insights-notification-schemas-java.version>
```
josejulio, line 49: Removing extra line
```suggestion
        <insights-notification-schemas-java.version>0.5</insights-notification-schemas-java.version>
```

backend/src/main/java/com/redhat/cloud/notifications/recipients/RecipientResolver.java
```diff
@@ -34,17 +31,31 @@
...
+
+                    return users;
+                }).onItem().transform(users -> {
+                    if (request.isIgnoreUserPreferences()) {
+                        return Set.copyOf(users);
+                    }
+
+                    return filterUsers(users, subscribers);
+                });
+    }
```
gwenneg, line 50: This is all clear to me, but I think it needs more comments for someone who wouldn't know that code at all:
```suggestion
        // The base list of recipients comes from RBAC.
        return usersUni
                .onItem().transform(Set::copyOf)
                .onItem().transform(users -> {
                    // If the request contains a list of users, then the recipients from RBAC who are not included in
                    // the request users list are filtered out.
                    if (request.getUsers().size() > 0) {
                        return filterUsers(users, request.getUsers());
                    }
                    // Otherwise, the full list of recipients from RBAC will be processed by the next step.
                    return users;
                }).onItem().transform(users -> {
                    // If the user preferences should be ignored, the recipients from RBAC (possibly filtered by the
                    // previous step) is returned without filtering out the users who didn't subscribe to the event type.
                    if (request.isIgnoreUserPreferences()) {
                        return Set.copyOf(users);
                    }
                    // Otherwise, the recipients from RBAC who didn't subscribe to the event type are filtered out.
                    return filterUsers(users, subscribers);
                });
    }
```

josejulio: Perfect!

gwenneg (approved):
LGTM

PR #900, opened 2022-01-11
split create-edit-modal from main app

aferd:


admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -158,54 +149,15 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
-                                                    onClick={ () => setShowModal(false) }>Cancel</Button>
-                                            </ActionGroup>
-                                        </Form>
-                                    </Modal>
+                                    <CreateEditModal
+                                        isEdit={ isEdit }
+                                        showModal={ showModal }
+                                        applicationName={ application?.displayName }
+                                        eventType={ eventType }
+                                        onClose={ onClose }
```
josejulio, line 157: This `onClose` function sets the `showDeleteModal` to false, not the `showModal`.
I suggest to create one `onCloseDeleteModal` and one `onCloseCreateEditModal` or something similar.

aferd:
The submit is no longer working, though in the main App nothing was changed as well as when going to edit the current fields do not come up..

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -120,12 +111,22 @@ export const ApplicationPage: React.FunctionComponent = () => {
         return true;
     }, [ deleteEventTypeMutation.mutate, eventType.id ]);
 
+    const handleChange = (value: string, event: React.FormEvent<HTMLInputElement> | React.FormEvent<HTMLTextAreaElement>) => {
+        const target = event.target as HTMLInputElement;
+        setEventType(prev => ({ ...prev, [target.name]: target.value }));
+    };
```
josejulio, line 117: I would preffer if this code lives in the modal

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -33,10 +29,10 @@ export const ApplicationPage: React.FunctionComponent = () => {
     const eventTypesQuery = useEventTypes(applicationId);
     const applicationTypesQuery = useApplicationTypes(applicationId);
     const deleteEventTypeMutation = useDeleteEventType();
+    const newEvent = useCreateEventType();
 
     const columns = [ 'Event Type', 'Name', 'Description', 'Event Type Id' ];
 
-    const newEvent = useCreateEventType();
     const [ eventType, setEventType ] = React.useState<Partial<EventType>>({});
```
josejulio, line 36: I would preffer to have this in the modal as well.

admin-console/src/main/webapp/src/components/CreateEditModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+import { useCreateEventType } from '../services/CreateEventTypes';
+
+interface CreateEditModalProps {
+    isEdit: boolean;
+    showModal: boolean;
+    applicationName?: string;
+    eventTypeName?: string;
+    eventTypeDisplayName?: string;
+    eventTypeDescription?: string;
```
josejulio, line 13: Instead of receiving each attribute, we could receive an initializer for the evenType: i.e.:
`initialEventType?: EventType`

and it could be used as:
```
const [ eventType, setEventType ] = React.useState<Partial<EventType>>(props.initalEventType ?? {});
```

admin-console/src/main/webapp/src/components/CreateEditModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+                        <TextArea
+                            type='text'
+                            value={ props.eventTypeDescription }
+                            onChange={ props.onChange }
+                            id='description'
+                            name="description"
+                        /></FormGroup>
+                    <ActionGroup>
+                        <Button variant='primary' type='submit'
+                            { ...(createNewEvent.loading || createNewEvent.payload?.status !== 200) ?
```
josejulio, line 64: I don't think this is doing what we need. `createEvent` in this component is never used. We could pass a `loading` prop to this component.

admin-console/src/main/webapp/src/components/CreateEditModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+    eventTypeDescription?: string;
+    onClose: () => void;
+    onSubmit: () => void;
+    eventTypeQuery: unknown;
+    onChange: (value: string, event: React.FormEvent<HTMLInputElement> | React.FormEvent<HTMLTextAreaElement>) => void;
+
+}
+
+export const CreateEditModal: React.FunctionComponent<CreateEditModalProps> = (props) => {
+    const createNewEvent = useCreateEventType();
```
josejulio, line 22: Not used here, lets delete it from here.

josejulio (changes requested):

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
aferd: this currently is not being used in the modal, but instead here in the handleSubmit...

admin-console/src/main/webapp/src/components/CreateEditModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+interface CreateEditModalProps {
+    isEdit: boolean;
+    showModal: boolean;
+    applicationName?: string;
+    eventTypeName?: string;
+    eventTypeDisplayName?: string;
+    eventTypeDescription?: string;
+    onClose: () => void;
+    onSubmit: () => void;
```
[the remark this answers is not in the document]
aferd: is this? meaning the onClose and onSubmit?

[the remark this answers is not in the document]
josejulio: I was reffering to onSubmit: i.e. `onSubmit: (eventType: EventType) => void`

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
josejulio: Yes, that's what i mean. instead of using this variable here, just pass it to the `onSubmit`  (from the modal)

admin-console/src/main/webapp/src/components/CreateEditModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+    initialEventType?: EventType;
+    isLoading?: boolean;
+    onClose: () => void;
+    onSubmit: () => void;
+    eventTypeQuery: unknown;
+
+}
+
+export const CreateEditModal: React.FunctionComponent<CreateEditModalProps> = (props) => {
+    const [ eventType, setEventType ] = React.useState<Partial<EventType>>(props.initialEventType ?? {});
```
josejulio, line 20: This is good.
Now we need to pass it to the `onSubmit` by doing something like:
```
const onSubmitLocal = React.useCallback(() => { props.onSubmit(eventType)  }, [ eventType, props.onSubmit]) // might require to make this more verbose to make the linter happy
```

josejulio: Something else that we might want to do is to initialize this value everytime the isOpen or maybe the initialEventType changes:
```
React.useEffect(() => {
  setEventType(props.initialEventType);
}, [ props.initialEventType ]);
```

admin-console/src/main/webapp/src/components/CreateEditModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+    const handleChange = (value: string, event: React.FormEvent<HTMLInputElement> | React.FormEvent<HTMLTextAreaElement>) => {
+        const target = event.target as HTMLInputElement;
+        setEventType(prev => ({ ...prev, [target.name]: target.value }));
+    };
+
+    const onSubmitLocal = React.useCallback(() => {
+        props.onSubmit(eventType); }, [ eventType, props ]);
+
+    React.useEffect(() => {
+        setEventType(props.initialEventType);
```
aferd, line 35: I am getting an error here `Argument of type 'Partial<EventType> | undefined' is not assignable to parameter of type 'SetStateAction<Partial<EventType>>'.` on the props.initialEventType

josejulio: I suppose it's complaining because it could be undefined. So you would have to provide a fallback value
```suggestion
        setEventType(props.initialEventType ?? {});
```

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
aferd: I moved this to the modal, but trying to understand. You're saying pass it to the new onsubmitLocal from the modal?

admin-console/src/main/webapp/src/components/CreateEditModal.tsx
```diff
@@ -0,0 +1,88 @@
...
+        const target = event.target as HTMLInputElement;
+        setEventType(prev => ({ ...prev, [target.name]: target.value }));
+    };
+
+    const onSubmitLocal = React.useCallback(() => {
+        props.onSubmit(eventType);
+    }, [ eventType, props ]);
+
+    React.useEffect(() => {
+        setEventType(props.initialEventType ?? {});
```
aferd, line 35: this made it happy. not sure what happened to my original comment

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
josejulio: The new `onSubmitLocal` should pass it to the `onSubmit` that is in `ApplicationPage` so it will be something like:
In ApplicationPage:

// ```
onSubmit(eventType: EventType) {
  // ...
}
// ....
<CreateEditModal onSubmit={ onSubmit }  />
```

and in CreateEditModal
```
// ...
onSubmitLocal = () => {
  // ...
  props.onSubmit(eventType);
}
```

aferd:
when you get a chance, can you circle back and check this out?

josejulio (approved):

PR #916, opened 2022-01-18
Add 'changeme' value

theute:


.rhcicd/clowdapp-aggregator.yaml
```diff
@@ -80,7 +80,7 @@ parameters:
   required: true
 - name: PROMETHEUS_PUSHGATEWAY
   description: Prometheus Pushgateway URL
-  value: ""
+  value: changeme
```
gwenneg, line 83:
```suggestion
  value: CHANGE_ME
```

theute: Oups, was too fast, will change if that worked

gwenneg: Don't worry about it, it's not that important.

PR #918, opened 2022-01-19
PROMETHEUS_PUSHGATEWAY should be optional

theute:
Trying again as it should be


gwenneg:
Can this be merged safely?

theute:
I think we should try again, I will do a stage + ephemeral deployment and see
This is how it should be. Hopefully the dummy value isn't needed.

PR #923, opened 2022-01-19
NOTIF-429 Expose subtypes

josejulio:
Exposes subtypes in the `Endpoint`. This allows to better filter the camel related subtypes to show in the UI

With this new addition, the `subType` is expected on the `Endpoint` level, for backwards compatibility, we are going to keep it on the Camel for a short time to allow migrations. 

The filter now accepts the subtypes of a specifc type separated by a colon i.e. `camel:splunk`
```
GET /api/integrations/v1.0/endpoints?limit=10&offset=0&type=webhook&type=camel%3Asplunk&type=camel%3Aanycamel
```

It has some backwards compatbility changes. We still have the `sub_type` on `CamelProperties` but it's readonly and backed by the endpoint.sub_type.

Includes work of: https://github.com/RedHatInsights/notifications-backend/pull/892

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
@@ -9,7 +9,6 @@ import com.redhat.cloud.notifications.models.EmailSubscriptionProperties;
 import com.redhat.cloud.notifications.models.Endpoint;
 import com.redhat.cloud.notifications.models.EndpointProperties;
 import com.redhat.cloud.notifications.models.EndpointType;
-import com.redhat.cloud.notifications.models.EventType;
 import com.redhat.cloud.notifications.models.WebhookProperties;
 import io.smallrye.mutiny.Uni;
 import org.hibernate.reactive.mutiny.Mutiny;
@@ -161,39 +160,6 @@ public class EndpointResources {
                 activeOnly,
                 useStatelessSession
         );
-    }
-
-    // Note: This method uses a stateless session
-    public Uni<List<Endpoint>> getTargetEndpoints(String tenant, EventType eventType) {
-        String query = "SELECT DISTINCT e FROM Endpoint e JOIN e.behaviorGroupActions bga JOIN bga.behaviorGroup.behaviors b " +
-                "WHERE e.enabled = TRUE AND b.eventType = :eventType AND bga.behaviorGroup.accountId = :accountId";
-
-        return sessionFactory.withStatelessSession(statelessSession -> {
-            return statelessSession.createQuery(query, Endpoint.class)
-                    .setParameter("eventType", eventType)
-                    .setParameter("accountId", tenant)
-                    .getResultList()
-                    .onItem().call(endpoints -> loadProperties(endpoints, true));
-        });
-    }
-
-    // Note: This method uses a stateless session
-    public Uni<List<Endpoint>> getTargetEndpointsFromType(String tenant, String bundleName, String applicationName, String eventTypeName, EndpointType endpointType) {
-        String query = "SELECT DISTINCT e FROM Endpoint e JOIN e.behaviorGroupActions bga JOIN bga.behaviorGroup.behaviors b " +
-                "WHERE e.enabled = TRUE AND b.eventType.name = :eventTypeName AND bga.behaviorGroup.accountId = :accountId " +
-                "AND b.eventType.application.name = :applicationName AND b.eventType.application.bundle.name = :bundleName " +
-                "AND e.compositeType.type = :endpointType";
-
-        return sessionFactory.withStatelessSession(statelessSession -> {
-            return statelessSession.createQuery(query, Endpoint.class)
-                    .setParameter("applicationName", applicationName)
-                    .setParameter("eventTypeName", eventTypeName)
-                    .setParameter("accountId", tenant)
-                    .setParameter("bundleName", bundleName)
-                    .setParameter("endpointType", endpointType)
-                    .getResultList()
-                    .onItem().call(endpoints -> loadProperties(endpoints, true));
-        });
     }
 
     public Uni<List<Endpoint>> getEndpoints(String tenant, Query limiter) {
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
@@ -4,10 +4,12 @@ import com.redhat.cloud.notifications.db.builder.QueryBuilder;
 import com.redhat.cloud.notifications.db.builder.WhereBuilder;
 import com.redhat.cloud.notifications.db.session.CommonStateSessionFactory;
 import com.redhat.cloud.notifications.models.CamelProperties;
+import com.redhat.cloud.notifications.models.CompositeEndpointType;
 import com.redhat.cloud.notifications.models.EmailSubscriptionProperties;
 import com.redhat.cloud.notifications.models.Endpoint;
 import com.redhat.cloud.notifications.models.EndpointProperties;
 import com.redhat.cloud.notifications.models.EndpointType;
+import com.redhat.cloud.notifications.models.EventType;
 import com.redhat.cloud.notifications.models.WebhookProperties;
 import io.smallrye.mutiny.Uni;
 import org.hibernate.reactive.mutiny.Mutiny;
@@ -15,6 +17,7 @@ import org.jboss.logging.Logger;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
+import javax.ws.rs.BadRequestException;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
@@ -38,6 +41,24 @@ public class EndpointResources {
 
     public Uni<Endpoint> createEndpoint(Endpoint endpoint, boolean useStatelessSession) {
         return commonStateSessionFactory.withSession(useStatelessSession, session -> {
+            // Todo: NOTIF-429 backward compatibility change - Remove soon.
+            if (endpoint.getType() == EndpointType.CAMEL) {
+                CamelProperties properties = endpoint.getProperties(CamelProperties.class);
+
+                if (endpoint.getSubType() == null && properties.getSubType() == null) {
+                    throw new BadRequestException("endpoint.subtype must have a value");
+                }
+
+                if (endpoint.getSubType() != null && properties.getSubType() != null && !properties.getSubType().equals(endpoint.getSubType())) {
+                    throw new BadRequestException("endpoint.subtype must be equal to endpoint.properties.subtype. Consider removing endpoint.properties.subtype as it is deprecated");
+                }
+
+                if (endpoint.getSubType() == null) {
+                    endpoint.setSubType(properties.getSubType());
+                } else if (properties.getSubType() == null) {
+                    properties.setSubType(endpoint.getSubType());
+                }
+            }
             return session.persist(endpoint)
                     .onItem().call(session::flush)
                     .onItem().call(() -> {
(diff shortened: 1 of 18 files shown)


backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -75,7 +75,7 @@
         return commonStateSessionFactory.withSession(useStatelessSession, session -> {
```
josejulio, line 75: I suppose I need to do some changes here to still be backwards compatible

backend/src/test/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessorTest.java
```diff
@@ -115,15 +115,14 @@ void testCamelEndpointProcessing() {
         JsonObject notifMetadata = payload.getJsonObject(NOTIF_METADATA_KEY);
         assertEquals(properties1.getDisableSslVerification().toString(), notifMetadata.getString("trustAll"));
         assertEquals(properties1.getUrl(), notifMetadata.getString("url"));
-        assertEquals(properties1.getSubType(), notifMetadata.getString("type"));
```
josejulio, line 118: todo: still test this one for backwards compatibility

josejulio: todo: needs to be equal to `endpoint1.getSubType()`

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointServiceTest.java
```diff
@@ -342,6 +342,9 @@ void addCamelEndpoint() {
             JsonObject properties = responsePoint.getJsonObject("properties");
             assertNotNull(properties);
             assertTrue(endpoint.getBoolean("enabled"));
+            assertEquals("ansible", endpoint.getString("sub_type"));
+            // Todo: Backwards compatibility test - Remove
+            System.out.println(properties);
```
josejulio, line 347: whoops. I need to remove this one

backend/src/test/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessorTest.java
josejulio: This check was missing, checking that the "type" prop of notifMetadata is endpoint1.getSubType

gwenneg:
Did you have a chance to talk about this with @pilhuhn?

josejulio:
> Did you have a chance to talk about this with @pilhuhn?

yes.

The Idea is to be backwards compatible until others can migrate. 

The change (once we remove the backwards compatibility) is that the APIs will require the `sub_type` at the `endpoint` level (not the `properties`). Currently we are accepting and returning the `sub_type` at both levels (`endpoint` and `properties`). Like wise we are also sending it to the Camel Integration in the same way.

pilhuhn:
Can you also try with the python scripts if adding an integration still works?

josejulio:
> Can you also try with the python scripts if adding an integration still works?

They are still working

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -97,23 +98,34 @@
...
             Uni<List<Endpoint>> endpoints;
             Uni<Long> count;
 
             if (targetType != null && targetType.size() > 0) {
-                Set<EndpointType> endpointType;
+                Set<CompositeEndpointType> compositeType;
                 try {
-                    endpointType = targetType.stream().map(s -> EndpointType.valueOf(s.toUpperCase())).collect(Collectors.toSet());
+                    compositeType = targetType.stream().map(s -> {
+                        String[] pieces = s.split(":", 2);
```
gwenneg, line 116: Before splitting, we could use a regex to validate that the target type is valid and throw a `BadRequestException` otherwise.

common/src/main/java/com/redhat/cloud/notifications/models/CompositeEndpointType.java
```diff
@@ -0,0 +1,71 @@
...
+import javax.persistence.Column;
+import javax.persistence.Convert;
+import javax.persistence.Embeddable;
+import javax.validation.constraints.NotNull;
+import java.util.Objects;
+
+@Embeddable
+public class CompositeEndpointType {
+    @NotNull
+    @Column(name = "endpoint_type")
```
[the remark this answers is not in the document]
josejulio: It is already like that in the DB (is not a new field). We could do a cleanup (I don't think we need the `endpoint_` prefix in here

common/src/main/java/com/redhat/cloud/notifications/models/CompositeEndpointType.java
```diff
@@ -0,0 +1,71 @@
...
+    @Convert(converter = EndpointTypeConverter.class)
+    private EndpointType type;
+
+    @Column(name = "endpoint_sub_type")
```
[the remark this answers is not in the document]
josejulio: Not really, just added the `endpoint_` for consistency. I could name it `sub_type` and we can rename the other either here or in a follow up

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -36,6 +41,24 @@
 
     public Uni<Endpoint> createEndpoint(Endpoint endpoint, boolean useStatelessSession) {
         return commonStateSessionFactory.withSession(useStatelessSession, session -> {
+            // Todo: NOTIF-429 backward compatibility change - Remove soon.
+            if (endpoint.getType() == EndpointType.CAMEL) {
+                CamelProperties properties = endpoint.getProperties(CamelProperties.class);
```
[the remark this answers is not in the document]
josejulio: Maybe, is there a case where we support an endpoint without properties?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
josejulio: It would fail below when doing `EndpointType.valueOf`, do you think we would still need to validate here?

common/src/main/java/com/redhat/cloud/notifications/models/EndpointType.java
```diff
@@ -20,7 +20,14 @@
      */
     @JsonProperty("default")
     @Deprecated
-    DEFAULT, // 2
+    DEFAULT(false), // 2
     @JsonProperty("camel")
-    CAMEL // 3
+    CAMEL(true); // 3
+
+    final boolean requiresSubType;
```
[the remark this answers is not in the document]
josejulio: I'm not sure - right now we don't need to know this information, but if we do, it can be added. The only other place using this would be `Endpoint` and it can already access this value.

engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
```diff
@@ -86,14 +86,14 @@ public void init() {
...
-        metaData.put("type", properties.getSubType());
+        metaData.put("type", endpoint.getSubType());
 
         if (properties.getSecretToken() != null && !properties.getSecretToken().isBlank()) {
             metaData.put(TOKEN_HEADER, properties.getSecretToken());
         }
 
         BasicAuthentication basicAuthentication = properties.getBasicAuthentication();
-        if (basicAuthentication != null) {
+        if (basicAuthentication != null && basicAuthentication.getUsername() != null && basicAuthentication.getPassword() != null) {
```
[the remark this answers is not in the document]
josejulio: That would work

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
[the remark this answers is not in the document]
josejulio: Going to create a separate PR to validate that our endpoints always include properties

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
josejulio: Updated the code so that the returned error is more meaningful to the user. It would return something like:
`Unknown endpoint type: [stuff:splunk]`

engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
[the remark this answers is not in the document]
josejulio: done

josejulio:
Rebased and updated the code

common/src/main/java/com/redhat/cloud/notifications/models/CompositeEndpointType.java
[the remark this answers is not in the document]
gwenneg: I didn't know it was already prefixed like that in the DB. Let's postpone the migration to another time 😃

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -122,33 +148,71 @@
...
+    public Uni<Long> getEndpointsCountPerType(String tenant, Set<EndpointType> type, Boolean activeOnly, boolean useStatelessSession) {
+        return getEndpointsCountPerCompositeType(
+                tenant,
+                type.stream().map(CompositeEndpointType::new).collect(Collectors.toSet()),
+                activeOnly,
+                useStatelessSession
+        );
+    }
+
+    // Note: This method uses a stateless session
```
gwenneg, line 166: All methods that rely only on the stateless session should be in `EndpointRepository` in the `engine`. Can you please move the code?

gwenneg: Or maybe that method already exists there and needs to be updated. I'm checking.

josejulio: I think I added this method back because of the rebase

josejulio: Actually I think I made some updates - will double check.

josejulio: Looks like these changes were already applied to the repository

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -302,13 +331,54 @@
...
+                        WhereBuilder.builder()
+                            .ifElse(
+                                    accountId == null,
+                                    WhereBuilder.builder().and("e.accountId IS NULL"),
+                                    WhereBuilder.builder().and("e.accountId = :accountId", "accountId", accountId)
+                            )
+                            .and(
+                                    WhereBuilder.builder()
+                                            .ifOr(basicTypes.size() > 0, "e.compositeType.type IN (:endpointType)", "endpointType", basicTypes)
+                                            .ifOr(compositeTypes.size() > 0, "e.compositeType IN (:compositeTypes)", "compositeTypes", compositeTypes)
```
gwenneg, line 376: I'm curious, how does that line translate into SQL? Do you have an example?

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -302,13 +331,54 @@
...
+                        endpoint.getProperties(CamelProperties.class).setSubType(endpoint.getSubType());
+                    }
                 }
             })).replaceWith(Uni.createFrom().voidItem());
         }
 
         return Uni.createFrom().voidItem();
     }
 
+    private <T extends EndpointProperties> Uni<List<T>> find(Class<T> typedEndpointClass, Set<UUID> endpointIds, boolean useStatelessSession) {
```
gwenneg, line 345: Where is that new method used?

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -122,33 +147,38 @@
...
-                    .buildCount(session::createQuery)
-                    .getSingleResult();
-        });
+    public Uni<Long> getEndpointsCountPerCompositeType(String tenant, Set<CompositeEndpointType> type, Boolean activeOnly, boolean useStatelessSession) {
+        return commonStateSessionFactory.withSession(useStatelessSession, session -> queryBuilderEndpointsPerType(tenant, type, activeOnly)
+                .buildCount(session::createQuery)
+                .getSingleResult());
+    }
+
+    public Uni<Long> getEndpointsCountPerType(String tenant, Set<EndpointType> type, Boolean activeOnly, boolean useStatelessSession) {
```
gwenneg, line 156: Is it still used?

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -70,24 +90,29 @@
...
+    }
 
-            return queryBuilderEndpointsPerType(accountId, type, activeOnly)
-                    .limit(limit)
-                    .sort(sort)
-                    .build(session::createQuery)
-                    .getResultList()
-                    .onItem().call(this::loadProperties);
-        });
+    public Uni<List<Endpoint>> getEndpointsPerType(String accountId, Set<EndpointType> type, Boolean activeOnly, Query limiter, boolean useStatelessSession) {
```
gwenneg, line 104:
```suggestion
    private Uni<List<Endpoint>> getEndpointsPerType(String accountId, Set<EndpointType> type, Boolean activeOnly, Query limiter, boolean useStatelessSession) {
```

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
josejulio: Aah -- no where, no idea if is more debris from the rebase or where it came from :D

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
josejulio: No, I'll remove some unused methods

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
josejulio: I'll add it as a test to see the actual SQL

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
josejulio: I deleted this method as part of a cleanup

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
josejulio: Removed

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
josejulio: Removed

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
josejulio: You can see the results here: https://github.com/RedHatInsights/notifications-backend/pull/923/files#diff-a75040d7ad093bffe7edab6fc316b111646bdbb1f132ea0f2dc7bddbb5b74fccR22

gwenneg: Thanks but that's the HQL, I was curious about the SQL 😃

josejulio: Ah! I can get it from psql too

gwenneg: Perfect, thank you!

PR #927, opened 2022-01-20
Remove HTTP config keys from aggregator

gwenneg:
Apparently all config keys starting with `quarkus.http` are no longer supported if the Quarkus app does depend on an HTTP-related extension. We received an alert on Sentry from stage because of that.


bennetelli:
I guess we never needed the access log filtering and we probably also never needed the port as the cronjob doesn't get queried from someone else.

bennetelli (approved):
lgtm

PR #934, opened 2022-01-25
NOTIF-450 Move code to the engine module

gwenneg:
This is preparation work for the notifications-backend split. Step 5: the `engine` module now contains the (almost) final runtime code. 

The Docker image built from the `backend` module still contains the entire application. However, the `engine` module is also a standalone Quarkus app that can be started with `./mvnw clean quarkus:dev` and will soon be published as a Docker image.

ℹ️ I created an app-interface MR today to request the creation of the `notfications-engine` repo on quay.io.

ℹ️ The `engine` tests are still located in `backend`. I will move them in the next step.


gwenneg:
Oh I created the PR in the wrong repo 😄 It was meant to be created on my fork.

backend/src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
```diff
@@ -182,19 +183,24 @@
...
-                                                }
-                                            });
+                                            .onItem().transformToMulti(Multi.createFrom()::iterable)
+                                            .onItem().transformToUniAndMerge(application -> {
+                                                ApplicationSettingsValue applicationSettingsValue = new ApplicationSettingsValue();
+                                                applicationSettingsValue.displayName = application.getDisplayName();
+                                                settingsValues.bundles.get(bundle.getName()).applications.put(application.getName(), applicationSettingsValue);
+                                                return Multi.createFrom().iterable(Arrays.asList(EmailSubscriptionType.values()))
+                                                        .onItem().transformToUniAndMerge(emailSubscriptionType -> {
+                                                            // TODO NOTIF-450 How do we deal with a failure here? What kind of response should be sent to the UI when the engine is down?
```
gwenneg, line 193: @josejulio WDYT?

josejulio: mmmmh. We can throw an error for now - in the future this info should live in the database and we won't have this problem.
Other option would be to put this check into common code

gwenneg:
The failed checked is unrelated (JS). @josejulio Do you have an idea on how to fix that?

gwenneg:
@josejulio Could you please review this PR? It was successfully tested by the IQE on ephemeral. Do not merge if you approve it though, I want to run a few manual tests first.

backend/src/main/java/com/redhat/cloud/notifications/routers/UserConfigService.java
gwenneg: Let's wait for the templates stored in the DB.

josejulio:
> The failed checked is unrelated (JS). @josejulio Do you have an idea on how to fix that?

It seems that is scanning the compiled js. I don't think it should do that.

gwenneg:
> It seems that is scanning the compiled js. I don't think it should do that.

Did we change something recently?

gwenneg:
This is not the only PR where this happened BTW so it's not related to the changes here.

josejulio:
> This is not the only PR where this happened BTW so it's not related to the changes in this one. See #935 for another example.

Probably a recent build triggered this issue: https://github.com/RedHatInsights/notifications-backend/commit/c4339d95167035a56d75d6042af965010f4f36fd

josejulio:
It doesn't appear on the original PR: https://github.com/RedHatInsights/notifications-backend/pull/900
I think we need to ignore the build files from the Code scanning

josejulio:
IIUC, this moves all the code to the engine module - but all the clases are still used in notifications-backend?

josejulio (approved):

gwenneg:
> IIUC, this moves all the code to the engine module - but all the clases are still used in notifications-backend?

Yes, `backend` has a `compile` dependency on `engine` so the classes are still used by `backend` after the move. And in the meantime, `engine` is a standalone Quarkus application that can be started without `backend`. My goal is to deploy both the old `notifications-backend` app (before the split) and the new `notifications-engine` pod side by side on prod, and then decide to put route some traffic to the new engine.

PR #935, opened 2022-01-25
Fix deprecated CronJob apiVersion

gwenneg:
Fixes
```
11:47:34 2022-01-25 10:47:34 [    INFO] [           pid-26005]  |stderr| W0125 10:47:34.765681   26005 warnings.go:70] batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
```


gwenneg:
The first PR check run should fail because of an unrelated transitive dependency issue.

gwenneg:
The failed check is unrelated (JS).

PR #943, opened 2022-01-27
NOTIF-450 Add RunOnStartup and LivenessService in engine

gwenneg:
This PR adds in `notifications-engine` the equivalent of `NotificationsApp` and `LivenessService` which already exist in `notifications-backend`.


gwenneg:
I need to check if notifications-backend is working as intended with two classes annotated `@Liveness` in its classpath.

gwenneg:
> I need to check if notifications-backend is working as intended with two classes annotated `@Liveness` in its classpath.

Looks good locally.

engine/src/main/java/com/redhat/cloud/notifications/health/LivenessService.java
```diff
@@ -0,0 +1,44 @@
...
+    @Inject
+    Mutiny.SessionFactory sessionFactory;
+
+    @Inject
+    KafkaConsumedTotalChecker kafkaConsumedTotalChecker;
+
+    @Override
+    public Uni<HealthCheckResponse> call() {
+        HealthCheckResponseBuilder response = HealthCheckResponse.named("Notifications readiness check");
+        if (kafkaConsumedTotalChecker.isDown()) {
```
josejulio, line 26: This assumes that we are always sure that at least 1 engine is processing kafka messages - is that true?

josejulio (approved):
lgtm - minor side comment for the deployment of this PR

engine/src/main/java/com/redhat/cloud/notifications/health/LivenessService.java
gwenneg: Not exactly. This is the exact same thing as [this current backend check](https://github.com/RedHatInsights/notifications-backend/blob/aa7369e12091e053f50dd0146aa94be8710555d7/backend/src/main/java/com/redhat/cloud/notifications/health/LivenessService.java#L34-L38). This class does not introduce anything new but replicates most of the current `backend` behavior into the `engine`.

gwenneg: To be more precise, if any `engine` fails to proceed Kafka messages for a while (delay is configuration-dependent) then it will be marked as unhealthy and restarted.

PR #953, opened 2022-02-01
Make notifications-engine ClowdApp dependency optional

gwenneg:
Since the `engine` hasn't been deployed yet on prod, the ClowdApp dependency has to be optional. Otherwise, we can't deploy the `backend` on prod.


theute:
Makes sense

PR #954, opened 2022-02-01
NOTIF-444 Add info logging for dynamic metrics

bennetelli:


engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
```diff
@@ -106,6 +106,8 @@ public void init() {
                             bopRequest,
                             getPayload(user, action, subject, body)
                     ).onItem().invoke(unused -> {
+                        logger.infof("Adding metric for bundle: [%s] application: [%s]", action.getBundle(), action.getApplication());
```
josejulio, line 109: This is going to spam the logs - I assume is only temporal ? :-D

Could we also expose the p8s endpoints to the internal API?  So that we would be able to see them for any diagnosis we need?

bennetelli: Yes, it's only temporal. Our / My expectation is, that the lines of code are not executed. I tested the same snippet locally, which was working as expected. So the only reason why the dynamic tag metrics are not being exposed is, that this snippet isn't executed.

josejulio (approved):

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
bennetelli: regarding the prometheus endpoints. If we do have access to the pod via http, we can access it by <pod-name...blabla..context-root>>/metrics. I think Gabor knows how to do that. 
I would prefer trying this instead of forwarding the metrics to a new  internal http endpoint. If the pod/http access isn't possible, we can still go for a forwarding way.

Wdyt?

theute: Should be easy to expose /metrics through turnpike

theute: Should be easy to expose /metrics through turnpike.
Gabor has terminal access to get the value (he accesses localhost), but we don't have that access.

bennetelli: So would it be okay if we merge this for now and work on the turnpike thing after merging?

PR #955, opened 2022-02-01
NOTIF-450 Introduce database module

gwenneg:
The `aggregator`, `backend` and `engine` modules all depend on the same DB schema which is why we need to share that schema between modules. `backend` is still the **only** module responsible for triggering the migrations at application startup.

The migration scripts can't be located in `common` because that module depends on `quarkus-hibernate-reactive` and `aggregator` uses the non-reactive Hibernate. As a consequence, we need a new `database` module for the DB migration scripts, at least temporarily. The scripts may be moved to `common` in the future if that module stops depending on reactive stuff.


bennetelli (approved):

backend/pom.xml
```diff
@@ -14,12 +14,17 @@
...
-        <!-- notifications-common module -->
+        <!-- notifications modules -->
         <dependency>
             <groupId>com.redhat.cloud.notifications</groupId>
             <artifactId>notifications-common</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.redhat.cloud.notifications</groupId>
+            <artifactId>notifications-database</artifactId>
```
josejulio, line 25: This dependency ensures that it's executed by the backend?

bennetelli:
lgtm

backend/pom.xml
josejulio: mmh, I see it's on the engine and the aggregator.

gwenneg: The dependency on `notifications-database` does not execute/trigger anything. It's what we do with Flyway that determines where and when the migration will happen.

gwenneg: And that didn't change in this PR:
- notifications-aggregator triggers a Flyway migration at startup in test mode only
- notifications-backend triggers a migration at startup all the time
- notifications-engine does not trigger a migration yet, but will do it in test mode only soon

josejulio (approved):

PR #956, opened 2022-02-02
Refactor EndpointRepository

gwenneg:
A significant part of the `EndpointRepository` code has become irrelevant after the split.


engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -35,83 +38,24 @@
...
+                    .onItem().transformToUni(emailEndpoints -> {
                         /*
-                         * If this line is reached, it means the endpoint properties are null or we don't support
-                         * persisting properties for the endpoint type. We still have to return something.
+                         * The purpose of the following code is to find or create an EMAIL_SUBSCRIPTION endpoint with
+                         * empty properties. This endpoint is used to aggregate and store in the DB the email actions
+                         * outcome, which will be used later by the event log. The recipients of the current email
+                         * action have already been resolved before this step, possibly from multiple endpoints and
+                         * recipients settings. The properties created below have no impact on the resolution of the
+                         * action recipients.
```
gwenneg, line 56: @josejulio Is that clear enough?

josejulio: The description seems good to me, but I'm unsure if the location is the best. 

The name of the method is the same as in the `EndpointResources`, and we use it there to create an EmailSubscription with parameters - it could be confusing to the reader.
So maybe moving the comment to where we actualy call this function or renaming the name of the method (getOrCreateDefaultEmailSubscription?)

gwenneg: I renamed the method as you suggested and moved the comment to the method signature.

josejulio (approved):
lgtm

PR #974, opened 2022-02-08
Remove endpoints subtype migration query

gwenneg:
That query caused a lock on stage twice because of the huge amount of history data.

If possible, it would be safer to remove it and wait for the `notification_history` table to be filled with records containing the subtype, which should take 31 days after the endpoints subtype code deployment.


josejulio (approved):
lgtm

PR #976, opened 2022-02-08
NOTIF-483 Speed up the last connection attempt frontend query

gwenneg:
This PR should hopefully speed up the query that is causing issues on stage with the QA account.

It will have an impact on this part of the code:

https://github.com/RedHatInsights/notifications-backend/blob/dd67d02115c1a661bb174dc262050a8d2eb4cdf3/backend/src/main/java/com/redhat/cloud/notifications/db/NotificationResources.java#L25-L56

See https://www.postgresql.org/docs/13/indexes-ordering.html for more details about the impact of indexes on sorting operations.


josejulio (approved):
lgtm

PR #980, opened 2022-02-10
Expose Git commit hash with /internal/version

gwenneg:
@dgaikwad With this PR, `GET /internal/version` will return the Git commit short hash that can be used to identify the version of the code that is tested.

For example, the endpoint could return `3b2e51c` which would identify https://github.com/RedHatInsights/notifications-backend/commit/3b2e51c. This short hash is also the tag of the Docker image we push on [Quay](https://quay.io/repository/cloudservices/notifications-backend?tab=tags).

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/InternalService.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/InternalService.java
@@ -1,5 +1,6 @@
 package com.redhat.cloud.notifications.routers;
 
+import com.redhat.cloud.notifications.StartupUtils;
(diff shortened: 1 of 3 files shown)


gwenneg:
If this change meets your needs with `notifications-backend`, I will put a similar endpoint in all apps in second phase.

dgaikwad:
It is okay to return at least a partial commit id. Just provide full id if possible.

gwenneg:
The full hash is not directly available right now in the app. This could be changed, but the short hash should be enough for you as it clearly identifies each commit in the sources.

PR #983, opened 2022-02-10
Index the events creation date to speed up ordering

gwenneg:
Similar to #976.


gwenneg:
@josejulio Let's test that on stage then prod ASAP.

PR #987, opened 2022-02-11
NOTIF-491 Merge getEventIds and count SQL queries

gwenneg:
This PR is based on the results from the analysis described in [NOTIF-491](https://issues.redhat.com/browse/NOTIF-491).

Postgre's `EXPLAIN ANALYZE` showed that half of the DB time was spent retrieving the events identifiers and the other half was spent counting the events. This PR merges these two operations into a single one, thanks to Postgre's `COUNT(*) OVER()` [window function](https://www.postgresql.org/docs/13/tutorial-window.html). When used in a limited query, that function returns the total count of records that would have been returned without the limit.

Hopefully, this will have some kind of impact on prod...

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
@@ -2,19 +2,30 @@ package com.redhat.cloud.notifications.db;
 
 import com.redhat.cloud.notifications.models.EndpointType;
 import com.redhat.cloud.notifications.models.Event;
+import com.redhat.cloud.notifications.routers.models.EventLogEntry;
+import com.redhat.cloud.notifications.routers.models.EventLogEntryAction;
+import com.redhat.cloud.notifications.routers.models.Meta;
+import com.redhat.cloud.notifications.routers.models.Page;
+import com.redhat.cloud.notifications.routers.models.PageLinksBuilder;
 import io.smallrye.mutiny.Uni;
 import org.hibernate.reactive.mutiny.Mutiny;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
+import java.math.BigInteger;
 import java.time.LocalDate;
+import java.time.LocalDateTime;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.regex.Matcher;
+import java.util.stream.Collectors;
 
+import static com.redhat.cloud.notifications.routers.EventService.PATH;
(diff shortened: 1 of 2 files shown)


backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
```diff
@@ -23,123 +34,144 @@
...
-                                        Integer limit, Integer offset, Optional<String> orderByCondition) {
-        return sessionFactory.withSession(session -> {
-            String hql = "SELECT e.id FROM Event e JOIN e.eventType et JOIN et.application a JOIN a.bundle b " +
-                    "WHERE e.accountId = :accountId";
+    private Mutiny.Query<Object[]> buildEventIdsAndCountQuery(Mutiny.Session session, String accountId, Set<UUID> bundleIds,
+                                                              Set<UUID> appIds, String eventTypeDisplayName, LocalDate startDate, LocalDate endDate, Set<EndpointType> endpointTypes,
+                                                              Set<Boolean> invocationResults, Integer limit, Integer offset, String sortBy) {
 
-            hql = addHqlConditions(hql, bundleIds, appIds, eventTypeDisplayName, startDate, endDate, endpointTypes, invocationResults);
+        String sql = "SELECT CAST(e.id AS VARCHAR), COUNT(*) OVER() FROM event e, event_type et, applications a, bundles b " +
```
gwenneg, line 117: `CAST(e.id AS VARCHAR)` is a workaround for an Hibernate Reactive issue/bug...

backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
```diff
@@ -158,25 +190,69 @@ private static void setQueryParams(Mutiny.Query<?> query, String accountId, Set<
...
-                return "et.displayName";
+                return nativeQuery ? "et.display_name" : "et.displayName";
             case "created":
                 return "e.created";
             default:
                 throw new IllegalArgumentException("Unknown sort field: " + field);
         }
     }
+
+    private static List<UUID> parseEventIds(List<Object[]> records) {
```
josejulio, line 215: Probably worth adding a note that records[0] has the stuff we need and records[1] has the same count.
Maybe we could even use the stream API to only pass the events and/or to create the result?

gwenneg: Good point, I added a comment.

Since we're getting rid of the reactiveness soon, I'd rather keep the number of streams as low as possible 😃 They come with many drawbacks (performances, stacktrace...).

backend/src/main/java/com/redhat/cloud/notifications/db/EventResources.java
gwenneg: I don't have a link and that workaround came from my own attempts to make the query work. Hopefully Hibernate ORM will work better and this will go away soon.

PR #993, opened 2022-02-14
NOTIF-491 Return events actions on demand

gwenneg:
This is a new attempt at improving the response time of the `GET /events` endpoint.

Last week, we deployed on stage a version with modified SQL queries (#987) but the results were really bad so that PR was reverted.

Log entries showed that we spend a significant amount of time outside of Postgres, so I am now focusing on Jackson's serialization time. Since we return the events and their actions in a single response, the important data volume could cause a slow serialization.

Depending on the results of this PR, we may split the `/events` endpoint into two endpoints:
- one would still return the events **_without_** the actions
- the other one would take an events identifiers list and return the corresponding actions


gwenneg:
The PR check will fail because of an unrelated issue:
```
MountVolume.SetUp failed for volume "keystore" : secret "nonprod-insightsnotifications" not found
```
I'm merging anyway as this does not affect prod and needs to be tested on stage.

gwenneg:
Response time of the Drift dashboard request on stage with the QA account before the merge: [813, 1092] ms (10 attempts, average around 900ms).

gwenneg:
Response time was slightly improved after the merge, but not nothing worth separating events from actions for good.

PR #995, opened 2022-02-14
NOTIF-491 Fix /events LazyInitializationException

gwenneg:


gwenneg:
PR check will fail because of an unrelated cause.

PR #997, opened 2022-02-14
NOTIF-491 Remove duplicate DB health check

gwenneg:
`SELECT * FROM pg_stat_activity` shows that 6 `SELECT 1` queries are executed each second on stage. This puts a lot of pressure on the datasource connection pool for nothing.

Our query isn't needed because Quarkus already runs it. Removing it will decrease the number from 6 to 3 on stage.

~The 6 queries on stage come from 3 different IP addresses. We could try to understand why 3 machines are calling our `/health` each second, but that will be for another time.~ That's actually one query per replica which is totally normal 😄


gwenneg:
This certainly contributed to the many health check failures that happened in the past, when the app was unable to acquire a connetion from the pool.

gwenneg:
Test fixed.

PR #1000, opened 2022-02-15
NOTIF-381 Speedup user fetching

bennetelli:


engine/src/test/java/com/redhat/cloud/notifications/processors/email/EmailTest.java
```diff
@@ -54,23 +55,22 @@
 @QuarkusTest
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 @QuarkusTestResource(TestLifecycleManager.class)
-public class EmailTest {
+public class EmailTest extends DbIsolatedTest {
```
bennetelli, line 58: how do we deal with those tests in engine?

gwenneg: `DbIsolatedTest` is gone in `engine`. You can simply remove it from the test.

bennetelli: done

bennetelli:
@gwenneg maybe you have an idea how to solve this issue.
in the application.properties I added this placeholder for example: 
com.redhat.cloud.notifications.recipients.itservice.ITUserService/mp-rest/url= ${IT_USER_SERVICE_URL}

this is a duplicate. we need to specify it this way because of this issue: https://github.com/quarkusio/quarkus/issues/8384

Now my question:
Right now the EmailTest is failing wth this message:
"SRCFG00011: Could not expand value IT_USER_SERVICE_URL in property com.redhat.cloud.notifications.recipients.itservice.ITUserService/mp-rest/url"

I tried to set the env variable with System.setProperty() in the EmailTest, but it didn't work. Do you have an idea how to fix it?

bennetelli:
@gburges I would like to run this one on ephemeral. Can we manage to do this together? 

The uncommented entries in application.properties could be needed, so I would like to test it in an OpenShift environment.

.rhcicd/clowdapp-backend.yaml
```diff
@@ -198,7 +198,7 @@ parameters:
 - name: NOTIFICATIONS_LOG_LEVEL
   description: Log level for com.redhat.cloud.notifications
   value: INFO
-- name: RBAC_AUTHENTICATION_READ_TIMEOUT
+- name: RBAC_AUTHENTICATION_READ_TIMEOUTIT_USER_SERVICE_REST_URL
```
theute, line 201: Just bumped on this accidentally, that sounds fishy ?

bennetelli: indeed. fixed. thanks!

josejulio:
Congrats on the PR #1000! 

and this one needs rebasing :D

bennetelli:
> Congrats on the PR #1000!
> 
> and this one needs rebasing :D

Yes... I started with it yesterday, but this is not a no brainer so I decided to do it earlier in the morning.

engine/src/main/java/com/redhat/cloud/notifications/recipients/itservice/ITUserServiceWrapper.java
```diff
@@ -0,0 +1,21 @@
...
+import com.redhat.cloud.notifications.recipients.itservice.pojo.request.ITUserRequest;
+import com.redhat.cloud.notifications.recipients.itservice.pojo.response.ITUserResponse;
+import org.eclipse.microprofile.rest.client.inject.RestClient;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+import java.util.List;
+
+@ApplicationScoped
+public class ITUserServiceWrapper {
```
gwenneg, line 12: I already asked that in the previous PR: why do we need a wrapper?

engine/src/main/java/com/redhat/cloud/notifications/recipients/itservice/ITUserService.java
```diff
@@ -0,0 +1,20 @@
...
+import javax.ws.rs.core.MediaType;
+import java.util.List;
+
+@RegisterRestClient(configKey = "it-s2s")
+public interface ITUserService {
+
+    @POST
+    @Path("/findUsers")
+    @Consumes(MediaType.APPLICATION_JSON)
+    List<ITUserResponse> getUsers(ITUserRequest itUserRequest);
```
gwenneg, line 19: `@Produces(MediaType.APPLICATION_JSON)` is missing here.

.rhcicd/clowdapp-engine.yaml
```diff
@@ -102,6 +102,23 @@ objects:
...
+        - name: QUARKUS_HTTP_SSL_CERTIFICATE_KEY_STORE_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              name: it-services
+              key: it-services-keystorepassword
+        - name: IT_S2S_MP_REST_URL
+          valueFrom:
+            secretKeyRef:
+              name: it-services
+              key: url
```
gwenneg, line 121: Does the secret exist on both stage and prod?

engine/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/ITUserRequest.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.LinkedList;
+import java.util.List;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class ITUserRequest {
+
+    public ITUserRequestBy by;
+    public Include include;
+
+    public ITUserRequest(String accountId, boolean adminsOnly) {
```
gwenneg, line 14: As I commented in the previous PR, many things in this constructor should be `static`. We shouldn't rebuild all objects for each request.

engine/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/response/ITUserResponse.java
```diff
@@ -0,0 +1,11 @@
+package com.redhat.cloud.notifications.recipients.itservice.pojo.response;
+
+import java.util.List;
+
+public class ITUserResponse {
```
gwenneg, line 5: It would be better if all classes from `com.redhat.cloud.notifications.recipients.itservice.pojo.response` were annotated with `@JsonIgnoreProperties(ignoreUnknown = true)`. On the other hand, that annotation isn't really useful in the `.request` package.

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -130,4 +129,29 @@ public void initCounters() {
         } while (rbacUsers.getData().size() == rbacElementsPerPage);
         return users;
     }
+
+    private List<User> transformToUser(List<ITUserResponse> itUserResponses, boolean adminsOnly) {
+        List<User> users = new LinkedList<>();
```
gwenneg, line 134: Why `LinkedList`?

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -130,4 +129,29 @@ public void initCounters() {
...
+        for (ITUserResponse itUserResponse : itUserResponses) {
```
gwenneg, line 135: Will all fields used in the following code always be populated? There are several potential NPE otherwise.

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -69,18 +79,7 @@ public void initCounters() {
...
-                Timer.Sample getUsersPageTimer = Timer.start(meterRegistry);
-                Page<RbacUser> rbacUsers = retryOnError(() ->
-                        rbacServiceToService.getUsers(accountId, adminsOnly, page * rbacElementsPerPage, rbacElementsPerPage));
-                getUsersPageTimer.stop(meterRegistry.timer("rbac.get-users.page", "accountId", accountId));
-                return rbacUsers;
-            }
-        );
-        getUsersTotalTimer.stop(meterRegistry.timer("rbac.get-users.total", "accountId", accountId, "users", String.valueOf(users.size())));
-        return users;
+        return transformToUser(itUserService.getUsers(accountId, adminsOnly), adminsOnly);
```
gwenneg, line 82: Wouldn't it be safer to start with the two implementations (RBAC and IT) side by side and switch from one to the other with a boolean from the configuration, instead of removing the old way already?

gwenneg (commented):
Runtime only review. I haven't checked the tests (yet).

.rhcicd/clowdapp-engine.yaml
bennetelli: yes, there is one thing to change. but I am also not planning to put it on prod today.
the keys are there, one value is probably not correct. but that's something we can change in vault and not here.

engine/src/main/java/com/redhat/cloud/notifications/recipients/itservice/ITUserServiceWrapper.java
bennetelli: I didn't manage to get it up and running with two RestClients in RbacRecipientUsersProvider. In the mid term we will remove the other rest client and then we can also remove the wrapper.

engine/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/response/ITUserResponse.java
bennetelli: i remember I did that already. It probably went away during merge.

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
bennetelli: Could also use ArrayList or something else. With that amount of entries it probably doesn't matter. Will add ArrayList

engine/src/main/java/com/redhat/cloud/notifications/recipients/itservice/ITUserService.java
bennetelli: added in #1045

engine/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/response/ITUserResponse.java
bennetelli: added in #1044

PR #1001, opened 2022-02-15
NOTIF-491 Change events purge and vacuum strategy

gwenneg:
The following results from the investigation conducted with `gabi` on stage.

`EXPLAIN ANALYZE` on the `COUNT(*)` query from the `/events` endpoint showed that Postgres is sometimes unable to perform [index-only scans](https://www.postgresql.org/docs/13/indexes-index-only-scans.html) during that query because it has to fetch records that have been deleted by the DB cleaning cron job.

```
EXPLAIN ANALYZE SELECT COUNT(*) FROM event e...
[...]
        ->  Index Only Scan using ix_event_service_index_only_scan_desc on event e  (cost=0.42..8138.76 rows=31651 width=16) (actual time=16.074..26.006 rows=3446 loops=11)
              Index Cond: ((account_id = 'HIDDEN'::text) AND (created >= '2022-02-03 00:00:00'::timestamp without time zone) AND (created <= '2022-02-10 00:00:00'::timestamp without time zone) AND (event_type_id = et.id))
              Heap Fetches: 51"
[...]
```

Queries on the `pg_locks` table also showed that Postgres locks the `event` table with an `AccessShareLock` when it has to fetch records during a `COUNT(*)`.

The idea of this PR is to synchronize the `event` purge with the `VACUUM` operation so that Postgres will never have to fetch any `event` record while running `COUNT(*)`. That way, Postgres should always use index-only scans and execute the query at maximum speed.

### DIFF
--- a/.rhcicd/clowdapp-backend.yaml
+++ b/.rhcicd/clowdapp-backend.yaml
@@ -94,10 +94,14 @@ objects:
     name: notifications-db-cleaner-config
   data:
     clean.sh: |
-      cat /notifications-db-cleaner/clean.sql | psql > /dev/null
+      cat /notifications-db-cleaner/clean.sql | psql
     clean.sql: |
+      \timing
       CALL cleanEventLog();
+      VACUUM event;
+      VACUUM notification_history;
       CALL cleanKafkaMessagesIds();
+      VACUUM kafka_message;
 - apiVersion: batch/v1
   kind: CronJob
   metadata:
@@ -160,7 +164,7 @@ parameters:
   value: 500m
 - name: DB_CLEANER_SCHEDULE
   description: Execution time specified in cron format
-  value: "*/10 * * * *"
+  value: "0 1 * * *"
 - name: DB_NAME
   description: Database name used by the notifications-db-cleaner CronJob
   value: notifications_backend
--- a/database/src/main/resources/db/migration/V1.42.0__NOTIF-491_initial_manual_vacuum.sql
+++ b/database/src/main/resources/db/migration/V1.42.0__NOTIF-491_initial_manual_vacuum.sql
@@ -1,0 +1,6 @@
+-- This script should allow Postgres to run index-only scans on the following tables immediately, without waiting for
+-- the next DB cleaning cron job execution.
+
+VACUUM ANALYZE event;
+VACUUM ANALYZE notification_history;
+VACUUM ANALYZE kafka_message;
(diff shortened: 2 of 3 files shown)


.rhcicd/clowdapp-backend.yaml
```diff
@@ -160,7 +160,7 @@ parameters:
   value: 500m
 - name: DB_CLEANER_SCHEDULE
   description: Execution time specified in cron format
-  value: "*/10 * * * *"
+  value: "0 1 * * *"
```
gwenneg, line 163: The DB cleaning cron job will now start at 1:00 AM (one hour after the aggregation process) each day.

.rhcicd/clowdapp-backend.yaml
```diff
@@ -97,7 +97,10 @@ objects:
       cat /notifications-db-cleaner/clean.sql | psql > /dev/null
     clean.sql: |
       CALL cleanEventLog();
+      VACUUM event;
```
gwenneg, line 100: VACUUM cannot be run from a Postgres procedure, so we have to do it here.

database/src/main/resources/db/migration/V1.42.0__NOTIF-491_initial_manual_vacuum.sql
```diff
@@ -0,0 +1,6 @@
+-- This script should allow Postgres to run index-only scans on the following tables immediately, without waiting for
```
gwenneg, line 1: We need two migration scripts because Flyway does not allow transactional (`ALTER TABLE`) and non-transactional (`VACUUM`) statements in the same file.

.rhcicd/clowdapp-backend.yaml
```diff
@@ -94,10 +94,14 @@ objects:
     name: notifications-db-cleaner-config
   data:
     clean.sh: |
-      cat /notifications-db-cleaner/clean.sql | psql > /dev/null
+      cat /notifications-db-cleaner/clean.sql | psql
     clean.sql: |
```
gwenneg, line 98: Here's the new output of this script:
```
Timing is on.
INFO:  2022-02-15 15:14:02.628547+00 Event log purge starting. Entries older than 31 days will be deleted.
INFO:  2022-02-15 15:14:02.628547+00 Event log purge ended. 0 entries were deleted from the database.
CALL
Time: 2.375 ms
VACUUM
Time: 11.020 ms
VACUUM
Time: 0.756 ms
INFO:  2022-02-15 15:14:02.642761+00 Kafka messages purge starting. Entries older than 1 day will be deleted.
INFO:  2022-02-15 15:14:02.642761+00 Kafka messages purge ended. 0 entries were deleted from the database.
CALL
Time: 0.678 ms
VACUUM
Time: 0.508 ms
```

.rhcicd/clowdapp-backend.yaml
gwenneg: Successfully tested on ephemeral.

josejulio (approved):

PR #1002, opened 2022-02-16
RHICOMPL-2722 consume action.events instead of action.context

skateman:
I'm not entirely sure if something like this would work, but I was told that I should pass data through the payload as it's semantically more correct than using the context for everything. For these specific templates we would always have just a single event, so the question is if the template would be consumed this way or not.

### DIFF
--- a/engine/src/main/resources/templates/Compliance/complianceBelowThresholdEmailBody.html
+++ b/engine/src/main/resources/templates/Compliance/complianceBelowThresholdEmailBody.html
@@ -35,10 +35,10 @@
                         <!-- Hi {{user.first_name}},
                         <br>
                         <br> -->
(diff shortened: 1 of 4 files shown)


josejulio (approved):
It's correct, but some comments/details.

The `context` payload is meant to have information global to this event - usually we put here information on what triggered the notification. In the events.payload we put information about the particular event.

For example, in the [policies application template](https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/main/resources/templates/Policies/instantEmailBody.html) we put the `host_id`, `display_name`, date in the `context` and each `policy_id` and `policy_name` in the payload.

skateman:
@josejulio thanks for your review. I need to discuss the `context` part with @vkrizan first, he already had a [similar comment](https://github.com/RedHatInsights/compliance-backend/pull/1169#discussion_r807823578) on the other side.

josejulio:
Merging as per our conversation

PR #1003, opened 2022-02-17
NOTIF-491 Denormalize events

gwenneg:
This PR adds a new (and disabled by default) way to retrieve the events data from the DB. The `event` table now contains denormalized data that can be used to answer to `GET /events` requests.

If `notifications.event-service.use-denormalized-events` is `true`, then the denormalized data will be used. Otherwise, the old normalized data will be used.

A data migration is required for the events that already exist in the DB. I expect it to be quite long which is why I didn't put that migration in a Flyway script. It will be run from an internal REST request instead.

Both the PR and the data migration can be easily reverted in case something goes wrong or if the performance impact isn't good enough.

@josejulio Let's try this on stage 😃

### DIFF
--- a/backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest.java
+++ b/backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest.java
@@ -539,7 +539,12 @@ public class EventServiceTest extends DbIsolatedTest {
     private Uni<Event> createEvent(String accountId, EventType eventType, LocalDateTime created) {
         Event event = new Event();
         event.setAccountId(accountId);
+        event.setBundleId(eventType.getApplication().getBundle().getId());
+        event.setBundleDisplayName(eventType.getApplication().getBundle().getDisplayName());
+        event.setApplicationId(eventType.getApplication().getId());
+        event.setApplicationDisplayName(eventType.getApplication().getDisplayName());
         event.setEventType(eventType);
+        event.setEventTypeDisplayName(eventType.getDisplayName());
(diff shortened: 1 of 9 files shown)


backend/src/test/java/com/redhat/cloud/notifications/routers/EventServiceTest.java
```diff
@@ -469,7 +470,16 @@ void testAllQueryParams() {
...
-                    return getEventLogPage(defaultIdentityHeader, null, null, null, null, null, null, null, null, null, null, false);
+                    getEventLogPage(defaultIdentityHeader, null, null, null, null, null, null, null, null, null, null, false);
+                    System.setProperty("notifications.event-service.legacy-mode", "true");
+                    given()
+                            .basePath(API_INTERNAL)
+                            .when()
+                            .get("/denormalizedEvents/migrate")
+                            .then()
+                            .statusCode(200);
+                    System.setProperty("notifications.event-service.use-denormalized-events", "true");
```
josejulio, line 481: Is this really changing the mode to use the normalized events? Are the queries indeed different in this test?

gwenneg: Yes, all queries are different with the boolean set to `true`.

gwenneg: FYI I ran all the tests locally with `notifications.event-service.use-denormalized-events=true` and made sure they passed.

PR #1009, opened 2022-02-23
NOTIF-488 Remove the reactiveness from backend

gwenneg:
:construction::construction::construction: WIP, do not review yet :construction::construction::construction:


backend/src/main/java/com/redhat/cloud/notifications/db/Query.java
```diff
@@ -1,9 +1,16 @@
...
+import javax.ws.rs.BadRequestException;
 import javax.ws.rs.QueryParam;
 import java.util.function.Function;
+import java.util.regex.Pattern;
+
+import static java.util.regex.Pattern.CASE_INSENSITIVE;
 
 public class Query {
+
+    private static final Pattern SORT_FIELD_PATTERN = Pattern.compile("^[a-z0-9_-]+$", CASE_INSENSITIVE);
```
gwenneg, line 12: @josejulio The backend will no longer allow sorting on fields that do not match this pattern. Please confirm that this won't break anything in the frontend.

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -33,193 +33,175 @@
     private static final Logger LOGGER = Logger.getLogger(EndpointResources.class);
 
     @Inject
-    Mutiny.SessionFactory sessionFactory;
+    EntityManager entityManager;
 
-    @Inject
-    CommonStateSessionFactory commonStateSessionFactory;
```
gwenneg, line 39: `EndpointResources` is no longer using `CommonStateSessionFactory` in this PR because it made the migration easier.

I wonder if this class still makes sense though because `backend` only uses stateful sessions and `engine` only uses stateless sessions. We can migrate it to Hibernate ORM later alongside with the `engine` module.

backend/src/main/java/com/redhat/cloud/notifications/ReactiveSessionFactoryWorkaround.java
```diff
@@ -0,0 +1,16 @@
+package com.redhat.cloud.notifications;
+
+import org.hibernate.reactive.mutiny.Mutiny;
+
+import javax.enterprise.inject.Produces;
+import javax.inject.Singleton;
+
+// TODO NOTIF-488 Remove this class when notifications-common no longer depends on Hibernate Reactive.
+public class ReactiveSessionFactoryWorkaround {
```
gwenneg, line 9: FYI Quarkus won't start `backend` unless this is present.

backend/src/main/java/com/redhat/cloud/notifications/db/Query.java
josejulio: IIRC there is at least one place where we use `something.created`. I'll try to search what endpoint is calling.

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
josejulio: It makes sense to remove it.

backend/src/main/java/com/redhat/cloud/notifications/routers/AdminService.java
```diff
@@ -31,15 +30,14 @@
...
 
-        Uni<RbacRaw> rbacRawUni = rbacServer.getRbacInfo("notifications,integrations", rhid);
-
-        return rbacRawUni
-                .onItem()
-                .transform(r -> Response.ok(r.data).build())
-                .onFailure()
-                .call(t -> Uni.createFrom().item(Response.serverError().entity("Rbac call failed -- see logs").build()));
+        try {
+            RbacRaw rbacRaw = rbacServer.getRbacInfo("notifications,integrations", rhid).await().indefinitely();
```
gwenneg, line 36: `RbacServer` still has a reactive signature because it looks like `quarkus-security` can only be used with reactive code. As a consequence, the entire `com.redhat.cloud.notifications.auth` package is still reactive. Maybe I'll find a way to get rid of the reactiveness there later. It doesn't harm to keep it for now.

backend/src/main/java/com/redhat/cloud/notifications/db/Query.java
josejulio: `/api/integrations/v1.0/endpoints/{id}/history` uses: `sortBy: 'nh.created:desc'`

backend/src/test/java/com/redhat/cloud/notifications/db/DbCleaner.java
```diff
@@ -59,42 +58,32 @@
     /**
      * Deletes all records from all database tables (except for flyway_schema_history) and restores the default records.
      * This method should be called from a method annotated with <b>both</b> {@link BeforeEach} and {@link AfterEach} in
-     * all test classes that involve SQL queries to guarantee the tests isolation in terms of stored data. Ideally, we
-     * should do that with {@link io.quarkus.test.TestTransaction} but it doesn't work with Hibernate Reactive, so this
```
gwenneg, line 63: Unfortunately, `@TestTransaction` does not work with RestAssured requests, so we'll keep `DbCleaner` permanently.

backend/src/main/java/com/redhat/cloud/notifications/db/Query.java
gwenneg: Good catch! I added `.` to the allowed characters in the regex.

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -49,345 +49,328 @@
...
-                        }
-
-                        return sessionQuery.executeUpdate();
-                    })
-                    .collect().asList()
-                    .replaceWith(OK)
-                    // The following exception will be thrown if the behavior group is not found with the first query.
-                    .onFailure(NoResultException.class).recoverWithItem(NOT_FOUND)
-                    // The following exception will be thrown if the behavior group contains more than one EMAIL_SUBSCRIPTION action.
-                    .onFailure(IllegalArgumentException.class).recoverWithItem(BAD_REQUEST);
```
josejulio, line 359: Is this still true? (i.e. Bad request if group contains more than one EMAIL_SUBSCRIPTION

backend/src/main/java/com/redhat/cloud/notifications/routers/AdminService.java
josejulio: How `indefinitely` is the `await().indefinitely()` ? we won't have requests waiting forever? Maybe we can set a reasonable time here.

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
gwenneg: I asked myself the same question and found the answer in the git history. The `IllegalArgumentException` was thrown in an older version but then we removed it. That `.onFailure(IllegalArgumentException.class)` is actually useless in the current code.

backend/src/main/java/com/redhat/cloud/notifications/routers/AdminService.java
gwenneg: Good question. I'm not sure if `rbac-authentication/mp-rest/connectTimeout` and `rbac-authentication/mp-rest/readTimeout` would protect us against all possible issues or not. I'll add a timeout here to be safe.

gwenneg: Done.

backend/src/test/java/com/redhat/cloud/notifications/db/DbCleaner.java
josejulio: At least we know that is permanent :D

backend/src/main/java/com/redhat/cloud/notifications/routers/AdminService.java
```diff
@@ -31,15 +31,15 @@
...
-        Uni<RbacRaw> rbacRawUni = rbacServer.getRbacInfo("notifications,integrations", rhid);
-
-        return rbacRawUni
-                .onItem()
-                .transform(r -> Response.ok(r.data).build())
-                .onFailure()
-                .call(t -> Uni.createFrom().item(Response.serverError().entity("Rbac call failed -- see logs").build()));
+        try {
+            RbacRaw rbacRaw = rbacServer.getRbacInfo("notifications,integrations", rhid)
+                    .await().atMost(Duration.ofSeconds(2L));
```
josejulio, line 38: I think we have seen requests taking more than 2s? or 2s was the max allowed?

josejulio (approved):
LGTM, did a smoke test with the UI and everything seemed fine.

backend/src/main/java/com/redhat/cloud/notifications/routers/AdminService.java
gwenneg: We currently time out after 2 seconds and the request is canceled. Also, this is only the `AdminService` which is rarely used.

PR #1010, opened 2022-02-23
NOTIF-491 Remove date conversion function in SQL queries

gwenneg:
I just realized we could get rid of the `DATE()` function is `/events` SQL queries. That function is slowing things down so it's better if we remove it.

### DIFF
--- a/common/src/main/java/com/redhat/cloud/notifications/models/Event.java
+++ b/common/src/main/java/com/redhat/cloud/notifications/models/Event.java
@@ -8,23 +8,29 @@ import javax.persistence.Id;
 import javax.persistence.JoinColumn;
 import javax.persistence.ManyToOne;
 import javax.persistence.OneToMany;
+import javax.persistence.PrePersist;
 import javax.persistence.Table;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/redhat/cloud/notifications/models/Event.java
```diff
@@ -8,24 +8,30 @@
...
 @Entity
 @Table(name = "event")
-public class Event extends CreationTimestamped {
+public class Event {
 
     @Id
     @GeneratedValue
     private UUID id;
 
+    private Timestamp created;
```
josejulio, line 33: how this worked before?

josejulio (approved):

common/src/main/java/com/redhat/cloud/notifications/models/Event.java
gwenneg: The `LocalDateTime created` field from `CreationTimestamped` was used but it required a date transformation with Postgres' `DATE()` function.

PR #1011, opened 2022-02-23
NOTIF-487 Using pagination for event types

josejulio:
Wraps event_types response in a page object, to be able to know the total number of elements.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/Query.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/Query.java
@@ -21,7 +21,6 @@ public class Query {
     private Integer pageNumber;
 
     @QueryParam("offset")
-    @DefaultValue("0")
     private Integer offset;
 
     @QueryParam("sort_by")


josejulio:
@gwenneg could you review? It's still a draft because it needs to go together with https://github.com/RedHatInsights/notifications-frontend/pull/240

backend/src/main/java/com/redhat/cloud/notifications/db/Query.java
```diff
@@ -10,6 +10,7 @@
 public class Query {
 
     private static final Pattern SORT_FIELD_PATTERN = Pattern.compile("^[a-z0-9._-]+$", CASE_INSENSITIVE);
+    private static final int DEFAULT_RESULTS_PER_PAGE  = 20;
```
gwenneg, line 13: This should probably be mentioned in the OpenAPI doc for all affected endpoints.

josejulio: Added the `@DefaultParam` annotation, that updates the openapi.json

gwenneg (approved):
LGTM

PR #1021, opened 2022-02-28
NOTIF-488 Remove ModelInstancesHolder from backend

gwenneg:
`ModelInstancesHolder` was initially introduced to flatten the reactive tests and was not removed in #1009 because it would have made the review even harder. It's time to say goodbye now!


josejulio (approved):
goodbye!

PR #1023, opened 2022-03-03
First cut at OpenBridge integration

pilhuhn:
This sets up Slack sending via OB. This is disabled by default. See  `ob.enabled=false` in application.properties (which is only to make it explicit, in code `false` is the default.

Engine code is 'pseudo' reactive (it works). There is a temp package with some OB interfaces, that should go away once engine is no longer reactive and then use the code from commons like ui-backend does.


common/src/main/java/com/redhat/cloud/notifications/openbridge/BridgeHelper.java
```diff
@@ -0,0 +1,107 @@
...
+    @Produces
+    public Bridge getBridgeIfNeeded() {
+
+        if (!obEnabled) {
+            return new Bridge("- OB not enabled -", "http://does.not.exist");
+        }
+
+        if (BridgeHolder.getInstance().getBridge() != null) {
+            return BridgeHolder.getInstance().getBridge();
+        }
```
gwenneg, line 51: Why is `BridgeHolder` needed? Can't we store the `Bridge` instance as a non-static field in this class after `@ApplicationScoped` has been added?

common/src/main/java/com/redhat/cloud/notifications/openbridge/BridgeHelper.java
```diff
@@ -0,0 +1,107 @@
...
+
+        String token = getAuthTokenInternal();
+
+        Map<String, Object> bridges = apiService.getBridges(token);
+        // List of bridges is in 'items'
+        List<Map<String, String>> bridgesList = (List<Map<String, String>>) bridges.get("items");
+
+        for (Map<String, String> b : bridgesList) {
+            if (b.get("name").equals(ourBridge)) {
```
gwenneg, line 60: I assume `b.get("name")` will never be `null`. The test could be reverted otherwise:
```suggestion
            if (ourBridge.equals(b.get("name"))) {
```

common/src/main/java/com/redhat/cloud/notifications/openbridge/BridgeHelper.java
```diff
@@ -0,0 +1,107 @@
...
+                String ep = b.get("endpoint");
+
+                Bridge bridge = new Bridge(bid, ep);
+                BridgeHolder.getInstance().setBridge(bridge);
+
+                return bridge;
+            }
+        }
+        // No luck
+        throw new IllegalStateException("Bridge not found");
```
gwenneg, line 71: What is the impact of this throw on the injection point?

engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
```diff
@@ -120,8 +136,13 @@ public NotificationHistory callCamel(Notification item, UUID historyId, JsonObje
         String accountId = item.getEndpoint().getAccountId();
         // the next could give a CCE, but we only come here when it is a camel endpoint anyway
         String subType = item.getEndpoint().getSubType();
+        CamelProperties camelProperties = (CamelProperties) item.getEndpoint().getProperties();
```
gwenneg, line 139:
```suggestion
        CamelProperties camelProperties = item.getEndpoint().getProperties(CamelProperties.class);
```

engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
```diff
@@ -120,8 +136,13 @@ public NotificationHistory callCamel(Notification item, UUID historyId, JsonObje
...
 
-        reallyCallCamel(payload, historyId, accountId, subType);
+        if (camelProperties.getSubType().equals("slack")) { // OpenBridge
```
gwenneg, line 141: `camelProperties.getSubType()` is deprecated. Should this and all other occurrences in this PR be changed?

engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
```diff
@@ -150,4 +171,34 @@ public boolean reallyCallCamel(JsonObject body, UUID historyId, String accountId
...
+
+        ce.put("id", id);
+        ce.put("source", "notifications"); // Source of original event?
+        ce.put("specversion", "1.0");
+        ce.put("type", "myType"); // Type of original event?
+        ce.put("rhaccount", accountId);
+        ce.put(PROCESSORNAME, extras.get(PROCESSORNAME));
+        // TODO add dataschema
+
+        LOGGER.infof("Sending Event with id(%s) for processor with name %s and id=%s\n",
```
gwenneg, line 188: Can we get rid of the new line char?
```suggestion
        LOGGER.infof("Sending Event with id(%s) for processor with name %s and id=%s",
```

engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
```diff
@@ -150,4 +171,34 @@ public boolean reallyCallCamel(JsonObject body, UUID historyId, String accountId
...
+        body.remove(NOTIF_METADATA_KEY); // Not needed on OB
+        ce.put("data", body.getMap());
+
+        BridgeEventService evtSvc = RestClientBuilder.newBuilder()
+                .baseUri(URI.create(bridge.getEventsEndpoint()))
+                .build(BridgeEventService.class);
+
+        evtSvc.sendEvent(ce, bridgeAuth.getToken());
+
+        return true;
```
gwenneg, line 200: The returned value is never used. Do we really need it?

common/src/main/java/com/redhat/cloud/notifications/openbridge/BridgeApiService.java
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Talk to the OpenBridge manager API to set up processors etc
+ */
+@Path("/api/v1/bridges")
+@Produces(MediaType.APPLICATION_JSON)
+@RegisterRestClient(configKey = "ob")
+public interface BridgeApiService {
+
+    @GET
+    @Path("/")
```
gwenneg, line 25: Does the remote server expect a trailing slash?

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -31,24 +37,33 @@
...
+
+    @Inject
+    @RestClient
+    BridgeApiService bridgeApiService;
+
+    @Inject
+    Bridge bridge;
+
+    @Inject
+    BridgeAuth bridgeAuth;
```
gwenneg, line 58: `EndpointResources` is supposed to be a repository for endpoints and only deal with DB queries. Can all the new code based on `BridgeApiService` be moved one layer above?

common/src/main/java/com/redhat/cloud/notifications/openbridge/BridgeEventService.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.Map;
+
+/**
+ * Deliver events to the OpenBridge ingress
+ */
+@Path("/events")
+public interface BridgeEventService {
+
+    @POST
+    @Path("/")
```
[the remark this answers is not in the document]
pilhuhn: It doesn't but interestingly enough it doesn't complain either

common/src/main/java/com/redhat/cloud/notifications/openbridge/BridgeHelper.java
pilhuhn: If this is null, we'll have bigger issues :-)
No friend of yoda-speak I am

pilhuhn:
Failing test is an issue in the test itself

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -35,20 +34,14 @@
...
         if (endpoint.getType() == EndpointType.CAMEL) {
             CamelProperties properties = endpoint.getProperties(CamelProperties.class);
 
-            if (endpoint.getSubType() == null && properties.getSubType() == null) {
-                throw new BadRequestException("endpoint.subtype must have a value");
-            }
-
-            if (endpoint.getSubType() != null && properties.getSubType() != null && !properties.getSubType().equals(endpoint.getSubType())) {
-                throw new BadRequestException("endpoint.subtype must be equal to endpoint.properties.subtype. Consider removing endpoint.properties.subtype as it is deprecated");
-            }
```
gwenneg, line 50: Is the removal of these lines a mistake or do you really mean to delete them?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -186,7 +232,22 @@ public Response deleteEndpoint(@Context SecurityContext sec, @PathParam("id") UU
...
+            EndpointProperties properties = e.getProperties();
+            if (properties instanceof CamelProperties) {
+                CamelProperties cp = (CamelProperties) properties;
+                // Special case wrt OpenBridge
+                if (e.getSubType().equals("slack")) {
+                    String processorId = cp.getExtras().get(OB_PROCESSOR_ID);
+                    bridgeApiService.deleteProcessor(bridge.getId(), processorId, bridgeAuth.getToken());
+                }
+            }
+        }
```
gwenneg, line 247: This should probably be done only if `obEnabled` is true.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -144,6 +172,24 @@ public Endpoint createEndpoint(@Context SecurityContext sec, @NotNull @Valid End
...
+            if (endpointSubType.equals("slack")) {
+                CamelProperties properties = endpoint.getProperties(CamelProperties.class);
+                String processorName = "p-" + endpoint.getAccountId() + "-" + UUID.randomUUID();
+                properties.getExtras().put(OB_PROCESSOR_NAME, processorName);
+                String processorId = setupOpenBridgeProcessor(endpoint, properties, processorName);
+
+                // TODO find a better place for these, that should not be
+                //       visible to users / OB actions
+                //       See also CamelTypeProcessor#callOpenBridge
+                properties.getExtras().put(OB_PROCESSOR_ID, processorId);
```
gwenneg, line 188: Did you check if Hibernate ORM saves the properties changes into the DB? It is supposed to work like that since we're talking about a persistent entity within a JTA transaction, but since we recently moved to Hibernate ORM, maybe it's worth checking.

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
pilhuhn: Yes, that is a mistake

backend/src/main/resources/application.properties
```diff
@@ -62,3 +62,16 @@ quarkus.log.cloudwatch.access-key-secret=placeholder
...
 
 template-engine/mp-rest/url=${clowder.endpoints.notifications-engine-service:http://localhost:8087}
+
+
+ob.enabled=false
+ob.kcUser=<key-cloak-user>
+ob.kcPass=<key-cload-password>
+ob.bridge.name=<name-of-the-bridge->
+ob.token.user=<ob-user-for-token>
+ob.token.pass=<ob-pass-for-token>
```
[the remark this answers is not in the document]
gwenneg: They could be commented then. But it's not a real issue, I can live with the current version 😃

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
pilhuhn: Actually the code works without them and assumes `endpoint.getSubType()` is valid, so I guess there is no real harm in keeping it removed

gwenneg: They were added for validation purposes. The code may work without them but we could process an invalid REST request without that validation. @josejulio since you added that, WDYT?

josejulio: they were added in the interim for camel users that were still sending to `properties.subType` instead of `subType` directly.

There are other changes needed.
The whole `// Todo: NOTIF-429 backward compatibility change - Remove soon.` block should go away and there are other blocks with this comment.

Are you aware of any team creating integrations directly via the API ? or is everyone using the UI directly?

pilhuhn: Currently Eventing is the only team using the API directly (afaik)

gwenneg (approved):
LGTM

common/src/main/java/com/redhat/cloud/notifications/openbridge/BridgeEventService.java
```diff
@@ -0,0 +1,25 @@
...
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
```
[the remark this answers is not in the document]
gwenneg: That's not what I understand from [the doc](https://github.com/eclipse/microprofile-rest-client/blob/master/spec/src/main/asciidoc/clientexamples.asciidoc#sample-definitions):
>Users may specify the media (MIME) type of the outbound request using the @Consumes annotation (this determine the Content-Type HTTP header), and the expected media type(s) of the response by using the @Produces annotation (the Accept HTTP header). This indicates that the client interface expects that the remote service consumes/produces the specified types. If no @Consumes or @Produces annotation is specified for a given request, it will default to jakarta.ws.rs.core.MediaType.APPLICATION_JSON ("application/json").

pilhuhn:
You are right.
I am doing exactly what you say in the Auth-interface.

gwenneg:
@pilhuhn Is this ready for a merge?

pilhuhn:
Yes, @gwenneg

PR #1025, opened 2022-03-03
NOTIF-488 Remove the reactiveness from engine

gwenneg:
This PR is the second and last phase of the Hibernate Reactive removal from all notifications modules. It also removes most of the reactiveness but there are a few small parts of the code that are still reactive. I will probably push things a bit further in a subsequent PR.

ℹ️ I tried to keep the code changes as minimal as possible to make the review easier. This can sometimes result in "weird" code in a few places but this PR isn't meant to replace the reactiveness with pretty imperative code. My short term goal is to have imperative code that is reviewable and without regressions (hopefully!). We can make the code prettier later (or not).

Before you start reviewing the changes, please have a look at the new `StatelessSessionFactory` class. Quarkus does not provide a way to inject an Hibernate ORM stateless session which is why I had to create that factory class. Each stateless session is bound to a thread (using `ThreadLocal`) and lives for the duration of `one` Kafka message processing in the `runtime` code. This means that the stateless session is created when we start processing the message and closed (and garbage collected) when the processing is finished. We can't use the same session for all messages because if a DB exception is thrown within the boundaries of a session, Hibernate will not accept to execute any more queries with that session.

:warning: All SQL queries executed from the test code are run using a `stateful` session with `EntityManager`. Even if it's technically possible, from now on the `stateless` session should never be injected and used in test code. Otherwise, because JUnit and SmallRye Reactive Messaging use the same thread during a test run, using a stateless session in test code will cause a lot of issues.


gwenneg:
I'm not done self-reviewing `EventConsumer` and `EmailSubscriptionTypeProcessor` but you can also start reviewing all other classes.

engine/src/main/java/com/redhat/cloud/notifications/session/StatelessSessionFactory.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * This method opens a new stateless session or reuses an existing one if available. That session is then passed to
+     * the given consumer which is executed. At the end of the execution, the session is closed.
+     * @param unitOfWork a unit of work
+     */
+    public void withSession(Consumer<StatelessSession> unitOfWork) {
+        try {
+            unitOfWork.accept(getOrCreateSession());
+        } finally {
+            closeSession();
```
josejulio, line 32: Wondering what will happen if there are multiple nested calls to `StatelessSessionFactory.withSession`, wouldn't the last one close it all for the others?

i.e.
```java
StatelessSessionFactory.withSession(s1 -> {
  // do stuff with s1
  StatelessSessionFactory.withSession(s2 -> {
    // do stuff with s2
  })
  // do more more stuff with s1 - but it's closed already?
})
```

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -97,127 +98,122 @@ void postConstruct() {
...
-    private Multi<Tuple2<NotificationHistory, EmailAggregationKey>> processAggregateEmailsByAggregationKey(EmailAggregationKey aggregationKey, LocalDateTime startTime, LocalDateTime endTime, EmailSubscriptionType emailSubscriptionType, boolean delete) {
+    private List<Tuple2<NotificationHistory, EmailAggregationKey>> processAggregateEmailsByAggregationKey(EmailAggregationKey aggregationKey, LocalDateTime startTime, LocalDateTime endTime, EmailSubscriptionType emailSubscriptionType, boolean delete) {
         final EmailTemplate emailTemplate = emailTemplateFactory.get(aggregationKey.getBundle(), aggregationKey.getApplication());
 
-        Multi<Tuple2<NotificationHistory, EmailAggregationKey>> doDelete = delete ?
-                emailAggregationRepository.purgeOldAggregation(aggregationKey, endTime)
-                        .onItem().transformToMulti(unused -> Multi.createFrom().empty()) :
-                Multi.createFrom().empty();
+        if (delete) {
+            // TODO NOTIF-488 To be changed
```
josejulio, line 194: What is this todo for?

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -97,127 +98,122 @@ void postConstruct() {
...
         }
-        return emailAggregator.getAggregated(aggregationKey, emailSubscriptionType, startTime, endTime)
-                .onItem().transform(Map::entrySet)
-                .onItem().transformToMulti(Multi.createFrom()::iterable)
-                .onItem().transformToMultiAndConcatenate(entries -> {
+
+        Set<Map.Entry<User, Map<String, Object>>> bla = emailAggregator.getAggregated(aggregationKey, emailSubscriptionType, startTime, endTime)
+                .entrySet();
+
+        return bla.stream()
```
josejulio, line 212: image

engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
```diff
@@ -109,72 +118,70 @@ private WebClient getWebClient(boolean disableSSLVerification) {
...
-                            details.put("error_message", t.getMessage()); // TODO This message isn't always the most descriptive..
-                            history.setDetails(details.getMap());
-                            return history;
-                        })
-                );
+        try {
+            return Failsafe.with(retryPolicy).get(() -> {
+
+                // TODO NOTIF-488 We may want to move to a non-reactive HTTP client in the future.
+                HttpResponse<Buffer> resp = req.sendJsonObject(payload).await().indefinitely();
```
josejulio, line 128: Should we have a reasonable timeout here?

engine/src/main/java/com/redhat/cloud/notifications/session/StatelessSessionFactory.java
josejulio: For what I have review so far - It looks like the assumption above is correct as the code calls the session create at the start of the flows.

I'm wondering if we should have a check in `withSession` to verify that there is no session (else above scenario would happen) - if that's the case, is better to fail early (`"hey, don't call this method multiple times, use StatelessSessionFactory.getSession()"`). 

Also, probably should split the `getOrCreateSession` into a private `createSession` and a public `getSession` so that we can check in `getSession` that there is already a session or warn/fail about the invalid usage.

engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateService.java
```diff
@@ -16,15 +14,14 @@
...
 
-    public Uni<TemplateInstance> compileTemplate(String template, String name) {
-        return Uni.createFrom().item(engine.parse(template, null, name))
-                .onItem().transform(Template::instance);
+    public TemplateInstance compileTemplate(String template, String name) {
+        return engine.parse(template, null, name).instance();
     }
 
-    public Uni<String> renderTemplate(User user, Action action, TemplateInstance templateInstance) {
+    public String renderTemplate(User user, Action action, TemplateInstance templateInstance) {
```
josejulio, line 21: Wondering if (when we have the templates in the database) this compile/render thing should be a common thing across the backend/engine

engine/src/main/java/com/redhat/cloud/notifications/session/StatelessSessionFactory.java
gwenneg: > // do more more stuff with s1 - but it's closed already?

Yes it's closed. I'll try to change things a bit and prevent that.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -97,127 +98,122 @@ void postConstruct() {
...
-                .onItem().transform(Set::copyOf)
-                .onItem().transformToUni(subscribers -> recipientResolver.recipientUsers(action.getAccountId(), requests, subscribers))
-        .onItem().transformToMulti(Multi.createFrom()::iterable)
-        .onItem().transformToUniAndConcatenate(user -> emailSender.sendEmail(user, event, subject, body));
+        Set<String> subscribers = Set.copyOf(emailSubscriptionRepository
+                .getEmailSubscribersUserId(action.getAccountId(), action.getBundle(), action.getApplication(), emailSubscriptionType));
+        return recipientResolver.recipientUsers(action.getAccountId(), requests, subscribers)
+                .stream()
+                .map(user -> emailSender.sendEmail(user, event, subject, body))
+                .filter(history -> history != null)
```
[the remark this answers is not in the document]
gwenneg: That's one of the unfinished things I'll fix tomorrow 😃

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
gwenneg: That's one of the unfinished things I'll fix tomorrow 😃

engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
gwenneg: Good catch! Yes we need a timeout here.

engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateService.java
gwenneg: Maybe... we'll see when the DB thing is done, but I like the idea of having only one app (currently the `engine`) responsible for the templates rendering (separation of concerns).

engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
gwenneg: I added a temporary timeout value (60 seconds) which can be changed with configuration. I'm not sure about the best value there... don't hesitate to suggest something else.

I'll try to remove that `await()` call soon so this should only be temporary hopefully.

engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
```diff
@@ -68,8 +67,7 @@ public void init() {
     }
 
     @Incoming(INGRESS_CHANNEL)
-    @Acknowledgment(PRE_PROCESSING)
```
gwenneg, line 71: Note for myself: to be confirmed.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
bennetelli: hahaha :-D

engine/pom.xml
```diff
@@ -104,6 +100,13 @@
             <version>${cloudwatch.version}</version>
         </dependency>
 
+        <!-- Retries -->
+        <dependency>
+            <groupId>dev.failsafe</groupId>
```
bennetelli, line 105: why don't we use resilience4j? The api is much cleaner imo and even Netflix deprecated their hystrix project and uses resilience4j instead.

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -45,106 +40,99 @@
...
+        EmailSubscriptionProperties properties = new EmailSubscriptionProperties();
+        Optional<Endpoint> endpointOptional = emailEndpoints
+                .stream()
+                .filter(endpoint -> properties.hasSameProperties(endpoint.getProperties(EmailSubscriptionProperties.class)))
+                .findFirst();
+        if (endpointOptional.isPresent()) {
+            return endpointOptional.get();
+        }
+
+        Endpoint endpoint = new Endpoint();
```
bennetelli, line 60: we could put this endpoint creation in a separate method.

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -45,106 +40,99 @@
...
-                        }
-                    });
-        });
+        List<Endpoint> endpoints = statelessSessionFactory.getCurrentSession().createQuery(query, Endpoint.class)
+                .setParameter("eventType", eventType)
+                .setParameter("accountId", tenant)
+                .getResultList();
+        loadProperties(endpoints);
+        for (Endpoint endpoint : endpoints) {
+            if (endpoint.getAccountId() == null) {
```
bennetelli, line 85: the null check looks like a check for a possible coding issue, but is actually a business check. Wrapping it in a separate method would make this more clear for others.

engine/pom.xml
gwenneg: I added failsafe here because I've used that lib in many projects for years and it never caused any production issue. I don't really know resilience4j so I don't have a preference.

Which part of the failsafe API used in this PR isn't clean enough?

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
gwenneg: This PR is only focused on dropping the reactiveness. I'm not changing old code unless it has a direct impact on the reactiveness drop.

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
gwenneg: This PR is only focused on dropping the reactiveness. I'm not changing old code unless it has a direct impact on the reactiveness drop.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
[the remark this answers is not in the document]
gwenneg: Updated.

> `history == null` are the ones that failed?

Yes, that's how `EmailSender` behaves (before this PR) in case of Qute exception.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
gwenneg: Removed.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
gwenneg: That wasn't meant to be pushed 😄 Removed.

gwenneg:
ℹ️ I'm done self-reviewing/polishing the changes.

engine/pom.xml
bennetelli: Just took a look at it. Version 3 of it looks much better than the previous one. I remember we updated from failsafe 1 to 2 some years ago and got a generic api which worked with Object instead of concrete classes. But nevermind, version 3 looks good. We can stay with failsafe. In general they are doing the same.

bennetelli (approved):

josejulio (approved):
lgtm

PR #1027, opened 2022-03-07
NOTIF-369 Forbid behavior groups display names duplicates

bennetelli:
…me display name

### DIFF
--- a/backend/src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java
+++ b/backend/src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java
@@ -85,6 +85,41 @@ public class LifecycleITest extends DbIsolatedTest {
     }
 
     @Test
+    void shouldReturn400AndBadRequestExceptionWhenDisplayNameIsAlreadyPresent() {
+        final String accountId = "tenant";
+        final String orgId = "someOrdId";
+        final String username = "user";
+
+        // Identity header used for all public APIs calls. Internal APIs calls don't need that.
+        Header identityHeader = initRbacMock(accountId, orgId, username, RbacAccess.FULL_ACCESS);
+        Header turnpikeIdentityHeader = createTurnpikeIdentityHeader("admin", adminRole);
+
+        // First, we need a bundle, an app and an event type. Let's create them!
+        String bundleId = createBundle(turnpikeIdentityHeader);
+
+        // Create first behavior group with name BG1
+        createBehaviorGroup(identityHeader, bundleId, "BG1");
+
+        // create a Behavior Group with the same name again
+        BehaviorGroup behaviorGroup = new BehaviorGroup();
+        behaviorGroup.setDisplayName("BG1");
+        behaviorGroup.setBundleId(UUID.fromString(bundleId));
+
+        String responseBody = given()
+                .header(identityHeader)
+                .contentType(JSON)
+                .body(Json.encode(behaviorGroup))
+                .when()
+                .post(API_NOTIFICATIONS_V_1_0 + "/notifications/behaviorGroups")
+                .then()
+                .statusCode(400)
+                .contentType(JSON)
+                .extract().asString();
+
+        assertEquals("A behavior group with display name [" + behaviorGroup.getDisplayName() + "] already exists", responseBody);
+    }
+
+    @Test
     void test() {
         /*
          * TODO
@@ -114,8 +149,8 @@ public class LifecycleITest extends DbIsolatedTest {
         checkAllEventTypes(identityHeader);
 
         // We also need behavior groups.
-        String behaviorGroupId1 = createBehaviorGroup(identityHeader, bundleId);
-        String behaviorGroupId2 = createBehaviorGroup(identityHeader, bundleId);
+        String behaviorGroupId1 = createBehaviorGroup(identityHeader, bundleId, "Behavior group 1");
+        String behaviorGroupId2 = createBehaviorGroup(identityHeader, bundleId, "Behavior group 2");
         String defaultBehaviorGroupId = createDefaultBehaviorGroup(turnpikeIdentityHeader, bundleId);
 
         // We need actions for our behavior groups.
@@ -349,9 +384,9 @@ public class LifecycleITest extends DbIsolatedTest {
         assertEquals(2, jsonEventTypes.size()); // One from the current test, one from the default DB records.
     }
 
-    private String createBehaviorGroupInternal(String path, Header identityHeader, String bundleId) {
+    private String createBehaviorGroupInternal(String path, Header identityHeader, String bundleId, String displayName) {
         BehaviorGroup behaviorGroup = new BehaviorGroup();
-        behaviorGroup.setDisplayName("Behavior group");
+        behaviorGroup.setDisplayName(displayName);
         behaviorGroup.setBundleId(UUID.fromString(bundleId));
 
         var request = given();
@@ -381,12 +416,12 @@ public class LifecycleITest extends DbIsolatedTest {
         return jsonBehaviorGroup.getString("id");
     }
 
-    private String createBehaviorGroup(Header identityHeader, String bundleId) {
-        return createBehaviorGroupInternal(API_NOTIFICATIONS_V_1_0 + "/notifications/behaviorGroups", identityHeader, bundleId);
+    private String createBehaviorGroup(Header identityHeader, String bundleId, String displayName) {
+        return createBehaviorGroupInternal(API_NOTIFICATIONS_V_1_0 + "/notifications/behaviorGroups", identityHeader, bundleId, displayName);
     }
 
     private String createDefaultBehaviorGroup(Header turnpikeIdentityHeader, String bundleId) {
-        return createBehaviorGroupInternal(API_INTERNAL + "/behaviorGroups/default", turnpikeIdentityHeader, bundleId);
+        return createBehaviorGroupInternal(API_INTERNAL + "/behaviorGroups/default", turnpikeIdentityHeader, bundleId, "Behavior group");
     }
 
     @Transactional
--- a/common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
+++ b/common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
@@ -13,9 +13,7 @@ import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
 import javax.persistence.JoinColumn;
 import javax.persistence.ManyToOne;
-import javax.persistence.NamedQuery;
 import javax.persistence.OneToMany;
-import javax.persistence.QueryHint;
 import javax.persistence.Table;
 import javax.persistence.Transient;
 import javax.validation.constraints.NotBlank;
@@ -30,27 +28,11 @@ import static com.fasterxml.jackson.annotation.JsonInclude.Include;
 import static com.fasterxml.jackson.annotation.JsonProperty.Access.READ_ONLY;
 import static com.fasterxml.jackson.databind.PropertyNamingStrategies.SnakeCaseStrategy;
 import static javax.persistence.FetchType.LAZY;
-import static org.hibernate.jpa.QueryHints.HINT_PASS_DISTINCT_THROUGH;
 
 @Entity
 @Table(name = "behavior_group")
 @JsonNaming(SnakeCaseStrategy.class)
 @JsonFilter(ApiResponseFilter.NAME)
-/*
- * When PostgreSQL sorts a BOOLEAN column in DESC order, true comes first. That's not true for all DBMS.
- *
- * When QueryHints.HINT_PASS_DISTINCT_THROUGH is set to false, Hibernate returns distinct results without passing the
- * DISTINCT keyword to the DBMS. This is better for performances.
- * See https://in.relation.to/2016/08/04/introducing-distinct-pass-through-query-hint/ for more details about that hint.
- */
-// TODO Move this query back to BehaviorGroupResources when hints are implemented in Mutiny.Query.
-@NamedQuery(
-        name = "findByBundleId",
-        query = "SELECT DISTINCT b FROM BehaviorGroup b LEFT JOIN FETCH b.actions a " +
-                "WHERE (b.accountId = :accountId OR b.accountId IS NULL) AND b.bundle.id = :bundleId " +
-                "ORDER BY b.created DESC, a.position ASC",
-        hints = @QueryHint(name = HINT_PASS_DISTINCT_THROUGH, value = "false")
-)
 public class BehaviorGroup extends CreationUpdateTimestamped {
 
     @Id
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
@@ -24,6 +24,7 @@ import java.util.UUID;
 
 import static javax.ws.rs.core.Response.Status.NOT_FOUND;
 import static javax.ws.rs.core.Response.Status.OK;
+import static org.hibernate.jpa.QueryHints.HINT_PASS_DISTINCT_THROUGH;
 
 @ApplicationScoped
 public class BehaviorGroupRepository {
@@ -43,6 +44,8 @@ public class BehaviorGroupRepository {
 
     @Transactional
     BehaviorGroup create(String accountId, BehaviorGroup behaviorGroup, boolean isDefaultBehaviorGroup) {
+        checkBehaviorGroupDisplayNameDuplicate(accountId, behaviorGroup, isDefaultBehaviorGroup);
+
         behaviorGroup.setAccountId(accountId);
         if (isDefaultBehaviorGroup != behaviorGroup.isDefaultBehavior()) {
             throw new BadRequestException(String.format(
@@ -81,9 +84,20 @@ public class BehaviorGroupRepository {
             throw new NotFoundException("Bundle not found");
         }
 
-        List<BehaviorGroup> behaviorGroups = entityManager.createNamedQuery("findByBundleId", BehaviorGroup.class)
+        /*
+         * When PostgreSQL sorts a BOOLEAN column in DESC order, true comes first. That's not true for all DBMS.
+         *
+         * When QueryHints.HINT_PASS_DISTINCT_THROUGH is set to false, Hibernate returns distinct results without passing the
+         * DISTINCT keyword to the DBMS. This is better for performances.
+         * See https://in.relation.to/2016/08/04/introducing-distinct-pass-through-query-hint/ for more details about that hint.
+         */
+        String query = "SELECT DISTINCT b FROM BehaviorGroup b LEFT JOIN FETCH b.actions a " +
+                "WHERE (b.accountId = :accountId OR b.accountId IS NULL) AND b.bundle.id = :bundleId " +
+                "ORDER BY b.created DESC, a.position ASC";
+        List<BehaviorGroup> behaviorGroups = entityManager.createQuery(query, BehaviorGroup.class)
                 .setParameter("accountId", accountId)
                 .setParameter("bundleId", bundleId)
+                .setHint(HINT_PASS_DISTINCT_THROUGH, false)
                 .getResultList();
         for (BehaviorGroup behaviorGroup : behaviorGroups) {
             behaviorGroup.filterOutBundle();
@@ -101,6 +115,8 @@ public class BehaviorGroupRepository {
 
     @Transactional
     boolean update(String accountId, BehaviorGroup behaviorGroup, boolean isDefaultBehavior) {
+        checkBehaviorGroupDisplayNameDuplicate(accountId, behaviorGroup, isDefaultBehavior);
+
         checkBehaviorGroup(behaviorGroup.getId(), isDefaultBehavior);
         String query = "UPDATE BehaviorGroup SET displayName = :displayName WHERE id = :id";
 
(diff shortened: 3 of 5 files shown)


common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
```diff
@@ -43,13 +44,16 @@
...
-)
+@NamedQueries({
+        @NamedQuery(
+                name = "findByBundleId",
+                query = "SELECT DISTINCT b FROM BehaviorGroup b LEFT JOIN FETCH b.actions a " +
+                        "WHERE (b.accountId = :accountId OR b.accountId IS NULL) AND b.bundle.id = :bundleId " +
+                        "ORDER BY b.created DESC, a.position ASC",
+                hints = @QueryHint(name = BehaviorGroup.HINT_PASS_DISTINCT_THROUGH, value = "false")),
+        @NamedQuery(name = "findByDisplayName",
+                query = "SELECT b FROM BehaviorGroup b WHERE b.displayName = :displayName"),
```
josejulio, line 55: This isn't wrong, but I'm not really sure if there is a different between using a named query or directly using the ORM to filter by a column.

I'm guessing we should stick to one and only use the named queries for specific cases.

@gwenneg WDYT?

josejulio: Also, this doesn't take into account the accountId. 

We should prevent from having the same (accountId,DisplayName). Note that accountId could be `null` and needs to be considered in the query (dynamic query) because `b.accountId = null` won't work well (`b.accountId is NULL)`

bennetelli: I like NamedQueries because they are "close" to the class where they belong to. So the SQL query isn't located somewhere in the code. IMO it's a bit like getters, setters or even business methods. They should be located in the class whose behavior they are changing instead of being somewhere in other classes or so.
But that's just my personal opinion :)

bennetelli: Another use case would be: if you search for "something which works with the behavior group feature" you would probably take a look at the behavior group first and find the queries immediately.

gwenneg: I think putting all queries in a similar place saves dev time.

We can't make all notifications queries named because of the many dynamic queries we already have. In the meantime, we can put all queries in the `*Resources`/`*Repository` classes.

So I think named queries aren't the best productivity option, at least for me.  I will always look for a query in the `*Resources`/`*Repository` class first before realizing the query is actually named.

common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
```diff
@@ -43,13 +44,16 @@
...
-)
+@NamedQueries({
+        @NamedQuery(
+                name = "findByBundleId",
+                query = "SELECT DISTINCT b FROM BehaviorGroup b LEFT JOIN FETCH b.actions a " +
+                        "WHERE (b.accountId = :accountId OR b.accountId IS NULL) AND b.bundle.id = :bundleId " +
+                        "ORDER BY b.created DESC, a.position ASC",
+                hints = @QueryHint(name = BehaviorGroup.HINT_PASS_DISTINCT_THROUGH, value = "false")),
+        @NamedQuery(name = "findByDisplayNameAndAccountId",
+                query = "SELECT b FROM BehaviorGroup b WHERE b.displayName = :displayName AND b.accountId = :accountId"),
```
gwenneg, line 55: We don't need the `BehaviorGroup` data so this could be replaced with a `COUNT(*)` query.

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -41,6 +41,13 @@ public BehaviorGroup createDefault(BehaviorGroup behaviorGroup) {
 
     @Transactional
     BehaviorGroup create(String accountId, BehaviorGroup behaviorGroup, boolean isDefaultBehaviorGroup) {
+        List<BehaviorGroup> behaviorGroupsSameDisplayNameAlreadyPresent = entityManager.createNamedQuery("findByDisplayNameAndAccountId", BehaviorGroup.class)
+                .setParameter("displayName", behaviorGroup.getDisplayName())
+                .setParameter("accountId", behaviorGroup.getAccountId())
+                .getResultList();
+        if (behaviorGroupsSameDisplayNameAlreadyPresent == null || !behaviorGroupsSameDisplayNameAlreadyPresent.isEmpty()) {
```
gwenneg, line 48: `getResultList()` never returns `null`.

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -41,6 +41,13 @@ public BehaviorGroup createDefault(BehaviorGroup behaviorGroup) {
 
     @Transactional
     BehaviorGroup create(String accountId, BehaviorGroup behaviorGroup, boolean isDefaultBehaviorGroup) {
+        List<BehaviorGroup> behaviorGroupsSameDisplayNameAlreadyPresent = entityManager.createNamedQuery("findByDisplayNameAndAccountId", BehaviorGroup.class)
+                .setParameter("displayName", behaviorGroup.getDisplayName())
+                .setParameter("accountId", behaviorGroup.getAccountId())
```
gwenneg, line 46:
```suggestion
                .setParameter("accountId", accountId)
```
(the behavior group `accountId` field should always be `null` at this point of the method execution)

gwenneg (commented):
Could you please fix the commit message / PR title?

gwenneg (changes requested):
Could you please add a test for this?

common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
bennetelli: In general we do have the same opinion about it :) 

In Spring Boot the Repositories are a good place to put the queries in. Quarkus / JPA probably doesn't have that feature, so *Resources/*Repository classes are something like Quarkus / JPA version of it.
I would also be fine with it.

common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
bennetelli: good catch. thanks.

common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
bennetelli: > Could you please add a test for this?

There is one. BehaviorGroupResourcesTest.shouldThrowExceptionWhenBehaviorGroupWithAccountIdAndDisplayNameAlreadyExists()

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
bennetelli: done

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
bennetelli: changed.

bennetelli:
> Could you please add a test for this?

done.

common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
bennetelli: @gwenneg I will open a new clean PR which moves the NamedQueries back in the BehaviorGroupResources.

backend/src/test/java/com/redhat/cloud/notifications/db/BehaviorGroupResourcesTest.java
```diff
@@ -48,6 +48,17 @@
     @Inject
     BehaviorGroupResources behaviorGroupResources;
 
+    @Test
+    void shouldThrowExceptionWhenBehaviorGroupWithAccountIdAndDisplayNameAlreadyExists() {
+        final IllegalArgumentException exception = assertThrows(IllegalArgumentException.class, () -> {
+            Bundle bundle = resourceHelpers.createBundle();
+            BehaviorGroup behaviorGroup = resourceHelpers.createBehaviorGroup(DEFAULT_ACCOUNT_ID, "displayName", bundle.getId());
```
josejulio, line 55: Could you add one for for a default behavior group too? 

Thanks!

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -41,14 +41,20 @@ public BehaviorGroup createDefault(BehaviorGroup behaviorGroup) {
 
     @Transactional
     BehaviorGroup create(String accountId, BehaviorGroup behaviorGroup, boolean isDefaultBehaviorGroup) {
+        List<Long> behaviorGroupsSameDisplayNameAlreadyPresent = entityManager.createNamedQuery("BehaviorGroup.countByDisplayNameAndAccountId", Long.class)
+                .setParameter("displayName", behaviorGroup.getDisplayName())
+                .setParameter("accountId", accountId)
+                .getResultList();
+        if (behaviorGroupsSameDisplayNameAlreadyPresent.get(0).intValue() > 0) {
```
gwenneg, line 48:
```suggestion
        Long behaviorGroupsSameDisplayNameAlreadyPresent = entityManager.createNamedQuery("BehaviorGroup.countByDisplayNameAndAccountId", Long.class)
                .setParameter("displayName", behaviorGroup.getDisplayName())
                .setParameter("accountId", accountId)
                .getSingleResult();
        if (behaviorGroupsSameDisplayNameAlreadyPresent > 0) {
```

common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
gwenneg: > @gwenneg I will open a new clean PR which moves the NamedQueries back in the BehaviorGroupResources.

Why not in this one?

common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
```diff
@@ -53,6 +53,8 @@
                 hints = @QueryHint(name = BehaviorGroup.HINT_PASS_DISTINCT_THROUGH, value = "false")),
         @NamedQuery(name = "BehaviorGroup.countByDisplayNameAndAccountId",
                 query = "SELECT COUNT(*) FROM BehaviorGroup b WHERE b.displayName = :displayName AND b.accountId = :accountId"),
+        @NamedQuery(name = "BehaviorGroup.countByDisplayAccountId",
+                query = "SELECT COUNT(*) FROM BehaviorGroup b WHERE b.displayName = :displayName"),
```
josejulio, line 57: We need to add a check for the accountId here
```suggestion
                query = "SELECT COUNT(*) FROM BehaviorGroup b WHERE b.displayName = :displayName" and b.accountId IS NULL),
```

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
bennetelli: I read about it and wanted to use getSingleResult() as well, but after reading these arguments against it, I decided to use getResultList() even when there is only one result in most of the cases: https://stackoverflow.com/questions/8500031/what-is-better-getsingleresult-or-getresultlist-jpa

common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
bennetelli: added. thanks!

common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
bennetelli: Thought it would be a bit out of scope as it's more or less a result task of the reactiveness removal.

I can also do it here. Give me some minutes.

bennetelli: I don't want to deal with merge conflicts in the evening :/

bennetelli: it's part of this PR since commit no. [7d1e79f](https://github.com/RedHatInsights/notifications-backend/pull/1027/commits/7d1e79faa2307f77d960707f66ced7fdb537f17b)

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
gwenneg: I don't see any argument in that discussion in favor of `getResultList()` when the query returns a `count`. Did I miss something?

common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
gwenneg: My question was only focused on the named query you added. The other one has to stay anyway for now.

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
bennetelli: good point. the situation changed after adding COUNT(*) to the query. Before this, I would have to add try and catch and wait for an exception to appear when there is either a result or no result. But with count, the situation is different.

Just changed it to getSingleResult()

common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
bennetelli: added again :D

bennetelli: done.

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -39,16 +40,36 @@ public BehaviorGroup createDefault(BehaviorGroup behaviorGroup) {
         return this.create(null, behaviorGroup, true);
     }
 
+    /*
+     * When PostgreSQL sorts a BOOLEAN column in DESC order, true comes first. That's not true for all DBMS.
+     *
+     * When QueryHints.HINT_PASS_DISTINCT_THROUGH is set to false, Hibernate returns distinct results without passing the
+     * DISTINCT keyword to the DBMS. This is better for performances.
+     * See https://in.relation.to/2016/08/04/introducing-distinct-pass-through-query-hint/ for more details about that hint.
+     */
```
josejulio, line 49: Could you move this comment to `findByBundleId` ?

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -41,14 +42,27 @@ public BehaviorGroup createDefault(BehaviorGroup behaviorGroup) {
...
+                    .setParameter("displayName", behaviorGroup.getDisplayName())
+                    .getSingleResult();
+        } else {
+            behaviorGroupsSameDisplayNameAlreadyPresent = entityManager.createQuery("SELECT COUNT(*) FROM BehaviorGroup b WHERE b.displayName = :displayName AND b.accountId = :accountId", Long.class)
+                    .setParameter("displayName", behaviorGroup.getDisplayName())
+                    .setParameter("accountId", accountId)
+                    .getSingleResult();
+        }
+        if (behaviorGroupsSameDisplayNameAlreadyPresent > 0) {
+            throw new IllegalArgumentException(String.format("A behavior group with accountId [%s] and display name [%s] is already present", accountId, behaviorGroup.getDisplayName()));
```
gwenneg, line 57: What is the HTTP status code returned when this exception is thrown? We need a 400 here and the HTTP response should also give the reason (same message than the exception here). Finally, the 400 status needs to be tested with a RestAssured test.

gwenneg:
Could you please fix the commit message? (too long)

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
bennetelli: it is a 400. Testing it is a bit tricky as the exception is thrown very deep in the code. I just added a throw new IllegalArgumentException("test"); in the /internal endpoint and this got mapped to a http 400.

josejulio: You could try updating the function at https://github.com/RedHatInsights/notifications-backend/pull/1027/files#diff-59f18caa249dea80d3c35f5a7d8723c28e5c9a5f3c6e820cedbecf807d48e30eR359 to receive the http-code as a param.

josejulio: I think this is what I meant: https://github.com/RedHatInsights/notifications-backend/blob/f23f77df47a905a106c365572f5a4364538c395a/backend/src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java#L369

Passing that value as a param, so you can check the 4xx status in there.

josejulio: ah - actually there are helpers in this class: https://github.com/RedHatInsights/notifications-backend/blob/f23f77df47a905a106c365572f5a4364538c395a/backend/src/test/java/com/redhat/cloud/notifications/CrudTestHelpers.java#L404

backend/src/main/java/com/redhat/cloud/notifications/IllegalArgumentExceptionMapper.java
```diff
@@ -0,0 +1,21 @@
+package com.redhat.cloud.notifications;
+
+import org.eclipse.microprofile.rest.client.ext.ResponseExceptionMapper;
+
+import javax.ws.rs.BadRequestException;
+import javax.ws.rs.core.Response;
+
+// This mapper is required to forward the HTTP 400 error message.
+public class IllegalArgumentExceptionMapper implements ResponseExceptionMapper<IllegalArgumentException> {
```
bennetelli, line 9: @gwenneg this mapping is somehow strange. Shouldn't it be something like toResponse() instead of toThrowable() ?

What is a good way to add this exception mapper? Just add it to all REST endpoint classes?

bennetelli: That's somehow a bit ugly. We probably would have to change NotificationService and InternalService signatures so it returns Response and surround the method calls inside with try, catch and mapping as we are doing for BadRequestException

backend/src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java
```diff
@@ -81,6 +81,50 @@ private Header initRbacMock(String tenant, String username, MockServerClientConf
...
+    @Test
+    void shouldReturn400AndBadRequestExceptionWhenDisplayNameIsAlreadyPresent() {
+        final String accountId = "tenant";
+        final String username = "user";
+
+        RequestDefaultBehaviorGroupPropertyList defaultBehaviorGroupProperties = new RequestDefaultBehaviorGroupPropertyList();
+        defaultBehaviorGroupProperties.setOnlyAdmins(true);
+
+        // All events are stored in the canonical email endpoint
+        RequestEmailSubscriptionProperties userEmailEndpointRequest = new RequestEmailSubscriptionProperties();
```
gwenneg, line 93: This isn't used during the test.

gwenneg:
The new check during the BG creation looks good, but we'll need the same thing for the behavior groups update since it can be used to change the display name.

bennetelli:
> The new check during the BG creation looks good, but we'll need the same thing for the behavior groups update since it can be used to change the display name.

good catch.

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
```diff
@@ -94,6 +109,10 @@ public boolean updateDefault(BehaviorGroup behaviorGroup) {
 
     @Transactional
     boolean update(String accountId, BehaviorGroup behaviorGroup, boolean isDefaultBehavior) {
+        if (behaviorGroupWithSameDisplayNameAlreadyPresent(accountId, behaviorGroup, isDefaultBehavior) > 0) {
+            throw new BadRequestException(String.format("A behavior group with accountId [%s] and display name [%s] is already present", accountId, behaviorGroup.getDisplayName()));
+        }
```
gwenneg, line 114: Could you please add tests for this?

bennetelli: ah, shit. forgot that part.

backend/src/test/java/com/redhat/cloud/notifications/db/BehaviorGroupResourcesTest.java
```diff
@@ -48,6 +49,39 @@
     @Inject
     BehaviorGroupResources behaviorGroupResources;
 
+    @Test
+    void shouldThrowExceptionWhenBehaviorGroupWithAccountIdAndDisplayNameAlreadyExistsWhenUpdating() {
+        final BadRequestException exception = assertThrows(BadRequestException.class, () -> {
+            Bundle bundle = resourceHelpers.createBundle();
+            final BehaviorGroup behaviorGroup = resourceHelpers.createBehaviorGroup(DEFAULT_ACCOUNT_ID, "displayName", bundle.getId());
+            resourceHelpers.updateBehaviorGroup(behaviorGroup);
```
gwenneg, line 57: I was about to approve and then that test made me realize there's a problem: updating `behaviorGroupA` with the display name of `behaviorGroupB` should be forbidden (this works) but updating `behaviorGroupA` without changing its display name should not be forbidden (this doesn't work).

For the update only, the SQL query needs to be something like:
```
SELECT COUNT(*) FROM BehaviorGroup b WHERE b.displayName = :displayName and b.accountId IS NULL AND id != :id
```

backend/src/main/java/com/redhat/cloud/notifications/db/BehaviorGroupResources.java
bennetelli: added

backend/src/test/java/com/redhat/cloud/notifications/db/BehaviorGroupResourcesTest.java
josejulio: This is still a thing, isn't it? @bennetelli

gwenneg:
Squashed commits and rebased on `master`. Comments still need to be addressed.

gwenneg (approved):
LGTM

gwenneg:
Some PR check failures are related. Investigation in progress...

PR #1035, opened 2022-03-09
Rename liveness check correctly

bennetelli:


gwenneg (commented):
Could you please update the `LivenessService` from `engine` as well?

bennetelli:
> Could you please update the `LivenessService` from `engine` as well?

Done. I was just diing into the backend only after reactiveness removal. Would have updated it otherwise.

gwenneg (approved):
LGTM

PR #1036, opened 2022-03-09
Refactorings

bennetelli:


backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -228,30 +228,30 @@ public boolean updateEndpoint(Endpoint endpoint) {
             return false;
         } else if (endpoint.getProperties() == null) {
             return true;
-        } else {
```
gwenneg, line 231: Having `else` blocks in the code even if they could be omitted makes the code easier to read/review/modify, especially for someone who didn't write it.

bennetelli: imo every additional indentation makes code harder to read

josejulio: I'm +1 on keeping the else.

Below chunk is not trivial and is better to have it surronded in a block.

Now... if you would like to wrap below code into a function i wouldn't mind a `return the_new_function()`

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -274,19 +274,21 @@ public void loadProperties(List<Endpoint> endpoints) {
...
-                    .getResultList();
-            for (T props : propList) {
-                if (props != null) {
-                    Endpoint endpoint = endpointsMap.get(props.getId());
-                    endpoint.setProperties(props);
-                    // Todo: NOTIF-429 backward compatibility change - Remove soon.
-                    if (typedEndpointClass.equals(CamelProperties.class)) {
-                        endpoint.getProperties(CamelProperties.class).setSubType(endpoint.getSubType());
-                    }
+        if (endpointsMap.size() <= 0) {
```
josejulio, line 277: size can't be below 0. 

It would be fine with a `endpointsMap.size() == 0` to avoid adding confusion.

backend/src/main/java/com/redhat/cloud/notifications/db/EndpointResources.java
```diff
@@ -300,17 +302,17 @@ public void loadProperties(List<Endpoint> endpoints) {
...
+                                        accountId == null,
+                                        WhereBuilder.builder().and("e.accountId IS NULL"),
+                                        WhereBuilder.builder().and("e.accountId = :accountId", "accountId", accountId)
+                                )
+                                .and(
+                                        WhereBuilder.builder()
+                                                .ifOr(basicTypes.size() > 0, "e.compositeType.type IN (:endpointType)", "endpointType", basicTypes)
+                                                .ifOr(compositeTypes.size() > 0, "e.compositeType IN (:compositeTypes)", "compositeTypes", compositeTypes)
+                                )
+                                .ifAnd(activeOnly != null, "e.enabled = :enabled", "enabled", activeOnly)
```
josejulio, line 315: Why do we need extra indentation? 
It just pushes the code to the far right and makes it harder to read.

PR #1040, opened 2022-03-10
Pushes coverage to codecov.io

josejulio:
Do we have coverage already somewhere else?
This seems to be working.

https://app.codecov.io/gh/RedHatInsights/notifications-backend/branch/coverage-test

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,6 +45,8 @@ jobs:
         run: bash .github/scripts/push-admin-console.sh
       - name: Build with Maven
         run: ./mvnw clean verify --no-transfer-progress
+      - name: Pushes coverage
+        run: bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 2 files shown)


josejulio:
@pilhuhn I recall you did some thing with jacoco before. It didn't work? or maybe we have the coverage somewhere?

gwenneg:
Did you look for coverage in our internal tool?

josejulio:
Didn't found any reported coverage else where. Maybe this is something we can start using.

gwenneg (commented):
https://quarkus.io/guides/tests-with-coverage may contain useful information about this.

bennetelli (approved):
in comparison to this: https://github.com/codecov/example-java it looks good :)

let's give it a try :)

josejulio:
I think it's finally working - had issues with the dependency en `dependencyManagement` instead of `dependencies`

PR #1054, opened 2022-03-11
NOTIF-510 begin UI mock for email templates

aferd:
first look at UI mock for email templates
On the Application page I put an email template table that would hold stored templates with ability to view/edit/delete. 

- Does the view option actually make sense? 
- Would it make sense to divide the table into "title templates" and " body templates"? or by "Daily Templates" and "Instant Notification"?
- email templates would be linked to their Application, while event-types would be linked to an email template

From there, create a new template would take you to a new page where you would be able to create/edit the template. I am not sure how this will be handled but we talked about having to differentiate between a daily email and an instant notification, therefore I added a simple checkbox that could tag the template in such a way.

https://user-images.githubusercontent.com/86322239/157925845-5ebf2681-3595-4a06-96d3-8215b45d75ce.mov

### DIFF
--- a/admin-console/src/main/webapp/src/pages/CreateEmailTemplatePage.tsx
+++ b/admin-console/src/main/webapp/src/pages/CreateEmailTemplatePage.tsx
@@ -1,0 +1,57 @@
+import { CodeEditor, Language } from '@patternfly/react-code-editor';
+import { ActionGroup, Button, PageSection, Split, SplitItem, Title } from '@patternfly/react-core';
+import React from 'react';
+import { Link } from 'react-router-dom';
+
+import { linkTo } from '../Routes';
+
+export const CreateEmailTemplatePage: React.FunctionComponent = () => {
+    return (
+        <>
+
+            <PageSection>
+                <Split>
+                    <SplitItem isFilled>
+                        <Title headingLevel="h1">Create/Edit Email Template for $applicationDisplayName</Title>
+                    </SplitItem>
+                    <SplitItem>
+                        <Button variant="primary" component={ (props: any) =>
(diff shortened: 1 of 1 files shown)


josejulio:
> * Does the view option actually make sense?

I think it does make sense, but not sure how we are going to implement it. 
We would need a payload to render the template, and it differs from event-type.

> * Would it make sense to divide the table into "title templates" and " body templates"? 

I'm not really sure, I think that "Email templates" is fine. Internally (API) we could store two templates.

> or by "Daily Templates" and "Instant Notification"?

I don't think we need to make this distinction at the template level. We could have an option (not in the template) to link an event-type to an event-type (instant) or link it to the application (daily)

> * email templates would be linked to their Application, while event-types would be linked to an email template

Yes, see above. 
We would have the email templates as children of the application - that would help with RBAC - apps would be in charge of their own templates.

josejulio:
I think we could use a feature flag or something to decide if we want to show the in-dev table. Hopefully soon we will be able to let applications use this admin portal. Or maybe some warning that it is still in development. 

Maybe with the addition of #984 we could use the "isAdmin" to decide to show this part.

aferd:
Ok, as far as the video above, overall does it make sense for what we are trying to do? Any suggestions on things to add/take away?

I will look into feature flags, that will be new to me. 
Otherwise, looking forward to getting this all hooked up!

gwenneg:
Was that work moved to another branch?

aferd:
It was not, I am pretty sure I did something here, but not exactly sure what. I was rebasing this branch with the new auth stuff from Josejulio and while doing that my computer restarted and i can't find the branch now.

gwenneg:
Your IDE local history may help recovering the lost files. Don't hesitate to ask for help if you've never done that.

PR #1057, opened 2022-03-14
NOTIF-381 Add get users total metric

bennetelli:


engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -51,14 +51,14 @@
...
     Duration maxBackOff;
 
-    @Inject
     MeterRegistry meterRegistry;
 
     private Counter failuresCounter;
     private RetryPolicy<Object> retryPolicy;
 
-    public RbacRecipientUsersProvider(ITUserServiceWrapper itUserService) {
+    public RbacRecipientUsersProvider(ITUserServiceWrapper itUserService, MeterRegistry meterRegistry) {
```
gwenneg, line 59: Injections are almost always done with `@Inject` in the entire app. Can we please stick with that to keep the code consistent? This constructor injection also creates noise in the test code for no reason.

bennetelli: Well.. I don't know how to test it otherwise. If you have an idea how to do it, please let me know.

josejulio: You can use `@InjectMock` in the test. see [this](https://github.com/RedHatInsights/notifications-backend/blob/80fe1b5953044f235f92c457f8246844909bbf92/engine/src/test/java/com/redhat/cloud/notifications/recipients/RecipientResolverTest.java#L26).

gwenneg: Actually the entire test needs to be rewritten. CDI is not working because `RbacRecipientUsersProvider` is instantiated with a `new` instead of using the injection/mocking capabilities we use in other tests. Let's fix that later.

gwenneg (approved):
LGTM, we can refactor the tested class and the test itself later.

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
bennetelli: In case anyone of you fixes the test, please ping me. I would be interested in seeing the fix

PR #1058, opened 2022-03-14
Clowdapp files for OpenBridge use

pilhuhn:
Those externalise the entries from application.properties for Clowder usage.

### DIFF
--- a/.rhcicd/clowdapp-backend.yaml
+++ b/.rhcicd/clowdapp-backend.yaml
@@ -88,8 +88,8 @@ objects:
           value: ${ENV_NAME}
         - name: OB_ENABLED
           value: ${OB_ENABLED}
-        - name: OB_BRIDGE_NAME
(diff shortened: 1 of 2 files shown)


gwenneg (commented):
Shouldn't some of them (passwords in particular) come from the vault?

pilhuhn:
In the long(er) run yes. For stage and their 'stable' env it is less of a deal (IMO)

gwenneg (approved):
This PR doesn't leak anything so LGTM I guess.

We need to be careful with the params that will be put in app-interface though.

pilhuhn:
Test failure is an issue in the test itself.

pilhuhn:
Changing to a draft. I will change the way to 'find' a bridge to using its uuid instead of the name.

pilhuhn:
Follow up to #1060

PR #1059, opened 2022-03-14
NOTIF-482 Allowing app devs to add applications

josejulio:
Continuation of #984 to avoid adding more noise.

The only change here is to allow app devs to create new applications.
They need to specify a role to manage the application (and they must be part of that role).


josejulio:
Updated and rebased

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/models/AddApplicationRequest.java
```diff
@@ -0,0 +1,19 @@
+package com.redhat.cloud.notifications.routers.internal.models;
+
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
+import java.util.UUID;
+
+public class AddApplicationRequest {
+    @NotNull
+    @Pattern(regexp = "[a-z][a-z_0-9-]*")
+    public String name;
```
gwenneg, line 10: ⚠️ Needs import.
```suggestion
    @Size(max = 255)
    public String name;
```

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/models/AddApplicationRequest.java
```diff
@@ -0,0 +1,19 @@
...
+
+    @NotNull
+    public String displayName;
+
+    @NotNull
+    public UUID bundleId;
+
+    public String ownerRole;
```
gwenneg, line 18: Is it possible to add a regexp constraint on this?

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/models/AddApplicationRequest.java
```diff
@@ -0,0 +1,19 @@
+package com.redhat.cloud.notifications.routers.internal.models;
+
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
+import java.util.UUID;
+
+public class AddApplicationRequest {
```
gwenneg, line 7: ⚠️ Needs import.
```suggestion
@JsonNaming(SnakeCaseStrategy.class)
public class AddApplicationRequest {
```

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/models/AddApplicationRequest.java
gwenneg: It's not very important but I suppose there are restrictions in RBAC regarding the characters allowed in a role name. I was thinking about doing something similar on our side.

Unrelated: we should put `@Size(max = 200)` on this field since we did it already in `InternalRoleAccess` 😃

josejulio: Unsure about that, but we can check it.
On the plus side, users would only be able to use of their roles. Only us (admins) would be able to type out something here.

gwenneg: It's really not important/urgent right now, let's not block this PR because of this 😄

gwenneg (approved):
LGTM

PR #1060, opened 2022-03-14
Use the bridge uuid instead of bridge name.

pilhuhn:
This is quicker and more stable. The bridge list is a paged entity and there is no guarantee right now, that our bridge is in the first page.


gwenneg (approved):
LGTM

pilhuhn:
Close&reopen to run the pr_check again

PR #1062, opened 2022-03-15
NOTIF-381 Move timer one method call up in the hierarchy

bennetelli:


bennetelli:
@gwenneg before we were measuring crap. Now it should be comparable to the measuring of the old rbac calls.

gwenneg:
image

gwenneg (approved):
LGTM

PR #1067, opened 2022-03-17
NOTIF-428 Remove floorist stuff from engine

bennetelli:


bennetelli:
I guess this is still there because of a merge. I removed it already.

PR #1069, opened 2022-03-17
NOTIF-381 Switch recipients retrieval between RBAC and IT

gwenneg:


josejulio (approved):
lgtm

gwenneg:
I need to fix the tests 😄

gwenneg:
@josejulio The IT impl is still enabled in the tests (to avoid rewriting/duplicating them) but the RBAC impl will be used by default when the application runs.

josejulio (approved):

PR #1070, opened 2022-03-17
NOTIF-623 Introduce Endpoint status

pilhuhn:
This introduces a status field for Endpoints. For classical ones this will directly be set to READY,
while for OpenBridge it is set to PROVISIONING and then a checker job checks the status and
updates it accordingly.

Currently it has those status values defined

    READY, // 0 , default for existing ones
    UNKNOWN,
    NEW,
    PROVISIONING,
    DELETING,
    FAILED

Not sure if we need all of them. For OpenBridge an additional UPDATING may be helpful, but could be mimic'd by PROVISIONING again.

@jeromemarc

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -82,7 +82,7 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-quartz</artifactId>
+            <artifactId>quarkus-scheduler</artifactId>
         </dependency>
 
         <!-- Flyway specific dependencies -->
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -80,6 +80,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-hibernate-orm</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-quartz</artifactId>
+        </dependency>
 
         <!-- Flyway specific dependencies -->
         <dependency>
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
@@ -16,6 +16,7 @@ import com.redhat.cloud.notifications.models.EmailSubscriptionProperties;
 import com.redhat.cloud.notifications.models.EmailSubscriptionType;
 import com.redhat.cloud.notifications.models.Endpoint;
 import com.redhat.cloud.notifications.models.EndpointProperties;
+import com.redhat.cloud.notifications.models.EndpointStatus;
 import com.redhat.cloud.notifications.models.EndpointType;
 import com.redhat.cloud.notifications.models.NotificationHistory;
 import com.redhat.cloud.notifications.openbridge.Bridge;
@@ -54,6 +55,7 @@ import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
+import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.SecurityContext;
@@ -109,7 +111,6 @@ public class EndpointResource {
 
     @Inject
     BridgeAuth bridgeAuth;
-
 
     @GET
     @Produces(APPLICATION_JSON)
@@ -193,14 +194,17 @@ public class EndpointResource {
                 String processorId = null;
                 try {
                     processorId = setupOpenBridgeProcessor(endpoint, properties, processorName);
+                    endpoint.setStatus(EndpointStatus.PROVISIONING);
                 } catch (Exception e) {
                     Log.warn("Processor setup failed: " + e.getMessage());
+                    endpoint.setStatus(EndpointStatus.FAILED);
                     throw new InternalServerErrorException("Can't set up the endpoint");
                 }
 
                 // TODO find a better place for these, that should not be
                 //       visible to users / OB actions
                 //       See also CamelTypeProcessor#callOpenBridge
+                //            and ReadyCheck#execute
                 properties.getExtras().put(OB_PROCESSOR_ID, processorId);
 
             }
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
@@ -55,7 +55,6 @@ import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
-import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.SecurityContext;
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/ReadyCheck.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/ReadyCheck.java
@@ -7,18 +7,21 @@ import com.redhat.cloud.notifications.models.EndpointType;
 import com.redhat.cloud.notifications.openbridge.Bridge;
 import com.redhat.cloud.notifications.openbridge.BridgeApiService;
 import com.redhat.cloud.notifications.openbridge.BridgeAuth;
+import io.quarkus.logging.Log;
 import io.quarkus.scheduler.Scheduled;
 import org.eclipse.microprofile.rest.client.inject.RestClient;
-import org.jboss.logging.Logger;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 import javax.persistence.EntityManager;
-import javax.persistence.Query;
 import javax.transaction.Transactional;
 import javax.ws.rs.WebApplicationException;
 import java.util.List;
 import java.util.Map;
+
+import static io.quarkus.scheduler.Scheduled.ConcurrentExecution.SKIP;
+import static javax.persistence.LockModeType.PESSIMISTIC_WRITE;
+import static org.hibernate.LockOptions.SKIP_LOCKED;
 
 /**
  * Scheduled task to check (OpenBridge) Endpoints
@@ -29,8 +32,6 @@ import java.util.Map;
  */
 @ApplicationScoped
 public class ReadyCheck  {
-
-    private static final Logger LOGGER = Logger.getLogger(ReadyCheck.class);
 
     @Inject
     EntityManager em;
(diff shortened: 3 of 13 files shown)


backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -180,17 +200,58 @@ public Endpoint createEndpoint(@Context SecurityContext sec, @NotNull @Valid End
...
+                                .withIntervalInSeconds(10)
+                )
+                .build();
+
+        try {
+            quartz.scheduleJob(jd, trigger);
+        } catch (SchedulerException e) {
+            em.find(Endpoint.class, endpointId).setStatus(EndpointStatus.FAILED);
+            e.printStackTrace();
+        }
```
josejulio, line 253: I think we should probably move this code to other service. What worries me here is that it has a "repeatForever" - Maybe we should have some kind of timeout or check it less often?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
```diff
@@ -180,17 +200,58 @@ public Endpoint createEndpoint(@Context SecurityContext sec, @NotNull @Valid End
...
+        JobDataMap jdm = new JobDataMap();
+        jdm.put("epid", endpointId);
+        jdm.put("pid", processorId);
+        JobDetail jd = JobBuilder.newJob().ofType(ReadyCheck.class)
+                .withIdentity("myJob")
+                .usingJobData(jdm)
+                .build();
+
+        Trigger trigger = TriggerBuilder.newTrigger()
+                .withIdentity(new TriggerKey("myJobs", "obWorker"))
```
josejulio, line 239: I don't know how quartz work, is this the key for the trigger? won't other slacks  endpoints overwrite this job/trigger?

pilhuhn: I was also thinking of passing a unique name instead of `myJobs` for each scheduler.
For the moment I am glad this works at all :-)

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
pilhuhn: Yes, that could probably go into `BridgeHelper`.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
gwenneg: Wouldn't it be simpler to have one single `@Scheduled`-annotated method that would check the status of all endpoints that currently have a non-final status at period execution time?

gwenneg:
```
@Scheduled
void checkEndpointsStatus() {
  // Get from DB all endpoints in non-final status
  // Call OB for each endpoint
  // Update endpoint status in DB if needed
}
```

josejulio: Wouldn't it cause conflicts with the replicas?

gwenneg: Not if Postgres' `PESSIMISTIC_LOCK` and `SKIP LOCKED` are both used.

Each endpoint in non-final status would only be processed by one pod at a time.

pilhuhn: Good question about a single job. I expect those provisioning/updating scenarios to be rather rare, so I don't know how much such a continuously running job makes sense. I am thinking that it makes a DB trip every 10s "for nothing". But then this is mostly a single select which is not too heavy. I don't know

gwenneg: The impact on the DB should be reasonable enough if the status field is indexed.

gwenneg:
I rebased this PR to fix conflicts resulting from the https://github.com/RedHatInsights/notifications-backend/pull/1077 merge.

common/src/main/java/com/redhat/cloud/notifications/models/Endpoint.java
```diff
@@ -201,6 +211,7 @@ public String toString() {
                 ", enabled=" + enabled +
                 ", type=" + compositeType.getType() +
                 ", subType=" + compositeType.getSubType() +
+                ", status=" + status + '\'' +
```
gwenneg, line 214: Do we need that `\'` char?

common/src/main/java/com/redhat/cloud/notifications/models/Endpoint.java
```diff
@@ -49,6 +49,8 @@
 
     private Boolean enabled = Boolean.FALSE;
 
+    private EndpointStatus status = EndpointStatus.UNKNOWN;
```
gwenneg, line 52: Does Hibernate ORM automatically converts the enum members into their ordinal? We use `EndpointTypeConverter` for the type but maybe we don't need it.

database/src/main/resources/db/migration/V1.46.0__endpoint_stauts.sql
```diff
@@ -0,0 +1,7 @@
+-- add a new status column
+
+alter table endpoints add column status integer ;
+
+update endpoints set status = 0 where status is null;
+
+create index status_idx ON endpoints (status);
```
gwenneg, line 7: I have two questions about this index:
- Should it be partial? I would say yes.
```suggestion
create index status_idx ON endpoints (status) where status not in (0, 5);
```
- Should it allow index-only scans? That's probably overkill.
```suggestion
create index status_idx ON endpoints (status, id, account_id, endpoint_type, endpoint_sub_type, enabled, name, description, created, updated) where status not in (0, 5);
```

common/src/main/java/com/redhat/cloud/notifications/models/EndpointStatus.java
```diff
@@ -0,0 +1,13 @@
...
+/**
+ *
+ */
+public enum EndpointStatus {
+    READY, // 0 , default for existing ones
+    UNKNOWN,
+    NEW,
+    PROVISIONING,
+    DELETING,
+    FAILED
```
gwenneg, line 12: `NEW` and `DELETING` aren't used yet, I suppose it's normal.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -181,17 +182,25 @@ public Endpoint createEndpoint(@Context SecurityContext sec, @NotNull @Valid End
...
+                    String processorId = setupOpenBridgeProcessor(endpoint, properties, processorName);
+                    endpoint.setStatus(EndpointStatus.PROVISIONING);
+
+                    // TODO find a better place for these, that should not be
+                    //       visible to users / OB actions
+                    //       See also CamelTypeProcessor#callOpenBridge
+                    //            and ReadyCheck#execute
+                    properties.getExtras().put(OB_PROCESSOR_ID, processorId);
+                } catch (WebApplicationException we) {
+                    we.printStackTrace();
```
gwenneg, line 195: This needs to be replaced with a proper log entry.

backend/src/main/java/com/redhat/cloud/notifications/routers/ReadyCheck.java
```diff
@@ -0,0 +1,86 @@
...
+import javax.persistence.Query;
+import javax.transaction.Transactional;
+import javax.ws.rs.WebApplicationException;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Scheduled task to check (OpenBridge) Endpoints
+ * for their state. Those endpoints can go from
+ * 'accepted' to 'provisioning' and then 'ready' or 'failed'.
```
gwenneg, line 26: There's no `accepted` in the current status enum 😃

backend/src/main/java/com/redhat/cloud/notifications/routers/ReadyCheck.java
```diff
@@ -0,0 +1,86 @@
...
+    Bridge bridge;
+
+    @Inject
+    BridgeAuth bridgeAuth;
+
+    String endpointQueryString = "SELECT e FROM Endpoint e " +
+            "WHERE e.compositeType.type = :type AND e.compositeType.subType IN (:subTypes) " +
+            "AND e.status NOT IN (:ready, :failed) ";
+
+    @Scheduled(concurrentExecution = Scheduled.ConcurrentExecution.SKIP, every = "10s")
```
gwenneg, line 52: The execution period could be configurable.

backend/src/main/java/com/redhat/cloud/notifications/routers/ReadyCheck.java
```diff
@@ -0,0 +1,86 @@
...
+
+        Query query = em.createQuery(endpointQueryString);
+        query.setParameter("ready", EndpointStatus.READY);
+        query.setParameter("failed", EndpointStatus.FAILED);
+        query.setParameter("type", EndpointType.CAMEL);
+        query.setParameter("subTypes", "slack");
+        List<Endpoint> endpoints = query.getResultList();
+        for (Endpoint ep : endpoints) {
+
+            if (!ep.getType().equals(EndpointType.CAMEL)) {
```
gwenneg, line 64: Will this condition ever be `true`?

gwenneg: If it stays here:
```suggestion
            if (ep.getType() != EndpointType.CAMEL) {
```

backend/src/main/java/com/redhat/cloud/notifications/routers/ReadyCheck.java
```diff
@@ -0,0 +1,86 @@
...
+                    String status = (String) processor.get("status");
+                    LOGGER.debugf("  Status reported by OB for processor %s : %s", processorId, status);
+                    if (status.equals("ready")) {
+                        ep.setStatus(EndpointStatus.READY);
+                    }
+                    if (status.equals("failed")) {
+                        ep.setStatus(EndpointStatus.FAILED);
+                    }
+                } catch (WebApplicationException wae) {
+                    LOGGER.warn("Getting data from OB failed: " + wae.getMessage());
```
gwenneg, line 80:
```suggestion
                    LOGGER.warn("Getting data from OB failed", wae);
```

backend/src/main/java/com/redhat/cloud/notifications/routers/ReadyCheck.java
```diff
@@ -0,0 +1,86 @@
...
+        for (Endpoint ep : endpoints) {
+
+            if (!ep.getType().equals(EndpointType.CAMEL)) {
+                LOGGER.warn("Not a camel endpoint, ignoring " + ep);
+            } else {
+                CamelProperties cp = em.find(CamelProperties.class, ep.getId()); // TODO Fetch in one go
+                String processorId = cp.getExtras().get("processorId");
+                try {
+                    Map<String, Object> processor = bridgeApiService.getProcessorById(bridge.getId(), processorId, bridgeAuth.getToken());
+                    String status = (String) processor.get("status");
```
gwenneg, line 71: Can this be `null`?

backend/src/main/java/com/redhat/cloud/notifications/routers/ReadyCheck.java
```diff
@@ -0,0 +1,86 @@
...
+
+    String endpointQueryString = "SELECT e FROM Endpoint e " +
+            "WHERE e.compositeType.type = :type AND e.compositeType.subType IN (:subTypes) " +
+            "AND e.status NOT IN (:ready, :failed) ";
+
+    @Scheduled(concurrentExecution = Scheduled.ConcurrentExecution.SKIP, every = "10s")
+    @Transactional
+    public void execute() {
+
+        Query query = em.createQuery(endpointQueryString);
```
gwenneg, line 56: ⚠️ Won't compile, needs import.
```suggestion
        TypedQuery<Endpoint> query = em.createQuery(endpointQueryString, Endpoint.class);
```

backend/src/main/java/com/redhat/cloud/notifications/routers/ReadyCheck.java
```diff
@@ -0,0 +1,86 @@
...
+        query.setParameter("ready", EndpointStatus.READY);
+        query.setParameter("failed", EndpointStatus.FAILED);
+        query.setParameter("type", EndpointType.CAMEL);
+        query.setParameter("subTypes", "slack");
```
gwenneg, line 60: We need Postgres' `SELECT FOR UPDATE` and `SKIP LOCKED` here to avoid concurrency issues when the query will be executed by multiple pods at the same time:
```suggestion
        query.setParameter("subTypes", "slack");
        query.setLockMode(LockModeType.PESSIMISTIC_WRITE);
        query.setHint("javax.persistence.lock.timeout", LockOptions.SKIP_LOCKED);
```
⚠️ My suggestion is untested, we need to make sure that Hibernate ORM adds `FOR UPDATE` and `SKIP LOCKED` in the generated SQL.

backend/pom.xml
```diff
@@ -79,6 +79,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-hibernate-orm</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-quartz</artifactId>
+        </dependency>
```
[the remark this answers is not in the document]
bennetelli: I just removed it and it's needed at least from a technical perspective. Otherwise the Scheduled stuff doesn't compile.
Maybe you know more than I do? Is there is something better than these quartz-job?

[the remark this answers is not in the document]
gwenneg: This should work:
```suggestion
        <dependency>
            <groupId>io.quarkus</groupId>
            <artifactId>quarkus-scheduler</artifactId>
        </dependency>
```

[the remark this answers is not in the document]
bennetelli: aah! nice!

gwenneg:
Rebased on master.

backend/src/main/java/com/redhat/cloud/notifications/routers/ReadyCheck.java
gwenneg: To be checked in logs.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointService.java
gwenneg: The code changed. Comment no longer relevant.

gwenneg:
@pilhuhn I applied and resolved most review suggestions. There are only a few of them still unresolved now. Let's try to finish this next week.

database/src/main/resources/db/migration/V1.46.0__endpoint_stauts.sql
```diff
@@ -0,0 +1,7 @@
+-- add a new status column
+
+alter table endpoints add column status integer ;
```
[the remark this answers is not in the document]
pilhuhn: Integers allow for easy < comparison implementing fallback that way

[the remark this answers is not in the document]
pilhuhn: Forget my (meanwhile deleted) comment. We could indeed use the string version here.

common/src/main/java/com/redhat/cloud/notifications/models/Endpoint.java
pilhuhn: Yes it does the ordinal conversion by default.

backend/src/main/java/com/redhat/cloud/notifications/routers/ReadyCheck.java
gwenneg: The Hibernate query looks good:
```
select
    endpoint0_.id as id1_9_,
    endpoint0_.created as created2_9_,
    endpoint0_.updated as updated3_9_,
    endpoint0_.account_id as account_4_9_,
    endpoint0_.endpoint_sub_type as endpoint5_9_,
    endpoint0_.endpoint_type as endpoint6_9_,
    endpoint0_.description as descript7_9_,
    endpoint0_.enabled as enabled8_9_,
    endpoint0_.name as name9_9_,
    endpoint0_.status as status10_9_
from
    endpoints endpoint0_
where
    endpoint0_.endpoint_type=?
    and (
        endpoint0_.endpoint_sub_type in (
            ?
        )
    )
    and (
        endpoint0_.status not in  (
            ? , ?
        )
    ) for update
        of endpoint0_ skip locked
```

gwenneg (approved):
LGTM

gwenneg:
I wonder if we should add a limit to the number of endpoints retrieved from the DB by the scheduled job.

That job locks the endpoints (preventing any update) and depends on an external host. It could potentially lock the endpoints for a long time if the query is unlimited and there are lots of endpoints matching the conditions.

gwenneg:
@pilhuhn I'll let you decide if you want to add a limit or not. The endpoints are locked at row level. If a user tries to edit an endpoint from the UI while this endpoint is locked by the scheduled job, then the UI will have to wait for the end of the scheduled job.

pilhuhn:
TBH for the moment I don't see this as an issue. Updates are relatively rare operations.

PR #1072, opened 2022-03-18
NOTIF-381 add paging

bennetelli:


bennetelli:
sometimes GitHub tries to be "smart" and lowercases the title..

gwenneg:
> sometimes GitHub tries to be "smart" and lowercases the title..

It happens when you create a PR that contains multiple commits.

bennetelli:
> > sometimes GitHub tries to be "smart" and lowercases the title..
> 
> It happens when you create a PR that contains multiple commits.

ah, thanks. I didn't notice it so far.

engine/src/main/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/ITUserRequest.java
```diff
@@ -8,7 +8,7 @@
     public ITUserRequestBy by;
     public Include include;
 
-    public ITUserRequest(String accountId, boolean adminsOnly) {
+    public ITUserRequest(String accountId, boolean adminsOnly, int pagingStart, int pagingEnd) {
```
gwenneg, line 11: Can we stick with the IT API names?
```suggestion
    public ITUserRequest(String accountId, boolean adminsOnly, int firstResult, int maxResults) {
```

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -82,10 +83,25 @@ public void initCounters() {
...
+
         List<User> users;
         if (retrieveUsersFromIt) {
-            final List<ITUserResponse> itUserResponses = itUserService.getUsers(accountId, adminsOnly);
-            users = transformToUser(accountId, itUserResponses, adminsOnly);
+            List<ITUserResponse> usersPaging;
+            List<ITUserResponse> usersTotal = new LinkedList<>();
+
+            int pagingStart = 0;
+            int pagingEnd = 10000;
```
gwenneg, line 93: I wonder if making this value configurable could be useful to fine-tune the value.

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -82,10 +83,25 @@ public void initCounters() {
...
+            do {
+                usersPaging = itUserService.getUsers(accountId, adminsOnly, pagingStart, pagingEnd);
+                usersTotal.addAll(usersPaging);
+
+                pagingStart = pagingEnd + 1;
```
gwenneg, line 98: This will skip one record if the `firstResult` base index is `0`. For example, after the first iteration, index `10000` will never be retrieved.

gwenneg (changes requested):

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -82,10 +83,25 @@ public void initCounters() {
...
         Timer.Sample getUsersTotalTimer = Timer.start(meterRegistry);
+
         List<User> users;
         if (retrieveUsersFromIt) {
-            final List<ITUserResponse> itUserResponses = itUserService.getUsers(accountId, adminsOnly);
-            users = transformToUser(accountId, itUserResponses, adminsOnly);
+            List<ITUserResponse> usersPaging;
+            List<ITUserResponse> usersTotal = new LinkedList<>();
+
+            int pagingStart = 0;
```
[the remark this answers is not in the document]
bennetelli: Yes, it starts at 0. Here is the paging snippet from the request json: "withPaging": { "firstResultIndex": 0, "maxResults": 10000 }

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -82,10 +83,25 @@ public void initCounters() {
...
+            int pagingEnd = 10000;
+            do {
+                usersPaging = itUserService.getUsers(accountId, adminsOnly, pagingStart, pagingEnd);
+                usersTotal.addAll(usersPaging);
+
+                pagingStart = pagingEnd + 1;
+                pagingEnd = pagingEnd + 10;
```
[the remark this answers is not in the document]
bennetelli: But yea, 10 is way too less.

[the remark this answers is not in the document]
josejulio: Wasn't this already a parameter? (the increment)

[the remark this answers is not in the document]
gwenneg: Then we need [this](https://github.com/RedHatInsights/notifications-backend/pull/1072#discussion_r829854672) 😃

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
bennetelli: hm. we can. i don't think we definitely need to do this as we will have to do a deployment each time we change it anyway. 
as long as everyone of us is able to understand the code, I would also be fine with this solution, but also with a constant or something else. Don't have a strong opinion here

gwenneg: Deploying a configuration change has a much lower impact than deploying a new version of the app.

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
gwenneg: The variables names are misleading, that's why I made [that comment](https://github.com/RedHatInsights/notifications-backend/pull/1072#discussion_r829853893) which hasn't been applied in all methods BTW.

`pagingEnd` is not a paging end, it's a max results value. If you start at index 0 and ask for 10 results, the IT service will return indexes from `0` to `9`, not from `0` to `10`.

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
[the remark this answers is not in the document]
gwenneg: I just realized my initial comment here was based on the wrong asumption that `pagingEnd` is a paging end while it's not. That value should stay the same for each iteration.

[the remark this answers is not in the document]
bennetelli: now I am confused :-D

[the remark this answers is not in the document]
bennetelli: just renamed it to "firstResult" and "maxResults" everywhere. Maybe that's less confusing.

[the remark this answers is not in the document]
theute: The IT Service API needs:
Index of the first element & the paging size
Not
Index of the first element & index of the last element

So for a page of 100 will be:
(0, 100) --> returns element from 0 to 99
(100, 100)
(200, 100)
(300, 100)
...

The terminology here is correct:
https://github.com/RedHatInsights/notifications-backend/pull/1072/files#diff-b5ae501ce5cb4a5385ebc94b9b971aa33f01fea5afd1fc735f5b64a560c05af0R11
`public ITUserRequest(String accountId, boolean adminsOnly, int firstResult, int maxResults) {`

But not here (pagingEnd is not the correct wording, which leads to the confusion):
https://github.com/RedHatInsights/notifications-backend/pull/1072/files#diff-988f3c9b3009196abbe24f8e85b001aa8c3a60d953533c77d792ea19aae04443R19
`return itUserService.getUsers(new ITUserRequest(accountId, adminsOnly, pagingStart, pagingEnd));`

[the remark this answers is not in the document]
theute: Yes, not the question is why maxResults value would change on iterations ? you always want the same number of results in your page (except the last page)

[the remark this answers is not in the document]
gwenneg: `maxResults` has to be constant and ideally configurable.

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -82,10 +85,27 @@ public void initCounters() {
...
+
+            int firstResult = 0;
+            int maxResults = 10000;
+            int expectedUsersCount = firstResult + maxResults;
+
+            do {
+                usersPaging = itUserService.getUsers(accountId, adminsOnly, firstResult, maxResults);
+                usersTotal.addAll(usersPaging);
+
+                firstResult = maxResults + 1;
```
gwenneg, line 102: This should be increased by the value of `maxResults`

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -77,20 +93,47 @@ public void initCounters() {
...
+
+            do {
+                int finalFirstResult = firstResult;
+                usersPaging = retryOnItError(() -> itUserService.getUsers(accountId, adminsOnly, finalFirstResult, maxResultsPerPage));
+                usersTotal.addAll(usersPaging);
+
+                firstResult += maxResultsPerPage;
+            } while (usersPaging.size() == maxResultsPerPage);
+
+            getUsersTotalTimer.stop(meterRegistry.timer("rbac.get-users.total", "accountId", accountId, "users", String.valueOf(usersTotal.size())));
```
gwenneg, line 129: I just noticed this line is a duplicate of the same instruction at the end of this method:
```suggestion
```
The `transformToUser` execution time is close to zero so it doesn't matter if it is counted by the timer and we will refactor things anyway when the RBAC implementation will be removed.

engine/src/test/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/ITUserRequestTest.java
```diff
@@ -8,7 +8,7 @@
 
 class ITUserRequestTest {
 
-    private final ITUserRequest testee = new ITUserRequest("someAccountId", true);
+    private final ITUserRequest testee = new ITUserRequest("someAccountId", true, 0, 101);
```
gwenneg, line 11: The new params are not tested here.

engine/src/test/java/com/redhat/cloud/notifications/recipients/itservice/ITUserServiceTest.java
```diff
@@ -57,9 +57,12 @@ void shouldPickPrimaryEMailAsUsersEmail() {
         List<ITUserResponse> itUserResponses = List.of(itUserResponse);
 
         RbacRecipientUsersProvider rbacRecipientUsersProvider = new RbacRecipientUsersProvider(itUserServiceWrapper, new SimpleMeterRegistry());
```
gwenneg, line 59: `RbacRecipientUsersProvider` is instantiated with `new` instead of being injected as a CDI bean. As a consequence, the `@PostConstruct` method is never called and the retry policy is never initialized, which leads to a test failure.

engine/src/main/resources/application.properties
```diff
@@ -73,9 +73,9 @@ quarkus.rest-client.rbac-s2s.read-timeout=120000
 
 # IT User service
 # these entries are needed because of a bug in quarkus: https://github.com/quarkusio/quarkus/issues/8384
-quarkus.rest-client."com.redhat.cloud.notifications.recipients.itservice.ITUserService".url=${QUARKUS_REST_CLIENT_IT_S2S_URL:FILL_ME}/v2
-quarkus.rest-client."com.redhat.cloud.notifications.recipients.itservice.ITUserService".key-store=file:${QUARKUS_HTTP_SSL_CERTIFICATE_KEY_STORE_FILE:FILL_ME}
-quarkus.rest-client."com.redhat.cloud.notifications.recipients.itservice.ITUserService".key-store-password=${QUARKUS_HTTP_SSL_CERTIFICATE_KEY_STORE_PASSWORD:FILL_ME}
+%prod.quarkus.rest-client."com.redhat.cloud.notifications.recipients.itservice.ITUserService".url=${QUARKUS_REST_CLIENT_IT_S2S_URL:FILL_ME}/v2
+%prod.quarkus.rest-client."com.redhat.cloud.notifications.recipients.itservice.ITUserService".key-store=file:${QUARKUS_HTTP_SSL_CERTIFICATE_KEY_STORE_FILE:FILL_ME}
+%prod.quarkus.rest-client."com.redhat.cloud.notifications.recipients.itservice.ITUserService".key-store-password=${QUARKUS_HTTP_SSL_CERTIFICATE_KEY_STORE_PASSWORD:FILL_ME}
```
gwenneg, line 78: These config values were causing issues during tests so they will only be active with the `prod` profile from now on.

gwenneg:
@josejulio I made too many changes to approve that PR, we need an extra pair of eyes 😃 Could you please have a look at this?

gwenneg (approved):

gwenneg:
The GH actions won't start, I don't know why... Probably another GitHub breakdown.

engine/src/test/java/com/redhat/cloud/notifications/recipients/itservice/pojo/request/ITUserRequestTest.java
josejulio: Are these tested somewhere else?

gwenneg:
Pushed again (no change) to trigger GitHub actions.

theute:
What do we do with this one ? Ready to merge and push to stage ?

gwenneg:
I'd like @josejulio approval before we merge.

bennetelli:
interesting. For me it wasn't working with two properties having a RestClient annotation.

josejulio (approved):

PR #1075, opened 2022-03-18
Fixes for internal RBAC

josejulio:


.rhcicd/clowdapp-backend.yaml
```diff
@@ -315,6 +317,9 @@ parameters:
 - name: INTERNAL_RBAC_ENABLED
   description: Enable Internal RBAC
   value: "false"
+- name: INTERNAL_RBAC_ADMIN
+  description: Internal admin group
+  value: "crc-notifications-team"
```
josejulio, line 322: I'm adding the parameter to be able to test on stage with a non admin user.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/models/AddAccessRequest.java
```diff
@@ -1,7 +1,11 @@
 package com.redhat.cloud.notifications.routers.internal.models;
 
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+
 import java.util.UUID;
 
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
```
josejulio, line 8: `applicationId` as being received as such. Changed to receive it as `application_id`

backend/src/main/java/com/redhat/cloud/notifications/models/InternalRoleAccess.java
```diff
@@ -56,6 +57,7 @@ public void setApplicationId(UUID applicationId) {
         this.applicationId = applicationId;
     }
 
+    @JsonIgnore
     public String getInternalRole() {
         return InternalRoleAccess.getInternalRole(this.role);
```
josejulio, line 62: this was leaking in the response

.rhcicd/clowdapp-backend.yaml
gwenneg: This may be deployed on prod unintentionally, is that OK to commit that value?

josejulio: That's the default admin group. It should match what we have already in the properties file

gwenneg (approved):
LGTM

PR #1077, opened 2022-03-18
Rename repositories and JAX-RS resources

gwenneg:
In the JAX-RS world, a class that exposes REST endpoints is called a `resource`.

In the _modern_ design patterns world, what we do with our current `*Resources` classes matches the usual definition of the `Repository` pattern.

So let's give these classes the name they deserve 😁

ℹ️ I will not disrupt any exiting PR with conflicts that would result from merging this.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/ApplicationRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/ApplicationRepository.java
@@ -1,5 +1,6 @@
-package com.redhat.cloud.notifications.db;
+package com.redhat.cloud.notifications.db.repositories;
 
+import com.redhat.cloud.notifications.db.Query;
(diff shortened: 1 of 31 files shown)


gwenneg:
CodeQL is unhappy because of false positives. They resurfaced because of the names changes.

gwenneg:
False positive confirmed and CodeQL alerts marked as such.

We are protected against malicious inputs by [this](https://github.com/RedHatInsights/notifications-backend/blob/544609d72d7edad93e3d84ad152cf6662af18cc6/backend/src/main/java/com/redhat/cloud/notifications/db/Query.java#L13) and [that](https://github.com/RedHatInsights/notifications-backend/blob/544609d72d7edad93e3d84ad152cf6662af18cc6/backend/src/main/java/com/redhat/cloud/notifications/routers/EventService.java#L43).

PR #1078, opened 2022-03-18
Introduce warning alert on admin ui

josejulio:
- Will warn when using prod or when the environment could not be found

A little heads up to let us know that we are in prod.

Screenshot_2022-03-18_11-43-31


aferd (approved):
lgtm

PR #1081, opened 2022-03-21
Template for Advisor deactivated recommendation notifications

mhuth:
- https://issues.redhat.com/browse/ADVISOR-1989 for ADVISOR-1990

### DIFF
--- a/engine/src/main/resources/templates/Advisor/deactivatedRecommendationInstantEmailBody.html
+++ b/engine/src/main/resources/templates/Advisor/deactivatedRecommendationInstantEmailBody.html
@@ -122,7 +122,7 @@
             <tr>
                 <td class="rh-content__block">
                     <p>
-                        Hi {{user.first_name}},
+                        Hi {user.firstName},
                         <br>
                         <br>
                         Certain recommendations that were affecting your systems have been recently deactivated and hence are no longer affecting your systems.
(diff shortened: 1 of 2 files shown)


mhuth:
My only question is will this line work if I'm not actually sending any 'user' data in the event?
https://github.com/RedHatInsights/notifications-backend/pull/1081/files#diff-abcbebc6eb4492b8d850a2e9c3b084b481ed0a0d8a6ac52bdfca42c0d455207fR125

I'm only sending account information in the event.  Does notifications enumerate the users in an account?

engine/src/main/resources/templates/Advisor/deactivatedRecommendationInstantEmailBody.html
```diff
@@ -0,0 +1,140 @@
...
+<!-- Body section -->
+<tr>
+    <td class="rh-body">
+        <table class="rh-content" role="presentation" align="center" border="0" cellpadding="0" cellspacing="0" width="100%">
+            <tbody>
+            <!-- Content block -->
+            <tr>
+                <td class="rh-content__block">
+                    <p>
+                        Hi {{user.first_name}},
```
gwenneg, line 125: Yes, the user first name will be available because we retrieve that information from RBAC using the account number we receive in the Kafka message. However, the template needs a small change:
```suggestion
                        Hi {user.firstName},
```

gwenneg (changes requested):

josejulio (approved):

PR #1082, opened 2022-03-21
NOTIF-482 Allow devs to delete their own app

josejulio:
Allows apps to delete their own applications.

I'm unsure if we should allow this permission, but lets try on stage first for some time.

### DIFF
--- a/backend/src/test/java/com/redhat/cloud/notifications/routers/InternalServiceTest.java
+++ b/backend/src/test/java/com/redhat/cloud/notifications/routers/InternalServiceTest.java
@@ -389,6 +389,7 @@ public class InternalServiceTest extends DbIsolatedTest {
         assertTrue(responseBody.matches("^[0-9a-f]{7}$"));
     }
 
+    @Test
     void appUserAccessTest() {
         String appRole = "app-admin";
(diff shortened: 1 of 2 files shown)


backend/src/test/java/com/redhat/cloud/notifications/routers/InternalServiceTest.java
```diff
@@ -389,6 +389,7 @@ void testVersion() {
         assertTrue(responseBody.matches("^[0-9a-f]{7}$"));
     }
 
+    @Test
```
josejulio, line 392: Can't believe I missed this

gwenneg: 😱

gwenneg:
> I'm unsure if we should allow this permission, but lets try on stage first for some time.

Yeah... I'm not sure this is a good idea unless the application deletion is a soft delete (based on a boolean in the DB).

josejulio:
I'll close it for now - will send a separate PR about the missing `@Test` annotation. We can decide later how we want to handle the soft deletion.

PR #1086, opened 2022-03-21
POL-383 Expose environment in templates

josejulio:
Exposes environment variable and a function to build an url depending the environment.

It prepends the prod/stage url for those environments and returns as is on other envs.

Screenshot_2022-03-21_13-53-56


engine/src/main/java/com/redhat/cloud/notifications/templates/models/ServerConfig.java
```diff
@@ -0,0 +1,33 @@
...
+        switch (environment) {
+            case "prod":
+                base = "https://console.redhat.com";
+                break;
+            case "stage":
+                base = "https://console.stage.redhat.com";
+                break;
+            default:
+                base = "";
+        }
```
gwenneg, line 29: Let's make Sonar happy:
```suggestion
            break;
        }
```

engine/src/main/java/com/redhat/cloud/notifications/templates/models/ServerConfig.java
```diff
@@ -0,0 +1,33 @@
+package com.redhat.cloud.notifications.templates.models;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+import javax.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped
+public class ServerConfig {
+
+    @ConfigProperty(name = "env", defaultValue = "local-dev")
```
gwenneg, line 10: Shouldn't this be
```suggestion
    @ConfigProperty(name = "env.name", defaultValue = "local-dev")
```
to match the existing `ENV_NAME` environment variable that is already used on stage/prod?

engine/src/main/java/com/redhat/cloud/notifications/templates/models/ServerConfig.java
```diff
@@ -0,0 +1,33 @@
...
+                base = "https://console.redhat.com";
+                break;
+            case "stage":
+                base = "https://console.stage.redhat.com";
+                break;
+            default:
+                base = "";
+        }
+
+        return base + path;
```
gwenneg, line 31: What if `path` does not start with a `/`? Should we check that and add it if missing here?

engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateService.java
```diff
@@ -22,6 +26,7 @@ public String renderTemplate(User user, Action action, TemplateInstance template
         return templateInstance
                 .data("action", action)
                 .data("user", user)
+                .data("config", config)
```
gwenneg, line 29: `config` is a bit of a generic term that does not really carry the meaning of the available features behind it. I wonder if we could find something more "precise".

engine/src/main/java/com/redhat/cloud/notifications/templates/models/ServerConfig.java
josejulio: :facepalm: thanks!

engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateService.java
josejulio: I wonder that too - other option could be just: `helpers`

gwenneg: WDYT of simply using `environment`? The methods could then be named `environment#name` and `environment#absolueUrl`?

josejulio: Changed to `environment` but still prefering the short `environment.url`

gwenneg (approved):
LGTM

PR #1089, opened 2022-03-22
NOTIF-428 Add query for accounts with disabled Splunk connector

bennetelli:


.rhcicd/clowdapp-backend.yaml
```diff
@@ -121,12 +121,10 @@ objects:
               name: notifications-ephemeral-data
               key: ephemeral_data.json
               optional: true
-        - name: QUARKUS_REST_CLIENT_RBAC_AUTHENTICATION_READ_TIMEOUT
+        - name: RBAC_AUTHENTICATION_MP_REST_READTIMEOUT
```
josejulio, line 124: This is probably not intented - I recall we had to hadd QUARKUS_* ?

gwenneg: Yeah `*MP_REST*` is no longer working.

.rhcicd/clowdapp-backend.yaml
```diff
@@ -153,9 +151,9 @@ objects:
           value: ${OB_TOKEN_USER}
         - name: OB_TOKEN_PASS
           value: ${OB_TOKEN_PASS}
-        - name: QUARKUS_REST_CLIENT_OB_URL
+        - name: OB_MP_REST_URL
           value: ${OB_MP_REST_URL}
-        - name: QUARKUS_REST_CLIENT_KC_URL
+        - name: OB_KC_REST_URL
```
gwenneg, line 156:
```suggestion
        - name: QUARKUS_REST_CLIENT_KC_URL
```

.rhcicd/clowdapp-backend.yaml
```diff
@@ -317,9 +321,6 @@ parameters:
 - name: INTERNAL_RBAC_ENABLED
   description: Enable Internal RBAC
   value: "false"
-- name: INTERNAL_RBAC_ADMIN
```
gwenneg, line 320: This shouldn't be removed.

.rhcicd/clowdapp-backend.yaml
```diff
@@ -240,6 +238,12 @@ objects:
           FROM "applications" 
           INNER JOIN "endpoint_email_subscriptions" ON "applications"."id"="endpoint_email_subscriptions"."application_id" 
           GROUP BY "applications"."id" HAVING COUNT("applications"."id") > 1;
+      - prefix: insights/notifications/accounts
+        query: >-
+          SELECT
+          COUNT(distinct account_id) 
+          FROM "endpoints"
+          WHERE "endpoint_type" != 3;
```
gwenneg, line 246: What is the exact goal of this query?

.rhcicd/clowdapp-backend.yaml
```diff
@@ -121,12 +121,10 @@ objects:
               name: notifications-ephemeral-data
               key: ephemeral_data.json
               optional: true
-        - name: QUARKUS_REST_CLIENT_RBAC_AUTHENTICATION_READ_TIMEOUT
+        - name: RBAC_AUTHENTICATION_MP_REST_READTIMEOUT
           value: ${RBAC_AUTHENTICATION_READ_TIMEOUT}
         - name: INTERNAL_RBAC_ENABLED
           value: ${INTERNAL_RBAC_ENABLED}
-        - name: INTERNAL_ADMIN_ROLE
-          value: ${INTERNAL_RBAC_ADMIN}
```
josejulio, line 129: We still need this param over here

.rhcicd/clowdapp-backend.yaml
josejulio: Is possible to put some sort of documentation / comments to know what is each thing?

bennetelli: I can take a look at it again. But I don't think it's possible.

It's the answer for "Number of accounts that have disabled Splunk connector".

.rhcicd/clowdapp-backend.yaml
bennetelli: Let me check. I did not want to add this.. I didn't even want to change this..

.rhcicd/clowdapp-backend.yaml
bennetelli: Yea. I don't know where this came from.. will add it again

.rhcicd/clowdapp-backend.yaml
gwenneg: The current query will return the number of accounts that own at least one endpoint of any type but `CAMEL`. FYI, `splunk` is a subtype of the `CAMEL` endpoint type.

I would translate my understanding of the goal into that kind of SQL:
```suggestion
          SELECT
          COUNT(distinct account_id) 
          FROM endpoints
          WHERE endpoint_type = 3 AND endpoint_sub_type = 'splunk' AND enabled IS FALSE;
```

josejulio (approved):

PR #1090, opened 2022-03-22
Add email templates for rbac

astrozzc:
Add Java binding;
Add 3 templates: role, group, platform default event templates


engine/src/main/java/com/redhat/cloud/notifications/templates/Rbac.java
```diff
@@ -0,0 +1,93 @@
...
+
+// Name needs to be "Rbac" to read templates from resources/templates/Rbac
+public class Rbac implements EmailTemplate {
+
+    @Override
+    public TemplateInstance getTitle(String eventType, EmailSubscriptionType type) {
+        if (type == EmailSubscriptionType.INSTANT) {
+            switch (eventType) {
+                case "rh-new-role-available":
+                case "rh-platform-defaul-role-upated":
```
josejulio, line 15: Should be updated? 
```suggestion
                case "rh-platform-defaul-role-updated":
```

josejulio (changes requested):
Some possible typos i found

engine/src/main/resources/templates/Rbac/platformGroup2customEmailBody.html
```diff
@@ -0,0 +1,52 @@
...
+            <!-- Content block -->
+            <tr>
+                <td class="rh-content__block">
+                    <p>
+                        Hi,
+                        <br>
+                        <br>
+                        Platform default group is modified and the user is responsible for managing it from now on.
+                        <br>
+                        Check the custom platform defualt group here: <a class="rh-url" href="https://console.redhat.com/api/rbac/v1/groups/{action.events[0].payload.uuid}/"><b>{action.events[0].payload.name}</b></a>{/if}
```
josejulio, line 40:
```suggestion
                        Check the custom platform default group here: <a class="rh-url" href="https://console.redhat.com/api/rbac/v1/groups/{action.events[0].payload.uuid}/"><b>{action.events[0].payload.name}</b></a>{/if}
```

josejulio: Is this URL pointing directly to the API intended?

engine/src/main/java/com/redhat/cloud/notifications/templates/Rbac.java
astrozzc: Thanks for catching these!

engine/src/main/resources/templates/Rbac/platformGroup2customEmailBody.html
astrozzc: Yea, so it could show the customer what it looks like. Any concerns with it?

astrozzc: I updated with UI link, which might be better. Thanks!

josejulio: That's what I meant to ask here, thanks!

josejulio:
@astrozzc Could you add your template here as well? https://github.com/RedHatInsights/notifications-backend/blob/ebf8c7ad63347785d7d56d9bf3f8fd73aff9c790/engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateFactory.java#L25

Thanks!

astrozzc:
> @astrozzc Could you add your template here as well?
> 
> https://github.com/RedHatInsights/notifications-backend/blob/ebf8c7ad63347785d7d56d9bf3f8fd73aff9c790/engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateFactory.java#L25
> 
> Thanks!

Sure, updated 😉

josejulio (approved):
lgtm

PR #1094, opened 2022-03-23
NOTIF-523 make breadcrumb into link

aferd:


admin-console/src/main/webapp/src/components/Wrappers/BreadCrumbLinkItem.tsx
```diff
@@ -0,0 +1,18 @@
+import { BreadcrumbItem, BreadcrumbItemProps } from '@patternfly/react-core';
+import * as React from 'react';
+
+import { OuiaComponentProps, withoutOuiaProps } from '../../utils/Ouia';
```
josejulio, line 4: We can remove the Ouia stuff for now.

aferd: removing it breaks the link?

josejulio: Yes. You might need to remove the usage of it. Shouldn't impact on the functionality of the component

aferd: I did remove the usage.. this is what happens...

https://user-images.githubusercontent.com/86322239/159773766-1eeed873-0654-4855-9fb6-663f4de0e66f.mov

admin-console/src/main/webapp/src/components/Wrappers/BreadCrumbLinkItem.tsx
```diff
@@ -0,0 +1,18 @@
...
+export const BreadcrumbLinkItem: React.FunctionComponent<BreadcrumbLinkItemProps> = (props) => {
+    return (
+        <BreadcrumbItem
+            { ...withoutOuiaProps(props) }
+            component={ LinkAdapter }
+        >
+            { props.children }
+        </BreadcrumbItem>
+    );
+};
```
josejulio, line 18: Try it like this:

`withoutOuiaProps` removes the specific ouia properties from the props. But we still want the rest of the properties, so we still need to pass the `props` here.

[Ouia](https://ouia.readthedocs.io/en/latest/) is a specification used for testing, it adds some extra properties to the DOM to specify component types and make it easier for automation software to find and manage them. 

```suggestion
import { LinkAdapter } from './LinkAdapter';

type BreadcrumbLinkItemProps = Omit<BreadcrumbItemProps, 'component'>;

export const BreadcrumbLinkItem: React.FunctionComponent<BreadcrumbLinkItemProps> = (props) => {
    return (
        <BreadcrumbItem
            { ...props }
            component={ LinkAdapter }
        >
            { props.children }
        </BreadcrumbItem>
    );
};
```

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -43,7 +45,7 @@ export const ApplicationPage: React.FunctionComponent = () => {
             return applicationTypesQuery.payload.value.bundleId;
         }
 
-        return undefined;
+        return 'bundle';
```
josejulio, line 48: Why is this needed? If is because the link, perhaps you could change it to an empty string in the link.

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -138,11 +140,12 @@ export const ApplicationPage: React.FunctionComponent = () => {
                 <Title headingLevel="h1">
                     <Breadcrumb>
                         <BreadcrumbItem target='#'> Bundles </BreadcrumbItem>
-                        <BreadcrumbItem target='#'>{ bundle ? bundle.display_name : <Spinner /> }
+                        <BreadcrumbLinkItem to={ linkTo.bundle(getBundleId) }>
```
josejulio, line 143:
```suggestion
                        <BreadcrumbLinkItem to={ linkTo.bundle(getBundleId ?? '') }>
```

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
aferd: It was because of the link. Will change to your suggestion below.

josejulio (approved):

PR #1100, opened 2022-03-25
Fix some issues and add some testing.

pilhuhn:


backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -186,11 +189,17 @@ public Endpoint createEndpoint(@Context SecurityContext sec, @NotNull @Valid End
...
+            if (endpointSubType != null && endpointSubType.equals("slack")) {
                 CamelProperties properties = endpoint.getProperties(CamelProperties.class);
                 String processorName = "p-" + endpoint.getAccountId() + "-" + UUID.randomUUID();
                 properties.getExtras().put(OB_PROCESSOR_NAME, processorName);
-                String processorId = setupOpenBridgeProcessor(endpoint, properties, processorName);
+                String processorId = null;
+                try {
+                    processorId = setupOpenBridgeProcessor(endpoint, properties, processorName);
+                } catch (Exception e) {
+                    LOGGER.warn("Processor setup failed: " + e.getMessage());
```
gwenneg, line 200:
```suggestion
                    LOGGER.warn("Processor setup failed", e);
```

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -186,11 +189,17 @@ public Endpoint createEndpoint(@Context SecurityContext sec, @NotNull @Valid End
...
+                    throw new InternalServerErrorException(e.getMessage());
```
gwenneg, line 201: `e.getMessage()` will be forwarded to the client, is that safe?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -252,7 +261,14 @@ public Response deleteEndpoint(@Context SecurityContext sec, @PathParam("id") UU
                     // Special case wrt OpenBridge
                     if (e.getSubType().equals("slack")) {
                         String processorId = cp.getExtras().get(OB_PROCESSOR_ID);
-                        bridgeApiService.deleteProcessor(bridge.getId(), processorId, bridgeAuth.getToken());
+                        if (processorId != null) { // Should not be null under normal operations.
```
gwenneg, line 264: Should we log something if it is `null`?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -252,7 +261,14 @@ public Response deleteEndpoint(@Context SecurityContext sec, @PathParam("id") UU
...
+                            try {
+                                bridgeApiService.deleteProcessor(bridge.getId(), processorId, bridgeAuth.getToken());
+                            } catch (Exception ex) {
+                                LOGGER.warn("Removal of OB processor failed:" + ex.getMessage());
```
gwenneg, line 268:
```suggestion
                                LOGGER.warn("Removal of OB processor failed", ex);
```

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
```diff
@@ -358,6 +366,145 @@ void addCamelEndpoint() {
...
+            assertEquals("#notifications", channel);
+
+        } finally {
+
+            given()
+                    .header(identityHeader)
+                    .when().delete("/endpoints/" + id)
+                    .then()
+                    .statusCode(204)
+                    .extract().body().asString();
```
gwenneg, line 500:
```suggestion
                    .statusCode(204);
```

common/src/main/java/com/redhat/cloud/notifications/openbridge/Bridge.java
```diff
@@ -4,30 +4,33 @@
...
+        return endpoint;
     }
 
     @Override
     public String toString() {
         final StringBuilder sb = new StringBuilder("Bridge{");
         sb.append("id='").append(id).append('\'');
         sb.append(", name='").append(name).append('\'');
-        sb.append(", eventsEndpoint='").append(eventsEndpoint).append('\'');
+        sb.append(", eventsEndpoint='").append(endpoint).append('\'');
```
gwenneg, line 33:
```suggestion
        sb.append(", endpoint='").append(endpoint).append('\'');
```

common/src/main/java/com/redhat/cloud/notifications/openbridge/BridgeHelper.java
```diff
@@ -75,7 +80,13 @@ public BridgeAuth getAuthToken() {
             return new BridgeAuth("- OB not enabled token -");
         }
 
-        BridgeAuth ba = new BridgeAuth(getAuthTokenInternal());
+        BridgeAuth ba = null;
+        try {
+            ba = new BridgeAuth(getAuthTokenInternal());
+        } catch (Exception e) {
+            LOGGER.warn("Failed to get an auth token: " + e.getMessage());
```
gwenneg, line 87:
```suggestion
            LOGGER.warn("Failed to get an auth token", e);
```

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
```diff
@@ -358,6 +366,145 @@ void addCamelEndpoint() {
...
+        CamelProperties cAttr = new CamelProperties();
+        cAttr.setDisableSslVerification(false);
+        cAttr.setUrl(String.format("https://%s", mockServerConfig.getRunningAddress()));
+        cAttr.setBasicAuthentication(new BasicAuthentication("testuser", "secret"));
+        Map<String, String> extras = new HashMap<>();
+        extras.put("template", "11");
+        cAttr.setExtras(extras);
+
+        Endpoint ep = new Endpoint();
+        ep.setType(EndpointType.CAMEL);
```
gwenneg, line 388: Is it bogus because it doesn't have a subtype?

josejulio: subtypes are required for `EndpointType.Camel` - I wonder if we can enforce that now :thinking:

gwenneg:
General question about all the HTTP calls to OB/KC: should we include retries? When the reactiveness was removed recently, we started retrying with `failsafe`. Here's an example: [part 1](https://github.com/RedHatInsights/notifications-backend/blob/d3f68a52802d071b355aecbaa6011e467b9e142d/engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java#L78-L82) and [part 2](https://github.com/RedHatInsights/notifications-backend/blob/d3f68a52802d071b355aecbaa6011e467b9e142d/engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java#L128). This can be done in a subsequent PR.

pilhuhn:
Yes, I was thinking about that as well. I think sprinkling some MP annotations should be enough :)

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
pilhuhn: Yes

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
pilhuhn: I don't think the full trace is needed, message should be enough

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
pilhuhn: I don't think the full trace is needed, message should be enough

common/src/main/java/com/redhat/cloud/notifications/openbridge/BridgeHelper.java
pilhuhn: I don't think the full trace is needed, message should be enough

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
gwenneg: I would agree if that line logged a specific exception but we're catching many different kinds of exceptions here, most of them resulting from an HTTP call failure. Also, logging the message instead of the exception itself not only discards the stacktrace but also the exception type.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
gwenneg: Same answer as [here](https://github.com/RedHatInsights/notifications-backend/pull/1100#discussion_r836177503).

common/src/main/java/com/redhat/cloud/notifications/openbridge/BridgeHelper.java
gwenneg: Same answer as [here](https://github.com/RedHatInsights/notifications-backend/pull/1100#discussion_r836177503).

gwenneg (approved):
I'd prefer if the full exceptions were logged but feel free to merge as is if you disagree 😃

PR #1101, opened 2022-03-25
NOTIF-528 Add probes

gburges:


gburges:
This to solve https://issues.redhat.com/browse/NOTIF-528

theute:
Just a note: 
I guess it won't hurt to add those, but AFAIK all those are very short lived cron jobs.
So the liveness check won't run in normal situations. (Maybe it would help if it gets stuck, and then I guess a good thing)

gburges:
> Just a note: I guess it won't hurt to add those, but AFAIK all those are very short lived cron jobs. So the liveness check won't run in normal situations. (Maybe it would help if it gets stuck, and then I guess a good thing)

I'll scale back the initial check time, would you say 3-5 seconds would be enough?

theute:
Looking at logs of one aggregator job, it started in 12s (won't accept HTTP requests before) and stopped after 17s:
(I don't know if we have a /ready and /live)
```
2022-03-24 00:00:06,437 INFO [com.red.clo.com.clo.con.ClowderConfigSourceFactory] (main) Using ClowderConfigSource with config at /cdapp/cdappconfig.json
[...]
2022-03-24 00:00:14,829 INFO [io.quarkus] (main) notifications-aggregator 1.0.0-SNAPSHOT on JVM (powered by Quarkus 2.7.5.Final) started in 12.176s.
[...]
2022-03-24 00:00:17,536 INFO [io.quarkus] (main) notifications-aggregator stopped in 0.210s
```
And for the dbcleaner, it's a `quay.io/cloudservices/postgresql-rds:12-1` job, I also don't know if it comes with liveness and health check probes, but I would be surprised if there is anything on port 8080, it's just a psql client I think.
It also takes <1min

```
Timing is on.
INFO: 2022-03-23 01:00:07.796534+00 Event log purge starting. Entries older than 31 days will be deleted.
INFO: 2022-03-23 01:00:07.796534+00 Event log purge ended. 94730 entries were deleted from the database.
CALL
Time: 9372.802 ms (00:09.373)
VACUUM
Time: 12925.841 ms (00:12.926)
VACUUM
Time: 167.699 ms
INFO: 2022-03-23 01:00:30.263036+00 Kafka messages purge starting. Entries older than 1 day will be deleted.
INFO: 2022-03-23 01:00:30.263036+00 Kafka messages purge ended. 19272 entries were deleted from the database.
CALL
Time: 23.171 ms
VACUUM
Time: 12.161 ms
```

gwenneg:
`db-cleaner` does not listen to HTTP traffic indeed. It only calls Postgres from the shell to run a few SQL queries, then it is terminated.

gwenneg:
I don't think `aggregator` currently accepts any HTTP traffic. This could be easily changed though.

theute:
I don't believe we need readiness probes at all (since those pods aren't made to receive traffic).
For liveness probes:
 - SQL job: I don't know if there is a command to check the status of the running call...  `pg_isready` seems useful for the server but here we're just running a few SQL queries sequentially.
 - Aggregator job: I don't even think we would want Kube to restart again and again the pod if it fails

theute (approved):

PR #1104, opened 2022-03-28
NOTIF-381 Add correct admin mapping

bennetelli:


engine/src/test/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProviderTest.java
```diff
@@ -49,6 +52,59 @@
...
+
+        List<ITUserResponse> itUserResponses = createTestResponse("portal_download");
+
+        final List<User> users = testee.transformToUser(itUserResponses);
+
+        assertFalse(users.get(0).isAdmin());
+    }
+
+    private List<ITUserResponse> createTestResponse(String permissionCode) {
+        List<ITUserResponse> itUserResponses = new LinkedList<>();
```
gwenneg, line 78: Please do not use `LinkedList` as the default implementation when you need to create a `List` instance.

bennetelli: We build the second largest ecommerce shop in europe and didnt have any significant performance issues or improvements with one if this. But we can go with ArrayList if you want.

bennetelli: ok :)

gwenneg (approved):
The tests still contain `LinkedList` instances but it's only a minor issue, so LGTM

bennetelli:
> The tests still contain `LinkedList` instances but it's only a minor issue, so LGTM

with the latest commit I replaced every LinkedList with ArrayList.

gwenneg:
Thanks!

PR #1106, opened 2022-03-28
NOTIF-484 Introduce templates REST API

gwenneg:
🚧🚧🚧 WIP - Do not review 🚧🚧🚧


engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EmailTemplateRepository.java
```diff
@@ -15,15 +15,15 @@
...
+public class EmailTemplateRepository {
 
     @Inject
     StatelessSessionFactory statelessSessionFactory;
 
     public boolean isEmailSubscriptionSupported(String bundleName, String appName, EmailSubscriptionType subscriptionType) {
         if (subscriptionType == INSTANT) {
-            String hql = "SELECT COUNT(*) FROM InstantEmailTemplate " +
-                    "WHERE eventType.application.bundle.name = :bundleName AND eventType.application.name = :appName";
+            String hql = "SELECT COUNT(*) FROM InstantEmailTemplate WHERE enabled IS TRUE " +
```
gwenneg, line 25: Should we allow subscription for `instant` notifications when the event type is linked with a disabled template?

josejulio: I don't think we should allow that.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
```diff
@@ -0,0 +1,338 @@
...
+                .setParameter("subscriptionType", template.getSubscriptionType())
+                .setParameter("subjectTemplate", subjectTemplate)
+                .setParameter("bodyTemplate", bodyTemplate)
+                .setParameter("id", id)
+                .executeUpdate();
+        return rowCount > 0;
+    }
+
+    @Transactional
+    public boolean enableOrDisableAggregationEmailTemplate(UUID id, boolean enabled) {
```
josejulio, line 271: The name seems odd - but not really important

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
```diff
@@ -0,0 +1,338 @@
...
+        if (enabled) {
+            AggregationEmailTemplate emailTemplate = findAggregationEmailTemplate(id);
```
josejulio, line 273: Isn't this throwing if nothing is enabled?

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/TemplateResource.java
```diff
@@ -0,0 +1,267 @@
...
+                    @Content(schema = @Schema(title = "RenderEmailTemplateResponseError", implementation = RenderEmailTemplateResponse.Error.class))
+            })
+    })
+    @RolesAllowed(RBAC_INTERNAL_USER)
+    public Response renderEmailTemplate(@NotNull @Valid RenderEmailTemplateRequest renderEmailTemplateRequest) {
+        try {
+            return templateEngineClient.render(renderEmailTemplateRequest);
+        } catch (BadRequestException e) {
+            // The following line is required to forward the HTTP 400 error message.
+            return Response.status(BAD_REQUEST).entity(e.getMessage()).build();
```
josejulio, line 264: why is this required? I though the BadRequest exception was mapped correctly to the http error.

gwenneg: This is old code (already on `master`) that fixes a `quarkus-rest-client-reactive` oddity. It shouldn't be necessary, but it is...

I haven't checked the behavior of that extension since we bumped Quarkus. Maybe they fixed that thing.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
gwenneg: It is long and not pretty, but it describes exactly what the method does 😃 I'm OK to change it if you have a better idea.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
gwenneg: The idea here, which I wanted to discuss with you, is to automatically disable other templates before enabling a new one because we can't have multiple templates enabled at the same time. There is a similar code for instant email template.

Another way of dealing with multiple enabled templates would be an exception throw that would be forwarded to the frontend and cancel the user action, asking him to first disable the other template before enabling a new one.

I have a slight preference for the current code which would only require a warning in the frontend when the user is about to enable a template, saying something like:
```
Warning. If you enable this template, all other templates that are already associated with [app] / [subscription type] will be automatically disabled. Do you want to continue? Yes/No
```

Or, the frontend could first call the backend and run a `COUNT(*)` on templates to check if there is another enabled template linked to the app/subscriptionType and only display the warning if there is one.

gwenneg: Or maybe we don't need the `enabled` field at all and linking the template to the app/subscriptionType is enough to enable a template 😄

Plenty of options here, we can pick any of them and I'll adapt the code.

gwenneg:
Rebased on `master` to fix the tests instabilities.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
josejulio: I think it's better to link-to-enable, and only 1 can be linked at a time.

common/src/main/java/com/redhat/cloud/notifications/models/InstantEmailTemplate.java
```diff
@@ -0,0 +1,178 @@
...
+    private UUID eventTypeId;
+
+    @ManyToOne(fetch = LAZY, optional = false)
+    @JoinColumn(name = "subject_template_id")
+    @JsonProperty(access = READ_ONLY)
+    private Template subjectTemplate;
+
+    @NotNull
+    @Transient
+    private UUID subjectTemplateId;
```
josejulio, line 50: are these transient fields actually used? or could be replaced by a getter ? (I think jackson picks the fields from `getStuff` fields when serializing) - Not specific to this PR, but maybe we could refactor to remove these extra fields.

gwenneg: Jackson serializes fields by default AFAIK.

About this specific field, yes it is used in one of the REST responses.

josejulio: The point i tried to convey is that (IIUC) this `subjectTemplateId` is a synthethic field that we only add for the REST responses.
Removing this field and creating a method:
```
@JsonProperty
getSubjectTemplateId() {
  return subjectTemplate.getId()
}
```

would yield the same behavior without the extra field that we need to keep in sync with the `subjectTemplate.id`

gwenneg: We could do that indeed.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
gwenneg: This method no longer exists.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
gwenneg: I pushed a new commit and removed the `enabled` field from the email templates.

common/src/main/java/com/redhat/cloud/notifications/models/InstantEmailTemplate.java
gwenneg: I forgot that we need that field for the incoming REST request. 😄

josejulio: Lets explore alternatives for that in the future (and outside of this PR)

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
josejulio: That was a clever way to avoid naming something :)
image

josejulio (approved):
lgtm

PR #1109, opened 2022-03-28
NOTIF-381 Remove paging for RBAC

bennetelli:


josejulio:
I think we can remove it when we decide to switch permanently from rbac to it call.

PR #1110, opened 2022-03-28
NOTIF-510 begin UI build out for email templates

aferd:
Second go.

1. We talked through the view option being a good idea but unsure of how that would work due to it depending on a payload. Any more thoughts on this?
2. On the table, listing the template name makes sense, aside from that would listing the event-types or Id associated to the template be necessary?
3. On the emailTemplatePage, would it make sense to have placeholderText or helperText to help guide users? Having a default template was mentioned as well, in which case we could set that to be the "view" option on the table vs being able to view them all as a sort of guide for the user?
4. Any other thoughts or suggestions are welcomed!! 

https://user-images.githubusercontent.com/86322239/160458084-bdf45c9e-7adc-462d-9574-ea685fdf0907.mov


josejulio:
> 1. We talked through the view option being a good idea but unsure of how that would work due to it depending on a payload. Any more thoughts on this?

Not really sure yet - We will need a payload to be able to render this. Perhaps we can skip this one for the moment.

> 2. On the table, listing the template name makes sense, aside from that would listing the event-types or Id associated to the template be necessary?

That would be a nice to have. Being able to see where the template is being used.

> 3. On the emailTemplatePage, would it make sense to have placeholderText or helperText to help guide users? Having a default template was mentioned as well, in which case we could set that to be the "view" option on the table vs being able to view them all as a sort of guide for the user?

Yes, see the current "template" page with a sample.

> 4. Any other thoughts or suggestions are welcomed!!

could we hide this behind the `isAdmin` permission? 
See https://github.com/RedHatInsights/notifications-backend/blob/master/admin-console/src/main/webapp/src/pages/BundlePage.tsx#L24

admin-console/src/main/webapp/src/components/EmailTemplates/VIewTemplateModal.tsx
```diff
@@ -0,0 +1,30 @@
+import { ActionGroup, Button, Form, Modal, ModalVariant } from '@patternfly/react-core';
```
josejulio, line 1: Check the filename, it is `VIew` (capital I)

aferd:
Updated state
https://user-images.githubusercontent.com/86322239/160483732-5dbcab13-0043-4cd8-89db-cf9b2d6fd565.mov

gwenneg:
How will we differentiate in the UI the `instant` email templates from the `aggregation` ones? Will they be shown at the same place?

In the DB, they are currently stored in separate tables because they are not associated with the same thing:
- instant templates need a link with an event type
- aggregation templates need a link with an application

This can be easily changed though as we could store all templates in the same tables with a few backend changes and then link any template with any (event multiple) event type or application.

admin-console/src/main/webapp/src/components/EmailTemplates/VIewTemplateModal.tsx
josejulio: The filename still shows as capital `I`. i.e. `VIew` instead of `View`

admin-console/src/main/webapp/src/components/EmailTemplates/EmailTemplateTable.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { useApplicationTypes } from '../../services/EventTypes/GetApplication';
+import { ListEventTypes } from '../EventTypes/ListEventTypes';
+import { ViewTemplateModal } from './ViewTemplateModal';
+
+type ApplicationPageParams = {
+    applicationId: string;
+}
+
+export const EmailTemplateTable: React.FunctionComponent = () => {
+    const { hasPermission } = useUserPermissions();
```
josejulio, line 24: To use the `isAdmin`

```suggestion
    const { hasPermission, isAdmin } = useUserPermissions();
```

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -210,6 +211,7 @@ export const ApplicationPage: React.FunctionComponent = () => {
                     </Tbody>
                 </TableComposable>
             </PageSection>
+            <EmailTemplateTable />
```
josejulio, line 214: Could we hide this for non admins?
on top:
` { isAdmin }  = usePermissions();`

here:
```suggestion
            { isAdmin && <EmailTemplateTable /> }
```

josejulio:
> How will we differentiate in the UI the `instant` email templates from the `aggregation` ones? Will they be shown at the same place?
> 
> In the DB, they are currently stored in separate tables because they are not associated with the same thing:
> 
> * instant templates need a link with an event type
> * aggregation templates need a link with an application
> 
> This can be easily changed though as we could store all templates in the same tables with a few backend changes and then link any template with any (or even multiple) event type or application.

I was thinking into being able to create not-linked templates in one place and then being able to link to other event-types (instants) or application (daily). We could use templates from other templates ([include](https://quarkus.io/guides/qute-reference#include_helper)) or selecting one for the [title and description](https://issues.redhat.com/browse/NOTIF-538) of a notification.

gwenneg:
> I was thinking into being able to create not-linked templates in one place and then being able to link to other event-types (instants) or application (daily). We could use templates from other templates ([include](https://quarkus.io/guides/qute-reference#include_helper)) or selecting one for the [title and description](https://issues.redhat.com/browse/NOTIF-538) of a notification.

Ok, this was my initial plan in the backend (partly commited already) but then I started wondering if this wasn't going too far for the initial version of DB templates. 😄 Let's stick with that plan for now.

I'll have to check if inclusion works from DB templates though. We'll need to register the templates from the DB into the Qute engine, which means we must first detect that an inclusion is needed.

gwenneg:
Could work, yes. That's part the standalone version of Qute. I'll talk with its creator about what we want to do 😃

gwenneg:
He said `TemplateLocator` should work, I'll give it a try.

Sorry for polluting this PR with backend discussions 😄

admin-console/src/main/webapp/src/components/EmailTemplates/VIewTemplateModal.tsx
aferd: this is not showing on my local machine, it is showing lowercase, weird.

aferd: Screen Shot 2022-03-29 at 9 45 24 AM

aferd:
the build here is failing, any suggestions on how to troubleshoot this? I have no errors when I am running the backend locally.

`Error:  Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.12.1:yarn (yarn run build) on project notifications-admin-console: Failed to run task: 'yarn build' failed. org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1) -> [Help 1]`

josejulio:
> the build here is failing, any suggestions on how to troubleshoot this? I have no errors when I am running the backend locally.
> 
> `Error: Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.12.1:yarn (yarn run build) on project notifications-admin-console: Failed to run task: 'yarn build' failed. org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1) -> [Help 1]`

Its failing with the following error [see line 106](https://github.com/RedHatInsights/notifications-backend/runs/5737841983?check_suite_focus=true#step:8:106)
> [INFO] Cannot find file './ViewTemplateModal' in './src/components/EmailTemplates'.

I suspect it's related to the `VIew` thing. 
I suggest to rename the file to something else. Commit the change. Rename back and commit the change, maybe that would help.

josejulio (approved):
lgtm

aferd:
@mmenestr

mmenestr:
Did you use specific mocks to create these email templates? Or does this have nothing to do with how it'll visually look to a customer

aferd:
> 

I did not. This is an internal only application. In our sync meeting today, we touched base on UXD involvement here (with the admin UI) and for now we are not going to have you all involved. I had a bit of a misunderstanding from our demo meeting regarding the email templates, my apologies!

PR #1120, opened 2022-03-30
NOTIF-347 Allow to create endpoints with rbac groups

josejulio:
Requires doing a validation with rbac to confirm the group id really exists.
We might need a way to delete rbac groups that are deleted.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacGroup.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacGroup.java
@@ -2,7 +2,6 @@ package com.redhat.cloud.notifications.auth.rbac;
 
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.eclipse.microprofile.rest.client.inject.RestClient;
-import org.jboss.logging.Logger;
 import org.jboss.resteasy.reactive.ClientWebApplicationException;
 
 import javax.enterprise.context.ApplicationScoped;
--- a/backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacGroup.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacGroup.java
@@ -2,6 +2,8 @@ package com.redhat.cloud.notifications.auth.rbac;
 
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.eclipse.microprofile.rest.client.inject.RestClient;
+import org.jboss.logging.Logger;
+import org.jboss.resteasy.reactive.ClientWebApplicationException;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
@@ -21,10 +23,19 @@ public class RbacGroup {
 
     public boolean validate(UUID groupId, String rhIdentity) {
         if (isRbacEnabled) {
-            Response response = rbacServer.getGroup(groupId, rhIdentity)
-                    .await()
-                    .atMost(Duration.ofSeconds(2L));
-            return response.getStatus() == 200;
+            try {
+                Response response = rbacServer.getGroup(groupId, rhIdentity)
+                        .await()
+                        .atMost(Duration.ofSeconds(2L));
+                return response.getStatus() == 200;
+            } catch (ClientWebApplicationException errorException) {
+                if (errorException.getResponse().getStatus() == 404) {
+                    // This is fine, the group was not found
+                    return false;
+                }
+
+                throw errorException;
+            }
         }
 
         return true;
--- a/backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacGroup.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacGroup.java
@@ -11,14 +11,14 @@ import java.time.Duration;
 import java.util.UUID;
 
 @ApplicationScoped
-public class RbacGroup {
+public class RbacGroupValidator {
 
     @Inject
     @RestClient
     RbacServer rbacServer;
 
     @ConfigProperty(name = "rbac.enabled", defaultValue = "true")
-    Boolean isRbacEnabled;
+    boolean isRbacEnabled;
 
     public boolean validate(UUID groupId, String rhIdentity) {
         if (isRbacEnabled) {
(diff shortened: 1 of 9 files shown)


backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacGroup.java
```diff
@@ -0,0 +1,42 @@
...
+import org.jboss.resteasy.reactive.ClientWebApplicationException;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+import javax.ws.rs.core.Response;
+import java.time.Duration;
+import java.util.UUID;
+
+@ApplicationScoped
+public class RbacGroup {
```
gwenneg, line 14:
```suggestion
public class RbacGroupValidator {
```

backend/src/main/java/com/redhat/cloud/notifications/routers/models/RequestEmailSubscriptionProperties.java
```diff
@@ -4,11 +4,13 @@
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 
 import javax.validation.constraints.NotNull;
+import java.util.UUID;
 
 @JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
 public class RequestEmailSubscriptionProperties {
     @NotNull
     private boolean onlyAdmins;
+    private UUID groupId;
```
gwenneg, line 13: Does this restrict the number of RBAC groups to one per endpoint? The images on Jira suggest that more than one group may be selected for an email endpoint.

backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacServer.java
```diff
@@ -29,4 +33,9 @@
                              @HeaderParam(X_RH_IDENTITY_HEADER) String rhIdentity
 
     );
+
+    @GET
+    @Path("/groups/{groupID}/") // trailing slash is required by api
+    @Produces("application/json")
+    Uni<Response> getGroup(@PathParam("groupID") UUID groupId, @HeaderParam(X_RH_IDENTITY_HEADER) String rhIdentity);
```
gwenneg, line 40: We don't need the reactiveness here:
```suggestion
    Response getGroup(@PathParam("groupID") UUID groupId, @HeaderParam(X_RH_IDENTITY_HEADER) String rhIdentity);
```

backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacGroup.java
```diff
@@ -0,0 +1,42 @@
...
+    public boolean validate(UUID groupId, String rhIdentity) {
+        if (isRbacEnabled) {
+            try {
+                Response response = rbacServer.getGroup(groupId, rhIdentity)
+                        .await()
+                        .atMost(Duration.ofSeconds(2L));
+                return response.getStatus() == 200;
+            } catch (ClientWebApplicationException errorException) {
+                if (errorException.getResponse().getStatus() == 404) {
+                    // This is fine, the group was not found
```
gwenneg, line 32: Or we could have an RBAC host issue in the configuration 😄

backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacGroup.java
```diff
@@ -0,0 +1,42 @@
...
+
+@ApplicationScoped
+public class RbacGroup {
+
+    @Inject
+    @RestClient
+    RbacServer rbacServer;
+
+    @ConfigProperty(name = "rbac.enabled", defaultValue = "true")
+    Boolean isRbacEnabled;
```
gwenneg, line 21: Nitpick: could be a primitive.
```suggestion
    boolean isRbacEnabled;
```

backend/src/main/java/com/redhat/cloud/notifications/routers/models/RequestEmailSubscriptionProperties.java
josejulio: Yes, one per endpoint; multiple endpoints can be created (internally) to handle those combinations.

backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacGroup.java
josejulio: ummh - it returns a 404 if we have host issues? :D 
Would you like to add other validation here about the content?

gwenneg: No it's OK 😃 If the RBAC config was wrong, we would know it before reaching this point anyway.

backend/src/main/java/com/redhat/cloud/notifications/routers/models/RequestEmailSubscriptionProperties.java
gwenneg: So the frontend will call `POST /system/email_subscription` multiple times while the user is only adding one email integration? That sounds a bit weird. Is it impossible to handle several groups with a single email endpoint?

josejulio: This is done to allow (in the future) multiple configuration types and not making the endpoint type more complicated.
So one could do `send to this group, override user preferences and only admins`
```
{
  group: "some-id",
  sendToAdmins: true,
  overrideUserPreferences: true
}
```

Furthermore, each system/email_subscription are currently immutable and reused. So that when we create a behavior group with the action: "Send to all users". Only 1 endpoint is created and shared across all behavior groups that have the same action (send to all users).

The system endpoints are immutable. If we were to accept multiple groups in there, we could end with multiple combinations and maybe more complicated queries.

Having them as a single configuration for an email seemed more easier to manage and flexible to me.
See the [tests](https://github.com/RedHatInsights/notifications-backend/blob/master/backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java#L882) for info on how they behave.

gwenneg: I remember the system endpoints immutability 😃 

Makes sense, thanks for clarifying!

gwenneg (approved):
LGTM

PR #1130, opened 2022-04-04
Removing 'users' tag on rbac.get-users.total Prometheus metric

theute:
It was useful to have, but now it adds extra cardinality and makes queries more complex when comparing accounts

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
@@ -141,7 +141,7 @@ public class RbacRecipientUsersProvider {
 
         // Micrometer doesn't like when tags are null and throws a NPE.
         String accountIdTag = accountId == null ? "" : accountId;
-        getUsersTotalTimer.stop(meterRegistry.timer("rbac.get-users.total", "accountId", accountIdTag, "users", String.valueOf(users.size())));
+        getUsersTotalTimer.stop(meterRegistry.timer("rbac.get-users.total", "accountId", accountIdTag));
         rbacUsers.set(users.size());
 
         return users;
--- a/engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
@@ -150,8 +150,8 @@ public class RbacRecipientUsersProvider {
     }
 
     private AtomicInteger getRbacUsersGauge(String accountIdTag) {
-        return rbacUsers.computeIfAbsent(accountIdTag, /* accountIdTag */ key -> {
-            Set<Tag> tags = Set.of(Tag.of("accountId", key));
+        return rbacUsers.computeIfAbsent(accountIdTag, accountId -> {
+            Set<Tag> tags = Set.of(Tag.of("accountId", accountId));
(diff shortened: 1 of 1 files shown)


gwenneg:
I have the same issue in [another PR](https://github.com/RedHatInsights/notifications-backend/pull/1106). It could be another GH issue/outage.

josejulio:
> [2022-04-04 12:58:06] [build-stderr] Illegal use of '**' in exclude path

It was fine before, but maybe something changed?

https://github.com/RedHatInsights/notifications-backend/pull/972

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -151,7 +151,7 @@ public void initCounters() {
                 }
             );
         }
-        getUsersTotalTimer.stop(meterRegistry.timer("rbac.get-users.total", "accountId", accountId, "users", String.valueOf(users.size())));
```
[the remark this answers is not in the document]
gwenneg: [This](https://github.com/RedHatInsights/notifications-backend/blob/10259a481f89da2c894f80cd96265b98ad410685/engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java#L93) is the gauge I mentioned.

gwenneg:
@theute Do you have time to resolve the conflicts and finish this? Otherwise I can take care of it.

theute:
Sorry, I left it after we had hopes for a counter metrics with "dynamic" tags, but I think it didn't work, so we would again need to add a timer to add the account tag I guess.
Happy if you finish it, it stayed open for too long

gwenneg:
Ok I'll take a loot at it later.

I'm doing spring cleaning today, trying to merge/close old PRs (or at least make some progress) 😄

gwenneg:
I rebased and fixed the conflicts.

I'll push one more commit with something that should be some kind of gauge with dynamic tags.

gwenneg:
@josejulio Could you please review my latest commit? It's untested but I think it will work.

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -141,12 +143,19 @@ public List<User> getUsers(String accountId, String orgId, boolean adminsOnly) {
...
+        getRbacUsersGauge(accountIdTag).set(users.size());
 
         return users;
     }
 
+    private AtomicInteger getRbacUsersGauge(String accountIdTag) {
+        return rbacUsers.computeIfAbsent(accountIdTag, /* accountIdTag */ key -> {
+            Set<Tag> tags = Set.of(Tag.of("accountId", key));
+            return meterRegistry.gauge("rbac.users", tags, new AtomicInteger());
+        });
```
josejulio, line 156: Isn't it better to call it `accountIdTag` instead of `key`? or even `accountId` ? 

```suggestion
        return rbacUsers.computeIfAbsent(accountIdTag, accountId -> {
            Set<Tag> tags = Set.of(Tag.of("accountId", accountId));
            return meterRegistry.gauge("rbac.users", tags, new AtomicInteger());
        });
```

gwenneg: We can't repeat `accountIdTag` since it's already taken but `accountId` will work, yes.

josejulio: yep, any is fine :D

PR #1134, opened 2022-04-04
NOTIF-347 Check if the group exists

josejulio:
If the group no longer exists, just ignore it.

Now that we are going to accept groups, they could be deleted without us noticing it. So instead of failing, lets just ignore it.

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
@@ -19,6 +19,7 @@ import io.quarkus.cache.CacheResult;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.eclipse.microprofile.rest.client.inject.RestClient;
 import org.jboss.logging.Logger;
+import org.jboss.resteasy.reactive.ClientWebApplicationException;
 
 import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
@@ -160,7 +161,18 @@ public class RbacRecipientUsersProvider {
     @CacheResult(cacheName = "rbac-recipient-users-provider-get-group-users")
     public List<User> getGroupUsers(String accountId, boolean adminOnly, UUID groupId) {
         Timer.Sample getGroupUsersTotalTimer = Timer.start(meterRegistry);
-        RbacGroup rbacGroup = retryOnRbacError(() -> rbacServiceToService.getGroup(accountId, groupId));
+        RbacGroup rbacGroup;
+        try {
+            rbacGroup = retryOnRbacError(() -> rbacServiceToService.getGroup(accountId, groupId));
+        } catch (ClientWebApplicationException exception) {
+            // The group does not exist (or no longer exists - ignore)
+            if (exception.getResponse().getStatus() == 404) {
+                return List.of();
+            }
+
+            throw exception;
+        }
+
         List<User> users;
         if (rbacGroup.isPlatformDefault()) {
             users = getUsers(accountId, adminOnly);
--- a/engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
@@ -24,6 +24,7 @@ import org.jboss.resteasy.reactive.ClientWebApplicationException;
 import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
+import javax.ws.rs.core.Response;
 import java.io.IOException;
 import java.time.Duration;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -160,7 +161,18 @@ public void initCounters() {
     @CacheResult(cacheName = "rbac-recipient-users-provider-get-group-users")
     public List<User> getGroupUsers(String accountId, boolean adminOnly, UUID groupId) {
         Timer.Sample getGroupUsersTotalTimer = Timer.start(meterRegistry);
-        RbacGroup rbacGroup = retryOnRbacError(() -> rbacServiceToService.getGroup(accountId, groupId));
+        RbacGroup rbacGroup;
+        try {
+            rbacGroup = retryOnRbacError(() -> rbacServiceToService.getGroup(accountId, groupId));
+        } catch (ClientWebApplicationException exception) {
```
bennetelli, line 167: I searched for this exception in the jira, kibana (prod) and code, but did not find any occurrence. I guess the Quarkus rest client wraps every exception/failure as a ClientWebApplicationException, but did not find hints in the docs for it.

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -160,7 +161,18 @@ public void initCounters() {
...
+            // The group does not exist (or no longer exists - ignore)
+            if (exception.getResponse().getStatus() == 404) {
```
bennetelli, line 169: Maybe a bit too scared, but I would maybe flip the left and right part of the equals.
We could also replace the 404 with a constant. I think we don't have to as close to everyone knows what 404 is, but replacing 404 with something "readable" like "not found" would improve readability.

This would do the same: 

if (HTTP_NOT_FOUND == exception.getResponse().getStatus()) {
    return List.of();
}

You can decide if you want to replace it or not. I am fine with both :-)

bennetelli: PS: the import is from java.net.HttpURLConnection.

engine/src/test/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProviderTest.java
```diff
@@ -122,6 +123,16 @@ public void getAllUsersFromDefaultGroup() {
         }
     }
 
+
```
bennetelli, line 126: that newline could be removed :)

engine/src/test/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProviderTest.java
```diff
@@ -122,6 +123,16 @@ public void getAllUsersFromDefaultGroup() {
...
+    @Test
+    public void getUsersFromNonExistentGroupResultsInNoUsers() {
```
bennetelli, line 128: I personally prefer the "pattern" should<<what I expect>>When<<What I do>> as test name. If the test fails and people read the failed test in the console or in the IDE, they immediately know what is going on in the test, what's expected and so on.

Something like this as test name maybe "shouldReturnNoUsersWhenGroupNotFound". 

This is again up to you to change it :-)

bennetelli (changes requested):
If you don't want to change it to my prefered "styl", just request a review again and I will improve it :)

engine/src/test/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProviderTest.java
bennetelli: It's just my personal preference. There are lots of different patterns for naming unit tests: https://dzone.com/articles/7-popular-unit-test-naming

Maybe that's something we should talk about with the other devs and then decide if we prefer one strategie? 
With one naming strategy the codebase would look a bit "cleaner"

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
josejulio: I don't think we had this exception yet, but I'm preparing because the client can indeed return that (as show in the tests) when we get a 404.

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
josejulio: I think the order is fine, but will add the NOT_FOUND constant

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
bennetelli: Ok.

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
bennetelli: Yea, the flip would save us if the exception object is null or the getResponse() returns null. But this probably never gonna happen.

bennetelli (approved):

PR #1136, opened 2022-04-05
Bump MockServer to 5.13.2

gwenneg:
Our build is unstable on GitHub when the `mockserver-client-java` version is higher than `5.5.4`.

This branch will be used to try and stabilize the build with the latest mockserver version.

PR #1137, opened 2022-04-05
Update mockserver java client version

bennetelli:


bennetelli:
Let's see if this is also working with GitHub Actions.

bennetelli:
WTF?! This is green on my machine :-D

bennetelli:
Version in MockServerLifecycleManager is different. So that exception is ok. Anyway. this should have been failed on my machine already

bennetelli:
@gwenneg the last GitHub Pull Request build failed because of the usual suspects.. time out and not-called-within one. 

If you have got some minutes of time, could you test this branch on your machine as well maybe? Just to be sure it's working on my machine, here and on your machine.

gwenneg:
⚠️ This PR (or #1136) must not be merged as `master` will be unstable again.

> If you have got some minutes of time, could you test this branch on your machine as well maybe? Just to be sure it's working on my machine, here and on your machine.

Yes, mockserver 5.13.x works on most of the team's machines AFAIK. `master` depended on that version until I downgraded it today.

bennetelli:
@gwenneg ok.. neither my nor your version update leads to stable builds. I triggered this GH Build three times and it was two times red and one time green. We should stay with the version we are using right now. 5.5.4

gwenneg:
I know my PR leads to unstable build, that's the reason why it exists and will remain a Draft for a while 😄

bennetelli:
I will try to remove evey mockserver/testcontainers dependency from my machine again and start with a clean project and empty .m2/repository folder. Maybe that fixes my local hickup.

gwenneg:
All I know is that `jamesdbloom` is the guy who created mockserver and still maintains it. Now why are we using one image instead of the other? No idea. Maybe the `mockserver/mockserver` one didn't exist at first.

bennetelli:
Nevermind the issue is still there.. 
The mockserver/mockserver one is the official one. I will open a PR so we will use the official one in the future.

gwenneg:
They are both official. The `mockserver/mockserver` image was first published with version 5.5.0 while `jamesdbloom/mockserver` started with 3.10.5. I suppose they chose to keep supporting the old image path to avoid breaking existing builds. There's no rush changing anyway, this can be done as part of the stabilization effort.

PR #1138, opened 2022-04-05
Use official Mockserver Docker image

bennetelli:


bennetelli:
I agree there is no rush to change the Docker image, but we have to do it anyway somewhere in the future. Why not just merge this PR now that it is open already and then we can't forget to change it and we also can't miss it and get a failing build becuase the jamesdbloom image isnt present anymore.

gwenneg (approved):
LGTM

PR #1139, opened 2022-04-05
NOTIF-444 Change the way of creating dynamic counters

bennetelli:


bennetelli:
as written to you in Slack. I just want to test it on stage quickly. If it's working then it's fine. If not, I will revert it so others are not affected by it.

josejulio (approved):
OK, lets try.

gwenneg:
What was the conclusion of the tests on stage? Is it safe to be deployed on prod?

bennetelli:
> What was the conclusion of the tests on stage? Is it safe to be deployed on prod?

It's "safe" in that way that the tags for the processor_email_processed metrics are still not present on stage and prod.
So it's safe to be deployed on prod. We didn't have these tags before this PR and with this one we will also not have it on prod. So it doesn't add something useful for customers, but at least it doesn't destroy something.

gwenneg:
Sounds safe to me, thanks!

PR #1140, opened 2022-04-05
Update sentry to the rcommend version in quarkiverse

bennetelli:


pom.xml
```diff
@@ -53,6 +53,7 @@
         <clowder-quarkus-config-source.version>1.0.0</clowder-quarkus-config-source.version>
 
         <cloudwatch.version>4.2.0</cloudwatch.version>
+        <sentry.version>1.1.0</sentry.version>
```
gwenneg, line 56: `sentry.version` is a bit misleading. This would help avoiding confusion:
```suggestion
        <quarkus-logging-sentry.version>1.1.0</quarkus-logging-sentry.version>
```

theute, line 56: Should be renamed something like `quarkus.sentry.version` because this is not the Sentry version here. (Which would be something like 5.x.x)

theute: Looks like we would upgrade from 5.5.0 to 5.6.3 (a good thing).
https://repo1.maven.org/maven2/io/quarkiverse/loggingsentry/quarkus-logging-sentry-parent/1.1.0/quarkus-logging-sentry-parent-1.1.0.pom
Sentry will still complain that we are not at 5.7.1 though :-)

gwenneg: This needs to be applied in all pom.xml files now 😃

bennetelli:
> Unrelated question: I have limited experience with the Quarkiverse. Is there a general guideline about the versioning for the Quarkiverse repos? In this specific case, quarkus-logging-sentry `1.1.0` depends on Quarkus `2.7.4.Final` which is fine since we're using `2.7.5.Final`, but it could depend on Quarkus `x.y.z` which could not be compatible with our Quarkus version. How do we avoid incompatibility issues between our Quarkus version and the one pulled by a Quarkiverse extension?

No, as far as know the developers can release how ever they like. You could even release version 1337 as the initial release.

Regarding the second question -> I need to dive into the GitHub Actions build. I would expect that they are doing a matrix build and build with every possible version. But I need to check that and I can also ask someone then as it would be interesting for the cloudwatch extension as well

bennetelli:
I will merge this now and dive into the quarkus quarkiverse compatibility question by Gwenneg

PR #1141, opened 2022-04-05
NOTIF-456 adding screens in admin UI for system(default) behavior groups

aferd:
WIP

### DIFF
--- a/admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
+++ b/admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
@@ -161,9 +161,9 @@ export const ApplicationPage: React.FunctionComponent = () => {
                             ))}
                         </Tr>
                     </Thead>
-                    <Tbody>{ (eventTypesQuery.payload.value.length === 0 ? 'There are no event types found for this application' : '') }</Tbody>
+                    <Tbody>{ (eventTypesQuery.data?.length === 0 ? 'There are no event types found for this application' : '') }</Tbody>
                     <Tbody>
-                        { eventTypesQuery.payload.value.map((e: EventType) => (
+                        { eventTypesQuery.data?.map((e: EventType) => (
                             <Tr key={ e.id }>
                                 <Td>{ e.displayName }</Td>
                                 <Td>{ e.name }</Td>
@@ -182,7 +182,7 @@ export const ApplicationPage: React.FunctionComponent = () => {
                 </TableComposable>
             </PageSection>
             { isAdmin && <EmailTemplateTable
-                application={ application?.displayName }
+                application={ application?.displayName ?? '' }
             /> }
             <CreateEditModal
                 isEdit={ isEdit }
--- a/admin-console/src/main/webapp/src/services/SystemBehaviorGroups/GetBehaviorGroups.tsx
+++ b/admin-console/src/main/webapp/src/services/SystemBehaviorGroups/GetBehaviorGroups.tsx
@@ -6,7 +6,7 @@ import { Operations } from '../../generated/OpenapiInternal';
 import { BehaviorGroup } from '../../types/Notifications';
 
 const validateResponse = validationResponseTransformer(
-    (payload: Operations.InternalServiceGetDefaultBehaviorGroups.Payload) => {
+    (payload: Operations.InternalResourceGetDefaultBehaviorGroups.Payload) => {
         if (payload.status === 200) {
             const systemBehaviorGroups: ReadonlyArray<BehaviorGroup> = payload.value.map(value => ({
                 bundleId: value.bundle_id,
@@ -27,17 +27,22 @@ const validateResponse = validationResponseTransformer(
     }
 );
 
-export const useSystemBehaviorGroups = () => {
-    const query = useQuery(Operations.InternalServiceGetDefaultBehaviorGroups.actionCreator());
+export const useSystemBehaviorGroups = (bundleId: string) => {
+    const query = useQuery(Operations.InternalResourceGetDefaultBehaviorGroups.actionCreator());
 
     const queryPayload = useMemo(() => {
         const payload = query.payload;
         if (payload) {
-            return validateResponse(payload);
+            const response = validateResponse(payload);
+            if (response.status === 200) {
+                response.value = response.value.filter(bg => bg.bundleId === bundleId);
+            }
+
+            return response;
         }
 
         return undefined;
-    }, [ query.payload ]);
+    }, [ query.payload, bundleId ]);
 
     return useMemo(() => ({
         ...query,
--- a/admin-console/src/main/webapp/src/pages/BundlePage/BehaviorGroupTable.tsx
+++ b/admin-console/src/main/webapp/src/pages/BundlePage/BehaviorGroupTable.tsx
@@ -6,10 +6,11 @@ import * as React from 'react';
 
 import { CreateEditBehaviorGroupModal } from '../../components/SystemBehaviorGroups/CreateEditBehaviorGroupModal';
 import { DeleteBehaviorGroupModal } from '../../components/SystemBehaviorGroups/DeleteBehaviorGroupModal';
-import { useBundleTypes } from '../../services/Applications/GetBundleById';
+import { Schemas } from '../../generated/OpenapiInternal';
 import { useCreateSystemBehaviorGroup } from '../../services/SystemBehaviorGroups/CreateSystemBehaviorGroup';
 import { useDeleteBehaviorGroup } from '../../services/SystemBehaviorGroups/DeleteSystemBehaviorGroup';
 import { useSystemBehaviorGroups } from '../../services/SystemBehaviorGroups/GetBehaviorGroups';
+import { useUpdateBehaviorGroupActionsMutation } from '../../services/SystemBehaviorGroups/UpdateActions';
 import { BehaviorGroup } from '../../types/Notifications';
 
 interface BundlePageProps {
@@ -18,10 +19,10 @@ interface BundlePageProps {
 }
 
 export const BehaviorGroupsTable: React.FunctionComponent<BundlePageProps> = (props) => {
-    const getBehaviorGroups = useSystemBehaviorGroups();
-    const getBundles = useBundleTypes(props.bundleId);
+    const getBehaviorGroups = useSystemBehaviorGroups(props.bundleId);
     const newBehaviorGroup = useCreateSystemBehaviorGroup();
     const deleteBehaviorGroupMutation = useDeleteBehaviorGroup();
+    const updateBehaviorActions = useUpdateBehaviorGroupActionsMutation();
 
     const columns = [ 'System Behavior Group', 'Action' ];
 
@@ -53,25 +54,35 @@ export const BehaviorGroupsTable: React.FunctionComponent<BundlePageProps> = (pr
     const handleSubmit = React.useCallback((systemBehaviorGroup) => {
         setShowModal(false);
         const mutate = newBehaviorGroup.mutate;
+        const updateActionsMutate = updateBehaviorActions.mutate;
         mutate({
             id: systemBehaviorGroup.id,
             displayName: systemBehaviorGroup.displayName ?? '',
-            actions: systemBehaviorGroup.actions,
-            bundleId: systemBehaviorGroup.bundleId
-        }).then(getBehaviorGroups.query);
+            bundleId: props.bundleId
+        })
+        .then(response => {
+            if (response.payload?.status === 200 && (response.payload.value.id || systemBehaviorGroup.id)) {
+                return updateActionsMutate({
+                    behaviorGroupId: response.payload.value.id ?? systemBehaviorGroup.id,
+                    body: [
+                        {
+                            ignore_preferences: false,
+                            only_admins: systemBehaviorGroup.actions === 'email-admin'
+                        }
+                    ]
+                });
+            }
+        })
+        .finally(getBehaviorGroups.query);
 
-    }, [ getBehaviorGroups.query, newBehaviorGroup.mutate ]);
+    }, [ getBehaviorGroups.query, newBehaviorGroup.mutate, props.bundleId, updateBehaviorActions.mutate ]);
 
     const handleDelete = React.useCallback(async () => {
         setShowDeleteModal(false);
         const deleteBehaviorGroup = deleteBehaviorGroupMutation.mutate;
-        const response = await deleteBehaviorGroup(systemBehaviorGroup.id);
-        if (response.error) {
-            return false;
-        }
-
-        return true;
-    }, [ deleteBehaviorGroupMutation.mutate, systemBehaviorGroup.id ]);
+        const response = await deleteBehaviorGroup(systemBehaviorGroup.id).finally(getBehaviorGroups.query);
+        return !response.error;
+    }, [ deleteBehaviorGroupMutation.mutate, systemBehaviorGroup.id, getBehaviorGroups.query ]);
 
     const onClose = () => {
         setShowModal(false);
@@ -97,8 +108,7 @@ export const BehaviorGroupsTable: React.FunctionComponent<BundlePageProps> = (pr
                 <Title headingLevel='h1'>
                     <Breadcrumb>
                         <BreadcrumbItem target='#'> Bundles </BreadcrumbItem>
-                        <BreadcrumbItem target='#' >{ (getBundles.loading || getBundles.payload?.status !== 200)
-                            ? <Spinner /> : getBundles.payload.value.displayName }
+                        <BreadcrumbItem target='#' >{ props.bundle }
                         </BreadcrumbItem>
                         <BreadcrumbItem target='#'> System Behavior Groups </BreadcrumbItem>
                     </Breadcrumb>
@@ -122,7 +132,16 @@ export const BehaviorGroupsTable: React.FunctionComponent<BundlePageProps> = (pr
                         { getBehaviorGroups.payload.value.map(b =>
                             <Tr key={ b.id }>
                                 <Td>{ b.displayName }</Td>
-                                <Td>{ b.actions }</Td>
+                                <Td>{ b.actions?.map(action => {
+                                    const properties = action.endpoint?.properties as Schemas.EmailSubscriptionProperties;
+                                    if (properties) {
+                                        if (properties.only_admins) {
+                                            return 'Admins';
+                                        } else {
+                                            return 'All users';
+                                        }
+                                    }
+                                }) }</Td>
                                 <Td>
                                     <Button className='edit' type='button' variant='plain'
                                         onClick={ () => editSystemBehaviorGroup(b) }
--- a/admin-console/src/main/webapp/src/services/SystemBehaviorGroups/CreateSystemBehaviorGroup.tsx
+++ b/admin-console/src/main/webapp/src/services/SystemBehaviorGroups/CreateSystemBehaviorGroup.tsx
@@ -1,21 +1,18 @@
 import { useMutation } from 'react-fetching-library';
 
 import { Operations } from '../../generated/OpenapiInternal';
-import { BehaviorGroupAction, UUID } from '../../types/Notifications';
+import { UUID } from '../../types/Notifications';
 
 export type CreateSystemBehaviorGroup = {
-    actions?: Array<BehaviorGroupAction> | undefined | null;
     bundleId: UUID;
-    isDefault?: boolean | undefined | null;
     displayName: string;
     id?: UUID | undefined | null;
 }
 
 const actionCreator =  (params: CreateSystemBehaviorGroup) => {
     if (params.id === undefined) {
-        return Operations.InternalServiceCreateDefaultBehaviorGroup.actionCreator({
+        return Operations.InternalResourceCreateDefaultBehaviorGroup.actionCreator({
             body: {
-                actions: params.actions,
                 bundle_id: params.bundleId,
                 display_name: params.displayName,
                 id: params.id
@@ -23,10 +20,9 @@ const actionCreator =  (params: CreateSystemBehaviorGroup) => {
         });
     }
 
-    return Operations.InternalServiceUpdateDefaultBehaviorGroup.actionCreator({
+    return Operations.InternalResourceUpdateDefaultBehaviorGroup.actionCreator({
         id: params.id ?? '',
         body: {
-            actions: params.actions,
             bundle_id: params.bundleId,
             display_name: params.displayName,
             id: params.id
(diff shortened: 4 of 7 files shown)


aferd:
@josejulio can you check this out when you have a moment? This is where I am currently, my selects aren't working properly and when you hit submit, it is adding them to every bundle, not just the current bundle. Could you give some direction? Mostly feeling stuck on why its updating each bundle...

https://user-images.githubusercontent.com/86322239/162997003-4c1d483a-3ec9-4180-8388-514cad503360.mov

admin-console/src/main/webapp/src/services/SystemBehaviorGroups/GetBehaviorGroups.tsx
```diff
@@ -0,0 +1,46 @@
...
+                payload.errors
+            );
+        }
+
+        return payload;
+    }
+);
+
+export const useSystemBehaviorGroups = () => {
+    const query = useQuery(Operations.InternalServiceGetDefaultBehaviorGroups.actionCreator());
```
josejulio, line 31: I guess the reason of why it's appearing on all the bundles is because we are fetching all the system behavior groups - does this action accepts a bundleId ? If not we might need to filter it.

admin-console/src/main/webapp/src/components/EventTypes/CreateEditModal.tsx
```diff
@@ -58,16 +66,21 @@ export const CreateEditModal: React.FunctionComponent<CreateEditModalProps> = (p
...
-                    <FormGroup label='Description' fieldId='description'
-                        helperText='Optional short description that appears in the UI
-                                                to help admin decide how to notify users.'>
-                        <TextArea
-                            type='text'
-                            value={ eventType.description }
-                            onChange={ handleChange }
-                            id='description'
-                            name="description"
-                        /></FormGroup>
```
josejulio, line 70: I gues we can leave this field in here, even if we don't show it on the table.

admin-console/src/main/webapp/src/components/EventTypes/CreateEditModal.tsx
```diff
@@ -58,16 +66,21 @@ export const CreateEditModal: React.FunctionComponent<CreateEditModalProps> = (p
...
+                    <FormGroup label='System Behavior Group' fieldId='system behavior group'
+                        helperText='Behavior groups are made up of action/recipient pairings that allow you to configure which
+                        notification actions different users will be able to receive. Once you have created
+                        a behavior group, you can assign it to an event type.'>
```
josejulio, line 72: Wording needs to be updated - here is where we assign the behavior group to an event type

admin-console/src/main/webapp/src/components/EventTypes/CreateEditModal.tsx
```diff
@@ -58,16 +66,21 @@ export const CreateEditModal: React.FunctionComponent<CreateEditModalProps> = (p
...
+                        <FormSelect
```
josejulio, line 73: We will need to use other component - it's possible to assign multiple behavior groups to a single event type - although maybe we don't want to overuse this. Lets leave it for now with 1 default behavior group per event.

admin-console/src/main/webapp/src/components/EventTypes/CreateEditModal.tsx
```diff
@@ -58,16 +66,21 @@ export const CreateEditModal: React.FunctionComponent<CreateEditModalProps> = (p
...
+                            value={ systemBehaviorGroup?.displayName }
+                            onChange={ handleSelect }
+                            id='system behavior group'
+                            name='system behavior group'
+                        >
+                            <FormSelectOption
+                                key={ systemBehaviorGroup?.id }
+                                value={ systemBehaviorGroup?.displayName }
```
josejulio, line 81: shouldn't the value be the id ?

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/CreateEditBehaviorGroupModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+                            aria-label="Select recipiect"
+                            placeholderText="Select a recipient"
+                            onToggle={ toggle }
+                            isOpen={ isOpen }
+                            selections={ selected }
+                            onSelect={ onSelect }
+                            menuAppendTo={ document.body }>
+                            { recipientOption }
+                        </Select>
+                    </FormGroup>
```
josejulio, line 96: We are not going to have a lot of available actions/recipients in here; maybe we can simplify this to a single Select ? 
`Send email to admin`, `Send email to all users`. In the future we might have some other options.

aferd: For sure.

admin-console/src/main/webapp/src/components/EventTypes/CreateEditModal.tsx
```diff
@@ -16,12 +19,19 @@ interface CreateEditModalProps {
...
     const [ eventType, setEventType ] = React.useState<Partial<EventType>>(props.initialEventType ?? {});
+    const [ systemBehaviorGroup, setSystemBehaviorGroup ] = React.useState<Partial<BehaviorGroup>>();
+    const getBehaviorGroups = useSystemBehaviorGroups();
 
     const handleChange = (value: string, event: React.FormEvent<HTMLInputElement> | React.FormEvent<HTMLTextAreaElement>) => {
         const target = event.target as HTMLInputElement;
         setEventType(prev => ({ ...prev, [target.name]: target.value }));
     };
 
+    const handleSelect = (value: string, event: React.FormEvent<HTMLSelectElement>) => {
```
aferd, line 30: would this be where I need to utilize the linking BG api?

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/CreateEditBehaviorGroupModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+                        </HelperTextItem></HelperText> : 'Enter a name for your group' }>
+                        <TextInput
+                            type='text'
+                            value={ systemBehaviorGroup.displayName }
+                            onChange={ handleChange }
+                            id='displayName'
+                            name="displayName"
+                        /></FormGroup>
+                    <FormGroup label='Action' fieldId='actions' isRequired>
+                        <Select
```
aferd, line 68: this won't dropdown..

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/DeleteBehaviorGroupModal.tsx
```diff
@@ -0,0 +1,58 @@
...
+    bundleName?: string;
+    systemBehaviorGroupName?: string;
+    isOpen: boolean;
+    onClose: () => void;
+
+}
+export const DeleteBehaviorGroupModal: React.FunctionComponent<DeleteBehaviorGroupModalProps> = (props) => {
+    const [ errors, setErrors ] = React.useState(true);
+
+    const onDelete = React.useCallback(async () => {
```
aferd, line 17: I will need to add checks here for unlinking any event-types before deleting.

aferd:
I still need to set something up so that when I create and get behavior groups, they are only being assigned and coming from said bundle. Would this make sense to do in the create and get BG services?

josejulio:
> I still need to set something up so that when I create and get behavior groups, they are only being assigned and coming from said bundle. Would this make sense to do in the create and get BG services?

I guess that for now you can filter the bg in the UI
i.e. `bgs.filter(bg => bg.bundle_id === current_bundle_id)`

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/DeleteBehaviorGroupModal.tsx
josejulio: You don't need to do the unlinking - upon deletion, they are unlinked.

admin-console/src/main/webapp/src/components/EventTypes/CreateEditModal.tsx
josejulio: No, I think you would do that on the `onSave`.

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/BehaviorGroupTable.tsx
```diff
@@ -0,0 +1,182 @@
...
+        setSystemBehaviorGroup(b);
+
+    };
+
+    const deleteBehaviorGroupModal = (b: BehaviorGroup) => {
+        setShowDeleteModal(true);
+        setSystemBehaviorGroup(b);
+    };
+
+    const handleUpdate = React.useCallback((systemBehaviorGroup) => {
```
aferd, line 65: I don't feel confident in this doing what I think it should be doing. I would need to add the updateActions to the rest, but I can't quite put the pieces together.

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/BehaviorGroupTable.tsx
```diff
@@ -0,0 +1,182 @@
...
+                            </ToolbarContent>
+                        </Toolbar>
+                        <Tr>
+                            {columns.map((column, columnIndex) => (
+                                <Th key={ columnIndex }>{column}</Th>
+                            ))}
+                        </Tr>
+                    </Thead>
+                    <Tbody>
+                        { getBehaviorGroups.payload.value.map(b =>
```
aferd, line 161: Filtering the bg's I tried add it here, but it broke everything. I tried similar to what you suggested but I think my placement is wrong. `getBehaviorGroups.payload.value.filter( bg => bg.bundleId === bundleId (utilizing the useParams above)`

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/CreateEditBehaviorGroupModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+                        </HelperTextItem></HelperText> : 'Enter a name for your group' }>
+                        <TextInput
+                            type='text'
+                            value={ systemBehaviorGroup.displayName }
+                            onChange={ handleChange }
+                            id='displayName'
+                            name="displayName"
+                        /></FormGroup>
+                    <FormGroup label='Action' fieldId='actions' isRequired>
+                        <FormSelect
```
aferd, line 63: Still not allowing me to actually select an item and it show up in the table.

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -196,7 +212,9 @@ export const ApplicationPage: React.FunctionComponent = () => {
                             <Tr key={ e.id }>
                                 <Td>{ e.displayName }</Td>
                                 <Td>{ e.name }</Td>
-                                <Td>{ e.description }</Td>
+                                <Td>
+                                    <ListSystemBehaviorGroups />
```
aferd, line 216: I would need to also filter this by bundle and event type.. to only list the current associated behaviors.

aferd:
This is visually where things are .. might help with some of my questions above..

https://user-images.githubusercontent.com/86322239/164070312-564888ed-1832-48b8-9c19-2bd0daf9929c.mov

aferd:
This is visually where things are, it might help with some of my issues above.

https://user-images.githubusercontent.com/86322239/164070600-6349b6c1-ca5c-4204-a352-fde209dfe950.mov

aferd:
last commit fixes the error when editing the event type, it now loads properly, though the ability to add/remove event type is not there.

https://user-images.githubusercontent.com/86322239/164072515-b37a81e7-6310-42bf-b22b-9004336e2165.mov

aferd:
@josejulio could you check this out when you get a moment?

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/BehaviorGroupTable.tsx
josejulio: Could you explain ?

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/BehaviorGroupTable.tsx
josejulio: It could be easier if you split the data from the ui components. You can abstract the logic of the filtering in a separate component and just passing params to  a `BehaviorGroupTable` component.

Try to not mix the UI and the logic

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/CreateEditBehaviorGroupModal.tsx
josejulio: Could you elaborate? What is actually doing and what do you want here?

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
josejulio: You could pass the bundle id to the component to only request those components.

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/CreateEditBehaviorGroupModal.tsx
aferd: I want to be able to select an action from the drop down list of actions and it fill the field with the selected action. The dropdown shows list of actions but upon clicking on the action, it does not fill the field with selected action.

aferd: You can see it in the video below.

josejulio: Is it doing the API call to update the server?

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/BehaviorGroupTable.tsx
```diff
@@ -0,0 +1,170 @@
...
+    const newBehaviorGroup = useCreateSystemBehaviorGroup();
+    const deleteBehaviorGroupMutation = useDeleteBehaviorGroup();
+
+    const columns = [ 'System Behavior Group', 'Action' ];
+
+    const [ showModal, setShowModal ] = React.useState(false);
+    const [ showDeleteModal, setShowDeleteModal ] = React.useState(false);
+
+    const [ isEdit, setIsEdit ] = React.useState(false);
+    const [ isOpen, setIsOpen ] = React.useState(false);
```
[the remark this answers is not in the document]
josejulio: I also noticed that the line below specifies the behavior group to edit (if any). So maybe this should also include that attribute.

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/BehaviorGroupTable.tsx
```diff
@@ -0,0 +1,170 @@
...
+                                        onSubmit={ handleSubmit }
+                                        isLoading={ false }
+                                    />
+                                    <DeleteBehaviorGroupModal
+                                        onDelete={ handleDelete }
+                                        bundleName={ bundle?.displayName }
+                                        systemBehaviorGroupName={ systemBehaviorGroup.displayName }
+                                        isOpen={ showDeleteModal }
+                                        onClose={ onDeleteClose }
+                                    />
```
josejulio, line 138: While there is no requirement to put them in a specific place (as they are modals). I usually put them at the end of the component. 
For me it's easier to find them if I know that the modals are by the end - as they don't really share the visual place anywhere else in the DOM.

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/CreateEditBehaviorGroupModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <React.Fragment>
+            <Modal
+                variant={ ModalVariant.medium }
+                title={ `${ props.isEdit ? 'Update' : 'Create'} your System Behavior Group` }
+                isOpen={ props.showModal }
+                onClose={ props.onClose }
+            ><Form isHorizontal>
+                    <FormGroup label='Group Name' fieldId='displayName' isRequired
+                        helperText={ props.isEdit ? <HelperText><HelperTextItem variant="warning" hasIcon>
+                                                    If this field is modified it may affect exisiting behavior.
```
josejulio, line 53: This is not really true for behavior groups. The problem with event types, applications and bundles is that we reffer them by name in the events - but behavior groups are always reffered by ID.

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/DeleteBehaviorGroupModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+    };
+
+    return (
+        <React.Fragment>
+            <Modal variant={ ModalVariant.small } titleIconVariant="warning" isOpen={ props.isOpen }
+                onClose={ props.onClose }
+                title={ `Permanently delete ${ props.systemBehaviorGroupName }` }>
+                { <b>{ props.systemBehaviorGroupName }</b> } {`from ${ props.bundleName ? props.bundleName :
+                    <Spinner /> } will be deleted. You will no longer be able to assign this system behavior
+                    group to event types. `}
```
josejulio, line 43: We should add something like: `All event types will stop using this behavior group.`

josejulio, line 43: Switching to use the markup instead of the template literal
 ```suggestion
                { <b>{ props.systemBehaviorGroupName }</b> } from { props.bundleName ? props.bundleName :
                    <Spinner /> } will be deleted. You will no longer be able to assign this system behavior
                    group to event types.
```

File: admin-console/src/main/webapp/src/components/SystemBehaviorGroups/DeleteBehaviorGroupModal.tsx [hunk 18]
```diff
@@ -0,0 +1,59 @@
+import { ActionGroup, Button, Modal, ModalVariant, Spinner, TextInput } from '@patternfly/react-core';
+import React from 'react';
+
+import { BehaviorGroup } from '../../types/Notifications';
+
+interface DeleteBehaviorGroupModalProps {
+    onDelete: (systemBehaviorGroup?: BehaviorGroup) => Promise<boolean>;
+    bundleName?: string;
+    systemBehaviorGroupName?: string;
```
josejulio, line 9: These 2 should always be required. If we are going to delete something we should already know.

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/CreateEditBehaviorGroupModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export const CreateEditBehaviorGroupModal: React.FunctionComponent<CreateEditModalProps> = (props) => {
+
+    const [ systemBehaviorGroup, setSystemBehaviorGroup ] = React.useState<Partial<BehaviorGroup>>(props.initialSystemBehaviorGroup ?? {});
+
+    const actionOption = [
+        <FormSelectOption key={ 0 } isPlaceholder label='Choose an action' />,
+        <FormSelectOption key={ systemBehaviorGroup?.id } label='Send an email to Users: All' value={ systemBehaviorGroup?.id } />,
+        <FormSelectOption key={ systemBehaviorGroup?.id } label='Send an email to Users: Admins' value={ systemBehaviorGroup?.id } />
```
josejulio, line 26: Should not reuse keys for the same array. This key uniquely identifies **each** element in the array.
Also - I'm unsure what the value means here, but i guess it should have the intent of the action i.e.

```suggestion
        <FormSelectOption key="choose action" isPlaceholder label='Choose an action' />,
        <FormSelectOption key="email-all" label='Send an email to Users: All' value="email-all" />,
        <FormSelectOption key="email-admin" label='Send an email to Users: Admins' value="email-admin" />
```

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/CreateEditBehaviorGroupModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+                            id='displayName'
+                            name="displayName"
+                        /></FormGroup>
+                    <FormGroup label='Action' fieldId='actions' isRequired>
+                        <FormSelect
+                            id='actions'
+                            name='actions'
+                            value={ systemBehaviorGroup.actions }
+                            open={ props.isOpen }
+                            onChange={ handleChange }
```
josejulio, line 68: I guess this case should be handled differently. Depending on the selected value we should set the properties of the action (maybe in a separate object as is not part of the `create behavior group request`)
email-all = `{ ignore_preferences: false, only_admins: false }`
email-admins = `{ ignore_preferences: false, only_admins: true }`

Other combinations would be
email-all-ignore-preferences = `{ ignore_preferences: true, only_admins: false }`
email-admin-ignore-preferences = `{ ignore_preferences: true, only_admins: true }`

josejulio (changes requested):

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/DeleteBehaviorGroupModal.tsx
aferd: would the following make sense - `{ behaviorGroup } will be deleted. You will no longer be able to assign this system behavior group to event types and any associated event types will stop using this behavior group.`

admin-console/src/main/webapp/src/components/SystemBehaviorGroups/CreateEditBehaviorGroupModal.tsx
aferd: do you mean having a displayName field does not really make sense? because BG are referred by ID. Should we auto-generate an ID here instead maybe?

aferd: @josejulio thoughts on the above?

josejulio: Sorry for late reply - I meant that we don't need to display the warning here. 
It's fine being able to edit the `displayName`, lets just remove the warning.

josejulio (approved):
lgtm
Some details we need to check across the admin ui is to prevent repeating the schemas declaration. IMO we should only create a new type if we are going to change it.

aferd:
create, update(needs 1 small fix in separate PR) and delete system behavior group in admin UI.

https://user-images.githubusercontent.com/86322239/178593510-c23ba5b2-9305-430c-8311-49aa360484e5.mov

PR #1147, opened 2022-04-06
WIP: Add working state for dynamic metric tags

bennetelli:


bennetelli:
> Drafts can't be merged smiley

I wasn't sure if an admin could convert the draft to a normal PR. Just wanted to be sure :D

josejulio:
Updated the name and added the don't-merge label :grin: 
but no one should undraft a PR and merge it without the author's letting us know that is ready for review/merge :)

bennetelli:
safety first :P

bennetelli:
probably not important, but better putting it in here in case someone wants to try it.
when you start  prometheus in a docker container and the application with mvn compile quarkus:dev, then the metrics wont be exposed. You either need to change the /etc/docker/daemon.json file and the prometheus.yml config. Or even download prometheus the normal way and edit the prometheus.yml.

the easiest thing you can do is to add this job here:
  - job_name: "quarkus"

    # metrics_path defaults to '/metrics'
    # scheme defaults to 'http'.

    static_configs:
      - targets: ["localhost:8087"]      
      
hint: change the port from 8087 to your quarkus application url. otherwise the quarkus app would expose the metrics to dev/null

bennetelli:
With that config above, the metrics will be exposed

image

bennetelli:
* We are exposing the metrics to prometheus prod and stage (when you click on prometheus prod then "status/targets" there are the three pods/urls listed for notifications-backend. You can't ping the url and watch the metrics in the browser, but they are exposed
* But when you do the same as I did locally, on prometheus prod or stage. Then the tags which are exposed locally, are not present anymore.

bennetelli:
app-sre is just figuring out if the metrics/tags are also visible at the application/metrics logs of our pods. ticket is ASIC-203.

bennetelli:
and that's what is exposed on stage or prod:

# HELP processor_email_processed_total  
# TYPE processor_email_processed_total counter
processor_email_processed_total XXXXXX

gwenneg:
Did you try to use dynamic tags like the one we have in `EventConsumer`? I know it's not elegant because it requires a `Timer`, but at least it works out of the box.

bennetelli:
@gwenneg should we keep this WIP? I could add a jira to have it in mind that we need to change the timer to a counter later in the future and add the link to this (closed) PR as well, so everyone could reproduce it if needed.

PR #1148, opened 2022-04-07
NOTIF-484 Prevent deletion of included templates

gwenneg:


josejulio (approved):

josejulio:
When we have more data from qute, we could also implement the other way - prevent from creating a template if it includes non existent templates.

PR #1155, opened 2022-04-11
Update SSO for OpenBridge integration

pilhuhn:


common/src/main/java/com/redhat/cloud/notifications/openbridge/BridgeAuthService.java
```diff
@@ -14,12 +14,17 @@
...
 
     @POST
     @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
     @Produces(MediaType.APPLICATION_JSON)
-    Map<String, Object> getTokenStruct(String body,
-         @HeaderParam("Authorization") String authHeader);
+    Map<String, Object> getTokenStructWithClientPassword(String body,
+                                                         @HeaderParam("Authorization") String authHeader);
+
+    @POST
```
gwenneg, line 26: Does this method and the other one target the same remote endpoint? I'm asking because there's no `@Path` on any of the methods here.

pilhuhn: It is the same endpoint. One of the two will go away in the future.

common/src/main/java/com/redhat/cloud/notifications/openbridge/BridgeAuth.java
```diff
@@ -13,4 +13,15 @@ public BridgeAuth(String tmp) {
...
         return token;
     }
+
+    @Override
+    public String toString() {
+        final StringBuilder sb = new StringBuilder("BridgeAuth{");
+        sb.append("token='");
+        sb.append(token.substring(0, 10));
+        sb.append("...").append(token.substring(token.length() - 10))
+                        .append('\'');
```
gwenneg, line 23: How long is the token? Does this code hide enough chars to avoid leaking it? I'm sure the answer is yes but I still have to ask 😃

.rhcicd/clowdapp-backend.yaml
```diff
@@ -145,14 +145,10 @@ objects:
...
-        - name: OB_KCPASS
-          value: ${OB_KCPASS}
-        - name: OB_TOKEN_USER
-          value: ${OB_TOKEN_USER}
-        - name: OB_TOKEN_PASS
-          value: ${OB_TOKEN_PASS}
+        - name: OB_TOKEN_CLIENT_ID
+          value: ${OB_TOKEN_CLIENT_ID}
+        - name: OB_TOKEN_CLIENT_SECRET
+          value: ${OB_TOKEN_CLIENT_SECRET}
```
gwenneg, line 151: Gentle reminder (for later): this should be stored in the vault.

common/src/main/java/com/redhat/cloud/notifications/openbridge/BridgeAuth.java
pilhuhn: Something like 2k.

.rhcicd/clowdapp-backend.yaml
pilhuhn: Yes, I agree

gwenneg (approved):
LGTM

PR #1162, opened 2022-04-12
NOTIF-598 Add EventTypeBehavior integrity check

gwenneg:
This PR adds an integrity check which verifies that an event type and a behavior group have the same bundle parent before associating them in the `event_type_behavior` table.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
@@ -192,7 +192,6 @@ public class BehaviorGroupRepository {
      * Returns false if the event type was not found.
      * If an exception other than NoResultException is thrown during the update, the DB transaction will be rolled back.
      */
-
     @Transactional
     public boolean updateEventTypeBehaviors(String accountId, UUID eventTypeId, Set<UUID> behaviorGroupIds) {
         // First, let's make sure the event type exists.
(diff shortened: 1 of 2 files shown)


backend/src/test/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepositoryTest.java
```diff
@@ -174,6 +174,12 @@ void testAddAndDeleteEventTypeBehavior() {
         updateAndCheckEventTypeBehaviors(DEFAULT_ACCOUNT_ID, eventType.getId(), true, behaviorGroup2.getId());
         updateAndCheckEventTypeBehaviors(DEFAULT_ACCOUNT_ID, eventType.getId(), true, behaviorGroup1.getId(), behaviorGroup2.getId(), behaviorGroup3.getId());
         updateAndCheckEventTypeBehaviors(DEFAULT_ACCOUNT_ID, eventType.getId(), true);
+
+        // If the event type bundle and the behavior group bundle do not match, the behavior should not be added.
+        Bundle otherBundle = resourceHelpers.createBundle();
+        Application otherApp = resourceHelpers.createApplication(otherBundle.getId());
+        EventType otherEventType = createEventType(otherApp.getId());
+        updateAndCheckEventTypeBehaviors(DEFAULT_ACCOUNT_ID, otherEventType.getId(), true);
```
gwenneg, line 182: WIP

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
```diff
@@ -226,7 +225,11 @@ public boolean updateEventTypeBehaviors(String accountId, UUID eventTypeId, Set<
                  */
                 String insertQuery = "INSERT INTO event_type_behavior (event_type_id, behavior_group_id, created) " +
                         "SELECT :eventTypeId, :behaviorGroupId, :created " +
-                        "WHERE EXISTS (SELECT 1 FROM behavior_group WHERE account_id = :accountId AND id = :behaviorGroupId) " +
+                        "WHERE EXISTS (" +
+                            "SELECT 1 FROM behavior_group bg, applications a, event_type et " +
+                            "WHERE bg.account_id = :accountId AND bg.id = :behaviorGroupId " +
+                            "AND bg.bundle_id = a.bundle_id AND a.id = et.application_id AND et.id = :eventTypeId" +
```
gwenneg, line 231: This is a short-term fix for the issue raised on Slack but I think we should revisit this fix later. Ideally, the endpoint should return a 400 with a message containing the exact cause (i.e. the behaviorGroupIds that weren't accepted here).

However, the validation has to be smart enough to avoid putting too much pressure on the DB. If possible, we should avoid iterating over the behaviorGroupIds list and run a SQL query to validate each one of them.

pilhuhn:
I am not convinced that this is the correct behaviour overall. A BG should be the m:n relation of M*(b/a/et) <-> N*(integration) meaning that if I want to associate an integration with all incoming events (for a tenant) it should be doable with one BG and not one per Bundle.

gwenneg:
@pilhuhn What you wrote makes sense, but that change would raise some questions:
- Where do we put the BGs in the frontend if not in the bundles? A new separate page?
- How do we deal with default/system BGs with were designed to be set at bundle level?

gwenneg:
@josejulio ⬆️

pilhuhn:
For the 1st bullet, I would keep as is. Meaning the full set of BGs would be displayed in each Bundle. For the 2nd item, I believe there are others that are more qualified to answer :)

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
josejulio: Could we just do a query like (outside the loop) `SELECT COUNT(*) FROM behavior_group bg WHERE bg.id IN (:behaviorGroupId)` and throw a bad request if the count is different than the input.

Additionaly, if we want to provide a more detailed error we could use a similar query fetching the behavior group.

josejulio:
I guess is OK to first fix the issue and then we can discuss if we want to change the way the behavior groups work now.

I'm not against the idea - but our overall design is right now based on behavior groups being per bundle.

vkrizan:
Created a Jira card out of my report from Slack: NOTIF-598

josejulio (approved):
lgtm - a separate PR can be added to give a more detailed error
NOTIF-598

gwenneg:
@josejulio I pushed a better integrity check which relies on a single SQL query and returns a 400 error with a detailed message.

josejulio (approved):
nice !

PR #1166, opened 2022-04-13
Add valid json check

bennetelli:
can't execute the tests in engine locally.. this one will probably fail.


bennetelli:
doesnt make sense to change things here. I can't execute tests locally.

PR #1168, opened 2022-04-13
Support external API path for frontend

BlakeHolifield:
Adds `apiPath` to fix frontend <--> backend networking in Ingress. Without this field, Clowder will use the app name `notifications-backend-service` by default.


BlakeHolifield:
@josejulio this should fix the demo for Gabor. Good catch!

josejulio (approved):

PR #1170, opened 2022-04-14
NOTIF-381 Remove RBAC user service calls

bennetelli:


engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -123,36 +121,21 @@ public List<User> getUsers(String accountId, boolean adminsOnly) {
...
+        do {
+            ITUserRequest request = new ITUserRequest(accountId, adminsOnly, firstResult, maxResultsPerPage);
+            usersPaging = retryOnItError(() -> itUserService.getUsers(request));
+            usersTotal.addAll(usersPaging);
 
-                firstResult += maxResultsPerPage;
+            firstResult += maxResultsPerPage;
+        } while (usersPaging.size() == maxResultsPerPage);
 
-                getUsersPageTimer.stop(meterRegistry.timer("rbac.get-users.page", "accountId", accountId));
```
gwenneg, line 141: Is this no longer needed?

bennetelli: Thought nobody would see any advantages in having it. If you would like to keep it, then I can add it again :)

gwenneg (approved):
LGTM

PR #1174, opened 2022-04-19
Fix NPE in endpoints REST API

gwenneg:
This PR should fix the following exception which reached our Sentry channel on stage:
```
java.lang.NullPointerException: null
    at java.util.Objects.requireNonNull(Objects.java:221)
    at java.util.ImmutableCollections$AbstractImmutableList.indexOf(ImmutableCollections.java:170)
    at java.util.ImmutableCollections$AbstractImmutableList.contains(ImmutableCollections.java:201)
    at com.redhat.cloud.notifications.routers.EndpointResource.checkSystemEndpoint(EndpointResource.java:447)
    at com.redhat.cloud.notifications.routers.EndpointResource.deleteEndpoint(EndpointResource.java:269)
```
ℹ️ I checked the DB with `gabi`, all endpoints have a non-null `endpoint_type` field on stage.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
@@ -93,18 +93,6 @@ public class EndpointRepository {
         return endpoints;
     }
 
-    public EndpointType getEndpointTypeById(String accountId, UUID endpointId) {
-        String query = "Select e.compositeType.type from Endpoint e WHERE e.accountId = :accountId AND e.id = :endpointId";
-        try {
-            return entityManager.createQuery(query, EndpointType.class)
-                    .setParameter("accountId", accountId)
-                    .setParameter("endpointId", endpointId)
-                    .getSingleResult();
-        } catch (NoResultException e) {
-            return null;
-        }
-    }
-
     @Transactional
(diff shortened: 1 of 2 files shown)


backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -93,18 +93,6 @@ public List<Endpoint> getEndpointsPerCompositeType(String accountId, @Nullable S
...
-        try {
-            return entityManager.createQuery(query, EndpointType.class)
-                    .setParameter("accountId", accountId)
-                    .setParameter("endpointId", endpointId)
-                    .getSingleResult();
-        } catch (NoResultException e) {
-            return null;
-        }
-    }
-
```
josejulio, line 107: Does this method has a problem? or why it was removed / replaced? 

This method prevent's a `loadProperties` call. The same `endpointType == null` could be used or this could even throw a `NotFoundException`.

gwenneg: Good catch, I didn't notice that the other one was loading the properties.

I'll update the PR with a better fix.

josejulio (approved):
lgtm

PR #1173, opened 2022-04-14
Add more templates for Rbac

astrozzc:
Email templates added for each of the event types.

PR #1178, opened 2022-04-19
NOTIF-484 Update migration code for RBAC templates

gwenneg:
Related to #1173.


gwenneg:
ℹ️ I noticed a few issues, fixing them.

PR #1184, opened 2022-04-20
Add gauge for the actual count of users used

bennetelli:
coming from a discussion with @josejulio some minutes ago. We would like to get to know how many users we are fetching from IT User Service and how many of them we are actually using after fetching. With this gauge we can compare both.

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/recipients/RecipientResolver.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/recipients/RecipientResolver.java
@@ -60,7 +60,7 @@ public class RecipientResolver {
             users = filterUsers(users, subscribers);
         }
 
-        usersCount.set(users.size());
+        updateUsersUsedGauge(users.size());
 
         return users;
     }
@@ -73,5 +73,9 @@ public class RecipientResolver {
                             .anyMatch(requested -> requested.equalsIgnoreCase(user.getUsername()))
                 )
                 .collect(Collectors.toSet());
+    }
+
+    private void updateUsersUsedGauge(int users) {
+        usersCount.set(users);
     }
 }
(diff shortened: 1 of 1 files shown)


engine/src/main/java/com/redhat/cloud/notifications/recipients/RecipientResolver.java
```diff
@@ -47,6 +60,8 @@ private Set<User> recipientUsers(String accountId, RecipientSettings request, Se
             users = filterUsers(users, subscribers);
         }
 
+        updateUsersUsedGauge(users.size());
```
josejulio, line 63: could we add the accountId somehow? I would like to compare how many we fetched on particular accounts. IIRC the rbac~ metrics do that too?

bennetelli: hmm.. that's actually not really comparable I would say.

comparing the new metric from this PR with this one is more fair IMO: https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java#L95

So could we maybe leave it as it is and see if it makes sense?
Otherwise I would have to add a timer.start and timer.stop probably which we would not use.

josejulio (approved):

engine/src/main/java/com/redhat/cloud/notifications/recipients/RecipientResolver.java
bennetelli: Thanks :) otherwise it would have become a bit hacky. Let's give it a try and if it doesnt do the job, we can still change it.

PR #1186, opened 2022-04-21
NOTIF-195 Encode endpoint property

gwenneg:
ℹ️ DB data before this PR:
| Environment | `endpoint_webhooks` | `camel_properties` |
| - | - | - |
| Stage | 0 record with `basic_authentication` data  | 29 records with empty/null `basic_authentication` data and 2 records with non empty `basic_authentication` data |
| Prod | 0 record with `basic_authentication` data | 13 records with empty/null `basic_authentication` data |

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacRestClientRequestFilter.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacRestClientRequestFilter.java
@@ -1,12 +1,11 @@
 package com.redhat.cloud.notifications.auth.rbac;
 
+import com.redhat.cloud.notifications.Base64Utils;
+
 import javax.ws.rs.client.ClientRequestContext;
 import javax.ws.rs.client.ClientRequestFilter;
 import java.io.IOException;
 import java.net.URI;
-import java.util.Base64;
-
-import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
  * Filter to optionally add data on outgoing requests to the RBAC service.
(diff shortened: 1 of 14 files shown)


database/src/main/resources/db/migration/V1.49.0__NOTIF-195_encoded_basic_auth.sql
```diff
@@ -0,0 +1,6 @@
+-- The following field can be safely wiped on stage and prod.
+-- Prod does not contain any 'basic_authentication' value at the time when this script is written.
+-- Stage has two records with 'basic_authentication' data but these records where created during tests from the REST API.
+-- The frontend does not allow (yet) to change the basic authentication settings.
+
+UPDATE camel_properties SET basic_authentication = null;
```
gwenneg, line 6: Waiting for the confirmation that this is a safe query.

gwenneg, line 6: Waiting for confirmation that this is a safe move.

gwenneg: Save move confirmed on Slack ✅

backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/RbacRestClientRequestFilter.java
```diff
@@ -19,7 +18,7 @@ public class RbacRestClientRequestFilter implements ClientRequestFilter {
     public RbacRestClientRequestFilter() {
         String tmp = System.getProperty("develop.exceptional.user.auth.info");
         if (tmp != null && !tmp.isEmpty()) {
-            authInfo = new String(Base64.getEncoder().encode(tmp.getBytes(UTF_8)), UTF_8);
+            authInfo = Base64Utils.encode(tmp);
```
josejulio, line 21: Should we remove the if ? I see that the `Base64Utils` already considers nulls

gwenneg: An empty String is encoded/decoded to an empty String with Base64 so we don't really need a special condition for that in Base64Utils.

josejulio: I didn't meant that. I was questioning if we need to remove the outer `if` (line 21-20), but current code is fine

josejulio (approved):

PR #1191, opened 2022-04-26
NOTIF-628 Return 404 when REST resource is not found

gwenneg:


josejulio (approved):

josejulio:
Did a smoke test with the uI and didn't see anything wrong .

gwenneg:
Thanks!

PR #1192, opened 2022-04-26
Allow to update a processor and use templates from DB.

pilhuhn:
A first cut at updating of OB-Processors. 
This changeset also includes a rewrite on OB handling as well as getting the transformation template from DB.
Updating the action part of a processor is unfortunately not yet supported by OB, so changing the slack-channel is not yet possible.


backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -452,51 +539,65 @@ private static void checkSystemEndpoint(EndpointType endpointType) {
...
 
-    private String setupOpenBridgeProcessor(Endpoint endpoint, CamelProperties properties, String processorName) {
 
-        Map<String, Object> out = new HashMap<>();
-        out.put("name", processorName);
+    /*
+     * Creates a template processor object from the provided
+     * endpoint.
+     */
+    private Processor createProcessor(Endpoint endpoint) {
```
gwenneg, line 547: How will processors work with user-defined templates? If each user is allowed to chose his template, does that mean that we will create one processor for each user template?

common/src/main/java/com/redhat/cloud/notifications/models/GenericTemplate.java
```diff
@@ -0,0 +1,102 @@
...
+    /*
+     * Kind of a template. Ordering is from fallback/default to most specific.
+     * The idea is that the caller can request a kind and if there is no
+     * template available, the one of a less specific type is taken.
+     */
+    public enum TemplateKind {
+        DEFAULT, // A fallback if nothing more specific is defined
+        APPLICATION, // specific for one (sending) application
+        EVENT_TYPE, // specific for an even_type of an application
+        USER  // user defined
```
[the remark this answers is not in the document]
pilhuhn: Renamed USER to ORG to clarify that this is per org.  A USER one could later be added below to make this even more spefific

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
```diff
@@ -295,4 +298,46 @@ private Template findTemplate(UUID id, String notFoundMessage) {
...
+            query.setParameter("appName", appName);
+        }
+        if (accountId != null) {
+            query.setParameter("accountId", accountId);
+        }
+
+        List<GenericTemplate> templates = query.getResultList();
+
+        if (templates.isEmpty()) {
+            if (templateKind.equals(GenericTemplate.TemplateKind.USER) && accountId != null) {
```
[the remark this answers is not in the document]
pilhuhn: For the other ones it would already fall back by the way it is done. for USER (now ORG) we need to remove the accountId before feeding back into the query for the fallbacks

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
pilhuhn: That may be the consequence.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
```diff
@@ -295,4 +298,46 @@ private Template findTemplate(UUID id, String notFoundMessage) {
...
+
+
+    public Optional<GenericTemplate> findGenericTemplate(String appName,
+                                                         String accountId,
+                                                         GenericTemplate.TemplateKind templateKind,
+                                                         String integrationType) {
+        String hql = "FROM GenericTemplate gt JOIN FETCH gt.theTemplate " +
+                "WHERE gt.templateKind <= :templateKind " +
+                "AND gt.integrationType = :iType " +
+                (accountId != null ? "AND gt.accountId = :accountId " : " ") +
```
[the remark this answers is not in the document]
pilhuhn: 👍

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -173,40 +192,27 @@ public Endpoint createEndpoint(@Context SecurityContext sec, @NotNull @Valid End
...
-                properties.getExtras().put(OB_PROCESSOR_ID, processorId);
+                UnaryOperator<Processor> execFun = in -> bridgeApiService.addProcessor(bridgeId, token, in);
+                UnaryOperator<Processor> updateFun = in -> { endpoint.setStatus(EndpointStatus.PROVISIONING); return in; };
+                Processor out = executeObAction(endpoint, updateFun, execFun);
+                // We need to record the id of that processor, as we need it later
 
+                cp.getExtras().put(OB_PROCESSOR_ID, out.getId());
+            } catch (Exception e) {
+                Log.error("Adding a processor failed ", e);
+                throw new InternalServerErrorException();
```
gwenneg, line 213: Should we add an exception message here to explain the client why an HTTP 500 status was returned?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -264,30 +276,20 @@ public Response deleteEndpoint(@Context SecurityContext sec, @PathParam("id") UU
...
+
+            try {
+                UnaryOperator<Processor> execFun = in -> {
+                    bridgeApiService.deleteProcessor(bridge.getId(), in.getId(), bridgeAuth.getToken());
+                    return null;
+                };
+                UnaryOperator<Processor> updateFun = in -> { ep.setStatus(EndpointStatus.DELETING); return in; };
+                executeObAction(ep, updateFun, execFun);
+            } catch (Exception ex) {
+                Log.warn("Deletion of processor failed: ", ex);
```
gwenneg, line 290: This one is logged at `warn` level while the other two exceptions (during creation and update) are logged at `error`. Is it because the deletion is less important?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -155,6 +165,15 @@ public EndpointPage getEndpoints(
             count = endpointRepository.getEndpointsCount(principal.getAccount(), name);
         }
 
+        for (Endpoint endpoint: endpoints) {
+            if (isForOpenBridge(endpoint)) {
+                // Don't return the processor info, it is internal only
```
gwenneg, line 170: Do we have a test that verifies these fields are never returned to the client?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -337,10 +339,98 @@ public Response updateEndpoint(@Context SecurityContext sec, @PathParam("id") UU
...
+                UnaryOperator<Processor> updateFun = in -> {
+                    in.getAction().getParameters().put(SLACK_CHANNEL, ref.channel);
+                    in.getAction().getParameters().put(SLACK_WEBHOOK_URL, ref.url);
+                    in.setTransformationTemplate(in.getTransformationTemplate() + " ---- updated"); // TODO remove
+                    ep.setStatus(EndpointStatus.PROVISIONING);
+                    return in;
+                };
+                UnaryOperator<Processor> execFun = in -> bridgeApiService.updateProcessor(bridge.getId(), in.getId(), bridgeAuth.getToken(), in);
+                Processor out = executeObAction(ep, updateFun, execFun);
+                Log.infof("Endpoint updated to processor %s ", out.toString());
```
gwenneg, line 380: `executeObAction` can return `null` so it looks like this could throw a NPE.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -264,30 +276,20 @@ public Response deleteEndpoint(@Context SecurityContext sec, @PathParam("id") UU
...
+        Endpoint ep = endpointRepository.getEndpoint(principal.getAccount(), id);
+        if (isForOpenBridge(ep)) {
+
+            try {
+                UnaryOperator<Processor> execFun = in -> {
+                    bridgeApiService.deleteProcessor(bridge.getId(), in.getId(), bridgeAuth.getToken());
+                    return null;
+                };
+                UnaryOperator<Processor> updateFun = in -> { ep.setStatus(EndpointStatus.DELETING); return in; };
+                executeObAction(ep, updateFun, execFun);
```
gwenneg, line 288: The code would be easier to understand/maintain if it looked like this:
```suggestion
                ep.setStatus(EndpointStatus.DELETING);
                executeObAction(ep, null, execFun); // TODO Handle null updateFun in executeObAction
```

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -173,40 +192,27 @@ public Endpoint createEndpoint(@Context SecurityContext sec, @NotNull @Valid End
...
+                cp.getExtras().put(OB_PROCESSOR_NAME, processorName);
 
-                // TODO find a better place for these, that should not be
-                //       visible to users / OB actions
-                //       See also CamelTypeProcessor#callOpenBridge
-                //            and ReadyCheck#execute
-                properties.getExtras().put(OB_PROCESSOR_ID, processorId);
+                UnaryOperator<Processor> execFun = in -> bridgeApiService.addProcessor(bridgeId, token, in);
+                UnaryOperator<Processor> updateFun = in -> { endpoint.setStatus(EndpointStatus.PROVISIONING); return in; };
+                Processor out = executeObAction(endpoint, updateFun, execFun);
```
gwenneg, line 207: The code would be easier to understand/maintain if it looked like this:
```suggestion
                endpoint.setStatus(EndpointStatus.PROVISIONING);
                Processor out = executeObAction(endpoint, updateFun, execFun); // TODO Handle null updateFun in executeObAction
```

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -337,10 +339,98 @@ public Response updateEndpoint(@Context SecurityContext sec, @PathParam("id") UU
...
+                    ref.url = cp.getUrl();
+                }
+            }
+
+            try {
+                UnaryOperator<Processor> updateFun = in -> {
+                    in.getAction().getParameters().put(SLACK_CHANNEL, ref.channel);
+                    in.getAction().getParameters().put(SLACK_WEBHOOK_URL, ref.url);
+                    in.setTransformationTemplate(in.getTransformationTemplate() + " ---- updated"); // TODO remove
+                    ep.setStatus(EndpointStatus.PROVISIONING);
```
gwenneg, line 375: Similarly to my two previous comments, this line could be moved out of the `UnaryOperator`.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -337,10 +339,98 @@ public Response updateEndpoint(@Context SecurityContext sec, @PathParam("id") UU
...
+     */
+    Processor executeObAction(Endpoint endpoint, UnaryOperator<Processor> updateFunc, UnaryOperator<Processor> executeFunc) {
+        if (endpoint != null) {
+            EndpointProperties properties = endpoint.getProperties();
+            if (properties != null && properties instanceof CamelProperties) {
+                CamelProperties cp = (CamelProperties) properties;
+                // Special case wrt OpenBridge
+                if (endpoint.getSubType().equals(SLACK)) {
+                    String processorId = cp.getExtras().get(OB_PROCESSOR_ID);
+                    Processor p = createProcessor(endpoint);
```
gwenneg, line 409: Could we avoid creating a processor here when this method is called to delete the processor on the OpenBridge side? The only data that is used in that case is the `processorId`, everything else is unused.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -448,51 +538,65 @@ private static void checkSystemEndpoint(EndpointType endpointType) {
...
 
-        String token = bridgeAuth.getToken();
+        List<Processor.Filter> filters = out.getFilters();
+        filters.add(mkFilter("StringEquals", "rhaccount", endpoint.getAccountId()));
+        filters.add(mkFilter("StringEquals", OB_PROCESSOR_NAME, processorName));
 
-        Map<String, Object> pInfo = bridgeApiService.addProcessor(bridge.getId(), token, out);
+        Processor.Action action = new Processor.Action(SLACK_ACTION);
+        Map<String, String> props = action.getParameters();
+        props.put(SLACK_CHANNEL, properties.getExtras().getOrDefault(SLACK_CHANNEL, "#general"));
```
gwenneg, line 564: Should the default channel be configurable?

pilhuhn: A fresh slack instance always comes with #random and #general   I don't think picking either on notifcations global level via config entry makes a change to hardcoding it here. And per customer seems like overkill to me. 
We could/should perhaps add some validation rules in out API that for slack the channel always needs to be provided. Likewise for other upcoming types.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
pilhuhn: I don't see deletes happen too often, so the memory/garbage collection overhead should be low. 
Or did you mean something else?

common/src/main/java/com/redhat/cloud/notifications/openbridge/Processor.java
```diff
@@ -0,0 +1,213 @@
...
+ *     "type": "string",
+ *     "parameters": {
+ *       "additionalProp1": "string"
+ *     }
+ *   }
+ * }
+ */
+public class Processor {
+
+    @JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
```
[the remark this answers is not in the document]
pilhuhn: Rather the other way around. We retrieve them from OB, but do never send them that way.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
pilhuhn: I did not add a message as to not give an attacker a hint.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
pilhuhn: Yes. A failed delete will worst case consume resources. We also had cases in the past where OB-internal confusion prevented the delete and a human needed to clean them up by hand.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
pilhuhn: No. As they are removed here, they should not be returned. I am probably missing something?

common/src/main/java/com/redhat/cloud/notifications/openbridge/BridgeEventService.java
```diff
@@ -6,16 +6,15 @@
...
 /**
  * Deliver events to the OpenBridge ingress
  */
-@Path("/events")
+@Path("/")
 public interface BridgeEventService {
 
     @POST
-    @Consumes(MediaType.APPLICATION_JSON)
+    @Consumes("application/cloudevents+json")
```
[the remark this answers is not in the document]
gwenneg: My bad. This is correct indeed.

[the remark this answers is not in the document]
pilhuhn: I think the semantics of this annotation (that came from the server side) is confusing, as one would expect mirrored behaviour, as I also did in the past.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
gwenneg: The current code is fine, you didn't miss anything. What I meant is that we could add a test to make sure this information isn't leaked unintentionally in the future.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
gwenneg: I'm more worried by the code complexity than the memory/GC 😃.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
gwenneg: When I wrote "configurable", I was thinking of something coming from the DB as it would indeed make no sense to configure that from `application.properties`.

Enforcing a non null/empty Slack channel sounds more reasonable 👍

common/src/main/java/com/redhat/cloud/notifications/models/GenericTemplate.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /*
+     * Kind of a template. Ordering is from fallback/default to most specific.
+     * The idea is that the caller can request a kind and if there is no
+     * template available, the one of a less specific type is taken.
+     */
+    public enum TemplateKind {
+        DEFAULT, // A fallback if nothing more specific is defined
+        APPLICATION, // specific for one (sending) application
+        EVENT_TYPE, // specific for an even_type of an application
```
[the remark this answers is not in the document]
pilhuhn: Something to consider for the future. Not in the 1st round though

gwenneg:
I fixed conflicts with https://github.com/RedHatInsights/notifications-backend/pull/1329.

gwenneg (approved):
LGTM

PR #1194, opened 2022-04-27
Put old Junit 4 Test annotation on the forbidden list.

pilhuhn:
This is to help developers to make sure the Jupiter `@Test` is used and not the one from Junit4


bennetelli (changes requested):
I guess we should also add these entries:

org.junit.Before
org.junit.After
org.junit.BeforeClass
org.junit.AfterClass
org.junit.Assert

not sure if we can use wildcards here. If yes, then I would use 
org.junit.Assert* so we cover everything.

bennetelli (approved):

PR #1196, opened 2022-04-27
Log which Garbage Collector is used

bennetelli:


gwenneg:
Isn't `Xlog:gc` used to print the GC logs rather than logging the GC name itself?

docker/Dockerfile.notifications-aggregator.jvm
```diff
@@ -26,7 +26,7 @@ RUN microdnf install openssl curl ca-certificates ${JAVA_PACKAGE} \
     && chown 1001:root /deployments \
     && echo "securerandom.source=file:/dev/urandom" >> /etc/alternatives/jre/lib/security/java.security
 
-ENV JAVA_OPTIONS="-Djava.util.logging.manager=org.jboss.logmanager.LogManager -XX:+ExitOnOutOfMemoryError"
+ENV JAVA_OPTIONS="-Djava.util.logging.manager=org.jboss.logmanager.LogManager -XX:+ExitOnOutOfMemoryError -Xlog:gc"
```
pilhuhn, line 29: I think this may not be what you want.
(I think) This flag makes the jvm print all(?) gc related events to stdout, not only the used collector

e.g.

```
❯ java -Xlog:gc target/notifications-common-1.0.0-SNAPSHOT-tests.jar
[0.014s][info][gc] Using G1
[0,738s][info][gc] GC(0) Pause Young (Normal) (G1 Evacuation Pause) 24M->2M(512M) 3,527ms
```

-XX:+PrintCommandLineFlags is probably what you are looking for

```
❯ java -XX:+PrintCommandLineFlags target/notifications-common-1.0.0-SNAPSHOT-tests.jar
-XX:G1ConcRefinementThreads=10 -XX:GCDrainStackTargetSize=64 -XX:InitialHeapSize=536870912 -XX:MaxHeapSize=8589934592 -XX:+PrintCommandLineFlags -XX:ReservedCodeCacheSize=251658240 -XX:+SegmentedCodeCache -XX:+UseCompressedClassPointers -XX:+UseCompressedOops -XX:+UseG1GC
```

pilhuhn (changes requested):

bennetelli:
> Isn't `Xlog:gc` used to print the GC logs rather than logging the GC name itself?



> Isn't `Xlog:gc` used to print the GC logs rather than logging the GC name itself?

the gc dev said:"this will run permanently, but with close to zero overhead".
As far as I know, we can't attach to prod via JMX or JFR/JMC. So this is the only option I know how we could find out which GC is used.

gwenneg:
This will only be deployed on stage, right?

I'm OK with merging this as long as the commit is reverted as soon as we get the GC name.

bennetelli:
> This will only be deployed on stage, right?
> 
> I'm OK with merging this as long as the commit is reverted as soon as we get the GC name.

I am fine with deploying it on stage only and reverting it after we know the answer of which gc we are using.
The pattern in Grafana looks like we are using the serial gc and this is just to check that again. 

As Roman said, this has close to zero overhead. So basically we could also run it on prod if we want.
I could be that different GCs are used on stage and prod, but not sure if that's important for us to know.

pilhuhn:
Are we sure that stage is close enough to prod that it uses the same memory/cpu settings? If so we could deploy+revert.
The overhead is not only on the jvm, but also on the container logging+ kubernetes log gathering etc.

gwenneg:
Yeah I'm not convinced about the "close to zero" overhead, but we could still deploy on prod if it's for a short duration.

PR #1202, opened 2022-05-02
Fix SQL endpoint queries conditions

gwenneg:
`IS TRUE` should always be preferred over `= TRUE`:
| Field value | `= TRUE` evaluation result | `IS TRUE` evaluation result |
| - | - | - |
| null | null | false |
| true | true | true |
| false | false | false |


josejulio:
Is there a case where `enabled` is null ?  or this is a preventive thing?

gwenneg:
Preventive :smiley: Jerome raised an issue on Slack earlier today which made me look at this code. It turns out that his issue was actually caused by a user error.

josejulio (approved):
lgtm

vkrizan:
Why not set the field to be `NOT NULL` instead?

gwenneg:
It may already be `NOT NULL`, I didn't check. Faulty SQL must be fixed anyway :smiley:

PR #1211, opened 2022-05-05
NOTIF-628 Return 404 when id not found

Mishrasubha:
The endpoints when passed an invalid ID with the correct "ID" format return an empty list with a 200 status code. should return a 404 "Not Found" exception.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
@@ -3,6 +3,7 @@ package com.redhat.cloud.notifications.db.repositories;
 import com.redhat.cloud.notifications.db.Query;
 import com.redhat.cloud.notifications.models.BehaviorGroup;
 import com.redhat.cloud.notifications.models.Bundle;
+import com.redhat.cloud.notifications.models.Endpoint;
 import com.redhat.cloud.notifications.models.EventType;
 
 import javax.enterprise.context.ApplicationScoped;
@@ -117,7 +118,7 @@ public class BehaviorGroupRepository {
             q = q.setParameter("accountId", accountId);
         }
 
-        return q.executeUpdate()  > 0;
+        return q.executeUpdate() > 0;
     }
 
     public boolean delete(String accountId, UUID behaviorGroupId) {
@@ -356,6 +357,11 @@ public class BehaviorGroupRepository {
     }
 
     public List<BehaviorGroup> findBehaviorGroupsByEndpointId(String accountId, UUID endpointId) {
+        Endpoint endpoint = entityManager.find(Endpoint.class, endpointId);
+        if (endpoint == null) {
+            throw new NotFoundException("Endpoint not found");
+        }
+
         String query = "SELECT bg FROM BehaviorGroup bg LEFT JOIN FETCH bg.bundle JOIN bg.actions a WHERE bg.accountId = :accountId AND a.endpoint.id = :endpointId";
         List<BehaviorGroup> behaviorGroups = entityManager.createQuery(query, BehaviorGroup.class)
                 .setParameter("accountId", accountId)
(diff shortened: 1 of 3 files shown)


theute:
The test should be in https://github.com/RedHatInsights/notifications-backend/blob/master/backend/src/test/java/com/redhat/cloud/notifications/routers/NotificationResourceTest.java

Very cool to see a PR from you !

Mishrasubha:
> 

Ahh! My bad.
Thanks :) it was a small change so, thought I could give it a try,

backend/src/test/java/com/redhat/cloud/notifications/routers/NotificationResourceTest.java
```diff
@@ -547,4 +547,16 @@ void testFindBehaviorGroupsByUnknownEventTypeId() {
...
+    @Test
+    void testBehaviorGroupsAffectedByRemovalOfUnknownEndpointId() {
+        Header identityHeader = initRbacMock("tenant", "user", FULL_ACCESS);
+        given()
+                .header(identityHeader)
+                .pathParam("endpointId", UUID.randomUUID())
+                .when()
+                .get("/notifications/behaviorGroups/affectedByRemovalOfEndpoint/{endpointId}")
+                .then()
+                .statusCode(404);
```
gwenneg, line 560: This test fails because the change we talked is not in this PR yet. This is the method that needs to be changed:
https://github.com/RedHatInsights/notifications-backend/blob/5fb0654876102b626f06ec8a1575fc43f0b16f30/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java#L358-L369

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
```diff
@@ -186,7 +186,7 @@ public boolean updateEventTypeBehaviors(String accountId, UUID eventTypeId, Set<
         // First, let's make sure the event type exists.
         EventType eventType = entityManager.find(EventType.class, eventTypeId);
         if (eventType == null) {
-            return false;
+            throw new NotFoundException("Event type not found");
```
gwenneg, line 189: This does not change the application behavior as we already returned a 404 if the event type was not found. However, it's still a good change because we return a better error message and it uses the same pattern we use in several other places. As a consequence, it's normal if this doesn't break any existing test.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -179,7 +179,7 @@ public Endpoint getEndpoint(String tenant, UUID id) {
             loadProperties(endpoint);
             return endpoint;
         } catch (NoResultException e) {
-            return null;
+            throw new NotFoundException("Endpoint not found");
```
gwenneg, line 182: This will break existing code that is probably not tested right now:
https://github.com/RedHatInsights/notifications-backend/blob/5fb0654876102b626f06ec8a1575fc43f0b16f30/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java#L267-L273

We already return from all REST endpoints that depend on that code, so it doesn't really need to be changed.

The only possible improvement I see would consist in adding a message that says that the endpoint was not found here:
https://github.com/RedHatInsights/notifications-backend/blob/5fb0654876102b626f06ec8a1575fc43f0b16f30/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java#L248-L250

gwenneg (approved):
LGTM

Thanks for that contribution!

gwenneg:
@josejulio Could you please confirm that the change to `GET /notifications/behaviorGroups/affectedByRemovalOfEndpoint/{endpointId}` (which now returns a 404 when the endpoint does not exist) will not cause any issue in the frontend?

PR #1212, opened 2022-05-05
NOTIF-338 Allow to filter by event type name

josejulio:


gwenneg (approved):
LGTM

We'll need tests for this (now or later).

gwenneg (approved):
Thanks for the tests!

PR #1213, opened 2022-05-05
NOTIF-646 Emit Camel metrics with subtype as tag.

pilhuhn:
Metrics after sending 2 events to slack

```
# HELP processor_camel_processed_total  
# TYPE processor_camel_processed_total counter
processor_camel_processed_total{subType="slack",} 2.0
```

Not sure why there is an additional `,` in the output, but that is what `curl localhost:8087/metrics | grep processor  ` shows

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
@@ -75,11 +75,8 @@ public class CamelTypeProcessor implements EndpointTypeProcessor {
     @Inject
     BridgeAuth bridgeAuth;
 
-    private Counter processedCount;
-
     @PostConstruct
     public void init() {
-        processedCount = registry.counter(PROCESSED_COUNTER_NAME);
     }
 
     @Override
@@ -93,8 +90,12 @@ public class CamelTypeProcessor implements EndpointTypeProcessor {
     }
 
     private NotificationHistory process(Notification item) {
-        processedCount.increment();
         Endpoint endpoint = item.getEndpoint();
+        String subType = endpoint.getSubType();
+
+        Counter processedCount = registry.counter(PROCESSED_COUNTER_NAME, "subType", subType);
+        processedCount.increment();
+
         CamelProperties properties = (CamelProperties) endpoint.getProperties();
 
         Map<String, String> metaData = new HashMap<>();
@@ -102,7 +103,7 @@ public class CamelTypeProcessor implements EndpointTypeProcessor {
         metaData.put("trustAll", String.valueOf(properties.getDisableSslVerification()));
 
         metaData.put("url", properties.getUrl());
-        metaData.put("type", endpoint.getSubType());
+        metaData.put("type", subType);
 
         if (properties.getSecretToken() != null && !properties.getSecretToken().isBlank()) {
             metaData.put(TOKEN_HEADER, properties.getSecretToken());
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
@@ -28,7 +28,6 @@ import org.eclipse.microprofile.reactive.messaging.Message;
 import org.eclipse.microprofile.rest.client.RestClientBuilder;
 import org.jboss.logging.Logger;
 
-import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 import java.net.URI;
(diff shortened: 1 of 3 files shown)


pilhuhn:
I wonder if we should chose a different name for the tag like `iType` or just `type`.

engine/src/test/java/com/redhat/cloud/notifications/MicrometerAssertionHelper.java
```diff
@@ -33,6 +33,12 @@ public void saveCounterValuesBeforeTest(String... counterNames) {
         }
     }
 
+    public void assertCounterIncrement(String counterName, double expectedIncrement, String... tags) {
+        double actualIncrement = registry.counter(counterName, tags).count() - counterValuesBeforeTest.getOrDefault(counterName + tags, 0D);
+        assertEquals(expectedIncrement, actualIncrement);
+    }
+
+
```
theute, line 41: nitpick just to pretend I carefully reviewed but there in an extra \n here ;-)

theute (approved):

engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
```diff
@@ -93,16 +90,20 @@ public List<NotificationHistory> process(Event event, List<Endpoint> endpoints)
     }
 
     private NotificationHistory process(Notification item) {
-        processedCount.increment();
         Endpoint endpoint = item.getEndpoint();
+        String subType = endpoint.getSubType();
+
+        Counter processedCount = registry.counter(PROCESSED_COUNTER_NAME, "subType", subType);
```
gwenneg, line 96: I'm confused. Do dynamic tags work with Micrometer's counter or not? :smiley: Last time I heard about that, it didn't work.

gwenneg: Maybe that was fixed upstream and it's finally working because we bumped Quarkus?

pilhuhn: They are probably not really dynamic. `registry.counter()` either registers a new counter or gives one back that has the passed name and tags. Meaning I am not using a dynamic tag at the time of `counter.increment()` but create a new counter for each of the passed tags (and re-use later).

theute: I *think* it works when the label is added that way, but not if the tags are declared after.
something.tag("blah", "blih") --> didn't work

The question is if `registry.counter(PROCESSED_COUNTER_NAME, "subType", subType)` is a costly operation or not, it used to be declared once.

engine/src/test/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessorTest.java
```diff
@@ -98,7 +100,9 @@ void testCamelEndpointProcessing() {
         // Two endpoints should have been processed.
         assertEquals(2, result.size());
         // Metrics should report the same thing.
-        micrometerAssertionHelper.assertCounterIncrement(PROCESSED_COUNTER_NAME, 2);
+        micrometerAssertionHelper.assertCounterIncrement(PROCESSED_COUNTER_NAME, 2, SUB_TYPE_KEY, SUB_TYPE);
```
gwenneg, line 103: The behavior of MicrometerAssertionHelper will be unreliable if `saveCounterValuesBeforeTest` isn't called before the tests (see `beforeEach()`) with the new `counter name + tags` combination.

engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
```diff
@@ -75,11 +75,8 @@ public class CamelTypeProcessor implements EndpointTypeProcessor {
     @Inject
     BridgeAuth bridgeAuth;
 
-    private Counter processedCount;
-
     @PostConstruct
     public void init() {
```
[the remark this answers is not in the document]
gwenneg: I meant to target the `init()` method :smiley:

pilhuhn:
❯ curl localhost:8087/metrics | grep processor
# HELP processor_input_endpoint_processed_total  
# TYPE processor_input_endpoint_processed_total counter
processor_input_endpoint_processed_total 6.0
# HELP processor_input_processed_total  
# TYPE processor_input_processed_total counter
processor_input_processed_total 6.0
# HELP processor_camel_processed_total  
# TYPE processor_camel_processed_total counter
processor_camel_processed_total{subType="slack",} 4.0
processor_camel_processed_total{subType="splunk",} 2.0

engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
pilhuhn: Internally there is a map <Id, Meter>  where Id = <name, Tags>  so that may be more costly on the 1st event, but cheap afterwards.

gwenneg (approved):
LGTM

PR #1221, opened 2022-05-10
Migrate edge-management templates files to DB

gwenneg:


gwenneg:

PR #1226, opened 2022-05-11
NOTIF-603 Add orgId to IT User Service requests

bennetelli:
It's still a pita to get a certificate, but we will have to do the change of the query anyway. I could not integration test (only comparing sample requests so far) the new query against the real IT User Service API for now.

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
@@ -41,6 +41,11 @@ public class RbacRecipientUsersProvider {
 
     public static final String ORG_ADMIN_PERMISSION = "admin:org:all";
 
+    public static final String USE_ORG_ID = "notifications.use-org-id";
+
+    @ConfigProperty(name = USE_ORG_ID, defaultValue = "false")
+    public boolean useOrgId;
+
     @Inject
     @RestClient
     RbacServiceToService rbacServiceToService;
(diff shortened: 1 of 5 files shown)


bennetelli:
> 

@josejulio wtf?! :-D I added tests that have not been there before. How the heck can I decrease the coverage by doing this?

bennetelli:
this always red  **ci.ext.devshift.net PR build** is getting a bit annoying..

bennetelli:
the response should stay the same so we don't need to change the response objects

pilhuhn:
@bennetelli The "should" in your last sentence worries me a bit. Otherwise I think it it ok.

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -41,6 +41,11 @@ public class RbacRecipientUsersProvider {
 
     public static final String ORG_ADMIN_PERMISSION = "admin:org:all";
 
+    public static final String USE_ORG_ID = "notifications.use-org-id";
```
josejulio, line 44: Is this the same config used in the other PR? Maybe we could put it in a single place to prevent any surprise.

bennetelli:
> @bennetelli The "should" in your last sentence worries me a bit. Otherwise I think it it ok.

removed the should :D

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
bennetelli: yes, that's the same. Thought about it as well and decided not to do this as I will have to remove the "notifications.use-org-id" from the application.properties anywy and then I will also remove each accountId and these entries. 

I think it's not needed in this step to be honest.

josejulio (approved):

PR #1227, opened 2022-05-11
Use prod image url

astrozzc:
The preprod image url will not work because google will proxy the url
in email and will not show the logo image.

### DIFF
--- a/engine/src/main/resources/templates/Rbac/insightsEmailBody.html
+++ b/engine/src/main/resources/templates/Rbac/insightsEmailBody.html
@@ -561,7 +561,7 @@
                                 <tr>
                                     <!--Red Hat logo-->
                                     <td align="center" class="rh-masthead__brand" bgcolor="#151515">
-                                        <a href="https://console.redhat.com/" target="_blank" class="rh-masthead__brand-link">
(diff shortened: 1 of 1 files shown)


engine/src/main/resources/templates/Rbac/insightsEmailBody.html
```diff
@@ -561,8 +561,8 @@
                                 <tr>
                                     <!--Red Hat logo-->
                                     <td align="center" class="rh-masthead__brand" bgcolor="#151515">
-                                        <a href="{environment.url('/')}" target="_blank" class="rh-masthead__brand-link">
-                                            <img src="{environment.url('/apps/frontend-assets/email-assets/logo_insights.jpg')}" alt="Red Hat logo" width="182" height="90" class="rh-masthead__brand-img" />
+                                        <a href="https://console.redhat.com/" target="_blank" class="rh-masthead__brand-link">
```
josejulio, line 564: I think this link is OK as is, IMO we only need to change the img src.

Peek 2022-05-11 12-29

astrozzc: Done 👍

josejulio (approved):
lgtm!

PR #1230, opened 2022-05-12
NOTIF-579 Bundle and application filter work well together

josejulio:
- It wasn't working correctly when the application was part of other bundle


gwenneg (approved):
LGTM

PR #1231, opened 2022-05-13
SPM-1483: Add instant notification email templates for Patch Advisories

michalslomczynski:


josejulio:
I think we have some issues with the itests - I'll review the PR.

Could you try your template with a payload in here? 
https://internal.console.stage.redhat.com/api/notifications/internal/#/email

josejulio (approved):

michalslomczynski:
Thanks @josejulio. Yes, all works like intended on that tool.

PR #1239, opened 2022-05-16
NOTIF-652 require non-admin to select role

aferd:
Admins have option to edit and are not required to select role.


josejulio:
The edit of role is missing a backend implementation or it works already?

PR #1242, opened 2022-05-17
typo fix in openapi doc

dgaikwad:


backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -120,7 +120,7 @@ public class EndpointResource {
         @Parameter(
                 name = "limit",
                 in = ParameterIn.QUERY,
-                description = "Number of items per page, if not specified or 0 is used, returns all elements",
+                description = "Number of items per page. If the value is 0, it will return all elements",
                 schema = @Schema(type = SchemaType.INTEGER)
```
theute, line 124: You should still add the defaultValue here:
`@Schema(type = SchemaType.INTEGER, defaultValue="20")`

dgaikwad:
@theute Test fix is here: https://gitlab.cee.redhat.com/insights-qe/iqe-notifications-plugin/-/merge_requests/179
Please update regularly if PR check fails.

PR #1243, opened 2022-05-17
Reverting OpenAPI schema change

theute:
Not needed, and may lead to confusion later


josejulio:
Default value already documented in https://github.com/RedHatInsights/notifications-backend/blob/e3008ef1a5b366083c02b38beaf96ef270a86198/backend/src/main/java/com/redhat/cloud/notifications/db/Query.java#L14-L18

josejulio (approved):

PR #1251, opened 2022-05-19
SPM-1484 Change New Advisories event name

michalslomczynski:
I changed event name in the patchman-engine and here is the adjustment for it.


gwenneg (approved):
LGTM

PR #1262, opened 2022-05-26
Add dummy probes to continue onboarding

gburges:
We need to implement a dummy probe now, as a workaround so we can progress onboarding. Clowder code change to implement the annotations functionality is in progress also


gburges:
FYI https://github.com/RedHatInsights/clowder/pull/621 this will be a solution, just not sure when it will be merged

.rhcicd/clowdapp-aggregator.yaml
```diff
@@ -47,6 +44,26 @@ objects:
...
+        readinessProbe:
+          httpGet:
+            path: /health/ready
+            port: 8000
+            scheme: HTTP
+          initialDelaySeconds: 1
+          periodSeconds: 1
+          timeoutSeconds: 1
+          successThreshold: 1
+          failureThreshold: 3
```
josejulio, line 56: I'm not sure i understand, but are these considering the usual time the cronjob takes to execute? 
Won't these affect the running process?

gburges: the timing is so that it succeeds near instantly, so the cronjob will be ready and healthy.  It's a workaround for kubelinter to not alert on it. Shouldn't affect runtime, and can be reverted when the above clowder change rolls out

josejulio (approved):

PR #1269, opened 2022-05-30
NOTIF-674 Limit fields of sort expressions.

pilhuhn:
Limit the fields that are acceptable for sorting to fixed lists of field names.

Open questions:
- should we accept any case?
- should we move the definitions to models and then also list them e.g. in the OpenApi's operation?
- What fields are needed beyond what is in this PR?


pilhuhn:
I wonder how we could best get a list of allowed sort criteria. So far I made them up (and tests pass), but UI probably uses a lot of columns that are not listed in tests. Or I could try to use some kind of introspection to get valid column names from the declared fields of a model type. That would be more than we need, but would at least prevent sending of random crap.

pilhuhn:
I have added a warning log for unknown fields. That may help us detect forgotten ones :)

PR #1271, opened 2022-05-31
NOTIF-603 Add orgId to missing classes

bennetelli:


bennetelli:
@gwenneg with this PR I also replaced the orgIdConf.isUseOrgId() checks with some simple checks like

event.getOrgId() != null && !event.getOrgId().isEmpty() and so on. 

Do we still need the flag in RbacRecipientUsersProvider? I am wondering if this could crash on prod when we replace the flag with the simple checks. Your opinion would be great to brainstorm a bit

engine/src/main/java/com/redhat/cloud/notifications/OrgIdConfig.java
```diff
@@ -0,0 +1,23 @@
...
+    public static final String USE_ORG_ID = "notifications.use-org-id";
+
+    @ConfigProperty(name = USE_ORG_ID, defaultValue = "false")
+    public boolean useOrgId;
+
+    public boolean isUseOrgId() {
+        return useOrgId;
+    }
+
+    public void setUseOrgId(boolean useOrgId) {
```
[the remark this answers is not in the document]
bennetelli: yup. it's deleted now. thanks for taking a look at it :)

bennetelli:
the default methods for Behavior Groups are still missing. And I need to split the ResourceHelpers.. That class will become a big ball of mud otherwise.

backend/src/main/java/com/redhat/cloud/notifications/OrgIdConfig.java
```diff
@@ -0,0 +1,23 @@
...
+    public static final String USE_ORG_ID = "notifications.use-org-id";
+
+    @ConfigProperty(name = USE_ORG_ID, defaultValue = "false")
+    public boolean useOrgId;
+
+    public boolean isUseOrgId() {
+        return useOrgId;
+    }
+
+    public void setUseOrgId(boolean useOrgId) {
```
josejulio, line 20: I still see this one here :D

backend/src/main/java/com/redhat/cloud/notifications/routers/NotificationResource.java
```diff
@@ -64,14 +66,25 @@ public class NotificationResource {
     @Inject
     BehaviorGroupRepository behaviorGroupRepository;
 
+    @Inject
+    BehaviorGroupRepositoryOrgId behaviorGroupRepositoryOrgId;
```
josejulio, line 70: Why is this one needed? Can't the `BehaviorGroupRepository` use the configuration to decide what field to use?

common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
```diff
@@ -93,8 +108,12 @@ public class BehaviorGroup extends CreationUpdateTimestamped {
 
     @JsonInclude
     @JsonProperty(access = READ_ONLY, value = "default_behavior")
-    public boolean isDefaultBehavior() {
-        return accountId == null;
+    public boolean isDefaultBehavior(boolean orgIdUsage) {
```
josejulio, line 111: Is this one working properly when returning behavior groups and default behavior groups?
IIUC this one is called automatically by the json encoder. Could we instead inject the configuration (either as a boolean field that is `@JsonIgnore'd`  or the OrgIDConfig object.

backend/src/main/java/com/redhat/cloud/notifications/routers/NotificationResource.java
bennetelli: I started with duplicating the methods in the BehaviorGroupRepository first. But this end in a god class where it was horrible to find the methods to be deleted after the migration has been finished. That's why I decided to split it, so we have to remove the old BehaviorGroupRepository which deals with accountIds only. 

Methods above the call hierarchy I am having a switch for accountId -> BehaviorGroupRepository or BehaviorGroupRepositoryOrgId -> orgId.

backend/src/main/java/com/redhat/cloud/notifications/OrgIdConfig.java
bennetelli: removed

backend/src/main/java/com/redhat/cloud/notifications/routers/NotificationResource.java
bennetelli: The NotificationResource has got both, the BehaviorGroupRepository and BehaviorGroupRepositoryOrgId and decides what to use. 

When we switch to orgId, we have to remove the if/else blocks in this class, delete the BehaviorGroupRepository class and use BehaviorGroupRepositoryOrgId everywhere. And rename BehaviorGroupRepositoryOrgId to BehaviorGroupRepository :)

common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
bennetelli: either we don't have tests covering this or it's working properly.

backend/src/main/java/com/redhat/cloud/notifications/OrgIdConfig.java
bennetelli: this setter haunts me :D

common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
bennetelli: wtf.. :D what a "feature". when you add a parameter to this method, the "default_behavior" field in the JSON response get's ignored. It's just not there. And we did not have a test checking if default_behavior is part of the JSON response.

PR #1275, opened 2022-06-01
NOTIF-680 Add exception counter

bennetelli:
I would like to change the Grafana Dashboard after this counter is available in prometheus.


bennetelli:
that test is green in intelliJ, but red here.

bennetelli:
still don't understand how that test could ever be green in intelliJ..

pilhuhn (approved):

PR #1279, opened 2022-06-02
First cut at OB error backchannel handling.

pilhuhn:
OB will send a webhook message when an event could not be sent due to errors.
There may be different ways of dealing with it later.


manstis:
Hi.. I'm on RHOSE, another Managed Service.

We elected to put DLQ info in the header (Kafka record header or HTTP headers) vs Cloud Event body.

I guess we should talk and make the planets align?

PR #1273, opened 2022-05-31
VULN-2291 add vulnerability daily mail template

yungbender:

PR #1283, opened 2022-06-02
Updating Enviroment.url methods from templates

josejulio:
- environment.url() no longer adds a trailing slash
- environment.url(string) no longer exists
- Updates all the templates to take into account the new usage.

These changes make it easier to read the templates (IMO).
A major problem with `environment.url(string)` is that it didn't allow to concatenate strings inside the method, thus:
`{environment.url('stuff/' + variable)}` would yield: `https://console.redhat.com/stuff/ + variable` (yes, literally that)

Users had to do: `{environment.url("stuff/")}{variable}` to get the desired output, which is harder to read and defeats the purpose of the argument anyway.

With the latest changes it can to be now: `{environment.url}/stuff/{variable}` which feels more natural.

This change was motivated by the review of https://github.com/RedHatInsights/notifications-backend/pull/1273

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/templates/models/Environment.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/templates/models/Environment.java
@@ -15,28 +15,14 @@ public class Environment {
     }
 
     public String url() {
-        return url("");
-    }
-
-    public String url(String path) {
-        final String base;
 
         switch (environment) {
             case "prod":
-                base = "https://console.redhat.com";
-                break;
+                return "https://console.redhat.com";
             case "stage":
-                base = "https://console.stage.redhat.com";
-                break;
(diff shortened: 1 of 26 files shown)


engine/src/main/java/com/redhat/cloud/notifications/templates/models/Environment.java
```diff
@@ -15,28 +15,14 @@ public String name() {
...
-    }
-
-    public String url(String path) {
-        final String base;
 
         switch (environment) {
             case "prod":
-                base = "https://console.redhat.com";
-                break;
+                return "https://console.redhat.com";
```
bennetelli, line 21: shouldn't we put this in application.properties or something like that?

bennetelli (approved):

PR #1291, opened 2022-06-09
Renaming db cleaner cronjob, and fixing volumemounts

gburges:


.rhcicd/clowdapp-backend.yaml
```diff
@@ -153,6 +153,7 @@ objects:
           - name: notifications-db-cleaner-volume
             configMap:
               name: notifications-db-cleaner-config
+        volumeMounts:
```
bennetelli, line 156: what is this good for?

gburges: where to mount it in the container, the configmap above is what to mount

bennetelli:
let's wait for the fucking mini yoda this time

PR #1292, opened 2022-06-09
Add back original cronjob until i fix the clowderized version

gburges:


theute (approved):

gburges:
Reverting it because the cronjob seems to be working

PR #1293, opened 2022-06-09
Roll back clowder config source to 1.0.0

pilhuhn:
1.0.1 has some issues - especially the log.type field is not populated in the provided config files and thus makes the evaluation of cloudwatch entries fail.


gwenneg (approved):
LGTM

gwenneg:
@pilhuhn Where did that issue happen? Was it on ephemeral or stage?

PR #1303, opened 2022-06-15
NOTIF-681 : list ALL templates in table for all applications

aferd:
https://user-images.githubusercontent.com/86322239/173889166-3e11def9-4f38-4392-a839-0891a3a6fba6.mov


admin-console/src/main/webapp/src/components/EmailTemplates/ListAssociatedEventTypes.tsx
```diff
@@ -1,10 +1,16 @@
...
 import { useInstantEmailTemplates } from '../../services/EmailTemplates/GetInstantTemplates';
 
+interface EmailTemplateParams {
+    applicationId: string;
+}
+
 export const ListAssociatedEventTypes: React.FunctionComponent = () => {
-    const getAssociatedEventTypes = useInstantEmailTemplates();
+    const { applicationId } = useParams<EmailTemplateParams>();
+    const getAssociatedEventTypes = useInstantEmailTemplates(applicationId);
```
josejulio, line 13: Instead of showing the instant templates here, we should show the `raw` templates, given by [this](https://github.com/RedHatInsights/notifications-backend/blob/master/admin-console/src/main/webapp/src/generated/OpenapiInternal.ts#L1739) API call.

It's currently open (list all without filters), but we can take care of that in a follow up.

Just to clarify, there are the following template types:
- `raw` which are actually the real template contents
- `instant templates` are actually composed of 2 `raw` templates for the subject and body and is assigned to an event type.
- `aggregation templates` are also composed of 2 `raw` templates, is assigned to the application itself.

Instant templates are already present in the top table.

aferd: so, I think the initial purpose of this was to list the event-types associated to the templates in the email template table. Is this something that you would be interested in keeping? otherwise I can delete this file. Its not currently shown anywhere.

josejulio: I don't think so, it has been moved to the top table.
We can have a list of the `raw` templates though. Currently they are not scoped to the application - it's something we should change in the API.

aferd: Yes, so currently in the template table(shown in the video above) is the list of raw templates. This "ListAssociatedEventTypes" component was initially made to list the event types that were attached to each template in the template table(this was done before back end implementation and then i think i tried it but we didn't have it complete.) and this component is currently used anywhere.

josejulio: Ok - so we can remove it, but the same design should also work for the raw templates, only changing the data that is being used by the raw templates.

aferd:
@josejulio is this good to merge? I removed the unused file.

PR #1310, opened 2022-06-22
Prevent error if an endpoint is gone when the history is written

gwenneg:
See the migration script comments for details about this PR.

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/db/repositories/NotificationHistoryRepository.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/db/repositories/NotificationHistoryRepository.java
@@ -1,6 +1,8 @@
 package com.redhat.cloud.notifications.db.repositories;
 
 import com.redhat.cloud.notifications.db.StatelessSessionFactory;
+import com.redhat.cloud.notifications.db.converters.EndpointTypeConverter;
+import com.redhat.cloud.notifications.db.converters.NotificationHistoryDetailsConverter;
 import com.redhat.cloud.notifications.models.Endpoint;
 import com.redhat.cloud.notifications.models.NotificationHistory;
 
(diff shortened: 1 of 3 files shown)


theute (approved):
Less alerts will be very much appreciated :)

database/src/main/resources/db/migration/V1.51.0__fk_notification_history_endpoint_id.sql
```diff
@@ -0,0 +1,13 @@
+-- This script fixes the following issue:
+--
+-- If an endpoint is deleted (from the REST API) while it is also being used to send a notification
+-- (notifications-engine received an action and is processing it, but the history hasn't been written yet),
+-- then the constraint will be violated and the insertion of the history record into the DB will fail.
+--
+-- This caused many Sentry alerts on prod because of a race condition in the IQE tests.
+--
+-- After this migration, some of the notification_history table records will contain an endpoint_id value
+-- that doesn't match any of the existing endpoints in the DB.
```
josejulio, line 10: I'm wondering if we should worry about it in the event log

gwenneg: It depends on what the frontend does with this:
https://github.com/RedHatInsights/notifications-backend/blob/c3047ecf192ed3e81ed329c5bc4b5f1122ece106/backend/src/main/java/com/redhat/cloud/notifications/routers/EventResource.java#L97

josejulio: It will try to load the information of the endpoint or fail with sort of "hey - couldn't find the endpoint"

gwenneg: Ok then I guess this PR won't cause any serious issue.

Maybe the frontend should have the same behavior (from a user perspective) when the endpoint is not found (that's new) and when the `endpointId` is `null` (which could already happen).

josejulio: Previously the foreign key constraint was in charge of changing the id to `null`. Now that we are dropping it, are we going to do that manually somehow?

I'm wondering if it wouldn't be better to check if the endpoint exists right before adding it to  the history - if it exists uses the id - if not, uses null.

gwenneg: Good point, new entries will never have a `null` value now.

gwenneg: We could check if the endpoint exists before inserting its id into the history, but that would require a transaction and locking the endpoint (select for update) to make sure it's isn't deleted before the end of the transaction. That's a lot of pressure on the DB but from a data integrity perspective, that's a better approach.

gwenneg: Yes, that should work.

josejulio:
Query looks good - I'm wondering if we could add a test for this scenario.

gwenneg:
> Query looks good - I'm wondering if we could add a test for this scenario.

Yes, that's why I converted the PR into a draft. I'll take the time to do manual test tomorrow and add one or more unit test(s).

gwenneg:
I compared the `notification_history` data before and after this PR, it looks alike. I'm adding a unit test now.

josejulio (approved):
lgtm

PR #1324, opened 2022-06-29
NOTIF-603 Add orgId to DB entities

gwenneg:
#1271 is too big and hard to rebase, so I'm breaking it down into smaller PRs.

This is step 1: adding the `orgId` field to all  DB entities where it was missing.

### DIFF
--- a/common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
+++ b/common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
@@ -44,6 +44,10 @@ public class BehaviorGroup extends CreationUpdateTimestamped {
     @JsonIgnore
     private String accountId;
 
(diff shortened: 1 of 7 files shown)


gwenneg:
I'll check the build failure shortly.

gwenneg:
@josejulio This is ready for a review.

common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
```diff
@@ -44,6 +44,10 @@ public class BehaviorGroup extends CreationUpdateTimestamped {
     @JsonIgnore
     private String accountId;
```
josejulio, line 45: Wonder if we should start to mark them as `@Deprecated` or something

josejulio (approved):
lgtm

PR #1333, opened 2022-07-01
NOTIF-603 Polish the orgId migration in engine

gwenneg:
This PR adds the `orgId` in most places where it was missing in `notifications-engine`.

After that, `CamelTypeProcessor` should be the only class where the `orgId` is missing. As explained below, it will be updated separately.

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
@@ -41,6 +41,7 @@ import static com.redhat.cloud.notifications.events.KafkaMessageDeduplicator.MES
 import static com.redhat.cloud.notifications.models.NotificationHistory.getHistoryStub;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
+// TODO NOTIF-603 Migrate to orgId (not started in this class)
 @ApplicationScoped
 public class CamelTypeProcessor implements EndpointTypeProcessor {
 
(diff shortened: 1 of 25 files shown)


engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
```diff
@@ -41,6 +41,7 @@
 import static com.redhat.cloud.notifications.models.NotificationHistory.getHistoryStub;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
+// TODO NOTIF-603 Migrate to orgId (not started in this class)
 @ApplicationScoped
 public class CamelTypeProcessor implements EndpointTypeProcessor {
```
gwenneg, line 46: This will be done in a separate PR. It requires changes in our integration with RHOSE.

gwenneg: https://issues.redhat.com/browse/NOTIF-706

engine/src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java
```diff
@@ -461,7 +463,7 @@ void addEventTypeBehavior(UUID eventTypeId, UUID behaviorGroupId) {
 
     @Transactional
     void clearEventTypeBehaviors(EventType eventType) {
-        entityManager.createQuery("DELETE EventTypeBehavior WHERE eventType = :eventType")
+        entityManager.createQuery("DELETE FROM EventTypeBehavior WHERE eventType = :eventType")
```
josejulio, line 466: Was this wrong before? Or is just better (style-wise) to use the `FROM` ?

gwenneg: I think Hibernate automatically added the missing `FROM` but IntelliJ complained about this and I wanted him to be happy 😄

PR #1339, opened 2022-07-05
NOTIF-696 create view/edit modal for instant templates

aferd:


aferd:
Screen Shot 2022-07-05 at 10 46 11 AM
the onClick is not working currently, could you take a look? I think its in the EventtypeTable component, its not connecting as it should but I can't figure out what I am missing.

josejulio:
> Also, thoughts on if user go to edit "event Type" should we show templates at all there? or just continue to keep them separate as we talked(and as is currently being done).

The thing of showing them, is that we require some sample data so that they will render correctly- probably better to have a separate screen for now.

In the future we can fetch data from the event log to render with.

admin-console/src/main/webapp/src/components/EmailTemplates/CreateViewInstantTemplateButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+    Button
+} from '@patternfly/react-core';
+import React from 'react';
+
+interface InstantTemplateButtonProps {
+    hasTemplate: boolean;
+    createInstantTemplateButton: React.ReactNode
+
+}
+export const InstantTemplateButton: React.FunctionComponent<InstantTemplateButtonProps> = (props) => {
```
josejulio, line 11: Is this used anywhere?

admin-console/src/main/webapp/src/components/EventTypes/EventTypeTable.tsx
```diff
@@ -101,7 +103,8 @@ const EventTypeTableImpl: React.FunctionComponent<EventTypeTableImplProps> = pro
                     />
                     <Td>{ eventType.displayName }</Td>
                     <Td>{ eventType.name }</Td>
-                    <Td><InstantEmailCell eventType={ eventType } /></Td>
+                    <Td><InstantEmailCell eventType={ eventType }
+                        onClick={ () => props.onCreateInstantTemplate } /></Td>
```
josejulio, line 107: This is always passing `onCreateInstantTemplate` to the onclick - I guess it should pass both somehow? Or have a single function that accepts an optional instantTemplate.

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -79,12 +85,26 @@ export const ApplicationPage: React.FunctionComponent = () => {
...
     const createEventType = () => {
         setShowModal(true);
         setIsEdit(false);
         setEventTypes({});
     };
 
+    const createInstantTemplate = () => {
+        setShowModal(true);
+        setIsEdit(false);
+        setInstantTemplates({});
```
josejulio, line 105: Try to have all these variables condensed into a single one

admin-console/src/main/webapp/src/components/EmailTemplates/CreateViewInstantTemplateButton.tsx
aferd: ah, no. i decided to just implement the button in the instantEmailCell. Will remove.

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
aferd: like 
`  const [ createEditInstantTemplateModal, setCreateEditInstantTemplateModal ] =
    React.useState({ showModal: false, isEdit: false, instantTemplates: {}});`

josejulio: Somewhat like that sounds good.

admin-console/src/main/webapp/src/components/EventTypes/EventTypeTable.tsx
```diff
@@ -101,7 +103,8 @@ const EventTypeTableImpl: React.FunctionComponent<EventTypeTableImplProps> = pro
                     />
                     <Td>{ eventType.displayName }</Td>
                     <Td>{ eventType.name }</Td>
-                    <Td><InstantEmailCell eventType={ eventType } /></Td>
+                    <Td><InstantEmailCell eventType={ eventType }
+                        onClick={ () => props.onCreateEditInstantTemplate } /></Td>
```
aferd, line 107: combined here.

still can not get the modal to open. It is currently in the App page component. Should it live in this component?

josejulio: Not required - I'll try to pull the code tomorrow and see what part is missing.

aferd:
This is possibly related but on the event type table, when I go to create new event type or edit it tries to redirect me (though the url does not change) and throws this error `TypeError: Cannot read properties of undefined (reading 'value')`

admin-console/src/main/webapp/src/components/EmailTemplates/InstantEmailTemplateModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+} from '@patternfly/react-core';
+import React from 'react';
+
+import { InstantTemplate } from '../../types/Notifications';
+
+interface InstantTemplateModalProps {
+    showModal: boolean;
+    isEdit: boolean;
+    onClose: () => void;
+    templates: string[] | undefined;
```
josejulio, line 18: You would need `Template[]` object here, with the name and id of the template.

admin-console/src/main/webapp/src/components/EmailTemplates/InstantEmailTemplateModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+    initialInstantTemplate?: Partial<InstantTemplate>;
+    onSubmit: (instantTemplate: Partial<InstantTemplate>) => void;
```
josejulio, line 20: This could be changed to a non partial InstantTemplate, would need to validate and cast to non partial InstantTemplate
```suggestion
    onSubmit: (instantTemplate: InstantTemplate) => void;
```

admin-console/src/main/webapp/src/components/EmailTemplates/InstantEmailTemplateModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+}
+
+export const InstantTemplateModal: React.FunctionComponent<InstantTemplateModalProps> = (props) => {
+
+    const [ instantTemplate, setInstantTemplate ] = React.useState<Partial<InstantTemplate>>(props.initialInstantTemplate ?? {});
+
+    const handleChange = (value: string, event: React.FormEvent<HTMLFormElement> | React.FormEvent<HTMLSelectElement>) => {
+        const target = event.target as HTMLSelectElement;
+        setInstantTemplate(prev => ({ ...prev, [target.name]: target.value }));
+    };
```
josejulio, line 30: This code is pretty generic, but could work nonetheless. Need to ensure that the values are being set on the form are what the instant template is expecting.

admin-console/src/main/webapp/src/components/EmailTemplates/InstantEmailTemplateModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+            ><Form isHorizontal>
+                    <FormGroup label='Subject template' fieldId='subject-template'>
+                        <FormSelect
+                            id="subject-template"
+                            name="subject-template"
+                            aria-label="Subject template"
+                            onChange={ handleChange }
+                            isRequired
+                        >
+                            { props.templates }
```
josejulio, line 56: Need to use `<Options>`s here

admin-console/src/main/webapp/src/components/EmailTemplates/InstantEmailTemplateModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+                    </FormGroup>
+                    <FormGroup label='Body template' fieldId='body-template'>
+                        <FormSelect
+                            id="body-template"
+                            name="body-template"
+                            aria-label="Body template"
+                            onChange={ handleChange }
+                            isRequired
+                        >
+                            { props.templates }
```
josejulio, line 67: Need to use `<Options>`s here

admin-console/src/main/webapp/src/components/EventTypes/EventTypeTable.tsx
```diff
@@ -24,6 +24,7 @@ interface EventTypeTableBaseProps {
     onCreateEventType: () => void;
     onEditEventType: (eventType: EventType) => void;
     onDeleteEventTypeModal: (eventType: EventType) => void;
+    onCreateEditInstantTemplate: (i: InstantTemplate) => void
```
josejulio, line 27: The second parameter should be optional 
```suggestion
    onCreateEditInstantTemplate: (i?: InstantTemplate) => void
```

admin-console/src/main/webapp/src/components/EventTypes/EventTypeTable.tsx
josejulio: This part is never executed, should be:
```suggestion
                        onClick={ () => props.onCreateEditInstantTemplate(eventType.instantEmail.id ? eventType.instantEmail : undefined ) } /></Td>
```

josejulio (changes requested):

admin-console/src/main/webapp/src/components/EventTypes/EventTypeTable.tsx
aferd: im unable to access the id on instantEmail... its only letting me use isLoading.

aferd: any ideas here?

PR #1340, opened 2022-07-05
NOTIF-711 Return the event types associated to a behavior group

josejulio:


gwenneg (commented):
The PR title mentions returning event types but the PR only filters out / removes data from the REST responses. Is there some code missing?

common/src/main/java/com/redhat/cloud/notifications/models/BehaviorGroup.java
```diff
@@ -74,9 +74,12 @@ public class BehaviorGroup extends CreationUpdateTimestamped {
     private boolean filterOutActions;
 
     @OneToMany(mappedBy = "behaviorGroup", cascade = CascadeType.REMOVE)
-    @JsonIgnore
```
josejulio, line 77: This is the one that allows to `return the event types associated to a behavior group`.
The rest mimics filtering out in the same places we filter out actions.

gwenneg: I didn't see that one, sorry! The PR makes sense now 😄

PR #1351, opened 2022-07-07
Always log orgId of incoming actions

gwenneg:
@brantleyr asked on Slack if we could confirm that Advisor sends an orgId value. This will give an answer in the engine log.


josejulio (approved):
lgtm

josejulio:
This could also be checked using the event log API

gwenneg:
No we don't return the accountId/orgId in that API.

josejulio:
> No we don't return the accountId/orgId in that API.

Do we filter the payload? I though we had an API to retrieve the payloads

gwenneg:

PR #1352, opened 2022-07-11
Fix Sentry alerts related to Arjuna transactions

gwenneg:


engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -296,10 +280,10 @@ private void processAggregateEmailsByAggregationKey(EmailAggregationKey aggregat
 
                 emailSender.sendEmail(aggregation.getKey(), event, subject, body);
             }
-        } finally {
-            if (delete) {
-                emailAggregationRepository.purgeOldAggregation(aggregationKey, endTime);
-            }
```
gwenneg, line 302: I suspect that running `purgeOldAggregation` from a `finally` block does not work if an exception is thrown during a preceding DB operation. Indeed, the transaction is already dead and interactions with the DB are no longer possible.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -296,10 +280,10 @@ private void processAggregateEmailsByAggregationKey(EmailAggregationKey aggregat
 
                 emailSender.sendEmail(aggregation.getKey(), event, subject, body);
```
gwenneg, line 281: This will never throw any exception raised during the email sending. They are catched in `emailSender`.

PR #1355, opened 2022-07-11
NOTIF-603 Disable ingress channel from the Clowder configuration

gwenneg:


gwenneg:
The default value doesn't affect prod. I'm merging to test it on stage.

PR #1356, opened 2022-07-11
NOTIF-603 Add orgId tests

gwenneg:
After this PR, most of the notifications tests will be run twice:
- with `notifications.use-org-id=false`
- with `notifications.use-org-id=true`

The test code is almost unchanged and there is only one runtime code change highlighted below.


backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
```diff
@@ -539,7 +539,7 @@ public void updateBehaviorGroupActions(String accountId, String orgId, UUID beha
             if (featureFlipper.isUseOrgId()) {
                 upsertQuery = "INSERT INTO behavior_group_action (behavior_group_id, endpoint_id, position, created) " +
                         "SELECT :behaviorGroupId, :endpointId, :position, :created " +
-                        "WHERE EXISTS (SELECT 1 FROM endpoints WHERE orgId " +
+                        "WHERE EXISTS (SELECT 1 FROM endpoints WHERE org_id " +
```
gwenneg, line 542: This is the only runtime change. It fixes a regression.

gwenneg:
ℹ️ I did a full self-review. I'm not waiting for an approval before the merge as this barely changes anything in the project code.

PR #1357, opened 2022-07-12
Lowering character count so k8s won't cry

gburges:


gwenneg (commented):
Shouldn't [this file](https://github.com/RedHatInsights/notifications-backend/blob/master/.rhcicd/run_org_id_populator.yaml) be updated as well?

gburges:
> Shouldn't [this file](https://github.com/RedHatInsights/notifications-backend/blob/master/.rhcicd/run_org_id_populator.yaml) be updated as well?

riight, i didn't update because i'm using a local copy paste :sweat: update inc

gwenneg (approved):
LGTM

PR #1359, opened 2022-07-12
NOTIF-603 Add dedicated orgId feature flag for /events

gwenneg:
`GET /events` cannot rely on the `org_id` that comes from the user authentication headers yet because some onboarded apps still send Kafka messages without an `org_id` value. If we want these messages to be returned by `/events`, then we need to retrieve all events from the DB using the `account_id` for now.

`notifications.use-org-id` will be set to `true` shortly on stage and prod.
`notifications.events.use-org-id` will be set to `true` a bit later when all the Kafka messages we receive will contain an `org_id`.


josejulio (approved):
lgtm

PR #1360, opened 2022-07-12
Remove QA workaround for the Update behavior group endpoint

josejulio:
This removes the workaround that was added to the Update behavior group.

The `bundle_id` was never used, but there are QA tests checking that the value has to be a valid UUID.


josejulio:
@dgaikwad not urgent - but this is the change I was talking about - That param is not being used anywhere and we are removing it.

josejulio:
Checking if QA tests are still using this field.

gwenneg (approved):
LGTM

PR #1368, opened 2022-07-13
use clowder_smoke env

digitronik:
For now, let's use clowder smoke env.
We need further investigation into why in ephemeral env we are not able to authenticate with stage user and now changes for default user iqe-core need to apply in all other dependent plugins.


gwenneg (approved):
LGTM

PR #1378, opened 2022-07-19
Fix the helper text

theute:
I think I got that right, it was confusing to me when I demoed this to QE and I want to test the autodeployment to stage so I will merge myself


theute:
Error is unrelated

PR #1399, opened 2022-08-02
NOTIF-733 ability to create templates in admin UI

aferd:


aferd:
current state - crashes when I go to type in editor

https://user-images.githubusercontent.com/86322239/183966287-8985d4ca-1e96-430d-bf81-c89b6d8383d2.mov

.

josejulio:
> current state - crashes when I go to type in editor

Is this the latest code?

aferd:
> 

should be now! With that change the editor no longer crashes.

josejulio:
This OK to start with, but we should probably have a preview in the next iteration (could use the same mechanism as the `Email templates` page.

josejulio (approved):

aferd:
> This OK to start with, but we should probably have a preview in the next iteration (could use the same mechanism as the `Email templates` page.

yes, this is the first iteration. I wanted to just get the ability to add(and edit WIP) up and going. Next I will work on showing a preview.

PR #1405, opened 2022-08-04
NOTIF-744 Translate account_id to org_id using BOP

gwenneg:
The translation is disabled by default and can be enabled with `notifications.translate-account-id-to-org-id=true`.

If the call to BOP fails for any reason, the Kafka message processing will continue (with a missing org ID) and a warn message will be logged and trigger a Sentry alert.

Org IDs are cached indefinitely because the translation result never changes for a given EAN.

### DIFF
--- a/engine/src/test/java/com/redhat/cloud/notifications/events/EventConsumerTest.java
+++ b/engine/src/test/java/com/redhat/cloud/notifications/events/EventConsumerTest.java
@@ -1,6 +1,7 @@
 package com.redhat.cloud.notifications.events;
 
 import com.redhat.cloud.notifications.MicrometerAssertionHelper;
+import com.redhat.cloud.notifications.MockServerConfig;
 import com.redhat.cloud.notifications.TestLifecycleManager;
 import com.redhat.cloud.notifications.config.FeatureFlipper;
 import com.redhat.cloud.notifications.db.repositories.EventRepository;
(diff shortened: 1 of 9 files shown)


engine/src/test/java/com/redhat/cloud/notifications/events/EventConsumerTest.java
```diff
@@ -186,10 +189,13 @@ void testValidPayloadWithoutOrgId() {
                 MESSAGE_ID_INVALID_COUNTER_NAME,
                 MESSAGE_ID_MISSING_COUNTER_NAME
         );
+        // The action org ID is set by OrgIdTranslator after the Kafka message is received.
+        action.setOrgId(DEFAULT_ORG_ID);
```
pilhuhn, line 193: This confuses me. Shouldn't this have happened already with the same value for orgId?

gwenneg: The `action` object from this test is used twice:
- it is serialized and sent as a Kafka payload (with `org_id=null`)
- it is compared with the action (another instance) that is processed by `EndpointProcessor` (this one has its `org_id` field populated with `OrgIdTranslator`)

The line you highlighted is meant to verify that `OrgIdTranslator` populated the `org_id` as intended. Since the original action object from the test didn't have an `org_id` value, it has to be set here, before the check.

gwenneg: This line will go away as soon as we're done with the orgId migration.

pilhuhn: Thanks for the explanation

pilhuhn (approved):

PR #1408, opened 2022-08-04
NOTIF-744 Replace BOP with IT users service

gwenneg:
Calling BOP to translate an `account_id` to an `org_id` didn't work on stage because of certificate issues.

This PR replaces the BOP call with an IT users service call, which should work because we already configured the certificate for that service. I reverse engineered BOP to determine how to call the IT users service and translate the `account_id`. It may not work on first try but it's still disabled by default with the feature flag so it won't do any harm.

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/events/orgid/ITUserServiceV2.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/events/orgid/ITUserServiceV2.java
@@ -3,27 +3,19 @@ package com.redhat.cloud.notifications.events.orgid;
 import org.eclipse.microprofile.rest.client.inject.RegisterRestClient;
 
 import javax.ws.rs.Consumes;
-import javax.ws.rs.HeaderParam;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
-import java.util.List;
-import java.util.Map;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 
 // TODO NOTIF-744 Remove this as soon as all onboarded apps include the org_id field in their Kafka messages.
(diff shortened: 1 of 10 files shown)


engine/src/main/java/com/redhat/cloud/notifications/events/orgid/OrgIdTranslator.java
```diff
@@ -2,39 +2,30 @@
...
-    String bopEnv;
-
     @Inject
     @RestClient
-    Bop bop;
+    ITUserServiceV2 itUserServiceV2;
 
     @CacheResult(cacheName = "account-id-to-org-id")
     public String translate(String accountId) {
         Log.debugf("Calling BOP to translate EAN %s to an org ID", accountId);
```
pilhuhn, line 20: Change log message?

gwenneg: Good catch. Fixed.

engine/src/main/java/com/redhat/cloud/notifications/events/orgid/ITUserServiceV2.java
```diff
@@ -3,27 +3,19 @@
...
 import javax.ws.rs.Produces;
-import java.util.List;
-import java.util.Map;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 
 // TODO NOTIF-744 Remove this as soon as all onboarded apps include the org_id field in their Kafka messages.
-@RegisterRestClient(configKey = "bop")
-public interface Bop {
+@RegisterRestClient(configKey = "it-s2s")
```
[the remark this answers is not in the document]
gwenneg: I know of the other one, I don't think they will clash. I kept them separated because the new one will be deleted very soon.

PR #1412, opened 2022-08-05
NOTIF-603 Add org_id migration REST API for events

gwenneg:
Migrating the events with the REST API should be easier/faster than using the `org-id-populator` job.


gwenneg:
TODO:
- Drop the `event` indexes before the migration.
- Test the migration API in this PR.
- Create new `event` indexes based on the `org_id` after the migration.

josejulio (approved):
Lgtm

PR #1416, opened 2022-08-10
NOTIF-181 Adding default template for preprod environments

josejulio:
Adds a new flag to decide if we use a default email template. This is meant for pre-production as it will include the json content of the notification.

The default email template is only going to be used if the application didn't provide (already) a template themselves.

Todo:
 - [x] Use the flag to allow to subscribe to user preferences 
 - [x] Add tests

For a fake notification with bundle: `my-bundle`, application: `my-app` and event type: `test-email-subscription-instant`

The title is: `03 Aug 2020 15:22 UTC - my-bundle/my-app/test-email-subscription-instant triggered`

And the content of the email is:

Screenshot_2022-08-19_07-22-55

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
@@ -162,14 +162,13 @@ public class EmailSubscriptionTypeProcessor implements EndpointTypeProcessor {
         if (featureFlipper.isUseTemplatesFromDb()) {
             Optional<InstantEmailTemplate> instantEmailTemplate = templateRepository
                     .findInstantEmailTemplate(event.getEventType().getId());
-            if (instantEmailTemplate.isPresent()) {
+            if (instantEmailTemplate.isEmpty()) {
+                return Collections.emptyList();
+            } else {
                 String subjectData = instantEmailTemplate.get().getSubjectTemplate().getData();
                 subject = templateService.compileTemplate(subjectData, "subject");
                 String bodyData = instantEmailTemplate.get().getBodyTemplate().getData();
                 body = templateService.compileTemplate(bodyData, "body");
-            } else {
-                subject = null;
-                body = null;
             }
         } else {
             if (!emailTemplate.isSupported(action.getEventType(), emailSubscriptionType)) {
@@ -187,7 +186,6 @@ public class EmailSubscriptionTypeProcessor implements EndpointTypeProcessor {
                 if (featureFlipper.isUseDefaultTemplate()) {
                     fileTemplateBody = Default.getBody(fileTemplateSubject != null, fileTemplateBody != null);
                     fileTemplateSubject = Default.getTitle();
-
                 } else {
                     return Collections.emptyList();
                 }
@@ -195,10 +193,6 @@ public class EmailSubscriptionTypeProcessor implements EndpointTypeProcessor {
 
             subject = fileTemplateSubject;
             body = fileTemplateBody;
-        }
-
-        if (subject == null || body == null) {
-            return Collections.emptyList();
         }
 
         Set<RecipientSettings> requests = Stream.concat(
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
@@ -2,6 +2,7 @@ package com.redhat.cloud.notifications.processors.email;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.redhat.cloud.notifications.config.FeatureFlipper;
 import com.redhat.cloud.notifications.db.StatelessSessionFactory;
 import com.redhat.cloud.notifications.db.repositories.EmailAggregationRepository;
 import com.redhat.cloud.notifications.db.repositories.EmailSubscriptionRepository;
@@ -23,6 +24,7 @@ import com.redhat.cloud.notifications.recipients.RecipientSettings;
 import com.redhat.cloud.notifications.recipients.User;
 import com.redhat.cloud.notifications.recipients.request.ActionRecipientSettings;
 import com.redhat.cloud.notifications.recipients.request.EndpointRecipientSettings;
+import com.redhat.cloud.notifications.templates.Default;
 import com.redhat.cloud.notifications.templates.EmailTemplate;
 import com.redhat.cloud.notifications.templates.EmailTemplateFactory;
 import com.redhat.cloud.notifications.templates.TemplateService;
@@ -33,7 +35,6 @@ import io.quarkus.logging.Log;
 import io.quarkus.qute.TemplateInstance;
 import io.smallrye.reactive.messaging.annotations.Blocking;
 import io.vertx.core.json.JsonObject;
-import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.eclipse.microprofile.reactive.messaging.Acknowledgment;
 import org.eclipse.microprofile.reactive.messaging.Incoming;
 
@@ -52,8 +53,6 @@ import java.util.UUID;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import static com.redhat.cloud.notifications.templates.TemplateService.USE_TEMPLATES_FROM_DB_KEY;
-
 @ApplicationScoped
 public class EmailSubscriptionTypeProcessor implements EndpointTypeProcessor {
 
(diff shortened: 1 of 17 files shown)


engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -181,7 +186,21 @@ private List<NotificationHistory> sendEmail(Event event, Set<Endpoint> endpoints
...
+            }
+
+            if (featureFlipper.isUseDefaultTemplate()) {
+                if (subject == null) {
+                    subject = Default.getTitle();
+                }
+
+                if (body == null) {
+                    body = Default.getBody();
+                }
```
gwenneg, line 200: I'm not sure about this.

Mixing a default title with a non-default body could go unnoticed from an onboarded app perspective.

Also, mixing a default body with a non-default title will result in an email notification with a message that is not accurate: there is actually an email template associated with the event type, but one part of the template is missing.

gwenneg (commented):
Here's a partial review. I'd like to discuss about my last comment before I continue reviewing.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
josejulio: You are right there, I guess we could overwrite both and put a message in the email body

"This event type had a subject (<or body>) template, but not <the other>"
The warning would be useful too.

engine/src/main/java/com/redhat/cloud/notifications/templates/extensions/ActionExtension.java
```diff
@@ -1,19 +1,38 @@
...
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import com.redhat.cloud.notifications.ingress.Action;
 import com.redhat.cloud.notifications.ingress.Context;
 import com.redhat.cloud.notifications.ingress.Payload;
 import io.quarkus.qute.TemplateExtension;
 
 // TODO NOTIF-484 Remove this annotation, it has no effect while using a standalone Qute engine.
```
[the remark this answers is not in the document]
josejulio: Whoops

engine/src/main/resources/templates/Default/instantEmailBody.html
```diff
@@ -0,0 +1,36 @@
...
+    <td class="rh-content__block">
+        <p>Please see <a class="rh-url" href="https://core-platform-apps.pages.redhat.com/notifications-docs/dev/user-guide/email-templates.html">documentation</a> for information about setting up email templates.</p>
+    </td>
+</tr>
+<tr>
+    <td class="rh-content__block">
+        <table class="rh-data-table rh-m-bordered">
+            <thead>
+                <tr>
+                    <th>RAW EVENT CONTENT</th>
```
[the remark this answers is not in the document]
josejulio: I just copied the template - but you are right, it's not an `event`. Updating to `ACTION`

engine/src/test/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessorTest.java
```diff
@@ -128,6 +130,24 @@ void shouldSuccessfullySendEmail() {
...
+                .withEventType("event-1")
+                .build()
+        );
+
+        testee.process(event, List.of(
+                new Endpoint()
+        ));
+        featureFlipper.setUseDefaultTemplate(false);
+    }
+
```
josejulio, line 150: Looks like i didn't finish this one.

josejulio:
Addressed all the comments, thanks!

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -181,7 +178,17 @@ private List<NotificationHistory> sendEmail(Event event, Set<Endpoint> endpoints
...
 
         if (subject == null || body == null) {
-            return Collections.emptyList();
+            if (subject != null || body != null) {
+                Log.warnf("Only one of the subject and body was found for event: %s/%s/%s", event.getBundleDisplayName(), event.getApplicationDisplayName(), event.getEventTypeDisplayName());
+            }
+
+            if (featureFlipper.isUseDefaultTemplate()) {
+                body = Default.getBody(subject != null, body != null);
+                subject = Default.getTitle();
```
gwenneg, line 187: Do you plan on keeping `Default` hardcoded as a filesystem-based template after we've enabled the DB templates? 

`Default` would definitely not be a conventional `InstantEmailTemplate` because it would never be associated with any event type, but the relation with the event type is optional so that could still work. It would probably require a new boolean field in `InstantEmailTemplate` to tag it as the default template and prevent linking it with a particular event type.

josejulio: I was thinking on keeping it on the filesystem, but then all the `annotations` required to work and the configuration would need to remain compatible for the filesystem templaltes....
So yes, maybe we will move it to the database as well. We could keep a separate class to do the "custom" configuration 
(i.e. adding the information if a explicit subject or body exists) to the template.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -158,26 +157,45 @@ private List<NotificationHistory> sendEmail(Event event, Set<Endpoint> endpoints
...
+        final TemplateInstance body;
 
-        if (useTemplatesFromDb) {
+        if (featureFlipper.isUseTemplatesFromDb()) {
             Optional<InstantEmailTemplate> instantEmailTemplate = templateRepository
                     .findInstantEmailTemplate(event.getEventType().getId());
-            if (instantEmailTemplate.isEmpty()) {
-                return Collections.emptyList();
-            } else {
+            if (instantEmailTemplate.isPresent()) {
```
gwenneg, line 166: The old "if template is empty then return Collection.emptyList()" could be restored here with your latest changes to the repository.

engine/src/main/java/com/redhat/cloud/notifications/templates/Default.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * This class handles the default emails by wrapping the original EmailTemplate.
+ * If the title or body is already supported, it will pass trough.
+ * If the title or body is not supported in the wrapped class AND the type is instant, it will return the default one.
+ * Else it will throw an exception (or reuse the one provided by the call to the wrapped class)
+ * This is only useful for the templates as files flow.
+ *
+ * For templates in database, it provides static methods for getting the TemplateInstance for the title and body.
+ */
+public class Default implements EmailTemplate {
```
gwenneg, line 18: The latest changes in `TemplateRepository` look good to me 👍

There is however something I don't understand in this class. What is the point of wrapping templates into a `Default` (when the feature is enabled) if the logic that determines [whether the default body/title should be used or not] is located in `EmailSubscriptionTypeProcessor`? Maybe I missed something but it feels like `EmailSubscriptionTypeProcessor` would keep working if you removed the wrapping here.

josejulio: Let me check again, the whole point of this started as a way to centralize that logic on this class, but surely that has diverged from the intention.

josejulio: As said, previously this `Default` also had the logic to decide if we use the original template or the default template. 

I had to move out this logic to take into account the case when only one of  the templates was present (i.e. not mixing default title with original and viceversa).

The logic that remains there is to check for the `isSupported*` methods, it will always support anything that has `INSTANT` regardless of the wrapped class value. This is used to avoid duplicating that logic in `user preferences` and `EmailSubscriptionTypeProcessor`.

josejulio:
Comments addressed and rebased to latest changes.

engine/src/main/java/com/redhat/cloud/notifications/templates/Default.java
gwenneg: Ok, I missed a few details, the wrapping is indeed necessary. Thanks for clarifying.

The current code is a bit hard to fully understand at first glance, but the tricky part is only temporary so merging it sounds reasonable to me.

josejulio: Yes - the idea is to get rid of it soon - Actually the part where the "warning" is added to the default body template only happens with in-file templates - as database templates will require both templates to exist.

engine/src/test/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessorTest.java
```diff
@@ -48,9 +73,30 @@ class EmailSubscriptionTypeProcessorTest {
...
     @Inject
     MicrometerAssertionHelper micrometerAssertionHelper;
 
+    @Inject
+    FeatureFlipper featureFlipper;
+
+    @BeforeEach
+    void beforeEach() {
+        featureFlipper.setUseTemplatesFromDb(false);
+        featureFlipper.setUseTemplatesFromDb(false);
```
gwenneg, line 97: Is this what you meant to write?
```suggestion
        featureFlipper.setUseTemplatesFromDb(false);
        featureFlipper.setUseDefaultTemplate(false);
```

josejulio: Whoops - yes

gwenneg (approved):
LGTM

PR #1421, opened 2022-08-17
Changes to Advisor deactivated recommendation email template

mhuth:


gwenneg:
A few changes are required to use `{environment.url}` in that kind of test as it's not available by default while testing.

First, you need to add an injection to the test class:
```
@Inject
Environment environment;
```

Then, before the template is rendered, the `environment` instance must be passed to the template instance:

**Before**
```
        String result = Advisor.Templates.newRecommendationInstantEmailTitle()
                .data("action", action)
                .render();
```

**After**
```
        String result = Advisor.Templates.newRecommendationInstantEmailTitle()
                .data("action", action)
                .data("environment", environment)
                .render();
```

gwenneg:
@mhuth How much time do you need to test this on stage? I could merge the PR now (the failed GH check is unrelated) but if I do, the changes may be deployed to prod shortly because we are deployed to prod today.

engine/src/main/resources/templates/Advisor/insightsEmailBody.html
```diff
@@ -560,7 +560,7 @@
                                 </tr>
                                 <tr>
                                     <td class="rh-masthead__subhead">
-                                        <h1><b>{#insert content-title}{/}</b></h1>
+                                        <h1><b style="line-height: 1.5;">{#insert content-title}{/}</b></h1>
```
gwenneg, line 563: I didn't know `style` worked on a `b` tag :smiley:

gwenneg (approved):
LGTM

engine/src/main/resources/templates/Advisor/insightsEmailBody.html
mhuth: It seemed to work :)  But I'll change it to the h1.

mhuth:
@gwenneg yes feel free to go ahead and merge it so it makes it into prod.  These notifications don't happen very often anyway.

gwenneg (approved):

mhuth:
@gwenneg I'm not sure about the failing tests though.  They seem to be unrelated to my changes.

gwenneg:
That's unrelated indeed. We have an issue with QA tests today.

PR #1426, opened 2022-08-19
Run ANALYZE after daily VACUUM

gwenneg:
`ANALYZE` updates Postgres' optimizer statistics. The query planner can only work better after that.

I don't know why I didn't add it to the cronjob initially. We already used `VACUUM ANALYZE` from Flyway scripts in the past.

### DIFF
--- a/.rhcicd/clowdapp-backend.yaml
+++ b/.rhcicd/clowdapp-backend.yaml
@@ -400,10 +400,10 @@ objects:
     clean.sql: |
       \timing
       CALL cleanEventLog();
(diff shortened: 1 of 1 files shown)


josejulio (approved):

josejulio:
Leaving a link to the [VACUUM docs](https://www.postgresql.org/docs/current/sql-vacuum.html#:~:text=VACUUM%20ANALYZE%20performs%20a%20VACUUM,it%20available%20for%20re%2Duse.)

PR #1354, opened 2022-07-11
NOTIF-603 Add org_id indexes for the Event Log

gwenneg:
Current `event` indexes retrieved on prod with gabi:
```
CREATE UNIQUE INDEX pk_event ON public.event USING btree (id)
CREATE INDEX ix_event_index_only_scan ON public.event USING btree (account_id, bundle_id, application_id, event_type_display_name, created DESC, id)
CREATE INDEX ix_event_account_id ON public.event USING btree (account_id, created DESC, id)
CREATE INDEX ix_event_account_id_application_id ON public.event USING btree (account_id, application_id, created DESC, id)
CREATE INDEX ix_event_org_id ON public.event USING btree (org_id)
```

PR #1428, opened 2022-08-19
NOTIF-603 Make Org ID mandatory and account optional in DB

gwenneg:
Depends on #1354.

:warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning:
This PR must be merged after #1354 and after the Event Log has been successfully tested with `NOTIFICATIONS_EVENTS_USE_ORG_ID=true` on prod.
:warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning:

This change is required to make the app compatible with anemic tenants.

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EmailSubscriptionRepository.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EmailSubscriptionRepository.java
@@ -19,7 +19,7 @@ public class EmailSubscriptionRepository {
 
     public List<String> getEmailSubscribersUserId(String accountId, String orgId, String bundleName, String applicationName, EmailSubscriptionType subscriptionType) {
         if (orgIdHelper.useOrgId(orgId)) {
-            String query = "SELECT es.id.userId FROM EmailSubscription es WHERE orgId = :orgId AND application.bundle.name = :bundleName " +
+            String query = "SELECT es.id.userId FROM EmailSubscription es WHERE id.orgId = :orgId AND application.bundle.name = :bundleName " +
                     "AND application.name = :applicationName AND id.subscriptionType = :subscriptionType";
             return statelessSessionFactory.getCurrentSession().createQuery(query, String.class)
                     .setParameter("orgId", orgId)
@@ -28,7 +28,7 @@ public class EmailSubscriptionRepository {
                     .setParameter("subscriptionType", subscriptionType)
                     .getResultList();
         } else {
-            String query = "SELECT es.id.userId FROM EmailSubscription es WHERE id.accountId = :accountId AND application.bundle.name = :bundleName " +
+            String query = "SELECT es.id.userId FROM EmailSubscription es WHERE accountId = :accountId AND application.bundle.name = :bundleName " +
                     "AND application.name = :applicationName AND id.subscriptionType = :subscriptionType";
             return statelessSessionFactory.getCurrentSession().createQuery(query, String.class)
                     .setParameter("accountId", accountId)
--- a/aggregator/src/test/java/com/redhat/cloud/notifications/db/EmailAggregationResourcesTest.java
+++ b/aggregator/src/test/java/com/redhat/cloud/notifications/db/EmailAggregationResourcesTest.java
@@ -10,15 +10,12 @@ import io.vertx.core.json.JsonObject;
 import org.junit.jupiter.api.Test;
 
 import javax.inject.Inject;
-import javax.persistence.EntityManager;
-import javax.transaction.Transactional;
 
 import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.util.List;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNull;
 
 @QuarkusTest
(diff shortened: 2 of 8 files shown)


database/src/main/resources/db/migration/V1.57.0__NOTIF-603_anemic_tenants.sql
```diff
@@ -0,0 +1,26 @@
+-- Table: behavior_group
+DROP INDEX ix_behavior_group_account_id;
+
+-- Table: email_aggregation
+ALTER TABLE email_aggregation
+    ALTER COLUMN account_id DROP NOT NULL,
+    ALTER COLUMN org_id SET NOT NULL;
```
gwenneg, line 7: This line depends on https://github.com/RedHatInsights/notifications-backend/pull/1425.

database/src/main/resources/db/migration/V1.57.0__NOTIF-603_anemic_tenants.sql
```diff
@@ -0,0 +1,26 @@
...
+-- Table: endpoints
+DROP INDEX ix_endpoints_account_id;
+
+-- Table: event
+DROP INDEX ix_event_account_id;
+DROP INDEX ix_event_account_id_application_id;
+DROP INDEX ix_event_index_only_scan;
+ALTER TABLE event
+    ALTER COLUMN account_id DROP NOT NULL,
+    ALTER COLUMN org_id SET NOT NULL;
```
gwenneg, line 26: This line won't work on prod for now. The DB contains old events that were not migrated because of an empty `account_id` field. They will be automatically purged eventually, but we could also purge them sooner manually.

josejulio: Are these events even valid? 
if not, we could clean them with this migration

gwenneg: No, they're not valid. One of the onboarded apps sent an empty `account_id` field by mistake for a while.

I know that Flyway sometimes has difficulties with migration scripts that contain both DML and DDL. I can't say for sure that it will work if we removing the events from this script.

gwenneg: I'll check later for Flyway :smiley:

josejulio: ack - could we create a new migration script for that then? or when are these events up for automatic cleaning?

gwenneg: Yes, we can add another DML script to this PR with a version lower than the DDL script.

I think they won't be automatically purged before one or maybe two weeks.

gwenneg: Done.

gwenneg:
:information_source: This PR is ready for a review, but will remain a draft until the time to merge it has come.

gwenneg:
The Event Log now relies on the Org ID on prod.

This PR can be reviewed and merged.

gwenneg:
Checked with `gabi` on prod today:

| Query | Result |
| - | - |
| select count(*) from behavior_group where org_id is null | 0 |
| select count(*) from email_aggregation where org_id is null | 0 |
| select count(*) from endpoint_email_subscriptions where org_id is null | 0 |
| select count(*) from endpoints where org_id is null | 1 (system endpoint) |
| select count(*) from event where org_id is null | 0 |

gwenneg:
:warning: Drafted to prevent a merge, but it can still be reviewed. :warning:

gwenneg:
:warning: Drafted to prevent a merge, but it can still be reviewed. :warning:

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EmailSubscriptionRepository.java
```diff
@@ -28,7 +28,7 @@ public List<String> getEmailSubscribersUserId(String accountId, String orgId, St
                     .setParameter("subscriptionType", subscriptionType)
                     .getResultList();
         } else {
-            String query = "SELECT es.id.userId FROM EmailSubscription es WHERE id.accountId = :accountId AND application.bundle.name = :bundleName " +
+            String query = "SELECT es.id.userId FROM EmailSubscription es WHERE accountId = :accountId AND application.bundle.name = :bundleName " +
```
josejulio, line 31: Are we keeping this branch?

josejulio (approved):
lgtm

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EmailSubscriptionRepository.java
josejulio: I'm guessing it will go away in a follow up cleanup :grin:

gwenneg: Exactly :smiley:

PR #1434, opened 2022-08-22
Remove unavailable settings from user preferences call

josejulio:
Remove (from user preferences) applications  that don't have any valid option.

e.g. We want to stop showing empty applications here such as `Storage broker` or `Internal`:

Screenshot_2022-08-22_10-36-24


gwenneg (approved):
LGTM

PR #1439, opened 2022-08-23
NOTIF-759 Clean up post-migration dead code in backend

gwenneg:
image


josejulio:
It's the `useOrgId` thing still in the `FeatureFlipper` ?

josejulio:
> It's the `useOrgId` thing still in the `FeatureFlipper` ?

I see is still in here: https://github.com/RedHatInsights/notifications-backend/blob/8de71b656313557a58974fcbef29d43b9b5d9bb1/common/src/main/java/com/redhat/cloud/notifications/config/FeatureFlipper.java#L55

Do we still depend on it? could remove it? (the useOrgId and friends)

josejulio (approved):
LGTM,
I assume featureFlipper and orgIdHelper are going to be updated in a follow up PR when all the modules are accountId-free ?

gwenneg:
> I assume featureFlipper and orgIdHelper are going to be updated in a follow up PR when all the modules are accountId-free ?

Yes, `engine` still depends on them after this PR. They'll be removed in the next one.

gwenneg:
@josejulio I'd like to move forward and merge this PR to finish the cleaning. I don't think we will ever roll things back to using the EAN now. WDYT? That question also applies to the cleaning PR in policies.

gwenneg:
I'll check the PR check failures before merging.

gwenneg:
The IQE tests are broken by this PR on ephemeral because the `org_id` value is not sent in the `x-rh-identity` header by the `iqe-notifications-plugin` project.

gwenneg:
The IQE tests were fixed on ephemeral and changes made to investigate that issue were reverted.

I'll merge as soon as all checks are successful.

PR #1441, opened 2022-08-24
NOTIF-759 Clean up post-migration dead code in engine

gwenneg:
Next step of #1439.


engine/src/main/java/com/redhat/cloud/notifications/models/EmailAggregation.java
```diff
@@ -26,9 +26,6 @@ public class EmailAggregation extends CreationTimestamped {
     @JsonProperty(access = READ_ONLY)
     private Integer id;
 
-    @Size(max = 50)
-    private String accountId;
```
gwenneg, line 30: This field will be removed from Postgres after this PR has been deployed on prod.

engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
```diff
@@ -114,24 +105,7 @@ public CompletionStage<Void> process(Message<String> message) {
...
-                String bundle = bundleName[0] == null ? "" : bundleName[0];
-                String application = appName[0] == null ? "" : appName[0];
-                Counter missingOrgIdCounter = registry.counter(MISSING_ORG_ID, "bundle", bundle, "application", application);
-                missingOrgIdCounter.increment();
-                if (featureFlipper.isUseOrgId()) {
-                    Log.info("The org ID migration is enabled but the orgId field is missing or blank in the action");
-                }
-                if (featureFlipper.isTranslateAccountIdToOrgId() && action.getAccountId() != null && !action.getAccountId().isBlank()) {
-                    try {
-                        String orgId = orgIdTranslator.translate(action.getAccountId());
```
gwenneg, line 128: We didn't translate anything since August 22 on prod. This can be safely removed.

aggregator/src/test/java/com/redhat/cloud/notifications/helpers/ResourceHelpers.java
```diff
@@ -33,9 +33,9 @@ public void purgeEmailAggregations() {
 
     @Transactional
     public Integer purgeOldAggregation(EmailAggregationKey key, LocalDateTime lastUsedTime) {
-        String query = "DELETE FROM EmailAggregation WHERE accountId = :accountId AND bundleName = :bundleName AND applicationName = :applicationName AND created <= :created";
+        String query = "DELETE FROM EmailAggregation WHERE orgId = :orgId AND bundleName = :bundleName AND applicationName = :applicationName AND created <= :created";
         return session.createQuery(query)
-                .setParameter("accountId", key.getAccountId())
+                .setParameter("orgId", key.getOrgId())
```
josejulio, line 38: Does all of our current aggregations have the `orgId` set? Just checking we are not ending up with data that will never go away (although we can check later if we have old data)

aggregator/src/test/java/com/redhat/cloud/notifications/models/EmailAggregation.java
```diff
@@ -26,9 +26,6 @@ public class EmailAggregation extends CreationTimestamped {
     @JsonProperty(access = READ_ONLY)
     private Integer id;
 
-    @Size(max = 50)
-    private String accountId;
```
josejulio, line 30: I suppose this is something we are deleting later - is there a list of the columns we are deleting?

aggregator/src/test/java/com/redhat/cloud/notifications/helpers/ResourceHelpers.java
gwenneg: Yes they do. The `org_id` field was changed to `NOT NULL` in https://github.com/RedHatInsights/notifications-backend/pull/1428.

aggregator/src/test/java/com/redhat/cloud/notifications/models/EmailAggregation.java
gwenneg: Yes, this is what I mentioned [here](https://github.com/RedHatInsights/notifications-backend/pull/1441#discussion_r959940772) (same DB field even though the Java class is different).

gwenneg: This will be the only field removed for now.

engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java
```diff
@@ -141,26 +137,26 @@ public List<User> getUsers(String accountId, String orgId, boolean adminsOnly) {
         users = transformToUser(usersTotal);
 
         // Micrometer doesn't like when tags are null and throws a NPE.
-        String accountIdTag = accountId == null ? "" : accountId;
-        getUsersTotalTimer.stop(meterRegistry.timer("rbac.get-users.total", "accountId", accountIdTag));
-        getRbacUsersGauge(accountIdTag).set(users.size());
+        String orgIdTag = orgId == null ? "" : orgId;
```
josejulio, line 140: Do we have cases where the `orgId` is actually `null` ?

gwenneg: It's not supposed to happen now but it still can because the `org_id` field isn't required yet in the notification schema.

gwenneg:
Rebased to fix conflicts.

josejulio (approved):
lgtm

PR #1443, opened 2022-08-24
Make do-nothing job clowderized

gburges:


gwenneg (approved):
LGTM

Do we need to update app-interface after this PR is merged or will that work out of the box?

gburges:
> 

Should work out of the box

PR #1449, opened 2022-08-29
Fix accountId and orgId constraints

gwenneg:
@josejulio I just realized that this should have been done in #1428. Please deploy this PR on prod ASAP (tomorrow is fine).


josejulio:
Thanks, will deploy tomorrow morning

PR #1451, opened 2022-08-30
make it cronjob, so it's automatic

gburges:


gburges:
we'll need to update the ref in app-interface so it deploys like this, but that should solve (finally) the issues

gwenneg:
Will that stop the spam in NOTIF-676?

gburges:
> Will that stop the spam in NOTIF-676?

That's the intention, yeah :) It also unblocks onboarding

gwenneg (approved):
LGTM

PR #1453, opened 2022-08-30
NOTIF-698 add create or edit for aggregation templates

aferd:


admin-console/src/main/webapp/src/pages/ApplicationPage/useApplicationTypes.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    const { query } = useClient();
+
+    useEffect(() => {
+        setAggregationRows(undefined);
+    }, [ applicationId ]);
+
+    useEffect(() => {
+        if (applicationTypesQuery.payload?.status === 200) {
+            const application = applicationTypesQuery.payload.value.map<AggregationCardRow>(e => ({
```
aferd, line 31: application here is showing that it is cast to any `const application: any` but it needs to show AggregationCardRow, could you take a look by chance?

josejulio: It's that type apparently (`Application1`) but you can't interate over it with a map, because it's a single element.
It should be:
```
const application = {
                ...applicationTypesQuery.payload.value,
                aggregationEmail: {
                    isLoading: true
                }
            };
```

admin-console/src/main/webapp/src/pages/ApplicationPage/useApplicationTypes.ts
```diff
@@ -0,0 +1,81 @@
...
+            setAggregationRows(aggregationRows);
+
+            if (application) {
+                let row;
+                query(Operations.TemplateResourceGetAggregationEmailTemplatesByApplication.actionCreator({
+                    appId: application.id ?? ''
+                }))
+                .then(result => {
+                    if (result.payload?.status === 200 || result.payload?.status === 401) {
+                        const value: Partial<AggregationTemplate> = result.payload.status === 200 ? {
```
aferd, line 47: error here - type response200 is not assignable to type string

aferd:
@josejulio if/when you have time, could you review this?

PR #1454, opened 2022-08-31
Support for account_id being null in the json string

josejulio:
Uses the version 0.16 of the schema, where the `account_id` can be the null value in the json string 
In a JSON a property can be `null`, `undefined` (explicitely or by not being present) or have other value. The java implementation considers a null java values for both the received `undefined` and `null` json values.

The only limitation is that `account_id` now is a java `Object` (by the way the json schema was made and maybe my poor knowledge of it, see [here](https://github.com/RedHatInsights/insights-schemas-java/pull/59))


josejulio:
@gwenneg The missing version should be released by the time you are back from PTO :)

gwenneg (approved):
It'd be great if we could remove all these casts in the future.

LGTM

PR #1459, opened 2022-09-05
NOTIF-764 Return 401 when org_id is missing in the x-rh-identity header

gwenneg:
If the `org_id` field is missing in the `x-rh-identity` header, the client will receive a 401 HTTP status in the response and an exception will be logged at `ERROR` level from this location:

https://github.com/RedHatInsights/notifications-backend/blob/5364662f96b485d735440c1373ddaa8b591196e5/backend/src/main/java/com/redhat/cloud/notifications/auth/ConsoleIdentityProvider.java#L160-L164

```
2022-09-05 10:37:42,201 ERROR [com.red.clo.not.aut.ConsoleIdentityProvider] (vert.x-eventloop-thread-0) Error while processing identity: io.quarkus.security.AuthenticationFailedException
	at com.redhat.cloud.notifications.auth.ConsoleIdentityProvider.lambda$authenticate$3(ConsoleIdentityProvider.java:115)
	[...]
Caused by: com.redhat.cloud.notifications.auth.principal.IllegalIdentityHeaderException: The org_id field is missing or blank in the x-rh-identity header
	at com.redhat.cloud.notifications.auth.principal.ConsolePrincipalFactory.fromIdentity(ConsolePrincipalFactory.java:16)
	at com.redhat.cloud.notifications.auth.ConsoleIdentityProvider.lambda$authenticate$7(ConsoleIdentityProvider.java:112)
	[...]
```
This is not supposed to happen while processing public HTTP requests because the upstream security layers should always include the `org_id` field in the header. It can happen though during an IQE test when the header is built by the test.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/auth/ConsoleIdentityProvider.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/auth/ConsoleIdentityProvider.java
@@ -5,6 +5,7 @@ import com.redhat.cloud.notifications.auth.principal.ConsoleIdentity;
 import com.redhat.cloud.notifications.auth.principal.ConsoleIdentityWrapper;
 import com.redhat.cloud.notifications.auth.principal.ConsolePrincipal;
 import com.redhat.cloud.notifications.auth.principal.ConsolePrincipalFactory;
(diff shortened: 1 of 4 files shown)


josejulio (approved):
lgtm

PR #1469, opened 2022-09-07
NOTIF-512 Disable endpoints in case of 4xx or 5xx HTTP response status

gwenneg:
Parts 1 and 2 will only be enabled if `notifications.disable-webhook-endpoints-on-failure=true` in the configuration (false by default).

▶️ **Part 1 of NOTIF-512**

After this PR, endpoints will be automatically disabled if:
- we receive more than `processor.webhook.max-server-errors` (default value: 10) server errors (5xx) from the remote server
- we receive one client error (4xx) from the remote server

The server errors are saved in the `server_errors` field from the `endpoints` DB table. That field is incremented on each server error and when its value exceeds `processor.webhook.max-server-errors`, the field value is reset to `0` and the endpoint is disabled. Each successful call (2xx) resets the `server_errors` field value to `0`, if its value was greater than zero.

The `processor.webhook.disabled.endpoints` counter and its `error_type` (client|server) tag will allow us to monitor the disabled endpoints from the metrics.

▶️ **Part 2 of NOTIF-512**

In a subsequent PR, we will send an email notification to org admins everytime an endpoint is automatically disabled.


engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
```diff
@@ -133,21 +151,68 @@ public NotificationHistory doHttpRequest(Notification item, HttpRequest<Buffer>
...
+                             * The target endpoint returned a 5xx status. That kind of error happens in case of remote
+                             * server failure, which is usually something temporary. Sending another notification to
+                             * the same endpoint may work in the future, so the endpoint is only disabled if the max
+                             * number of endpoint failures allowed from the configuration is exceeded.
+                             */
+                            boolean disabled = endpointRepository.incrementEndpointServerErrors(item.getEndpoint().getId(), maxServerErrors);
+                            if (disabled) {
+                                disabledEndpointsServerErrorCount.increment();
+                                Log.infof("Endpoint %s was disabled because we received too many 5xx status while calling it", item.getEndpoint().getId());
+                                // TODO NOTIF-512 Send a notification to the org admin explaining the situation.
```
josejulio, line 178: We could use the internal event type - and configure the recipients to force send to admins.

Something else we could explore in the future is the use of default settings. i.e. if nothing is configured, send to all admins.

gwenneg: > We could use the internal event type - and configure the recipients to force send to admins.

That's exactly my plan 😃

database/src/main/resources/db/migration/V1.59.0__NOTIF-512_endpoints_server_errors.sql
```diff
@@ -0,0 +1,6 @@
+-- This only affects future inserted rows but not the existing ones.
+ALTER TABLE endpoints
+    ADD COLUMN server_errors INTEGER NOT NULL DEFAULT 0;
+
+-- This updates existing rows.
+UPDATE endpoints SET server_errors = 0;
```
josejulio, line 6: shouldn't the value be already `0` as is the default and has `NOT NULL`?

gwenneg: No, adding a `DEFAULT` rule to an existing table does not update old records. It only affects new records.

However, your question makes me wonder if the `NOT NULL` constraint should be added after the `UPDATE`. I'll run a quick local check for that right away.

gwenneg: > No, adding a `DEFAULT` rule to an existing table does not update old records. It only affects new records.

This may be wrong. Running the `ALTER TABLE` query locally affects old records... It's not supposed to do that according to the Postgres doc but I guess the `NOT NULL` constraint changes things.

gwenneg: I'll remove the `UPDATE` query from the script and we'll see how this goes on stage.

gwenneg: Done.

josejulio: > but I guess the `NOT NULL` constraint changes things

That's what I recall - found this https://stackoverflow.com/a/42150852/688379

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -109,6 +112,95 @@ public List<Endpoint> getTargetEmailSubscriptionEndpoints(String orgId, String b
...
+        Optional<Endpoint> endpoint = lockEndpoint(endpointId);
+        /*
+         * The endpoint should always be present unless it's been deleted recently from another thread or pod.
+         * It may or may not have been disabled already from the frontend or because of a 4xx error.
+         */
+        if (endpoint.isPresent() && endpoint.get().isEnabled()) {
+            if (endpoint.get().getServerErrors() + 1 > maxServerErrors) {
+                /*
+                 * The endpoint exceeded the max server errors allowed from configuration.
+                 * It is therefore disabled and the errors counter is reset.
```
josejulio, line 136: I was wondering if we should leave the max server errors to verify that it was auto-disabled because of server errors.

We could reset this value when updating the integration - and that would make sense I guess - if we update the URL or any configuration, the response might be different.

gwenneg: We can verify the reason why it was disabled from the log entries in `WebhookTypeProcessor`.

I'd rather not reset this field from another place like `backend` REST APIs as it will make the whole thing more complex and harder to understand/maintain.

engine/src/test/java/com/redhat/cloud/notifications/processors/webhook/WebhookTest.java
```diff
@@ -146,6 +176,75 @@ private void testRetry(boolean shouldSucceedEventually) {
...
+            Event event = new Event();
+            event.setAction(action);
+            Endpoint ep = buildWebhookEndpoint(getMockServerUrl() + "/server-error");
+            persistEndpoint(ep);
+            assertTrue(ep.isEnabled());
+            statelessSessionFactory.withSession(statelessSession -> {
+                for (int i = 0; i < 4; i++) {
+                    /*
+                     * The processor retries 3 times in case of server error,
+                     * so the endpoint will actually be called 12 times in this test.
```
josejulio, line 220: Shouldn't it only be called it 10 times ? Or do we keep trying even if the endpoint was disabled by this?

gwenneg: Good question. With the current code, Failsafe will apply its retry policy until it's done retrying a particular call (3 times by default). We could probably interrupt the Failsafe retries somehow but in the meantime it doesn't really hurt if we don't. There is one unlikely yet possible bad scenario: if call #10 fails and call #11 succeeds, we will send an email while the endpoint will actually not be disabled. Maybe that a good enough reason to not let Failsafe complete its retry cycle.

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
josejulio: Not exaclty sure I like this behavior but we can always revisit this - maybe it's an unlikely scenario - besides if it keeps failing it should be disabled.

engine/src/test/java/com/redhat/cloud/notifications/processors/webhook/WebhookTest.java
josejulio: Lets see how this behaves

josejulio: We could create a JIRA to follow up this particular issue

josejulio (approved):
lgtm

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
gwenneg: If you don't like it then it makes me want to improve it 😄 I'll submit another PR with something better.

PR #1476, opened 2022-09-08
Fix email templates preview in admin UI

gwenneg:
The email templates preview currently shows an error message when the page is first loaded on stage because we made the `org_id` field mandatory and the default payload from that page doesn't have that field.


josejulio (approved):
lgtm

PR #1482, opened 2022-09-09
NOTIF-512 Notify org admins when a webhook is disabled

gwenneg:
▶️ **Part 2 of NOTIF-512**

When a webhook is automatically disabled because of 4xx or 5xx failures, an action is emitted on the `platform.notifications.ingress` topic with the following information:
- bundle: `console`
- application: `notifications`
- event type: `integration-disabled`

The action also contains additional information about the issue:
- the endpoint ID and name
- the type of the error (client or server)
- the HTTP status in case of client error
- the number of errors in case of server errors

Using a default behavior group, the event type described above will be used to send an email notification to the org admins of the account that owns the disabled integration. The email body will vary depending on the type of the error.


gwenneg:
First drafts for the email notifications:

image

image

gwenneg:
Rebased to fix conflicts.

josejulio (approved):
lgtm

PR #1483, opened 2022-09-09
Fixing up backend do-nothing

gburges:


.rhcicd/do-nothing.yaml
```diff
@@ -7,26 +7,22 @@ objects:
 - apiVersion: cloud.redhat.com/v1alpha1
   kind: ClowdApp
   metadata:
-    name: notifications-backend-${IMAGE_TAG}
+    name: notifications-do-nothing-${IMAGE_TAG}
     annotations:
       ignore-check.kube-linter.io/unset-cpu-requirements : "Not required for this job that does nothing"  
       ignore-check.kube-linter.io/unset-memory-requirements : "Not required for this job that does nothing"  
   spec:
-    envName: ${ENV_NAME}
```
gwenneg, line 15: Are you sure it's okay to remove this? The ClowdApp validation failed because that field was missing a few days ago.

gburges: i also removed the parameter on on line 27-29 because it will only run on stage

gwenneg: Ok but it did only run on stage before and still failed because of that missing field. See the description in https://github.com/RedHatInsights/notifications-backend/pull/1471.

gburges: pretty weird, the original, didn't have EnvName there either - i'll add it back just to be safe

gwenneg (approved):
LGTM

PR #1484, opened 2022-09-09
NOTIF-512 Reset endpoint server errors on endpoint update

gwenneg:
@josejulio Do you like this better? 😃

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
@@ -133,13 +133,13 @@ public class EndpointRepository {
             if (endpoint.get().getServerErrors() + 1 > maxServerErrors) {
                 /*
                  * The endpoint exceeded the max server errors allowed from configuration.
-                 * It is therefore disabled and the errors counter is reset.
+                 * It is therefore disabled.
                  */
-                String hql = "UPDATE Endpoint SET serverErrors = 0, enabled = FALSE WHERE id = :id";
-                statelessSessionFactory.getCurrentSession().createQuery(hql)
+                String hql = "UPDATE Endpoint SET enabled = FALSE WHERE id = :id AND enabled IS TRUE";
+                int updated = statelessSessionFactory.getCurrentSession().createQuery(hql)
                         .setParameter("id", endpointId)
                         .executeUpdate();
-                return true;
+                return updated > 0;
             } else {
                 /*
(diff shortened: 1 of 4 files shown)


engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -133,9 +133,9 @@ public boolean incrementEndpointServerErrors(UUID endpointId, int maxServerError
             if (endpoint.get().getServerErrors() + 1 > maxServerErrors) {
                 /*
                  * The endpoint exceeded the max server errors allowed from configuration.
-                 * It is therefore disabled and the errors counter is reset.
+                 * It is therefore disabled.
                  */
-                String hql = "UPDATE Endpoint SET serverErrors = 0, enabled = FALSE WHERE id = :id";
+                String hql = "UPDATE Endpoint SET enabled = FALSE WHERE id = :id";
```
josejulio, line 138: one question though, now that we don't reset and we still keep trying on failure (because the 3 retries) - maybe this is going to send 3 emails instead of one? (one for the 10 retry, 11 retry and 12 retry)

gwenneg: Correct, but there is an easy fix for that and I'll push it right away.

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -133,13 +133,13 @@ public boolean incrementEndpointServerErrors(UUID endpointId, int maxServerError
...
+                 * It is therefore disabled.
                  */
-                String hql = "UPDATE Endpoint SET serverErrors = 0, enabled = FALSE WHERE id = :id";
-                statelessSessionFactory.getCurrentSession().createQuery(hql)
+                String hql = "UPDATE Endpoint SET enabled = FALSE WHERE id = :id AND enabled IS TRUE";
+                int updated = statelessSessionFactory.getCurrentSession().createQuery(hql)
                         .setParameter("id", endpointId)
                         .executeUpdate();
-                return true;
+                return updated > 0;
```
gwenneg, line 142: Now the endpoint can't be disabled twice even if there are 11, 12 or 13 server errors.

The returned boolean determines whether we should send an email notification or not.

gwenneg:
I'll add a `backend` test before we merge this.

josejulio (approved):
lgtm

PR #1503, opened 2022-09-15
Fix the JsonProperty of isAdmin and isOrgAdmin

josejulio:
- SnakeCaseStrategy turns `isActive` to `active` - we are receiving `is_active` - adding the explicit property instead
 - The same applies for `isOrgAdmin`


josejulio:
Failing test is unrelated

gwenneg (approved):
LGTM

PR #1507, opened 2022-09-19
NOTIF-789 Prepare migration of ROS templates to DB

gwenneg:


gwenneg:

PR #1508, opened 2022-09-20
Refactor bridge handling.

pilhuhn:
Look the bridge up by name. Create if it does not exist.

There is one test in CamelTypeProcessorTest, that seems order dependent, but I can't figure out yet how/why. In isolation it passes


backend/src/main/resources/application.properties
```diff
@@ -67,7 +67,7 @@ quarkus.cache.caffeine.maintenance.expire-after-write=PT60s
 
 quarkus.rest-client.template-engine.url=${clowder.endpoints.notifications-engine-service:http://localhost:8087}
 
-ob.bridge.uuid=<uuid-of-the-bridge instance>
+ob.bridge.uuid=<name of the bridge>
```
pilhuhn, line 70: This should be `ob.bridge.name=`

pilhuhn:
This one should perhaps not be merged individually, but as part of #1509 
Compile failure comes from one API change that should have been in #1509 only

gwenneg:
Ok then let's close this one and keep the other PR.

gwenneg:
Everything will be merged in #1509.

PR #1509, opened 2022-09-20
Openbridge backchannel

pilhuhn:
Poll the backchannel and update history items accordingly

Builds on #1508


gwenneg:
FYI I have a review in progress, I'll try to finish it today.

common/src/main/java/com/redhat/cloud/notifications/openbridge/BridgeHelper.java
```diff
@@ -39,7 +43,7 @@ public class BridgeHelper {
 
     private Bridge bridgeInstance;
 
-    @ApplicationScoped
+    @RequestScoped
```
gwenneg, line 46: Did you make that change because of the token? I think this should still be `@ApplicationScoped` and the token should be injected at class level instead of calling `getAuthToken()` directly in this method.

engine/src/main/java/com/redhat/cloud/notifications/events/FromOpenBridgeHistoryFiller.java
```diff
@@ -0,0 +1,133 @@
...
+            return;
+        }
+
+        for (ProcessingError pe : items) {
+
+            // We may have seen an item before, skip it
+//            String historyId = pe.getHeaders().get(ORIGINAL_EVENT_ID_HEADER);
+//            if (alreadyProcessed(historyId)) {
+//                continue;
+//            }
```
gwenneg, line 94: I suppose this should be uncommented before the merge.

engine/src/main/java/com/redhat/cloud/notifications/events/FromOpenBridgeHistoryFiller.java
```diff
@@ -0,0 +1,133 @@
...
+
+            statelessSessionFactory.withSession(statelessSession -> {
+                Map<String, Object> decodedPayload = decodeItem(pe);
+                // TODO reinjectIf needed (?)
```
gwenneg, line 98: TODO now or later?

engine/src/main/java/com/redhat/cloud/notifications/events/FromOpenBridgeHistoryFiller.java
```diff
@@ -0,0 +1,133 @@
...
+                try {
+                    notificationHistoryRepository.updateHistoryItem(decodedPayload);
+                    messagesWithError.increment();
+                } catch (Exception e) {
+                    Log.info("|  Update Fail", e);
```
gwenneg, line 103: This seems worth a Sentry alert.

common/src/main/java/com/redhat/cloud/notifications/openbridge/BridgeHelper.java
```diff
@@ -19,11 +21,13 @@
 @ApplicationScoped
 public class BridgeHelper {
 
+    public static final String CLOUD_PROVIDER = "aws";
+    public static final String CLOUD_REGION = "us-east-1";
```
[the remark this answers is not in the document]
pilhuhn: Right now (and the foreseeable future): yes

engine/src/main/java/com/redhat/cloud/notifications/events/FromOpenBridgeHistoryFiller.java
pilhuhn: This is like what is in the Camel code path. I am not aware we are using it, but may be wrong. Could be done later still

gwenneg: It's never been enabled on prod so far indeed.

engine/src/main/java/com/redhat/cloud/notifications/events/FromOpenBridgeHistoryFiller.java
```diff
@@ -0,0 +1,133 @@
...
+                }
+            });
+        }
+    }
+
+    private Map<String, Object> decodeItem(ProcessingError pe) {
+        Map<String, Object> map = new HashMap<>();
+        Map<String, String> headers = pe.getHeaders();
+        map.put("historyId", headers.get(ORIGINAL_EVENT_ID_HEADER));
```
gwenneg, line 112: This small change would make it obvious that we're talking about a Smart Events `event` and not a Notifications `event`:
```suggestion
        map.put("historyId", headers.get(ORIGINAL_SMART_EVENT_ID_HEADER));
```

pilhuhn: Well technically it is a header from SE, but the content is the original ID we submitted from the Notifications side

common/src/main/java/com/redhat/cloud/notifications/config/FeatureFlipper.java
```diff
@@ -63,6 +63,9 @@ public class FeatureFlipper {
     @ConfigProperty(name = "notifications.disable-webhook-endpoints-on-failure", defaultValue = "false")
     boolean disableWebhookEndpointsOnFailure;
 
+    @ConfigProperty(name = "ob.backchannel-filler.enabled", defaultValue = "true")
```
gwenneg, line 66: Is it safe to have this enabled by default?

josejulio: We can have them disabled by default

gwenneg:
Last commit didn't produce the expected result. I'm fixing the tests.

common/src/main/java/com/redhat/cloud/notifications/config/FeatureFlipper.java
pilhuhn: We should enable by default in stage to check that it is working as intended

gwenneg: We'll do that with the flag disabled on `master` and enabled for stage in app-interface.

engine/src/main/resources/application.properties
```diff
@@ -120,13 +120,16 @@ quarkus.log.category."com.redhat.cloud.notifications.health.KafkaConsumedTotalCh
 reinject.enabled=false
 
 %test.ob.enabled=true
-ob.bridge.uuid=<uuid-of-the-bridge instance>
+# If the next is enabled for tests, the bridge creation may Clash with CamelTypeProcessorTest
+%test.ob.backchannel-filler.enabled=false
```
gwenneg, line 124:
```suggestion
```
No longer needed because it's disabled by default now.

backend/src/main/java/com/redhat/cloud/notifications/OBTest.java
```diff
@@ -0,0 +1,246 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * Helper during development
+ */
+@Path("/internal/ob-test")
+public class OBTest {
```
gwenneg, line 34: Note for myself:
```suggestion
@IfBuildProfile("dev")
public class OBTest {
```

gwenneg: Class moved to #1534.

engine/src/main/java/com/redhat/cloud/notifications/events/FromOpenBridgeHistoryFiller.java
```diff
@@ -0,0 +1,137 @@
...
+
+    boolean alreadyProcessed(String historyId) {
+        AtomicBoolean seen = new AtomicBoolean(true);
+        seenIds.get(historyId, k -> {
+            // If the value loader function is invoked, then we've never seen the current historyId.
+            seen.set(false);
+            return DUMMY_CACHE_VALUE;
+        }).await().indefinitely();
+        return seen.get();
+    }
```
gwenneg, line 136: @pilhuhn FYI I introduced this to cache the seen ids. It relies on a `Cache` from quarkus-cache.

engine/src/main/resources/application.properties
```diff
@@ -120,13 +120,14 @@ quarkus.log.category."com.redhat.cloud.notifications.health.KafkaConsumedTotalCh
...
 %test.ob.enabled=true
-ob.bridge.uuid=<uuid-of-the-bridge instance>
+ob.bridge.name=<name-of-the-bridge instance>
 ob.token.client.id=<id of the service account>
 ob.token.client.secret=<secret of the service account>
 
 quarkus.rest-client.ob.url=https://bridge.acme.org
 quarkus.rest-client.kc.url=https://keycloak.acme.org
 quarkus.cache.caffeine.kc-cache.expire-after-write=PT120s
+quarkus.cache.caffeine.from-open-bridge-history-filler.expire-after-write=PT24H
```
gwenneg, line 130: @pilhuhn How long do you think we should cache the seen ids?

gwenneg (approved):
LGTM

engine/src/main/resources/application.properties
pilhuhn: I have no idea and we should perhaps watch the returned a list for a while for the ratio of new entries vs existing ones. If most are new, we can lower that to 1h.
Perhaps @danielezonca knows if they expire entries after a fixed amount of time and what the queue length setting on their side is.

pilhuhn: Keep in mind though, that if there is a sporadic error, it will stay for a long time until the queue is full and the error would be overwritten. If we expire it on our side, we would need to do a little more work, but it would not harm too much.

danielezonca: Currently there is no expiration so every time you invoke the error endpoint, you get back the last N errors (default should be `100`) without any expiration time (so they can be quite old)

gwenneg: Thanks @danielezonca!

It might be better to rely on a DB table to store the seen ids on our side then.

PR #1513, opened 2022-09-21
rename app, add state table

r14chandra:


engine/src/main/resources/templates/ResourceOptimization/dailyEmailBody.html
```diff
@@ -0,0 +1,46 @@
+{#include ResourceOptimization/insightsEmailBody}
+{#content-title}Daily Resource Optimization Service Summary - {action.context.start_time.toStringFormat()}{/content-title}
```
gwenneg, line 2: I suppose this is needed too.
```suggestion
{#content-title}Daily Resource Optimization Summary - {action.context.start_time.toStringFormat()}{/content-title}
```

gwenneg:
The [dailyEmailTitle.txt](https://github.com/RedHatInsights/notifications-backend/pull/1513/files#diff-71ea8496651891fbb52194dbdfb75b528bfb4aaa450d89dfae99b911f81db34a) file still contains Resource Optimization `Service`.

gwenneg:
Don't worry about the build failure. Your template will soon be migrated from the current files to our DB and I prepared something for that which got broken by the app renaming. I'll take care of fixing that.

engine/src/main/resources/templates/ResourceOptimization/dailyEmailBody.html
r14chandra: Thanks for pointing!

gwenneg (approved):
LGTM

PR #1519, opened 2022-09-26
NOTIF-496 Return 400 when validating invalid b/a/et triplet

gwenneg:
This is part 1.

See also [part 2](https://github.com/RedHatInsights/notifications-gw/pull/258).


josejulio (approved):
lgtm

PR #1524, opened 2022-09-27
NOTIF-370 Prevent from creating integrations with the same name

josejulio:
We will need migrations to update the name of repeated endpoints

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
@@ -1,5 +1,6 @@
 package com.redhat.cloud.notifications.db.repositories;
 
+import com.redhat.cloud.notifications.config.FeatureFlipper;
 import com.redhat.cloud.notifications.db.Query;
 import com.redhat.cloud.notifications.db.builder.QueryBuilder;
 import com.redhat.cloud.notifications.db.builder.WhereBuilder;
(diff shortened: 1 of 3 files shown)


backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -36,8 +38,41 @@ public class EndpointRepository {
     @Inject
     EntityManager entityManager;
 
+    @Inject
+    FeatureFlipper featureFlipper;
+
+    public void checkEndpointDisplayNameDuplicate(Endpoint endpoint) {
```
gwenneg, line 44:
```suggestion
    public void checkEndpointNameDuplicate(Endpoint endpoint) {
```

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -36,8 +38,41 @@ public class EndpointRepository {
...
+        TypedQuery<Long> query = entityManager.createQuery(hql, Long.class)
+                .setParameter("name", endpoint.getName())
+                .setParameter("orgId", endpoint.getOrgId());
+
+        if (endpoint.getId() != null) {
+            query.setParameter("endpointId", endpoint.getId());
+        }
+
+        if (query.getSingleResult() > 0) {
+            throw new BadRequestException("An endpoint with display name [" + endpoint.getName() + "] already exists");
```
gwenneg, line 69:
```suggestion
            throw new BadRequestException("An endpoint with name [" + endpoint.getName() + "] already exists");
```

gwenneg (approved):
LGTM

PR #1525, opened 2022-09-27
NOTIF-370 Ensure the behavior group name is per bundle

josejulio:
This is what we were talking earlier - updated so it does the check per bundle

Will follow up with a migration PR to move rename all the repeated behavior groups and integrations

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
@@ -69,7 +69,7 @@ public class BehaviorGroupRepository {
                     .setParameter("id", behaviorGroupId)
                     .getSingleResult();
         } catch (NoResultException nre) {
-            throw new NotFoundException("Behavior group not found", nre);
+            throw new NotFoundException("Behavior group not found");
         }
     }
 
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
@@ -63,8 +63,14 @@ public class BehaviorGroupRepository {
         return this.create(null, null, behaviorGroup, true);
     }
 
-    public BehaviorGroup get(UUID behaviorGroupId) {
-        return entityManager.find(BehaviorGroup.class, behaviorGroupId);
+    public UUID getBundleId(UUID behaviorGroupId) {
+        try {
+            return entityManager.createQuery("SELECT bundle.id FROM BehaviorGroup WHERE id = :id", UUID.class)
+                    .setParameter("id", behaviorGroupId)
+                    .getSingleResult();
+        } catch (NoResultException nre) {
+            throw new NotFoundException("Behavior group not found", nre);
+        }
     }
 
     @Transactional
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
@@ -168,9 +168,7 @@ public class BehaviorGroupRepository {
             q = q.setParameter("orgId", orgId);
         }
 
-        if (q.executeUpdate() == 0) {
-            throw new BadRequestException("Behavior group not updated - probably the behavior group was not part of the orgId");
-        }
+        q.executeUpdate();
     }
 
     private void checkBehaviorGroupDisplayNameDuplicate(String orgId, BehaviorGroup behaviorGroup, boolean isDefaultBehaviorGroup) {
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
@@ -15,7 +15,6 @@ import javax.persistence.TypedQuery;
 import javax.transaction.Transactional;
 import javax.validation.Valid;
 import javax.ws.rs.BadRequestException;
-import javax.ws.rs.InternalServerErrorException;
 import javax.ws.rs.NotFoundException;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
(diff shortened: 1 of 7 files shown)


backend/src/main/java/com/redhat/cloud/notifications/routers/NotificationResource.java
```diff
@@ -275,8 +275,11 @@ public Response updateBehaviorGroup(@Context SecurityContext sec,
         String orgId = getOrgId(sec);
 
         if (request.displayName != null) {
-            BehaviorGroup behaviorGroup = new BehaviorGroup();
-            behaviorGroup.setId(id);
+            BehaviorGroup behaviorGroup = behaviorGroupRepository.get(id);
```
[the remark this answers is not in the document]
josejulio: The purpose of this fetch is to get the `bundleId` from the database - that one is needed now and is not provided in the request.

I guess we could also update the check to have something like:
```
if (behaviorGroup.getId() != null) {
   query += " AND bundle.id != (SELECT bundle_id FROM BehaviorGroup WHERE id = :behaviorGroupId)
} else {
   query += " AND bundle.id != :bundleId"
}
``` 

WDYT?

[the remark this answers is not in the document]
josejulio: Is not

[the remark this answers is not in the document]
gwenneg: Sorry, I got confused between the BG and the bundle id 😅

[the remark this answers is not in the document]
gwenneg: The managed bean could be a problem indeed. Since the `displayName` field is changed, the bean will be considered dirty and therefore be updated at the end of the JPA transaction, after the other intended update.

You could replace
```
public BehaviorGroup get(UUID behaviorGroupId) {
    return entityManager.find(BehaviorGroup.class, behaviorGroupId);
}
```
with
```
public UUID getBundleId(UUID behaviorGroupId) {
    return entityManager.createQuery("SELECT bundle.id FROM BehaviorGroup WHERE id = :id", UUID.class)
    [...]
}
```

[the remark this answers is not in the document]
josejulio: Done!

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
```diff
@@ -63,6 +63,16 @@ public BehaviorGroup createDefault(BehaviorGroup behaviorGroup) {
         return this.create(null, null, behaviorGroup, true);
     }
 
+    public UUID getBundleId(UUID behaviorGroupId) {
+        try {
+            return entityManager.createQuery("SELECT bundle.id FROM BehaviorGroup WHERE id = :id", UUID.class)
+                    .setParameter("id", behaviorGroupId)
+                    .getSingleResult();
+        } catch (NoResultException nre) {
+            throw new NotFoundException("Behavior group not found", nre);
```
gwenneg, line 72:
```suggestion
            throw new NotFoundException("Behavior group not found");
```

backend/src/main/java/com/redhat/cloud/notifications/routers/NotificationResource.java
```diff
@@ -275,11 +275,21 @@ public Response updateBehaviorGroup(@Context SecurityContext sec,
...
             BehaviorGroup behaviorGroup = new BehaviorGroup();
             behaviorGroup.setId(id);
             behaviorGroup.setDisplayName(request.displayName);
+            behaviorGroup.setBundleId(bundleId);
+
+            boolean didUpdate;
+            try {
+                didUpdate = behaviorGroupRepository.update(orgId, behaviorGroup);
+            } catch (BadRequestException badRequestException) {
+                // The duplicate name in bundle id validation failed - keep the responses consistent by returning false.
```
gwenneg, line 288: Which responses?

We are returning a 400 status with explanation in case of creation failure because of a duplicate name. Shouldn't we do the same for the update?

josejulio: Before this PR, the update failure returns a 200 with a `false` in the body - The only usual way AFAIK to trigger this status is to send a behavior group id of other org id.

I agree that `false` does not really provide any information - what do you think about changing to throwing bad request in both cases?

josejulio:
Updated

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
```diff
@@ -156,7 +166,9 @@ boolean update(String orgId, BehaviorGroup behaviorGroup, boolean isDefaultBehav
             q = q.setParameter("orgId", orgId);
         }
 
-        return q.executeUpdate() > 0;
+        if (q.executeUpdate() == 0) {
+            throw new BadRequestException("Behavior group not updated - probably the behavior group was not part of the orgId");
+        }
```
gwenneg, line 171: Unless I missed something, the only situation where `q.executeUpdate() == 0` would be `true` is if the current request happened concurrently with another request that deletes the BG. Even if that happened, I don't think we should throw any kind of exception. We could simply do this here:
```suggestion
        q.executeUpdate();
```
WDYT?

josejulio: It would also yield `0` if trying to update a behavior group that is not part of your org id - the update to `getBundleId` might cover that though.

josejulio: anyhow - I think it's fine to fail with an exception if nothing is updated- an update was requested, if nothing was updated there is something wrong.

In this case, the only posibilty is if the orgId and the id don't identify a behavior group - but a previous check was just added. I guess we can omit this check.

josejulio:
There is a test failure - I'll address that later today

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
gwenneg: > It would also yield `0` if trying to update a behavior group that is not part of your org id - the update to `getBundleId` might cover that though.

That can't happen now that you fixed my other comment.

gwenneg (approved):
LGTM

PR #1528, opened 2022-09-29
NOTIF-370 Migrate duplicated names

josejulio:


gwenneg:
What if an org has two endpoints named "Foo" but also an endpoint named "Foo 2" before the migration? I suppose we'll need to run the migration several times until it returns that nothing was updated.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/DuplicateNameMigrationResource.java
```diff
@@ -0,0 +1,123 @@
...
+            migrationReport.updatedBehaviorGroups += updateBehaviorGroups(repeatedValue);
+        }
+
+        Log.infof("Finished updating duplicates names for behavior groups");
+
+        return migrationReport;
+    }
+
+    @Transactional
+    private int updateEndpoints(Object[] repeatedValue) {
```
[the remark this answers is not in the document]
josejulio: > `@Transactional` doesn't work on `private` methods.

and I guess that's why I added the annotation to the other method.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/DuplicateNameMigrationResource.java
```diff
@@ -0,0 +1,123 @@
...
+
+        Log.infof("Starting migration for duplicated names on behavior groups and integrations.");
+
+        List<Object[]> repeatedIntegrationValues = entityManager.createNativeQuery(
+                // string_agg concatenates all the grouped values with the delimiter (, in this case)
+                "SELECT string_agg(CAST(id as character varying), ','), name FROM endpoints WHERE endpoint_type != :endpoint_type_email GROUP BY name, org_id HAVING count(*) > 1"
+        )
+            .setParameter("endpoint_type_email", EndpointType.EMAIL_SUBSCRIPTION.ordinal())
+            .getResultList();
+        Log.infof("%d organizations needs to migrate their integration names", repeatedIntegrationValues.size());
```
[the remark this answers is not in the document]
josejulio:
```suggestion
        Log.infof("Found %d different integration names across the  organizations that needs updating - each name could be used by multiple integrations", repeatedIntegrationValues.size());
```

josejulio:
> What if an org has two endpoints named "Foo" but also an endpoint named "Foo 2" before the migration? I suppose we'll need to run the migration several times until it returns that nothing was updated.

That's correct - we should run this until there are no changes.
I'll add a test for that scenario to confirm the exceptations.

josejulio:
@gwenneg updated the code to run multiple times until there are no changes and added a related test

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/DuplicateNameMigrationResource.java
```diff
@@ -0,0 +1,147 @@
...
+    public DuplicateNameMigrationReport migrateDuplicateNames(@QueryParam("ack") String ack) {
+        if (!Objects.equals(this.ack, ack)) {
+            throw new BadRequestException(String.format("Invalid ack provided - should be [%s]", ack));
+        }
+
+        DuplicateNameMigrationReport migrationReport = new DuplicateNameMigrationReport();
+
+        Log.infof("Starting migration for duplicated names on behavior groups and integrations.");
+
+        while (true) {
```
gwenneg, line 53: Could you add a log entry inside that loop? Even if that's very unlikely, I'd like to know if we are stuck in an infinite or very long loop.

josejulio: There is this log: 

https://github.com/RedHatInsights/notifications-backend/pull/1528/files#diff-5cf2fbcd4bdeaedc0cc1387624379dbfdbeeb9b2c97b98dd5ca0ae3a5b5abe45R71

gwenneg: My bad, I missed that one. Thanks!

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/DuplicateNameMigrationResource.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    DuplicateNameMigrationReport runMigrationStep() {
+        DuplicateNameMigrationReport migrationReport = new DuplicateNameMigrationReport();
+
+        Log.infof("Running a migration step");
+
+        List<Object[]> repeatedIntegrationValues = entityManager.createNativeQuery(
+                        // string_agg concatenates all the grouped values with the delimiter (, in this case)
+                        "SELECT string_agg(CAST(id as character varying), ','), name FROM endpoints WHERE endpoint_type != :endpoint_type_email GROUP BY name, org_id HAVING count(*) > 1"
```
gwenneg, line 75: I wonder if the endpoints should be ordered by creation date.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/DuplicateNameMigrationResource.java
```diff
@@ -0,0 +1,147 @@
...
+
+        for (Object[] repeatedValue : repeatedIntegrationValues) {
+            migrationReport.updatedIntegrations += updateEndpoints(repeatedValue);
+        }
+
+        Log.infof("Finished step updating duplicates names for integrations");
+
+        List<Object[]> repeatedBehaviorGroupValues = entityManager.createNativeQuery(
+                // string_agg concatenates all the grouped values with the delimiter (, in this case)
+                "SELECT string_agg(CAST(id as character varying), ',') as ids, display_name FROM behavior_group GROUP BY display_name, org_id, bundle_id HAVING count(*) > 1"
```
gwenneg, line 92: I wonder if the behavior groups should be ordered by creation date.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/DuplicateNameMigrationResource.java
josejulio: That would be a nice improvement

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/DuplicateNameMigrationResource.java
josejulio: That would be a nice improvement

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/DuplicateNameMigrationResource.java
josejulio: Done

backend/src/test/java/com/redhat/cloud/notifications/routers/internal/DuplicateNameMigrationResourceTest.java
```diff
@@ -0,0 +1,353 @@
...
+                if (i == 0) {
+                    assertEquals(behaviorGroup.getDisplayName(), other.getDisplayName());
+                } else {
+                    assertEquals(String.format("%s %d", behaviorGroup.getDisplayName(), i + 1), other.getDisplayName());
+                }
+            }
+        } else {
+            for (BehaviorGroup behaviorGroup : behaviorGroupList) {
+                BehaviorGroup other = resourceHelpers.getBehaviorGroup(behaviorGroup.getId());
+                assertEquals(other.getDisplayName(), other.getDisplayName());
```
gwenneg, line 295:
```suggestion
                assertEquals(behaviorGroup.getDisplayName(), other.getDisplayName());
```

gwenneg (approved):
LGTM

PR #1532, opened 2022-09-30
NOTIF-429 followup - removing properties from CamelProperties

josejulio:
Was checking the code and found this. Forgot this was still around


gwenneg (approved):
LGTM

PR #1536, opened 2022-10-03
NOTIF-803 New Integrations are created with correct status

josejulio:
Integrations that are under our control are created with `READY` status directly.


backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
```diff
@@ -572,6 +575,7 @@ void addOpenBridgeEndpoint() {
         ep.setDescription("I guess the camel is slacking");
         ep.setEnabled(true);
         ep.setProperties(cAttr);
+        ep.setStatus(EndpointStatus.DELETING); // Trying to set other status
```
gwenneg, line 578: What did you want to test with that line?

josejulio: The creator of the Endpoint does not decide the endpoint status.

This is more an API problem, it's accepting a whole Endpoint, while it should accept something simpler.

gwenneg (approved):
LGTM

PR #1540, opened 2022-10-05
Log 4xx errors at warn level while calling BOP

gwenneg:
4xx errors shouldn't go unnoticed when we call BOP.

From now on, they will be logged at `warn` level (and generate a Sentry alert) instead of `debug`.


MikelAlejoBR (approved):
LGTM!

PR #1542, opened 2022-10-05
Fix Integrations template body and test it

gwenneg:
Fixes:
```
io.quarkus.qute.TemplateException: Rendering error in template [Integrations/integrationDisabledBody.html] line 1: included template [ConsoleNotifications/insightsEmailBody] not found
```
which was thrown on stage.

Unfortunately, our Sentry alerts are broken right now (fix in progress) so we didn't notice that exception.


MikelAlejoBR (approved):
LGTM! I went through the changes and fetched the branch locally. Everything's green!

PR #1544, opened 2022-10-06
NOTIF-809 Adding status for notifications

josejulio:
- Camel now uses the new status for processing, processing-failed and creation-failed
 - RHOSE goes directly to `SENT`

The old invocationResult is keep for backwards compatibility -  a follow up to remove it from the database could be pursue and we could still compute it's value from the notification status (i.e. true if it's success or sent)


common/src/main/java/com/redhat/cloud/notifications/models/NotificationStatus.java
```diff
@@ -0,0 +1,20 @@
+package com.redhat.cloud.notifications.models;
+
+public enum NotificationStatus {
+    // The notification failed to be created for processing
+    // This could happen when sending the notification to a different processor and it was rejected before entering a processing phase
+    FAILED_CREATION,
+    // The notification is being processed
+    // A notification was sent to a different processor and we are waiting for it's output
+    PROCESSING,
```
josejulio, line 9: Pending to add a `SEND` status.

josejulio: Sent status added ; notifications sent trought RHOSE are now marked as `SENT`

josejulio:
This PR is having the same PR check failures as https://github.com/RedHatInsights/notifications-backend/pull/1560

common/src/main/java/com/redhat/cloud/notifications/models/NotificationStatus.java
```diff
@@ -0,0 +1,22 @@
+package com.redhat.cloud.notifications.models;
+
+public enum NotificationStatus {
+    // The notification failed to be created for processing
+    // This could happen when sending the notification to a different processor and it was rejected before entering a processing phase
+    FAILED_CREATION,
```
gwenneg, line 6: I wonder if instead of `FAILED_CREATION`/`FAILED_PROCESSING` we should have:

- `FAILED_INTERNAL` for any exception that is thrown in Notifications, including when we fail to send an event to RHOSE. This would tell us that we can find an explanation in our logs/metrics.
- `FAILED_EXTERNAL` for errors returned by Camel or RHOSE.

common/src/main/java/com/redhat/cloud/notifications/models/NotificationStatus.java
```diff
@@ -0,0 +1,22 @@
...
+    // The notification is being processed
+    // A notification was sent to a different processor and we are waiting for it's output
+    PROCESSING,
```
gwenneg, line 9: `PROCESSING` and `SENT` are very close. Could we only use `SENT` even from Camel and then change is to `SUCCESS` for Camel while it would remain `SENT` for RHOSE (for now)?

common/src/main/java/com/redhat/cloud/notifications/models/NotificationStatus.java
```diff
@@ -0,0 +1,22 @@
...
+    // The notification has been in processing state for too long
+    PROCESSING_TIMEOUT,
```
gwenneg, line 11: I'm not sure we need that for now. Later maybe (in another PR)?

gwenneg (commented):
I'm not done reviewing but here's a few comments.

common/src/main/java/com/redhat/cloud/notifications/models/NotificationStatus.java
josejulio: It sounds good

common/src/main/java/com/redhat/cloud/notifications/models/NotificationStatus.java
josejulio: We could add the state later when we are ready to implement it.

common/src/main/java/com/redhat/cloud/notifications/models/NotificationStatus.java
gwenneg: Okay, sounds good to me.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/NotificationRepository.java
```diff
@@ -23,7 +23,7 @@ public class NotificationRepository {
     EntityManager entityManager;
 
     public List<NotificationHistory> getNotificationHistory(String orgId, UUID endpoint, boolean includeDetails, Query limiter) {
-        String query = "SELECT NEW NotificationHistory(nh.id, nh.invocationTime, nh.invocationResult, nh.endpoint, nh.created";
+        String query = "SELECT NEW NotificationHistory(nh.id, nh.invocationTime, nh.status, nh.endpoint, nh.created";
```
gwenneg, line 26: Is the `invocationResult` no longer needed in the response of the REST endpoint that invokes this method?

backend/src/main/java/com/redhat/cloud/notifications/routers/models/EventLogEntry.java
```diff
@@ -32,7 +34,7 @@ public class EventLogEntry {
 
     private String payload;
 
-    @NotNull
+    @Enumerated(EnumType.STRING)
```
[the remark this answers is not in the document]
josejulio: I saw this annotation in here https://github.com/RedHatInsights/notifications-backend/blob/01f262eaa6a1fd40aed488aab6e1647e8873409f/common/src/main/java/com/redhat/cloud/notifications/models/Endpoint.java#L69

I immediatly though it was for storing the enum as a string as opposed to a number, but didn't really go into the details of it.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/NotificationRepository.java
josejulio: I was planning to remove the `invocationResult` column from the table and compute it from the `status` value - and then decided to leave it for a followup PR - I'll check if this was a remaining of that and revert the change.

backend/src/main/java/com/redhat/cloud/notifications/routers/models/EventLogEntry.java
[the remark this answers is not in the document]
gwenneg: Okay but that one is placed on `private List<EventLogEntryAction> actions;` which is not an enum 😄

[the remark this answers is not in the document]
josejulio: Ah - didn't see that :grin: Will check and fix

josejulio:
Remaining failing tests are unrelated (similar to  #1560)

common/src/main/java/com/redhat/cloud/notifications/models/NotificationHistory.java
```diff
@@ -187,9 +200,13 @@ public static NotificationHistory getHistoryStub(Endpoint endpoint, Event event,
         history.setEndpointType(endpoint.getType());
         history.setEndpointSubType(endpoint.getSubType());
         history.setEvent(event);
-        history.setInvocationResult(false);
+        history.setStatus(NotificationStatus.FAILED_INTERNAL);
```
gwenneg, line 203: I think we could/should get rid of that line.

engine/src/test/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessorTest.java
```diff
@@ -109,6 +111,7 @@ void afterEach() {
     }
 
     @Test
+    @Transactional
```
gwenneg, line 114: Is that really necessary?

common/src/main/java/com/redhat/cloud/notifications/models/NotificationHistory.java
josejulio: I'm wondering if we should have a default to prevent NPE - or maybe just validate that we have some status before saving it.

engine/src/test/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessorTest.java
josejulio: At some point I was actually running SQL in there, but not anymore - removing

common/src/main/java/com/redhat/cloud/notifications/models/NotificationHistory.java
gwenneg: Then let's make the `status` field `@NotNull` maybe.

engine/src/test/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessorTest.java
```diff
@@ -38,6 +39,7 @@
 
 import javax.enterprise.inject.Any;
 import javax.inject.Inject;
+import javax.transaction.Transactional;
```
gwenneg, line 42: This may cause a build failure.

Otherwise, LGTM!

gwenneg (approved):

PR #1547, opened 2022-10-07
Fix email templates

FabriciaDinizRH:
If no hosts were created, the errors don't appear in the service.
The email templates were updated accordingly.


gwenneg (approved):
LGTM

PR #1548, opened 2022-10-07
[NOTIF-753] feature: remove the scopes from the default security scheme

MikelAlejoBR:
When the "RolesAllowed" annotation is used in a method, smallrye generates a basic default security scheme for the operation that represents that method. The problem is that it also includes the scopes defined in the annotation, which OpenApi doesn't accept unless an "oAuth2" or "openIdConnect" authentication method is defined.

## Links

[[NOTIF-753]](https://issues.redhat.com/browse/NOTIF-753)


josejulio:
Could you add a test? 
https://github.com/RedHatInsights/notifications-backend/blob/3e1a9b69241c5cf8bd700bc33ec246d3eae4adb5/backend/src/test/java/com/redhat/cloud/notifications/openapi/OpenApiTest.java#L51

Thanks!

backend/src/test/java/com/redhat/cloud/notifications/openapi/OpenApiTest.java
```diff
@@ -63,6 +79,36 @@ public void validateOpenApi() throws Exception {
         }
     }
 
+    /**
+     * Tests that the default security schemes have all their scope definitions empty. Check the "@see" annotation blow
```
josejulio, line 83:
```suggestion
     * Tests that the default security schemes have all their scope definitions empty. Check the "@see" annotation below
```

josejulio (approved):
lgtm  

found  a small typo

backend/src/test/java/com/redhat/cloud/notifications/openapi/OpenApiTest.java
MikelAlejoBR: Great catch! :fishing_pole_and_fish:

PR #1550, opened 2022-10-07
Inventory - changes from request ID to display name

FabriciaDinizRH:
Christian Marineau requested that the request ID in the emails is removed and display name added in its place.


gwenneg (approved):
LGTM

PR #1551, opened 2022-10-07
[DRFT-906] Handle personification in email templates

jpramos123:
Handle personification in drift email templates


josejulio (approved):
lgtm

PR #1553, opened 2022-10-10
Fixes Inventory EmailAggregator

FabriciaDinizRH:
This change fixes the problem of the aggregator always having only one item and updates related tests.

### DIFF
--- a/engine/src/main/resources/templates/Inventory/dailyEmailBody.html
+++ b/engine/src/main/resources/templates/Inventory/dailyEmailBody.html
@@ -35,14 +35,12 @@
                     </tr>
                 </thead>
                 <tbody>
-                    {#for error in action.context.inventory.errors}
-                    {#if count < 11}
+                    {#each action.context.inventory.errors.take(10)}
                     <tr>
-                        <td>{#if error.display_name}{error.display_name}{#else}Not available{/if}</td>
-                        <td>{error.message}</td>
+                        <td>{#if it.display_name}{it.display_name}{#else}Not available{/if}</td>
+                        <td>{it.message}</td>
                     </tr>
-                    {/if}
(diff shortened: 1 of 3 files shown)


engine/src/main/resources/templates/Inventory/dailyEmailBody.html
```diff
@@ -35,11 +35,11 @@
                     </tr>
                 </thead>
                 <tbody>
-                    {#for key in action.context.inventory.errors.keySet()}
-                    {#if key_count < 11}
+                    {#for error in action.context.inventory.errors}
+                    {#if count < 11}
```
gwenneg, line 39: I'm not sure `count` exists here.

What do you want to count and what is the expected output depending on the result of that count?

FabriciaDinizRH: I want to only show the first 10 hosts, so that the email doesn't get too big.

FabriciaDinizRH: I read it here: https://github.com/quarkusio/qute/blob/master/guide.adoc

josejulio: It looks like you can also use `take` see [docs](https://es.quarkus.io/guides/qute-reference#arrays)

`{#for error in action.context.inventory.errors.take(10)}` and remove the `count`  which also [looks](https://es.quarkus.io/guides/qute-reference#loop_section) valid.

FabriciaDinizRH: Yeah, that seems like a better synthax

josejulio (approved):

PR #1554, opened 2022-10-11
NOTIF-804 Load RHOSE client_id and secret from actual secrets

gwenneg:
This PR depends on two app-interface MRs that I will not link here.

I'll undraft it when app-interface is ready.


gwenneg:
The failure on ephemeral is expected, the secret isn't available yet because of the pending app-interface MRs.

gwenneg:
The secrets are now available on ephemeral and stage. This MR can be merged.

MikelAlejoBR (approved):
LGTM! :+1:

PR #1555, opened 2022-10-11
render template when creating template

aferd:


aferd:
Screen Shot 2022-10-11 at 12 10 19 PM

josejulio:
> trying to find the problem here, post request failing with 400.

`Action parsing failed...` is because you are sending the html text instead of an `action` i.e. should be sending something like:
```
{
  "bundle": "rhel",
  "application": "policies",
  "event_type": "policy-triggered",
  "timestamp": "2021-08-05T16:21:14.243",
  "org_id": "5758117",
  "context": "{\"inventory_id\":\"80f7e57d-a16a-4189-82af-1d68a747c8b3\",\"system_check_in\":\"2021-08-05T16:21:12.953036\",\"display_name\":\"cool display name\"}",
  "events": [
    {
      "metadata": {},
      "payload": "{\"my_id\":\"3df53241-3e09-481b-a322-4892caaaaadc\",\"my_name\":\"Red color\"}"
    },
    {
      "metadata": {},
      "payload": "{\"my_id\":\"6c5e8451-a40a-4bb7-ab9a-0cb10a4c577d\",\"my_name\":\"Green color\"}"
    },
    {
      "metadata": {},
      "payload": "{\"my_id\":\"b4c6378a-c1fb-4d3e-8e9b-7e5bdfc09dd3\",\"my_name\":\"Blue color\"}"
    }
  ]
}
```

aferd:
https://user-images.githubusercontent.com/86322239/197225089-bdd20107-034b-4362-8f70-c556c8502029.mov

josejulio:
I see some strange things after watching the video: 
- We are creating a template, but it's specifying both the subject and the body. At that level, templates should only have 1 editable content, the concept of title/body appears when assigning them to an event type
- The user clicks on `create template` but the page says "Update template"
- The URL looks wrong, it is something like: `email-templates/[object%20Object]`

admin-console/src/main/webapp/src/components/EmailTemplates/EmailTemplateForm.tsx
```diff
@@ -18,53 +19,243 @@ const SkeletonIfLoading: React.FunctionComponent<SkeletonIfLoading> = props => (
...
+                            isLineNumbersVisible
+                            code={ defaultSubjectTemplate }
+                            onChange={ setSubjectTemplate }
+                            isMinimapVisible={ false }
+                            height="50px"
+                        />
+                    </SkeletonIfLoading>
+                </FormGroup>
+                <FormGroup
+                    label="Body"
```
josejulio, line 229: We don't need `subject` and `body` - only 1,- it can be called `template` or `content`

aferd:
https://user-images.githubusercontent.com/86322239/199796977-34ba5b8b-cbf1-4aa2-a2f6-f6e5da1d21c3.mov

the content is rendering showing twice, which I think it coming from the render func? as its sending the content twice in the payload but i am not 100%

aferd:
Question - Should the render only show when creating a new template and NOT when updating a template?

josejulio:
> Question - Should the render only show when creating a new template and NOT when updating a template? Since the payload is not saved, thinking it wouldn't make much sense to have the render show on the UPDATE screen.

I think it would make sense - to be able to see changes as we are updating - we might workaround that payload in the future by providing some real payloads from the server.

aferd:
I have this rendering when you are creating a new template but NOT when you are updating a template. From a user standpoint, when you go to update your template you would expect your payload and I feel it would be confusing to have a defaultpayload rendering when you are updating something specific to your user. I know eventually we mentioned storing payloads but until then.

let me know. 

https://user-images.githubusercontent.com/86322239/201181711-86ec274f-ac72-46fb-86b7-ef736290b8cb.mov

josejulio:
> From a user standpoint, when you go to update your template you would expect your payload and I feel it would be confusing to have a defaultpayload rendering when you are updating something specific to your user.

I think it's good as a first step - anyhow I feel like we should provide the default payload even if it fails initially, but at least lets the users to copy/paste some data to update the template

PR #1558, opened 2022-10-12
Fix daily email for inventory

josejulio:


engine/src/main/resources/templates/Inventory/dailyEmailBody.html
```diff
@@ -28,14 +28,14 @@
...
-                <thead>              
+                <thead>
                     <tr>
                         <th>Host Name</th>
                         <th>Error</th>
                     </tr>
                 </thead>
                 <tbody>
-                    {#each action.context.inventory.errors.take(10)}
+                    {#each action.context.inventory.errors.take(action.context.inventory.errors.size() > 10 ? 10 : action.context.inventory.errors.size())}
```
josejulio, line 38: Though we could use stuff like `java.lang.Math.min` or `Math.min` but looks like it does not come out of the box.

gwenneg (approved):
LGTM

PR #1564, opened 2022-10-17
NOTIF-832 Retrieve and persist dead-letter-reason header from RHOSE

gwenneg:
RHOSE relies on the [error handling strategies](https://quarkus.io/guides/kafka#error-handling) provided by Quarkus when a failure happens during the processing of one of our events.

Until now, we retrieved the `dead-letter-cause` field returned by RHOSE in case of error and it was persisted in the DB. Unfortunately, that field was empty most of the time and we were unable to understand why the error happened and how to fix it.

After adding some debug log entries on stage, I found out that RHOSE can also return the cause of the error with the `dead-letter-reason` field. This PR retrieves that field and persists it alongside with `dead-letter-cause`.


MikelAlejoBR (approved):
LGTM! :+1:

gwenneg:

PR #1565, opened 2022-10-18
NOTIF-835 Use the endpoint id as the RHOSE processor name

gwenneg:


gwenneg:

common/src/main/java/com/redhat/cloud/notifications/models/CreationTimestamped.java
```diff
@@ -34,5 +34,13 @@ public void prePersist() {
         if (created == null) {
             created = LocalDateTime.now(UTC);
         }
+        additionalPrePersist();
+    }
```
josejulio, line 38: I guess we can't override this method because the annotation doesn't work well?

I wonder if we should mark as final the `prePersist` method to prevent someone from overriding it.

MikelAlejoBR (approved):
LGTM!

common/src/main/java/com/redhat/cloud/notifications/models/CreationTimestamped.java
gwenneg: I tried to add an additional method annotated with `@PrePersist` (one in `CreationTimestamped`, the other in `Endpoint`) but it didn't work well. I didn't try to override `CreationTimestamped#prePersist` but that would be good as any change in the overridden method would be missed in the override.

Making `CreationTimestamped#prePersist` final is definitely a good idea and I'll add that to the PR right away, thanks!

gwenneg:
The PR check failures are still unrelated.

PR #1567, opened 2022-10-19
[RHCLOUD-22178] Include environment URLs when serializing Actions

MikelAlejoBR:
The PR:

- Removes the `transform` function in favor of `toJsonObject`, since we could use just a single function instead of having scattered calls to the two functions.
- Adds a test to make sure that, at least, the structure we've defined will remain consistent in the future.
- Adds the environment URL every time the action gets serialized, so that it can be associated with the environment it must be checked on.

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
@@ -272,7 +272,7 @@ public class CamelTypeProcessor extends EndpointTypeProcessor {
                 .build(BridgeEventService.class);
 
         // Include the environment's URL when sending the event.
-        ce.put("environment_url", this.environment.url());
+        body.put("environment_url", this.environment.url());
 
         JsonObject payload = JsonObject.mapFrom(ce);
         try {
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
@@ -15,6 +15,7 @@ import com.redhat.cloud.notifications.openbridge.BridgeAuth;
 import com.redhat.cloud.notifications.openbridge.BridgeEventService;
 import com.redhat.cloud.notifications.openbridge.RhoseErrorMetricsRecorder;
 import com.redhat.cloud.notifications.processors.EndpointTypeProcessor;
+import com.redhat.cloud.notifications.templates.models.Environment;
 import com.redhat.cloud.notifications.transformers.BaseTransformer;
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.MeterRegistry;
@@ -85,6 +86,12 @@ public class CamelTypeProcessor extends EndpointTypeProcessor {
 
     @Inject
     RhoseErrorMetricsRecorder rhoseErrorMetricsRecorder;
+
+    /**
+     * Used to send the environment's URL on OpenBridge events.
+     */
+    @Inject
+    Environment environment;
 
     private Bridge bridge;
 
@@ -256,6 +263,9 @@ public class CamelTypeProcessor extends EndpointTypeProcessor {
         BridgeEventService evtSvc = RestClientBuilder.newBuilder()
                 .baseUri(URI.create(bridge.getEndpoint()))
                 .build(BridgeEventService.class);
+
+        // Include the environment's URL when sending the event.
+        ce.put("environment_url", this.environment.url());
 
         JsonObject payload = JsonObject.mapFrom(ce);
         try {
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
@@ -63,6 +63,13 @@ public class CamelTypeProcessor extends EndpointTypeProcessor {
     public static final String CLOUD_EVENT_TYPE_PREFIX = "com.redhat.console.notification.toCamel.";
     public static final String CAMEL_SUBTYPE_HEADER = "CAMEL_SUBTYPE";
     public static final String PROCESSORNAME = "processorname";
+    /**
+     * Constants for the {@link CamelTypeProcessor#callOpenBridge(JsonObject, UUID, String, CamelProperties, String, String)}.
+     * function.
+     */
+    public static final String SOURCE = "notifications";
+    public static final String SPEC_VERSION = "1.0";
+    public static final String TYPE = "myType";
 
     @Inject
     FeatureFlipper featureFlipper;
@@ -246,9 +253,9 @@ public class CamelTypeProcessor extends EndpointTypeProcessor {
         Map<String, Object> ce = new HashMap<>();
 
         ce.put("id", id);
-        ce.put("source", "notifications"); // Source of original event?
-        ce.put("specversion", "1.0");
-        ce.put("type", "myType"); // Type of original event?
+        ce.put("source", SOURCE); // Source of original event?
+        ce.put("specversion", SPEC_VERSION);
+        ce.put("type", TYPE); // Type of original event?
         ce.put(ORG_ID_FILTER_NAME, orgId);
         ce.put(PROCESSORNAME, extras.get(PROCESSORNAME));
         ce.put("originaleventid", originalEventId);
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
@@ -155,7 +155,7 @@ public class CamelTypeProcessor extends EndpointTypeProcessor {
         }
         metaData.put("_originalId", originalEventId);
 
-        JsonObject payload = transformer.transform(event.getAction());
+        JsonObject payload = transformer.toJsonObject(event.getAction());
         UUID historyId = UUID.randomUUID();
 
         JsonObject metadataAsJson = new JsonObject();
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
@@ -88,7 +88,7 @@ public class CamelTypeProcessor extends EndpointTypeProcessor {
     RhoseErrorMetricsRecorder rhoseErrorMetricsRecorder;
 
     /**
-     * Used to send the environment's URL on OpenBridge events.
+     * Used to send the environment's URL on RHOSE events.
      */
     @Inject
     Environment environment;
--- a/engine/src/main/java/com/redhat/cloud/notifications/transformers/BaseTransformer.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/transformers/BaseTransformer.java
@@ -11,6 +11,18 @@ import java.util.stream.Collectors;
 @ApplicationScoped
 public class BaseTransformer {
 
+    // JSON property names' definition.
+    public static final String ACCOUNT_ID = "account_id";
+    public static final String APPLICATION = "application";
+    public static final String BUNDLE = "bundle";
+    public static final String CONTEXT = "context";
+    public static final String EVENT_TYPE = "event_type";
+    public static final String EVENTS = "events";
+    public static final String METADATA = "metadata";
+    public static final String ORG_ID = "org_id";
+    public static final String PAYLOAD = "payload";
+    public static final String TIMESTAMP = "timestamp";
+
     /**
      * Transforms the given action into a {@link JsonObject}.
      * @param action the {@link Action} to transform.
@@ -19,24 +31,24 @@ public class BaseTransformer {
     public JsonObject toJsonObject(final Action action) {
         JsonObject message = new JsonObject();
 
-        message.put("account_id", action.getAccountId());
-        message.put("application", action.getApplication());
-        message.put("bundle", action.getBundle());
-        message.put("context", JsonObject.mapFrom(action.getContext()));
-        message.put("event_type", action.getEventType());
+        message.put(ACCOUNT_ID, action.getAccountId());
+        message.put(APPLICATION, action.getApplication());
+        message.put(BUNDLE, action.getBundle());
+        message.put(CONTEXT, JsonObject.mapFrom(action.getContext()));
+        message.put(EVENT_TYPE, action.getEventType());
         message.put(
-            "events",
+            EVENTS,
             new JsonArray(
                 action.getEvents().stream().map(
                     event -> Map.of(
-                        "metadata", JsonObject.mapFrom(event.getMetadata()),
-                        "payload", JsonObject.mapFrom(event.getPayload())
+                        METADATA, JsonObject.mapFrom(event.getMetadata()),
+                        PAYLOAD, JsonObject.mapFrom(event.getPayload())
                     )
                 ).collect(Collectors.toList())
             )
         );
-        message.put("org_id", action.getOrgId());
-        message.put("timestamp", action.getTimestamp().toString());
+        message.put(ORG_ID, action.getOrgId());
+        message.put(TIMESTAMP, action.getTimestamp().toString());
 
         return message;
     }
--- a/engine/src/main/java/com/redhat/cloud/notifications/transformers/BaseTransformer.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/transformers/BaseTransformer.java
@@ -11,23 +11,32 @@ import java.util.stream.Collectors;
 @ApplicationScoped
 public class BaseTransformer {
 
-    public JsonObject transform(Action action) {
-        return toJsonObject(action);
-    }
-
-    protected JsonObject toJsonObject(Action action) {
+    /**
+     * Transforms the given action into a {@link JsonObject}.
+     * @param action the {@link Action} to transform.
+     * @return a {@link JsonObject} containing the given action.
+     */
+    public JsonObject toJsonObject(final Action action) {
         JsonObject message = new JsonObject();
-        message.put("bundle", action.getBundle());
-        message.put("application", action.getApplication());
-        message.put("event_type", action.getEventType());
+
         message.put("account_id", action.getAccountId());
+        message.put("application", action.getApplication());
+        message.put("bundle", action.getBundle());
+        message.put("context", JsonObject.mapFrom(action.getContext()));
+        message.put("event_type", action.getEventType());
+        message.put(
+            "events",
+            new JsonArray(
+                action.getEvents().stream().map(
+                    event -> Map.of(
+                        "metadata", JsonObject.mapFrom(event.getMetadata()),
+                        "payload", JsonObject.mapFrom(event.getPayload())
+                    )
+                ).collect(Collectors.toList())
+            )
+        );
         message.put("org_id", action.getOrgId());
         message.put("timestamp", action.getTimestamp().toString());
-        message.put("events", new JsonArray(action.getEvents().stream().map(event -> Map.of(
-                "metadata", JsonObject.mapFrom(event.getMetadata()),
-                "payload", JsonObject.mapFrom(event.getPayload())
-        )).collect(Collectors.toList())));
-        message.put("context", JsonObject.mapFrom(action.getContext()));
 
         return message;
     }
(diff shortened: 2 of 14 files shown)


engine/src/main/java/com/redhat/cloud/notifications/transformers/BaseTransformer.java
```diff
@@ -1,33 +1,52 @@
...
+
+        message.put("account_id", action.getAccountId());
         message.put("application", action.getApplication());
+        message.put("bundle", action.getBundle());
+        message.put("context", JsonObject.mapFrom(action.getContext()));
+        // This class is often manually instantiated within some "TestHelpers" classes. In order to avoid hitting null
+        // pointer exceptions, this null check is required.
+        if (this.environment != null) {
+            message.put("environment_url", this.environment.url());
+        }
```
gwenneg, line 35: Sorry, I did not anticipate that this would also be sent as part of email and webhook notifications which is not something we want. 99% of the changes in this PR still make sense, but now I'm not sure `BaseTransformer` is the best place to add `environment_url`.

`Environment` could be injected in `CamelTypeProcessor` and then this line
https://github.com/RedHatInsights/notifications-backend/blob/9f0526ddc6248dd334d0befc8e2cd99df4557c01/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java#L158
would become
```
JsonObject payload = transformer.transform(event.getAction()); 
payload.put("environment_url", this.environment.url());
```
Or the signature of this method could have a boolean arg to specify if the environment URL should be included or not in the payload. And maybe an overload signature without that arg that would set it to `false` to avoid passing the arg at all calling points.
```
public JsonObject toJsonObject(final Action action) {
    return toJsonObject(action, false);
}

public JsonObject toJsonObject(final Action action, boolean includeEnvironmentUrl) {
[...]
        if (includeEnvironmentUrl && this.environment != null) {
            message.put("environment_url", this.environment.url());
        }
[...]
}
```

gwenneg: Actually, even with what I just wrote, the URL will be sent to Camel which is again not something needed. It is only required for Slack notifications. Let me think a bit more about this... 😄

gwenneg: https://github.com/RedHatInsights/notifications-backend/blob/9f0526ddc6248dd334d0befc8e2cd99df4557c01/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java#L230

The `body` arg in that method is the payload that will be sent to RHOSE and transformed into a Slack notification. I think that arg is the one where `environment_url` should be added.

gwenneg: Our Camel and RHOSE integrations are mixed in a single processor (CamelTypeProcessor) and that creates a lot of confusion... This is something I've wanted to fix for a while but I never got the time to take care of that.

josejulio: Splitting Camel & Splunk code might be something we should do in the short term - but i wonder if we should send across all messages the information about the environment in which the notification was generated (i.e. `environment: { type: prod, url: https://console.... }`

gwenneg: FYI I created [RHCLOUD-22545](https://issues.redhat.com/browse/RHCLOUD-22545) for the split.

MikelAlejoBR: All right, I've addressed the comments, and added a test for when we send events to Open Bridge. Now if we change anything from what we send to them, we should get a test failure for it.

engine/src/test/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessorTest.java
```diff
@@ -76,6 +87,48 @@ class CamelTypeProcessorTest {
     public static final String SUB_TYPE_KEY = "subType";
     public static final String SUB_TYPE = "sub-type";
 
+    /**
+     * The following constants are hard coded in {@link CamelTypeProcessor#callOpenBridge(JsonObject, UUID, String, CamelProperties, String, String)}.
+     */
+    private static final String FIXTURE_TYPE = "myType";
+    private static final String FIXTURE_SOURCE = "notifications";
+    private static final String FIXTURE_SPEC_VERSION = "1.0";
```
gwenneg, line 95: Let's make them `public static final` fields in `CamelTypeProcessor` and use them from here.

engine/src/test/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessorTest.java
```diff
@@ -76,6 +87,48 @@ class CamelTypeProcessorTest {
...
+
+    /**
+     * Endpoint fixtures for the {@link #buildEvent()} function.
```
gwenneg, line 98: `Endpoint` or `Event`? :smiley:

engine/src/test/java/com/redhat/cloud/notifications/transformers/BaseTransformerTest.java
```diff
@@ -1,36 +1,153 @@
...
+        action.setEventType(FIXTURE_EVENT_TYPE);
+        action.setEvents(FIXTURE_EVENTS);
+        action.setOrgId(FIXTURE_ORG_ID);
+        action.setTimestamp(FIXTURE_TIMESTAMP);
+
+        // Call the function under test.
+        JsonObject result = this.baseTransformer.toJsonObject(action);
+
+        // Assert that all the properties from the Action are present in the JSON object.
+        assertTrue(result.containsKey(ACCOUNT_ID), "the account id is missing from the resulting JSON object");
```
gwenneg, line 121: Is it really necessary to check if a field is missing while there's another assertion below that checks the field value? If the field is missing, then the value check will fail and it will be obvious from JUnit's output that the failure was caused by a missing field.

This is a general question that applies to several "flat" fields in this test but also to several fields in `CamelTypeProcessorTest`. By "flat" fields, I mean field types like `String`, `Boolean`... I do not include fields that contain a JSON structure (`events`, `metadata` or `payload`) in that question.

engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
```diff
@@ -86,6 +87,12 @@ public class CamelTypeProcessor extends EndpointTypeProcessor {
     @Inject
     RhoseErrorMetricsRecorder rhoseErrorMetricsRecorder;
 
+    /**
+     * Used to send the environment's URL on OpenBridge events.
```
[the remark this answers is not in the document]
gwenneg: If you rename `OpenBridge` to `RHOSE`, please do that only for the lines you added in this PR. We _may_ (it's completely minor) take care of the renaming in the entire app later with a separate PR.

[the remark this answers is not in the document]
MikelAlejoBR: All right, done!

engine/src/test/java/com/redhat/cloud/notifications/transformers/BaseTransformerTest.java
```diff
@@ -1,36 +1,153 @@
...
     @Test
     void shouldNotRaiseAnExceptionWhenOrgIdIsMissing() {
         Action action = new Action();
         action.setTimestamp(LocalDateTime.of(2022, 12, 24, 12, 0));
 
-        final JsonObject jsonObject = testee.toJsonObject(action);
+        final JsonObject jsonObject = baseTransformer.toJsonObject(action);
+
+        assertNull(jsonObject.getString(ORG_ID));
+    }
```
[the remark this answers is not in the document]
MikelAlejoBR: Removed!

engine/src/test/java/com/redhat/cloud/notifications/transformers/BaseTransformerTest.java
```diff
@@ -1,36 +1,153 @@
...
+    private static final String ACCOUNT_ID = "account_id";
+    private static final String APPLICATION = "application";
+    private static final String BUNDLE = "bundle";
+    private static final String CONTEXT = "context";
+    private static final String EVENT_TYPE = "event_type";
+    private static final String EVENTS = "events";
+    private static final String METADATA = "metadata";
+    private static final String ORG_ID = "org_id";
+    private static final String PAYLOAD = "payload";
+    private static final String TIMESTAMP = "timestamp";
```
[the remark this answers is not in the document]
MikelAlejoBR: Good idea! Done!

engine/src/test/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessorTest.java
MikelAlejoBR: This is done too!

engine/src/test/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessorTest.java
MikelAlejoBR: Good catch!

engine/src/test/java/com/redhat/cloud/notifications/transformers/BaseTransformerTest.java
MikelAlejoBR: Yeah, I also thought that these tests might be a bit redundant in that sense... I wasn't sure, so I went with the full, verbose way of testing the JSON results. I am going to remove the checks for the primitive types, hang on...

engine/src/main/java/com/redhat/cloud/notifications/transformers/BaseTransformer.java
```diff
@@ -11,23 +11,44 @@
...
+    /**
+     * Transforms the given action into a {@link JsonObject}.
+     * @param action the {@link Action} to transform.
+     * @return a {@link JsonObject} containing the given action.
+     */
+    public JsonObject toJsonObject(final Action action) {
         JsonObject message = new JsonObject();
-        message.put("bundle", action.getBundle());
+
+        message.put("account_id", action.getAccountId());
```
gwenneg, line 34:
```suggestion
        message.put(ACCOUNT_ID, action.getAccountId());
```
Same for the other fields :smiley:

MikelAlejoBR: Oh, you're right... Changed!

engine/src/test/java/com/redhat/cloud/notifications/transformers/BaseTransformerTest.java
MikelAlejoBR: I've removed quite a few "primitive key checks" from the tests, and left only the "test the value for this key" tests. The only keys I'm looking for their presence for are the ones that represent another JSON object.

engine/src/test/java/com/redhat/cloud/notifications/transformers/BaseTransformerTest.java
```diff
@@ -1,36 +1,114 @@
...
+        // Assert the values.
+        assertEquals(FIXTURE_ACCOUNT_ID, result.getString(BaseTransformer.ACCOUNT_ID), "the account id isn't the same");
+        assertEquals(FIXTURE_APPLICATION, result.getString(BaseTransformer.APPLICATION), "the application isn't the same");
+        assertEquals(FIXTURE_BUNDLE, result.getString(BaseTransformer.BUNDLE), "the bundle isn't the same");
+        assertEquals(FIXTURE_EVENT_TYPE, result.getString(BaseTransformer.EVENT_TYPE), "the event type isn't the same");
+
+        // Assert that the keys that contain inner JSON arrays or objects are present.
+        assertTrue(result.containsKey(BaseTransformer.CONTEXT), "the context is missing from the resulting JSON object");
+        assertTrue(result.containsKey(BaseTransformer.EVENTS), "the events is missing from the resulting JSON object");
+        assertTrue(result.containsKey(BaseTransformer.ORG_ID), "the org id is missing from the resulting JSON object");
```
gwenneg, line 95: This could be removed since there's a value check below.
```suggestion
```

engine/src/test/java/com/redhat/cloud/notifications/transformers/BaseTransformerTest.java
```diff
@@ -1,36 +1,114 @@
...
+
+        final JsonArray events = result.getJsonArray(BaseTransformer.EVENTS);
+        if (events.size() != FIXTURE_EVENTS.size()) {
+            fail("the number of events in the JSON file is different from the test data");
+        }
+
+        final JsonObject event = events.getJsonObject(0);
+        final JsonObject metadata = event.getJsonObject(BaseTransformer.METADATA);
+        assertTrue(metadata.containsKey(FIXTURE_METADATA_ADDITIONAL_PROPERTY), "the metadata object doesn't contain the additional property's key");
```
gwenneg, line 104: This could be removed since there's a value check below.
```suggestion
```

engine/src/test/java/com/redhat/cloud/notifications/transformers/BaseTransformerTest.java
```diff
@@ -1,36 +1,114 @@
...
+        assertEquals(FIXTURE_METADATA_ADDITIONAL_PROPERTY_VALUE, metadata.getString(FIXTURE_METADATA_ADDITIONAL_PROPERTY), "the metadata's additional property's value isn't the same");
+
+        final JsonObject payload = event.getJsonObject(BaseTransformer.PAYLOAD);
+        assertTrue(payload.containsKey(FIXTURE_PAYLOAD_ADDITIONAL_PROPERTY));
```
gwenneg, line 108: This could be removed since there's a value check below.
```suggestion
```

engine/src/test/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessorTest.java
```diff
@@ -299,46 +351,185 @@ void oBEndpointProcessingGoodBridge() {
...
+        final JsonObject eventResult = events.getJsonObject(0);
+
+        Assertions.assertTrue(eventResult.containsKey("metadata"), "the \"metadata\" field is not present");
+        Assertions.assertTrue(eventResult.containsKey("payload"), "the \"payload\" field is not present");
+
+        // Assert the "payload" object in the event.
+        final JsonObject payload = eventResult.getJsonObject("payload");
+        Assertions.assertTrue(payload.containsKey(FIXTURE_PAYLOAD_ADDITIONAL_PROPERTY), String.format("the \"%s\" payload field is not present", FIXTURE_PAYLOAD_ADDITIONAL_PROPERTY));
+        Assertions.assertTrue(payload.containsKey(FIXTURE_PAYLOAD_ADDITIONAL_PROPERTY_2), String.format("the \"%s\" payload field is not present", FIXTURE_PAYLOAD_ADDITIONAL_PROPERTY_2));
+        Assertions.assertTrue(payload.containsKey(FIXTURE_PAYLOAD_ADDITIONAL_PROPERTY_3), String.format("the \"%s\" payload field is not present", FIXTURE_PAYLOAD_ADDITIONAL_PROPERTY_3));
```
gwenneg, line 472: These could be removed since there are value checks below.
```suggestion
```

engine/src/test/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessorTest.java
```diff
@@ -299,46 +351,185 @@ void oBEndpointProcessingGoodBridge() {
...
+        Assertions.assertEquals(this.environment.url(), json.getString("environment_url"), "the environment URL isn't the same");
+        Assertions.assertEquals(DEFAULT_ORG_ID, json.getString("rhorgid"), "the \"rhorgid\" values don't match");
+        Assertions.assertEquals(CamelTypeProcessor.SPEC_VERSION, json.getString("specversion"), "the \"specversion\" values don't match");
+        // The UUID is randomly generated, so the only way to test that the ID is valid is to check if it is a valid
+        // UUID.
+        UUID.fromString(json.getString("id"));
+        Assertions.assertEquals(CamelTypeProcessor.SOURCE, json.getString("source"), "the \"source\" values don't match");
+        Assertions.assertEquals(FIXTURE_CAMEL_EXTRAS_PROCESSOR_NAME, json.getString("processorname"), "the \"processorname\" values don't match");
+        Assertions.assertEquals(CamelTypeProcessor.TYPE, json.getString("type"), "the \"type\" values don't match");
+        Assertions.assertEquals(this.environment.url(), json.getString("environment_url"), "the \"environment_url\" values don't match");
```
gwenneg, line 433: Duplicate of line 424.
```suggestion
```

engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
```diff
@@ -257,6 +271,9 @@ private void callOpenBridge(JsonObject body, UUID id, String orgId, CamelPropert
                 .baseUri(URI.create(bridge.getEndpoint()))
                 .build(BridgeEventService.class);
 
+        // Include the environment's URL when sending the event.
+        ce.put("environment_url", this.environment.url());
```
gwenneg, line 275: Even though this _could_ work (I'm not sure about it), let's put the `environment_url` in the `body` variable of this method (after `body.remove(NOTIF_METADATA_KEY)` for example) so that the URL ends up in the `data` field that is used in RHOSE to generate the Slack notification from our template.

engine/src/test/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessorTest.java
```diff
@@ -299,46 +351,185 @@ void oBEndpointProcessingGoodBridge() {
...
+        Assertions.assertEquals(FIXTURE_ACTION_ACCOUNT_ID, data.getString("account_id"), "the \"account_id\" values don't match");
+        Assertions.assertEquals(FIXTURE_ACTION_APP, data.getString("application"), "the \"application\" values don't match");
+        Assertions.assertEquals(FIXTURE_ACTION_BUNDLE, data.getString("bundle"), "the \"bundle\" values don't match");
+        Assertions.assertEquals(FIXTURE_ACTION_EVENT_TYPE, data.getString("event_type"), "the \"event_type\" values don't match");
+        Assertions.assertEquals(FIXTURE_ACTION_ORG_ID, data.getString("org_id"), "the \"org_id\" values don't match");
+        Assertions.assertEquals(FIXTURE_ACTION_TIMESTAMP.toString(), data.getString("timestamp"), "the \"timestamp\" values don't match");
+
+        // Assert the fields and the values in the "context" object.
+        final JsonObject context = data.getJsonObject("context");
+        Assertions.assertTrue(context.containsKey(FIXTURE_ACTION_CONTEXT), String.format("the expected \"%s\" context field is not present", FIXTURE_ACTION_CONTEXT));
```
gwenneg, line 454: This could be removed since there's a value check below.
```suggestion
```

engine/src/main/java/com/redhat/cloud/notifications/transformers/BaseTransformer.java
MikelAlejoBR: Done!

engine/src/test/java/com/redhat/cloud/notifications/transformers/BaseTransformerTest.java
```diff
@@ -1,36 +1,114 @@
...
+        action.setEventType(FIXTURE_EVENT_TYPE);
+        action.setEvents(FIXTURE_EVENTS);
+        action.setOrgId(FIXTURE_ORG_ID);
+        action.setTimestamp(FIXTURE_TIMESTAMP);
+
+        // Call the function under test.
+        JsonObject result = this.baseTransformer.toJsonObject(action);
+
+        // Assert the values.
+        assertEquals(FIXTURE_ACCOUNT_ID, result.getString(BaseTransformer.ACCOUNT_ID), "the account id isn't the same");
```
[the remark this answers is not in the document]
MikelAlejoBR: You are right, yes, it makes more sense to just use the original object to make the comparisons. Changed!

engine/src/test/java/com/redhat/cloud/notifications/transformers/BaseTransformerTest.java
MikelAlejoBR: Done!

engine/src/test/java/com/redhat/cloud/notifications/transformers/BaseTransformerTest.java
MikelAlejoBR: This too!

engine/src/test/java/com/redhat/cloud/notifications/transformers/BaseTransformerTest.java
MikelAlejoBR: And this one!

engine/src/test/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessorTest.java
MikelAlejoBR: Thanks! Done!

engine/src/test/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessorTest.java
MikelAlejoBR: This too!

engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
MikelAlejoBR: All right! Done!

gwenneg (approved):
LGTM

PR #1571, opened 2022-10-20
RHCLOUD-22557 Using Failed instead of FailedInternal and FailedExternal

josejulio:


josejulio:
Marking as draft - I'm missing the inverse operation for a filter

backend/src/main/java/com/redhat/cloud/notifications/routers/EventResource.java
```diff
@@ -64,6 +66,7 @@ public Page<EventLogEntry> getEvents(@Context SecurityContext securityContext, @
 
         Set<EndpointType> basicTypes = Collections.emptySet();
         Set<CompositeEndpointType> compositeTypes = Collections.emptySet();
+        Set<NotificationStatus> notificationStatusSet = status == null ? Set.of() : toNotificationStatus(status);
```
josejulio, line 69: I noticed that `status` was always an empty set, even if the query param wasn't specified- Anyway added a null check.

backend/src/main/java/com/redhat/cloud/notifications/routers/EventResource.java
```diff
@@ -130,4 +133,48 @@ public Page<EventLogEntry> getEvents(@Context SecurityContext securityContext, @
...
+            default:
+                Log.warnf("Uncovered status found:[%s]. This is a bug", status);
+                return EventLogEntryActionStatus.UNKNOWN;
+        }
+    }
+
+    static Set<NotificationStatus> toNotificationStatus(Set<EventLogEntryActionStatus> statusSet) {
+        Set<NotificationStatus> notificationStatusSet = new HashSet<>();
+        statusSet.forEach(status -> {
+            switch (status) {
```
gwenneg, line 157: Could you please add a null-check on `status` here? RESTEasy allows `null` values in collections that are received as REST params. We'll probably need to throw a `BadRequestException` is status is `null`.

backend/src/main/java/com/redhat/cloud/notifications/routers/EventResource.java
```diff
@@ -130,4 +133,48 @@ public Page<EventLogEntry> getEvents(@Context SecurityContext securityContext, @
...
+            case FAILED_EXTERNAL:
+            case FAILED_INTERNAL:
+                return EventLogEntryActionStatus.FAILED;
+            default:
+                Log.warnf("Uncovered status found:[%s]. This is a bug", status);
+                return EventLogEntryActionStatus.UNKNOWN;
+        }
+    }
+
+    static Set<NotificationStatus> toNotificationStatus(Set<EventLogEntryActionStatus> statusSet) {
```
gwenneg, line 154: We could use a test for each switch case.

backend/src/main/java/com/redhat/cloud/notifications/routers/EventResource.java
josejulio: done

backend/src/main/java/com/redhat/cloud/notifications/routers/EventResource.java
josejulio: aaand done.

PR #1572, opened 2022-10-20
RHCLOUD-22573 Fix NPE when saving a notificationHistory

josejulio:
- If there was an exception in the sending mechanism, the history was created without status, causing the NPE


gwenneg (commented):
LGTM

PR #1573, opened 2022-10-20
RHCLOUD-22573 Add a test to verify NotificationHistory status

josejulio:
This adds a test to verify the status after an exception has been throw at any part of the processing.

It does so by mocking the `Failsafe.with` method to throw an exception.


gwenneg (approved):
LGTM

PR #1575, opened 2022-10-25
[RHCLOUD-22602] refactor: reformat the POMs using the Quarkus CLI

MikelAlejoBR:
This will standardize the POMs and make it easier to review the changes when we use the CLI to add or remove dependencies. The command used to reformat the POMs was the following one:

* quarkus extension remove quarkus-rest-client-mutiny

None of the POMs had that extension, so that way I made sure that the command was basically a NOP that reformatted the files.

## Links

[[RHCLOUD-22602]](https://issues.redhat.com/browse/RHCLOUD-22602)


MikelAlejoBR:
What do you all think? On the one hand I like that it is standardized, and that now issuing a `quarkus extension <action> <extension-name>` command simply changes a few lines. On the other hand, it removes all the comments or spaces that we might have and might make the POMs more readable...

gwenneg:
Could you try putting the old comments back at their new location and then adding a new extension (without committing it) to check if they are removed again or preserved?

MikelAlejoBR:
Yeah, I tried that but unfortunately the CLI removes them. It removes all the white spaces and the comments from the file.

gwenneg:
Let's put this on hold until https://github.com/quarkusio/quarkus/issues/28873 has been fixed.

MikelAlejoBR:
After discussing it in the team meeting, we've said that we are going to close this one for now. The Quarkus developer team is aware of the issue, but obviously this is more a "tech debt" or a very low priority issue than a very important one. Also, since we already have our POMs already set up, using the CLI is going to turn everything upside down and we probably won't benefit too much from using it at his point. We might come back to this in the future.

PR #1580, opened 2022-10-27
RHCLOUD-22686 Update internal template render REST API to support mul…

josejulio:
…tiple templates

Instead of specifying the subject and body, now you can send an array of the template contents.


MikelAlejoBR (approved):
LGTM!

PR #1583, opened 2022-10-28
[RHCLOUD-21733]feature: remove default endpoints from the database

MikelAlejoBR:
After researching we have found out that it is safe to remove the endpoints from the database.

# Links
[[RHCLOUD-21733]](https://issues.redhat.com/browse/RHCLOUD-21733)

### DIFF
--- a/database/src/main/resources/db/migration/V1.67.0__RHCLOUD-21733_remove-default-endpoints.sql
+++ b/database/src/main/resources/db/migration/V1.67.0__RHCLOUD-21733_remove-default-endpoints.sql
@@ -1,0 +1,5 @@
+DELETE FROM
+    endpoints
+WHERE
+    endpoint_type = 2 -- stands for default endpoint types.
+;


database/src/main/resources/db/migration/V1.67.0__RHCLOUD-21733_remove-default-endpoints.sql
```diff
@@ -0,0 +1,5 @@
+DELETE FROM
+    endpoints
+WHERE
+    endpoint_type = 2 -- stands for default endpoint types.
```
gwenneg, line 4: :+1:

https://github.com/RedHatInsights/notifications-backend/blob/8c03a313a6944ee19aa0da121e4bc7d8cd295f2b/common/src/main/java/com/redhat/cloud/notifications/models/EndpointType.java#L15-L23

gwenneg (approved):
LGTM

josejulio (approved):
lgtm

PR #1586, opened 2022-10-31
Moving aggregator models to common-aggregator

josejulio:
- These models were duplicated  in the aggregator and engine

Unsure if there was a good reason for that or maybe something that predates our common module.


gwenneg (commented):
Deduplicating model classes is great but this PR also introduces many unnecessary transitive dependencies into `notifications-aggregator`: cache, micrometer, rest-client-reactive, resteasy-reactive-jackson, smallrye-openapi, swagger-ui... I'm not sure that's a good thing.

josejulio:
> Deduplicating model classes is great but this PR also introduces many unnecessary transitive dependencies into `notifications-aggregator`: cache, micrometer, rest-client-reactive, resteasy-reactive-jackson, smallrye-openapi, swagger-ui... I'm not sure that's a good thing.

Maybe we should create a more focused model pom then: `notifications-aggregator-common`.

gwenneg:
Yes, introducing a new Maven module is the cleanest way to deal with the classes duplication. That module doesn't have to be published on Maven Central as a lib because everything's in the same repo.

gwenneg (approved):
LGTM

PR #1587, opened 2022-10-31
[RHCLOUD-21830] feature: annotation to disallow internal hosts

MikelAlejoBR:
This annotation should help in filtering the URLs that we are given that have internal IPs or hostnames that point to internal IPs. That way we can mitigate part of the SSRF attacks.

## Links

[[RHCLOUD-21830]](https://issues.redhat.com/browse/RHCLOUD-21830)

### DIFF
--- a/backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
+++ b/backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
@@ -1706,46 +1706,30 @@ public class EndpointResourceTest extends DbIsolatedTest {
     @Test
     void testEndpointInvalidUrls() {
         // Set up the RBAC access for the test.
-        final String orgId = "endpoint-invalid-urls";
-        final String userName = "user";
-
-        final String identityHeaderValue = TestHelpers.encodeRHIdentityInfo(orgId, orgId, userName);
+        final String identityHeaderValue = TestHelpers.encodeRHIdentityInfo("endpoint-invalid-urls", "endpoint-invalid-urls", "user");
         final Header identityHeader = TestHelpers.createRHIdentityHeader(identityHeaderValue);
 
         MockServerConfig.addMockRbacAccess(identityHeaderValue, MockServerConfig.RbacAccess.FULL_ACCESS);
 
-        // Set up the fixture data.
-        final var disableSslVerification = false;
-        final HttpType method = HttpType.POST;
-        final String password = "endpoint-invalid-urls-basic-authentication-password";
-        final String username = "endpoint-invalid-urls-basic-authentication-username";
-        final String secretToken = "endpoint-invalid-urls-secret-token";
-
         // Create the properties for the endpoint. Leave the URL so that we can set it afterwards.
         final var camelProperties = new CamelProperties();
-        camelProperties.setBasicAuthentication(new BasicAuthentication(username, password));
-        camelProperties.setDisableSslVerification(disableSslVerification);
-        camelProperties.setSecretToken(secretToken);
+        camelProperties.setBasicAuthentication(new BasicAuthentication("endpoint-invalid-urls-basic-authentication-username", "endpoint-invalid-urls-basic-authentication-password"));
+        camelProperties.setDisableSslVerification(false);
+        camelProperties.setSecretToken("endpoint-invalid-urls-secret-token");
 
         final var webhookProperties = new WebhookProperties();
-        webhookProperties.setBasicAuthentication(new BasicAuthentication(username, password));
-        webhookProperties.setDisableSslVerification(disableSslVerification);
-        webhookProperties.setMethod(method);
-        webhookProperties.setSecretToken(secretToken);
+        webhookProperties.setBasicAuthentication(new BasicAuthentication("endpoint-invalid-urls-basic-authentication-username", "endpoint-invalid-urls-basic-authentication-password"));
+        webhookProperties.setDisableSslVerification(false);
+        webhookProperties.setMethod(HttpType.POST);
+        webhookProperties.setSecretToken("endpoint-invalid-urls-secret-token");
 
         // Create an endpoint without the type and the properties set.
-        final var name = "endpoint-invalid-urls-name";
-        final var description = "endpoint-invalid-urls-description";
-        final var enabled = true;
-        final var serverErrors = 0;
-        final var subType = "slack";
-
         final var endpoint = new Endpoint();
-        endpoint.setDescription(description);
-        endpoint.setEnabled(enabled);
-        endpoint.setName(name);
-        endpoint.setServerErrors(serverErrors);
-        endpoint.setSubType(subType);
+        endpoint.setDescription("endpoint-invalid-urls-description");
+        endpoint.setEnabled(true);
+        endpoint.setName("endpoint-invalid-urls-name");
+        endpoint.setServerErrors(0);
+        endpoint.setSubType("slack");
 
         // Create a simple class to make testing easier.
         final class TestCase {
--- a/backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
+++ b/backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
@@ -19,6 +19,8 @@ import com.redhat.cloud.notifications.models.HttpType;
 import com.redhat.cloud.notifications.models.IntegrationTemplate;
 import com.redhat.cloud.notifications.models.Template;
 import com.redhat.cloud.notifications.models.WebhookProperties;
+import com.redhat.cloud.notifications.models.validation.ValidNonPrivateUrlValidator;
+import com.redhat.cloud.notifications.models.validation.ValidNonPrivateUrlValidatorTest;
 import com.redhat.cloud.notifications.openbridge.Bridge;
 import com.redhat.cloud.notifications.openbridge.BridgeHelper;
 import com.redhat.cloud.notifications.openbridge.BridgeItemList;
@@ -31,6 +33,7 @@ import io.restassured.http.ContentType;
 import io.restassured.http.Header;
 import io.restassured.response.Response;
 import io.vertx.core.json.JsonObject;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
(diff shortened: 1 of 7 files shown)


common/src/main/java/com/redhat/cloud/notifications/models/validation/HttpsNonInternalUrlValidator.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Checks that the URL is well-formed, that the URI syntax is correct, and finally that the host doesn't point to
+     * an internal IP.
+     * @param rawUrl the URL to check.
+     * @param constraintValidatorContext the validation context.
+     * @return true if the provided URL is a well-formed, HTTPS URL which points to an external IP.
+     */
+    @Override
+    public boolean isValid(String rawUrl, ConstraintValidatorContext constraintValidatorContext) {
```
josejulio, line 30: I don't think it's good to keep it all on the same validation, it's probably worth to split into more validators, but it's good now!

The reason is to provide a better error message:
- Invalid url
- Unable to resolve host
- Internal addresses are not allowed

common/src/main/java/com/redhat/cloud/notifications/models/validation/HttpsNonInternalUrlValidator.java
```diff
@@ -0,0 +1,49 @@
...
+        try {
+            final URL url = new URL(rawUrl);
+            final URI uri = url.toURI();
+
+            // Check that the scheme is "HTTPS".
+            if (!SCHEME_HTTPS.equals(uri.getScheme())) {
+                return false;
```
josejulio, line 37: We actually support non https calls - which probably we should warn the user in the UI instead of full blocking it.

josejulio (changes requested):

common/src/main/java/com/redhat/cloud/notifications/models/validation/HttpsNonInternalUrlValidator.java
MikelAlejoBR: Can you believe that I had developed three annotations, had them tested, and then be like "nah, probably I'm going to get told to put them all in one"? :rofl: 

All right, I'll work on that!

common/src/main/java/com/redhat/cloud/notifications/models/validation/HttpsNonInternalUrlValidator.java
MikelAlejoBR: Great, no problem. Way better!

josejulio (approved):
lgtm

common/src/main/java/com/redhat/cloud/notifications/models/validation/NotPrivateHostnameValidator.java
```diff
@@ -0,0 +1,36 @@
+package com.redhat.cloud.notifications.models.validation;
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import java.net.InetAddress;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.UnknownHostException;
+
+public class NotPrivateHostnameValidator implements ConstraintValidator<NotPrivateHostname, String>  {
```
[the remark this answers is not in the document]
MikelAlejoBR: I have included the IPs in the test and the annotation seems to pick them up too. The same thing for the `AllowedHttpSchemeValidator`, since it should reject URLs without the `http` or `https` schemes specified. So I guess it validates the IPs too... however it will also not be used just by itself, so I was also thinking on maybe grouping these three annotations into one or something...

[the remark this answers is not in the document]
josejulio: I saw that you added tests for the rejected case for the ips - which i think it's fine. I'm guessing that it won't accept non internal ips? i.e. `34.235.198.240`

common/src/main/java/com/redhat/cloud/notifications/models/validation/AllowedHttpsSchemeValidator.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public boolean isValid(final String rawUrl, final ConstraintValidatorContext constraintValidatorContext) {
+        try {
+            final var url = new URL(rawUrl);
+            final URI uri = url.toURI();
+
+            final String scheme = uri.getScheme();
+
+            return SCHEME_HTTP.equals(scheme) || SCHEME_HTTPS.equals(scheme);
+        } catch (MalformedURLException | URISyntaxException e) {
```
gwenneg, line 35: Will this constraint validator always be executed after `WellFormedUrlValidator`? Otherwise, this validator may return a validator error saying that the HTTP scheme is invalid while the actual problem is a malformed URL.

common/src/main/java/com/redhat/cloud/notifications/models/validation/AllowedHttpsScheme.java
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Defines the annotation to make sure that the provided {@link String} field is defined with an allowed HTTP scheme.
+ */
+@Constraint(validatedBy = AllowedHttpsSchemeValidator.class)
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.FIELD})
+public @interface AllowedHttpsScheme {
+    Class<?>[] groups() default {};
+    String message() default "The provided URL does not contain a valid HTTP scheme";
```
gwenneg, line 20: The error message could list the allowed schemes: http and https.

common/src/main/java/com/redhat/cloud/notifications/models/validation/AllowedHttpsScheme.java
```diff
@@ -0,0 +1,22 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Defines the annotation to make sure that the provided {@link String} field is defined with an allowed HTTP scheme.
+ */
+@Constraint(validatedBy = AllowedHttpsSchemeValidator.class)
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.FIELD})
+public @interface AllowedHttpsScheme {
```
gwenneg, line 18: Nitpick:
```suggestion
public @interface AllowedHttpScheme {
```
Same for the validator.

gwenneg: It could even be named:
```suggestion
public @interface AllowedScheme {
```

MikelAlejoBR: Renamed!

common/src/main/java/com/redhat/cloud/notifications/models/validation/AllowedHttpsSchemeValidator.java
MikelAlejoBR: We don't know. This is one of the reasons why I initially wrote the three annotations all in one, though that came with the drawback of not having a precise enough message.

I could probably create a validation group which enforces the order of these three validations, and then apply that in the corresponding "endpoint resource".

gwenneg: `MalformedURLException` is caught by the 3 validators which means that a malformed URL will trigger a validation error with any of the 3 validators randomly. Similarly, `UnknownHostException` is caught by two validators and is affected by the same kind of issue.

I think the validation of the URL should be deterministic and the API users should never receive misleading error messages.

MikelAlejoBR: I have modified the annotations so that more accurate error messages are returned if these exceptions are thrown... I thought of coming back to a "single annotation validation" with these custom error messages too, but I rather wait until I receive feedback from you, since I'm not very sure if what I sent is acceptable or not...

common/src/main/java/com/redhat/cloud/notifications/models/WebhookProperties.java
```diff
@@ -17,7 +20,11 @@
 @Table(name = "endpoint_webhooks")
 @JsonNaming(SnakeCaseStrategy.class)
 public class WebhookProperties extends EndpointProperties {
+
+    @AllowedScheme
     @NotNull
+    @NotPrivateHostname
+    @WellFormedUrl
```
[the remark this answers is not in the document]
MikelAlejoBR: I am going to check that right now and post the results in the ticket.

common/src/main/java/com/redhat/cloud/notifications/models/validation/AllowedHttpsSchemeValidator.java
MikelAlejoBR: All right, I've gone back to the single annotation with multiple error messages version.

common/src/main/java/com/redhat/cloud/notifications/models/validation/ValidNonPrivateUrlValidator.java
```diff
@@ -0,0 +1,99 @@
...
+import java.net.UnknownHostException;
+
+public class ValidNonPrivateUrlValidator implements ConstraintValidator<ValidNonPrivateUrl, String> {
+
+    // Error messages.
+    public static final String INVALID_SCHEME = "The provided URL does not contain a valid scheme. It should be either \"http\" or \"https\"";
+    public static final String INVALID_URI = "The provided URL cannot be converted to a valid URI";
+    public static final String INVALID_URL = "The provided URL is invalid";
+    public static final String PRIVATE_IP = "The provided URL's hostname resolves to a private IP";
+    public static final String UNKNOWN_HOST = "The IP address of the provided host cannot be determined";
```
gwenneg, line 19: These messages may be forwarded to the frontend and showed to the customers (org admins) soon. We could make them a bit more user-friendly:
```suggestion
    public static final String INVALID_SCHEME = "The endpoint URL must start with \"http\" or \"https\"";
    public static final String INVALID_URL = "The endpoint URL is invalid";
    public static final String PRIVATE_IP = "The endpoint URL's host resolves to a private IP";
    public static final String UNKNOWN_HOST = "The IP address of the endpoint URL's host cannot be determined";
```
From a customer perspective, there's no real difference between `INVALID_URI` and `INVALID_URL` which is why I removed `INVALID_URI` from my suggestion.

common/src/main/java/com/redhat/cloud/notifications/models/CamelProperties.java
```diff
@@ -21,6 +22,7 @@
 public class CamelProperties extends EndpointProperties implements SourcesSecretable {
 
     @NotNull
+    @ValidNonPrivateUrl
```
gwenneg, line 25: Could you please test this annotation with our REST API? Here's an example of something similar:
https://github.com/RedHatInsights/notifications-backend/blob/415d7ab35236b52fcba33bd9844f251830525d3f/backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java#L302-L325
Both `CamelProperties` and `WebhookProperties` will need to be tested that way.

common/src/main/java/com/redhat/cloud/notifications/models/validation/AllowedHttpsSchemeValidator.java
josejulio: Sorry for the going back and forth :D

MikelAlejoBR: Don't worry! It's actually my bad, since I made you all review this a few times... I learned about overriding the default messages, so I think it's been positive! Thanks @josejulio ! :smile:

common/src/main/java/com/redhat/cloud/notifications/models/WebhookProperties.java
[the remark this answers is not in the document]
MikelAlejoBR: I have updated the Jira ticket with details of the issue, but I want to leave notice here too. Basically we have got a few endpoints that don't comply with this new annotation. My idea was to disable all of them, because there are a few endpoints that are still enabled —and if they were used, they would be probably auto-disabled by our "disable after N retries" policy—, which would make the users that are interested in using that endpoint to reintroduce a valid URL when editing the endpoint.

common/src/main/java/com/redhat/cloud/notifications/models/validation/ValidNonPrivateUrlValidator.java
MikelAlejoBR: Done!

common/src/main/java/com/redhat/cloud/notifications/models/CamelProperties.java
MikelAlejoBR: Added a couple of tests!

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
```diff
@@ -1695,6 +1699,218 @@ void testSearchWithType() {
...
+        final var enabled = true;
+        final var serverErrors = 0;
+        final var subType = "slack";
+
+        final var endpoint = new Endpoint();
+        endpoint.setDescription(description);
+        endpoint.setEnabled(enabled);
+        endpoint.setName(name);
+        endpoint.setServerErrors(serverErrors);
+        endpoint.setSubType(subType);
```
gwenneg, line 1749: I think (and you can disagree with that 😄) that the test would be a bit easier to read/maintain if the fields were not declared as `vars` before being set in `endpoint`. Values that are not required could also be omitted. That's only my personal opinion but when I read a test, I like to spend as less time as possible on minor/irrelevant details and focus my brain on the thing that is being tested. This comment also applies to other objects built in this class.
```suggestion
        // Create an endpoint without the type and the properties set.
        final var endpoint = new Endpoint();
        endpoint.setDescription("endpoint-invalid-urls-description");
        endpoint.setName("endpoint-invalid-urls-name");
```

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
```diff
@@ -1695,6 +1699,218 @@ void testSearchWithType() {
...
+                ValidNonPrivateUrlValidatorTest.unknownHosts
+            )
+        );
+
+        // Test the URLs with both camel and webhook endpoints.
+        for (final var testCase : testCases) {
+            for (final var url : testCase.testUrls) {
+                // Test with a camel endpoint.
+                camelProperties.setUrl(url);
+                endpoint.setType(EndpointType.CAMEL);
```
gwenneg, line 1792: The "slack" subtype should be set here instead of above (it doesn't make sense for `webhook` endpoints).

common/src/main/java/com/redhat/cloud/notifications/models/validation/ValidNonPrivateUrlValidator.java
```diff
@@ -0,0 +1,94 @@
...
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.net.UnknownHostException;
+
+public class ValidNonPrivateUrlValidator implements ConstraintValidator<ValidNonPrivateUrl, String> {
+
+    // Error messages.
+    public static final String INVALID_SCHEME = "The endpoint URL must start with \"http\" or \"https\"";
+    public static final String INVALID_URL = "The endpoint's URL is invalid";
+    public static final String PRIVATE_IP = "The host of the endpoint's URL host resolves to a private IP";
```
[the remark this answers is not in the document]
MikelAlejoBR: Oops! Done, thanks!

common/src/test/java/com/redhat/cloud/notifications/models/validation/ValidNonPrivateUrlValidatorTest.java
```diff
@@ -0,0 +1,154 @@
...
+import javax.validation.Validator;
+import javax.validation.ValidatorFactory;
+
+public class ValidNonPrivateUrlValidatorTest {
+
+    public static final String[] internalHosts = {"https://192.168.0.1", "https://172.16.0.1", "https://10.0.0.1", "https://192.168.0.1", "https://172.16.0.1", "https://10.0.0.1"};
+    public static final String[] invalidSchemes = {"ftp://redhat.com"};
+    public static final String[] malformedUris = {"https://example.com /hello", "https:/\\/example.com"};
+    public static final String[] malformedUrls = {"htt:/example.com", "redhat.com", "redhat"};
+    public static final String[] validUrls = new String[]{"http://redhat.com", "https://redhat.com"};
```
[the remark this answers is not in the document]
MikelAlejoBR: Done!

common/src/test/java/com/redhat/cloud/notifications/models/validation/ValidNonPrivateUrlValidatorTest.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Sets up the validator.
+     */
+    @BeforeAll
+    public static void setUp() {
+        final ValidatorFactory validatorFactory = Validation.buildDefaultValidatorFactory();
+
+        validator = validatorFactory.getValidator();
+
+        validatorFactory.close();
```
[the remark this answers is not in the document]
MikelAlejoBR: All right, much better. Done!

common/src/test/java/com/redhat/cloud/notifications/models/validation/ValidNonPrivateUrlValidatorTest.java
```diff
@@ -0,0 +1,154 @@
...
+        for (final var unknownHost : unknownHosts) {
+            final var invalid = new SimpleDto(unknownHost);
+
+            final var constraintViolations = validator.validate(invalid);
+            Assertions.assertEquals(expectedNumberConstraintViolations, constraintViolations.size(), "unexpected number of constraint violations");
+
+            for (final var cv : constraintViolations) {
+                Assertions.assertEquals(ValidNonPrivateUrlValidator.UNKNOWN_HOST, cv.getMessage(), "unexpected error message received");
+            }
+        }
```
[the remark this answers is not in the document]
MikelAlejoBR: Yeah, good catch... I was coming from having all these tests on separate files, and I forgot to step back and think about the design... thank you!

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
MikelAlejoBR: Surely it makes it more readable. I think this comes from my habit of writing tests that then check that the values are the same... but this doesn't make sense here. Thanks for the suggestion!

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
MikelAlejoBR: Right, I'll make it more accurate. Thanks!

gwenneg (approved):
LGTM

gwenneg:

PR #1588, opened 2022-11-02
[RHCLOUD-22667] feature: return bundle's display names on Slack notifications

MikelAlejoBR:
The idea is to make the messages more user friendly, by allowing them to chose the more friendlier bundle display name, instead of it's schematic name.

I was initially going to have just one single `toJsonObject` which took an `Event`, but I was unsure about which approach was better. So I decided to have two instead, but I'm open to discussing this :eyes:

## Links

[[RHCLOUD-22667]](https://issues.redhat.com/browse/RHCLOUD-22667)

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/transformers/BaseTransformer.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/transformers/BaseTransformer.java
@@ -1,6 +1,7 @@
 package com.redhat.cloud.notifications.transformers;
 
 import com.redhat.cloud.notifications.ingress.Action;
+import com.redhat.cloud.notifications.models.Event;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 
@@ -16,20 +17,24 @@ public class BaseTransformer {
     public static final String APPLICATION = "application";
     public static final String BUNDLE = "bundle";
     public static final String CONTEXT = "context";
+    public static final String DISPLAY_NAME = "display_name";
     public static final String EVENT_TYPE = "event_type";
     public static final String EVENTS = "events";
     public static final String METADATA = "metadata";
     public static final String ORG_ID = "org_id";
     public static final String PAYLOAD = "payload";
+    public static final String SOURCE = "source";
     public static final String TIMESTAMP = "timestamp";
 
     /**
-     * Transforms the given action into a {@link JsonObject}.
-     * @param action the {@link Action} to transform.
-     * @return a {@link JsonObject} containing the given action.
+     * Transforms the given event into a {@link JsonObject}.
+     * @param event the {@link Event} to transform.
+     * @return a {@link JsonObject} containing the given event.
      */
-    public JsonObject toJsonObject(final Action action) {
-        JsonObject message = new JsonObject();
+    public JsonObject toJsonObject(final Event event) {
+        final JsonObject message = new JsonObject();
+
+        final Action action = event.getAction();
 
         message.put(ACCOUNT_ID, action.getAccountId());
         message.put(APPLICATION, action.getApplication());
@@ -40,9 +45,9 @@ public class BaseTransformer {
             EVENTS,
             new JsonArray(
                 action.getEvents().stream().map(
-                    event -> Map.of(
-                        METADATA, JsonObject.mapFrom(event.getMetadata()),
-                        PAYLOAD, JsonObject.mapFrom(event.getPayload())
+                    ev -> Map.of(
+                        METADATA, JsonObject.mapFrom(ev.getMetadata()),
+                        PAYLOAD, JsonObject.mapFrom(ev.getPayload())
                     )
                 ).collect(Collectors.toList())
             )
@@ -50,7 +55,24 @@ public class BaseTransformer {
         message.put(ORG_ID, action.getOrgId());
         message.put(TIMESTAMP, action.getTimestamp().toString());
 
+        // Include the display names of the different objects.
+        // Since "insights-notifications-schemas-java:0.20".
+        final JsonObject source = new JsonObject();
+
+        final JsonObject sourceAppDisplayName = new JsonObject();
+        sourceAppDisplayName.put(DISPLAY_NAME, event.getApplicationDisplayName());
+        source.put(APPLICATION, sourceAppDisplayName);
+
+        final JsonObject sourceBundleDisplayName = new JsonObject();
+        sourceBundleDisplayName.put(DISPLAY_NAME, event.getBundleDisplayName());
+        source.put(BUNDLE, sourceBundleDisplayName);
+
+        final JsonObject sourceEventTypeDisplayName = new JsonObject();
+        sourceEventTypeDisplayName.put(DISPLAY_NAME, event.getEventTypeDisplayName());
+        source.put(EVENT_TYPE, sourceEventTypeDisplayName);
+
+        message.put(SOURCE, source);
+
         return message;
     }
-
 }
--- a/engine/src/test/java/com/redhat/cloud/notifications/ComplianceTestHelpers.java
+++ b/engine/src/test/java/com/redhat/cloud/notifications/ComplianceTestHelpers.java
@@ -6,8 +6,6 @@ import com.redhat.cloud.notifications.ingress.Event;
 import com.redhat.cloud.notifications.ingress.Metadata;
 import com.redhat.cloud.notifications.ingress.Payload;
 import com.redhat.cloud.notifications.models.EmailAggregation;
-import com.redhat.cloud.notifications.transformers.BaseTransformer;
-import io.vertx.core.json.JsonObject;
 
 import java.time.LocalDateTime;
 import java.util.List;
@@ -16,8 +14,6 @@ import static com.redhat.cloud.notifications.TestConstants.DEFAULT_ORG_ID;
 import static java.time.ZoneOffset.UTC;
 
 public class ComplianceTestHelpers {
-
-    public static BaseTransformer baseTransformer = new BaseTransformer();
 
     public static EmailAggregation createEmailAggregation(String bundle, String application, String eventType, String policyId, String inventoryId) {
         EmailAggregation aggregation = new EmailAggregation();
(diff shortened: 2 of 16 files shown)


josejulio:
I think we should send a consistent message across all our services and not only to camel.

Right now we send 1:1 what we received - I guess we could start enhacing the message with these additional values:

- Display name of:
  - bundle
  - application
  - event_type
- console url (depending on the environment)

Webhooks could also be used to send messages as well, and they could benefit on the extra data.

MikelAlejoBR:
In this case I'll refactor the code to have a single `toJsonObject`, then, and make it receive an `Event` instead.

gwenneg:
> I think we should send a consistent message across all our services and not only to camel/slack.
> 
> Right now we send 1:1 what we received - I guess we could start enhacing the message with these additional values:
> 
> * Display name of:
>   
>   * bundle
>   * application
>   * event_type
> * console url (depending on the environment)
> 
> Webhooks could also be used to send messages as well, and they could benefit on the extra data.

I agree. Should we talk about the new data keys/structure with all these new fields before finalizing this PR? Changing the keys or structure may break existing integrations so our initial choice is important.

engine/src/main/java/com/redhat/cloud/notifications/transformers/BaseTransformer.java
```diff
@@ -24,33 +26,37 @@ public class BaseTransformer {
...
-                    event -> Map.of(
-                        METADATA, JsonObject.mapFrom(event.getMetadata()),
-                        PAYLOAD, JsonObject.mapFrom(event.getPayload())
-                    )
-                ).collect(Collectors.toList())
-            )
-        );
-        message.put(ORG_ID, action.getOrgId());
-        message.put(TIMESTAMP, action.getTimestamp().toString());
+        message.put(BUNDLE_DISPLAY_NAME, event.getBundleDisplayName());
```
gwenneg, line 36: From an RHCLOUD-22667 perspective, the app display name is missing here. But as discussed in other comments, we'll probably add more fields as well.

josejulio:
> Should we talk about the new data keys/structure with all these new fields before finalizing this PR?
+1

MikelAlejoBR:
Rebased to include the latest dependency changes that should enable us to finish this MR.

MikelAlejoBR:
All right the PR is ready to be reviewed. Basically the `BaseTransformer` receives an `Event` now, and if the event contains the `*displayName`s for the "application", "bundle" or "event type" properties, then it includes them inside the `"source": {}` key in the resulting JSON.

engine/src/test/java/com/redhat/cloud/notifications/DriftTestHelpers.java
```diff
@@ -52,7 +52,11 @@ public static EmailAggregation createEmailAggregation(String bundle, String appl
 
         emailActionMessage.setOrgId(DEFAULT_ORG_ID);
 
-        JsonObject payload = baseTransformer.toJsonObject(emailActionMessage);
+        // Create an empty event for the "toJsonObject" function.
+        final com.redhat.cloud.notifications.models.Event event = new com.redhat.cloud.notifications.models.Event();
+        event.setAction(emailActionMessage);
+
+        final JsonObject payload = baseTransformer.toJsonObject(event);
```
josejulio, line 59: Could we create a test helpers that does all of this?

engine/src/main/java/com/redhat/cloud/notifications/transformers/BaseTransformer.java
```diff
@@ -40,17 +45,61 @@ public JsonObject toJsonObject(final Action action) {
...
+     * present or, they are blank, then the key is not included in the payload.
+     * @param message the JSON object to append the "source" key to.
+     * @param event the event to get the display names from.
+     */
+    private void appendSourceToMessage(final JsonObject message, final Event event) {
+        final JsonObject source = new JsonObject();
+
+        JsonObject sourceAppDisplayName = null;
+        final String appDisplayName = event.getApplicationDisplayName();
+        if (appDisplayName != null && !appDisplayName.isBlank()) {
```
josejulio, line 76: IIRC all our event-types/apps/bundles should have a display name all the times- Is there a specific case where it does not?

engine/src/test/java/com/redhat/cloud/notifications/VulnerabilityTestHelpers.java
```diff
@@ -39,8 +38,8 @@ public static EmailAggregation createEmailAggregation(String bundle, String appl
                                                             .build())
                                     .build()
         ));
-        JsonObject payload = baseTransformer.toJsonObject(emailActionMessage);
-        aggregation.setPayload(payload);
+
+        aggregation.setPayload(TestHelpers.wrapActionToJsonObject(emailActionMessage));
```
gwenneg, line 42: `baseTransformer` is no longer needed in this class.

engine/src/test/java/com/redhat/cloud/notifications/InventoryTestHelpers.java
```diff
@@ -80,8 +79,7 @@ public static EmailAggregation createEmailAggregation(String tenant, String bund
 
         emailActionMessage.setOrgId(DEFAULT_ORG_ID);
 
-        JsonObject payload = baseTransformer.toJsonObject(emailActionMessage);
-        aggregation.setPayload(payload);
+        aggregation.setPayload(TestHelpers.wrapActionToJsonObject(emailActionMessage));
```
gwenneg, line 82: `baseTransformer` is no longer needed in this class.

engine/src/test/java/com/redhat/cloud/notifications/DriftTestHelpers.java
```diff
@@ -52,8 +51,7 @@ public static EmailAggregation createEmailAggregation(String bundle, String appl
 
         emailActionMessage.setOrgId(DEFAULT_ORG_ID);
 
-        JsonObject payload = baseTransformer.toJsonObject(emailActionMessage);
-        aggregation.setPayload(payload);
+        aggregation.setPayload(TestHelpers.wrapActionToJsonObject(emailActionMessage));
```
gwenneg, line 54: `baseTransformer` is no longer needed in this class.

engine/src/test/java/com/redhat/cloud/notifications/ComplianceTestHelpers.java
```diff
@@ -56,8 +55,7 @@ public static EmailAggregation createEmailAggregation(String bundle, String appl
 
         emailActionMessage.setOrgId(DEFAULT_ORG_ID);
 
-        JsonObject payload = baseTransformer.toJsonObject(emailActionMessage);
-        aggregation.setPayload(payload);
+        aggregation.setPayload(TestHelpers.wrapActionToJsonObject(emailActionMessage));
```
gwenneg, line 58: `baseTransformer` is no longer needed in this class.

engine/src/test/java/com/redhat/cloud/notifications/PatchTestHelpers.java
```diff
@@ -56,8 +55,7 @@ public static EmailAggregation createEmailAggregation(String bundle, String appl
         ));
         emailActionMessage.setOrgId(DEFAULT_ORG_ID);
 
-        JsonObject payload = baseTransformer.toJsonObject(emailActionMessage);
-        aggregation.setPayload(payload);
+        aggregation.setPayload(TestHelpers.wrapActionToJsonObject(emailActionMessage));
```
gwenneg, line 58: `baseTransformer` is no longer needed in this class.

engine/src/test/java/com/redhat/cloud/notifications/TestHelpers.java
```diff
@@ -72,8 +73,7 @@ public static EmailAggregation createEmailAggregation(String orgId, String bundl
 
         emailActionMessage.setOrgId(orgId);
 
-        JsonObject payload = baseTransformer.toJsonObject(emailActionMessage);
-        aggregation.setPayload(payload);
+        aggregation.setPayload(TestHelpers.wrapActionToJsonObject(emailActionMessage));
```
gwenneg, line 76: Is `TestHelpers#baseTransformer` still used?

engine/src/test/java/com/redhat/cloud/notifications/VulnerabilityTestHelpers.java
MikelAlejoBR: Removed!

engine/src/test/java/com/redhat/cloud/notifications/InventoryTestHelpers.java
MikelAlejoBR: Removed!

engine/src/test/java/com/redhat/cloud/notifications/DriftTestHelpers.java
MikelAlejoBR: Removed!

engine/src/test/java/com/redhat/cloud/notifications/ComplianceTestHelpers.java
MikelAlejoBR: Removed!

engine/src/test/java/com/redhat/cloud/notifications/TestHelpers.java
MikelAlejoBR: Nope, I removed it!

engine/src/main/java/com/redhat/cloud/notifications/transformers/BaseTransformer.java
```diff
@@ -40,17 +45,34 @@ public JsonObject toJsonObject(final Action action) {
...
+
+        final JsonObject sourceBundleDisplayName = new JsonObject();
+        sourceBundleDisplayName.put(DISPLAY_NAME, event.getBundleDisplayName());
+        source.put(BUNDLE, sourceBundleDisplayName);
+
+        final JsonObject sourceEventTypeDisplayName = new JsonObject();
+        sourceEventTypeDisplayName.put(DISPLAY_NAME, event.getEventTypeDisplayName());
+        source.put(EVENT_TYPE, sourceEventTypeDisplayName);
+
+        message.put(SOURCE, source);
```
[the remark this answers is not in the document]
josejulio: I'm good with these - we wil need to do something similar with the cloud events

gwenneg:
`TeamsProcessor` needs a small update 😃

gwenneg (approved):
LGTM

MikelAlejoBR:
I'm going to redeploy it with Bonfire to see what's up...

gwenneg:
This PR will need to be rebased onto `master`.

PR #1594, opened 2022-11-08
Shorten RHOSE log entry to fix Kibana dashboard

gwenneg:
Kibana has a [256 chars limit](https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html) for log entry messages that are indexed. If that limit is exceeded, the messages are still displayed in the Kibana UI but they can't be queried to build Kibana dashboards.

This PR removes the integration name from our log entry which guarantees that the log entry will never exceed 256 characters. We didn't really need that name in the first place. The `processorName` field of the log entry already contains the endpoint ID from the Notifications DB.


gwenneg:
New log entry:
```
2022-11-08T16:34:44.7314375Z 2022-11-08 16:34:44,688 INFO  [com.red.clo.not.pro.cam.CamelTypeProcessor] (main) SE: Sending Event with historyId=19b438ab-d3e9-4e54-b86c-a79366e73f2c, orgId=default-org-id, processorName=test-custom-processor-name, processorId=null, action=slack, origId=dcf62167-d7bc-41dc-94b2-0829619e19df
```

PR #1596, opened 2022-11-09
[RHCLOUD-22470] Feature test endpoint

MikelAlejoBR:
It sends a Kafka message for that endpoint to the "ingress" Kafka
channel, so that the engine can pick it up and act accordingly.

 ## Links

[[RHCLOUD-22470]](https://issues.redhat.com/browse/RHCLOUD-22470)


backend/pom.xml
```diff
@@ -124,6 +124,19 @@
             <version>${quarkus-logging-cloudwatch.version}</version>
         </dependency>
 
+        <!-- Kafka -->
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-reactive-messaging-kafka</artifactId>
```
gwenneg, line 130: FYI we have a "zero Kafka" policy in `notifications-backend` to make sure our frontend never goes down when there's a Kafka outage (which happened several times in the past). You can work around that by forwarding the REST request by `notifications-backend` to a `notifications-engine` REST API using `quarkus-rest-client-reactive`.

gwenneg:
In case you didn't notice it, there's a Jira story about this: https://issues.redhat.com/browse/RHCLOUD-22470

backend/pom.xml
MikelAlejoBR: I see. Then my idea was to:

* Create a `POST /endpoints/test` which takes an `{"orgId": "whatever"}` as the payload, since we don't have the `x-rh-identity` utils to extract the orgId from the headers. And instead of repeating code, I was thinking on just receiving that payload.
* Create a `POST /endpoints/<uuid>/test` on the backend, which extracts the orgId and makes sure that the endpoint exists, before sending the order to the engine.

What do you think?

josejulio: It makes sense, we don't need to verify the `x-rh-identity` header in the engine. It's ok to trust notifications-backend with the  `endpointId`. `notifications-backend` can verify that the `orgId` of the caller matches the one in the `endpoint`.

MikelAlejoBR: I've been talking to Gwenneg and we finally may go the Kafka route instead of doing it via REST endpoints on our side. We were thinking on adding the `endpointId` to the `Context` of the `Action`, so that we can pull that ID and just skip all the behaviour group dance. What do you think @josejulio?

josejulio: It makes sense to me - that way we can put this in a queue and let the engine process it normally - We might come up with mechanism to notify the user about the status of the endpoint - it could be maybe a notification in the drawer and I suppose this event should appear in the event log as "manual test" or something?

MikelAlejoBR: Hey @josejulio,

Sorry about the delay, I forgot about this PR. 

> I suppose this event should appear in the event log as "manual test" or something?

Do I have to set up the event or event type in some particular way to make it show that text? For now I've created a custom event type, but I don't know if that's enough.

Thank you!

josejulio: I'm not enterely sure. 

Right now our event log shows as:

Screenshot_2022-12-01_12-40-09

If we are going to go with the route of using the `endpointId` in the action, we might not even need a real event - we would only need some way to mark the notification history so that we know it was a manual run.

We will have to modify the UI to detect it was a manual run to show something different (Manual testing or so)

MikelAlejoBR: Mmmmm... I added a new bundle and an application via a migration, which tie to the event type. Can the UI use that to show it as a test event, maybe?

gwenneg:
@MikelAlejoBR Let's try to make progress on this PR. Could you please rebase it onto the latest `master`?

MikelAlejoBR:
> @MikelAlejoBR Let's try to make progress on this PR. Could you please rebase it onto the latest `master`?

All right! Done!

backend/pom.xml
gwenneg: > I've been talking to Gwenneg and we finally may go the Kafka route instead of doing it via REST endpoints on our side.

I don't remember what we said several months ago but unless there is a very strong reason that I forgot, we cannot introduce Kafka in `notifications-backend`.

database/src/main/resources/db/migration/V1.71.0__RHCLOUD-22470_insert_test_action_events_bundle_application_event_type.sql
```diff
@@ -0,0 +1,42 @@
+-- Insert a bundle, an application and an event type that will be specifically
+-- used for test actions and events.
+
+INSERT INTO bundles(
```
gwenneg, line 4: Instead of creating a new BAET hierarchy for this, we could add a new event type under `[bundle=console, app=notifications]`. This event type could be:
- name: `integration-test`
- display_name: `Integration Test`

engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
```diff
@@ -172,6 +173,20 @@ public CompletionStage<Void> process(Message<String> message) {
...
+                                final var isTestEvent = (Boolean) propertiesMap.get(TestEventConstants.TEST_ACTION_CONTEXT_TEST_EVENT);
+
+                                if (isTestEvent != null && isTestEvent) {
+                                    final UUID endpointUuid = UUID.fromString((String) propertiesMap.get(TestEventConstants.TEST_ACTION_CONTEXT_ENDPOINT_ID));
+
+                                    this.endpointProcessor.processTestEvent(event, endpointUuid);
+                                }
+                            }
+                        }
+
```
gwenneg, line 189: I'm not sure we need to change anything in `EventConsumer`, see my next comment for the reason.

engine/src/main/java/com/redhat/cloud/notifications/events/EndpointProcessor.java
```diff
@@ -96,4 +98,35 @@ public void process(Event event) {
...
+                case EMAIL_SUBSCRIPTION:
+                    processor = emailProcessor;
+                    break;
+                default:
+                    throw new IllegalArgumentException("Unexpected endpoint type: " + endpoint.getType());
+            }
+
+            processor.process(event, List.of(endpoint));
+        });
+    }
```
gwenneg, line 131: Instead of doing this, we could replace
https://github.com/RedHatInsights/notifications-backend/blob/a55d5e62e4268dac31fb6ac789d47aa1e07cd2d4/engine/src/main/java/com/redhat/cloud/notifications/events/EndpointProcessor.java#L56-L58
with
```java
    private static boolean isIntegrationTest(Event event) {
        return event.getAction().getBundle().equals("console") &&
                event.getAction().getApplication().equals("notifications") &&
                event.getAction().getEventType().equals("integration-test");
    }

    public void process(Event event) {
        processedItems.increment();

        List<Endpoint> endpoints;
        if (isIntegrationTest(event)) {
            Endpoint endpoint = endpointRepository.findByUuid(event.getOrgId(), /* pass the endpoint ID from the action context */);
            endpoints = List.of(endpoint);
        } else {
            endpoints = endpointRepository.getTargetEndpoints(event.getOrgId(), event.getEventType());
        }
```

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -160,6 +160,27 @@ public boolean incrementEndpointServerErrors(UUID endpointId, int maxServerError
...
+     * @return the endpoint found.
+     */
+    public Endpoint findByUuid(final UUID endpointUuid) {
+        final String query =
+                "SELECT " +
+                    "e " +
+                "FROM " +
+                    "Endpoint AS e " +
+                "WHERE " +
+                    "id = :uuid";
```
gwenneg, line 175: Please check the `orgId` as well here.

gwenneg:
I'll add more minor comments later, I am focusing on major topics for now.

backend/pom.xml
MikelAlejoBR: I vaguely recall that we mentioned that on the ticket there were two suggestions to implement, and that the suggestion number 2 was the one we should probably implement. But to be honest, I think I misunderstood what we talked back then. I'm sorry for the confusion, I'll go back to the REST API implementation for the test events.

MikelAlejoBR: Reworked!

database/src/main/resources/db/migration/V1.71.0__RHCLOUD-22470_insert_test_action_events_bundle_application_event_type.sql
MikelAlejoBR: All right! I am assuming that you meant `app=integrations`, right?

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
```diff
@@ -0,0 +1,62 @@
...
+         * Set the rest of the action's members.
+         */
+        testAction.setApplication(TestEventConstants.TEST_ACTION_APPLICATION);
+        testAction.setBundle(TestEventConstants.TEST_ACTION_BUNDLE);
+        testAction.setEvents(List.of(testEvent));
+        testAction.setEventType(TestEventConstants.TEST_ACTION_EVENT_TYPE);
+        testAction.setId(UUID.randomUUID());
+        testAction.setRecipients(List.of(recipient));
+        testAction.setOrgId(orgId);
+        testAction.setTimestamp(LocalDateTime.now());
```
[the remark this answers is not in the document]
MikelAlejoBR: Done, thanks!

engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
```diff
@@ -172,6 +173,20 @@ public CompletionStage<Void> process(Message<String> message) {
                      * The Event and the Action it contains are processed by all relevant endpoint processors.
                      */
                     try {
+                        final Context context = action.getContext();
+                        if (context != null) {
+                            final var propertiesMap = context.getAdditionalProperties();
```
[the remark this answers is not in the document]
MikelAlejoBR: Refactored to explicitly specify the type!

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
MikelAlejoBR: Done! I removed the repository from the engine since it is no longer needed, but added a similar method on the backend's repository.

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * Regular action test data.
+     */
+    public static final String TEST_ACTION_CONTEXT_TEST_EVENT = "test-action-context-test-event";
+    public static final boolean TEST_ACTION_CONTEXT_TEST_EVENT_VALUE = true;
+    public static final String TEST_ACTION_CONTEXT_ENDPOINT_ID = "test-action-context-endpoint-id";
+    public static final String TEST_ACTION_METADATA_KEY = "test-metadata-key";
+    public static final String TEST_ACTION_METADATA_VALUE = "test-metadata-value";
+    public static final String TEST_ACTION_PAYLOAD_KEY = "test-payload-key";
+    public static final String TEST_ACTION_PAYLOAD_VALUE = "test-payload-value";
```
gwenneg, line 32: As a first step, we could use data that is a bit more user friendly. Maybe a payload that contains something like `"message": "Congratulations, the integration you created on https://console.redhat.com was successfully tested!"`.

Later, we may provide a default payload and allow the org admin to change it from the frontend.

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
```diff
@@ -2012,6 +2030,65 @@ public void testEndpointSubtypeIsOnlyAllowedWhenRequired() {
...
+        MockServerConfig.addMockRbacAccess(identityHeaderValue, MockServerConfig.RbacAccess.FULL_ACCESS);
+
+        // Call the endpoint under test.
+        final String path = String.format("/endpoints/%s/test", UUID.randomUUID());
+        given()
+                .header(identityHeader)
+                .when()
+                .post(path)
+                .then()
+                .statusCode(404);
```
gwenneg, line 2089: If you apply my other comment which adds an error message, please check that message in this test.

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
```diff
@@ -0,0 +1,123 @@
...
+    public static final String TEST_ACTION_RECIPIENT = "test-recipient-1";
+    public static final String TEST_ACTION_VERSION = "0.0.0";
```
gwenneg, line 34: I'm not sure of the correct version we should use here, but `0.0.0` doesn't feel right. To be confirmed.

engine/pom.xml
```diff
@@ -167,6 +167,11 @@
             <version>${testcontainers.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jsr310</artifactId>
```
gwenneg, line 172: Is that really needed?

engine/src/test/java/com/redhat/cloud/notifications/routers/EndpointTestResourceTest.java
```diff
@@ -0,0 +1,125 @@
...
+            .then()
+            .statusCode(204);
+
+        // We should receive the action triggered by the REST call.
+        InMemorySink<String> actionsOut = this.inMemoryConnector.sink(FromCamelHistoryFiller.EGRESS_CHANNEL);
+        final var actionsList = actionsOut.received();
+
+        // Only one test action should have been sent to Kafka.
+        final var expectedActionsCount = 1;
+        Assertions.assertEquals(expectedActionsCount, actionsList.size(), "unexpected number of actions sent to Kafka");
```
gwenneg, line 67: Whenever you need to check the messages received by a sink, it is safer for test stability purposes to wait using Awaitility:
https://github.com/RedHatInsights/notifications-backend/blob/41ac032e469c30ab62cb311c7e47f950d4d3a054/engine/src/test/java/com/redhat/cloud/notifications/events/IntegrationDisabledNotifierTest.java#L77

engine/src/main/java/com/redhat/cloud/notifications/events/EndpointProcessor.java
```diff
@@ -96,4 +109,35 @@ public void process(Event event) {
             }
         });
     }
+
+    /**
+     * Processes the test event with the given endpoint's UUID.
+     * @param event the event to process.
+     * @param endpointUuid the endpoint's UUID.
+     */
+    public void processTestEvent(final Event event, final UUID endpointUuid) {
```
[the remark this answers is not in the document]
MikelAlejoBR: You're right, yes! Removed.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -103,9 +106,16 @@ public class EndpointResource {
     public static final String SLACK_WEBHOOK_URL = "slack_webhook_url";
     public static final String EMPTY_SLACK_CHANNEL_ERROR = "The channel field is required";
 
+    // The channel name we will send Kafka messages to for the "/test" endpoint.
+    public static final String INGRESS_CHANNEL = "ingress";
+
```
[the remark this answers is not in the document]
MikelAlejoBR: True, removed as well!

common/src/main/java/com/redhat/cloud/notifications/routers/endpoints/EndpointTestRequest.java
```diff
@@ -0,0 +1,19 @@
+package com.redhat.cloud.notifications.routers.endpoints;
+
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotNull;
+import java.util.UUID;
+
+public class EndpointTestRequest {
```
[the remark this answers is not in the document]
MikelAlejoBR: Done!

backend/src/main/java/com/redhat/cloud/notifications/routers/engine/EndpointTestService.java
```diff
@@ -0,0 +1,26 @@
...
+import com.redhat.cloud.notifications.routers.endpoints.EndpointTestRequest;
+import com.redhat.cloud.notifications.routers.sources.SourcesXRHIDHeaderFactory;
+import org.eclipse.microprofile.faulttolerance.Retry;
+import org.eclipse.microprofile.rest.client.annotation.RegisterClientHeaders;
+import org.eclipse.microprofile.rest.client.inject.RegisterRestClient;
+
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+
+@RegisterClientHeaders(SourcesXRHIDHeaderFactory.class)
```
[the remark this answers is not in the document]
MikelAlejoBR: Whoops... I made a mistake while copying the Sources service... Thanks for pointing it out!

backend/src/main/java/com/redhat/cloud/notifications/routers/engine/EndpointTestService.java
```diff
@@ -0,0 +1,26 @@
...
+@RegisterRestClient(configKey = "internal-endpoint-test")
+public interface EndpointTestService {
+
+    /**
+     * Sends a request to the engine to test the provided endpoint. This
+     * happens when the client wants to test their integration with a test
+     * event.
+     * @param endpointTestRequest the payload of the request.
+     */
+    @Path("/internal/endpoints/test")
```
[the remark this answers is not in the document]
MikelAlejoBR: Done!

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -658,6 +668,37 @@ public boolean unsubscribeEmail(
...
+                schema = @Schema(type = SchemaType.STRING)
+            )
+    })
+    @Produces(APPLICATION_JSON)
+    @RolesAllowed(ConsoleIdentityProvider.RBAC_READ_INTEGRATIONS_ENDPOINTS)
+    public Response testEndpoint(@Context SecurityContext sec, @RestPath UUID uuid) {
+        final String orgId = SecurityContextUtil.getOrgId(sec);
+
+        if (!this.endpointRepository.existsByUuidAndOrgId(uuid, orgId)) {
+            throw new NotFoundException();
```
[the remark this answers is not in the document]
MikelAlejoBR: Much better. Done!

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
```diff
@@ -0,0 +1,123 @@
...
+     * that it can be easily identified.
+     * @param endpointUuid the endpoint UUID that will be set in the context.
+     * @param orgId the org ID for the action.
+     * @return the created action.
+     */
+    public static Action createTestAction(final UUID endpointUuid, final String orgId) {
+        Action testAction = new Action();
+
+        final var context = new com.redhat.cloud.notifications.ingress.Context();
+        context.setAdditionalProperty(TEST_ACTION_CONTEXT_TEST_EVENT, TEST_ACTION_CONTEXT_TEST_EVENT_VALUE);
```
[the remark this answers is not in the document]
MikelAlejoBR: Yeah... I did kind of a double check, but we can get rid of it. Done!

database/src/main/resources/db/migration/V1.71.0__RHCLOUD-22470_insert_test_action_events_bundle_application_event_type.sql
```diff
@@ -0,0 +1,17 @@
...
+-- field so that it makes it easier to debug the event type.
+
+INSERT INTO event_type(
+    application_id,
+    id,
+    description,
+    display_name,
+    "name"
+) VALUES (
+    (SELECT id FROM applications WHERE "name"='integrations'),
```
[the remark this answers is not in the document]
MikelAlejoBR: I used an inner join instead, if you're okay with it! Done!

engine/src/main/java/com/redhat/cloud/notifications/routers/EndpointTestResource.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Creates a "endpoint integration test" action and sends it to the ingress
+     * Kafka channel, based on the received payload.
+     * @param endpointTestRequest the payload to create the test event from.
+     */
+    @Consumes(APPLICATION_JSON)
+    @Path("/test")
+    @POST
+    public Response testEndpoint(@Valid final EndpointTestRequest endpointTestRequest) {
```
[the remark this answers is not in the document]
MikelAlejoBR: Done!

engine/src/main/java/com/redhat/cloud/notifications/routers/EndpointTestResource.java
```diff
@@ -0,0 +1,46 @@
+package com.redhat.cloud.notifications.routers;
+
+import com.redhat.cloud.notifications.events.FromCamelHistoryFiller;
+import com.redhat.cloud.notifications.ingress.Action;
+import com.redhat.cloud.notifications.models.event.TestEventHelper;
+import com.redhat.cloud.notifications.routers.endpoints.EndpointTestRequest;
+import io.vertx.core.json.Json;
```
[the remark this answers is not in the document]
MikelAlejoBR: This class is part of the test package of the common module, so it is not accessible by the source classes. Do you think I should move it to the sources so that we can use it in the main classes as well?

engine/src/test/java/com/redhat/cloud/notifications/events/EventConsumerTest.java
```diff
@@ -88,6 +87,10 @@ public class EventConsumerTest {
     @Inject
     MeterRegistry registry;
 
+    @Inject
+    EventConsumer eventConsumer;
+
+
```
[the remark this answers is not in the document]
MikelAlejoBR: Removed!

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
```diff
@@ -0,0 +1,123 @@
...
+    public static boolean isIntegrationTestEvent(final com.redhat.cloud.notifications.models.Event event) {
+        try {
+            final Context context = event.getAction().getContext();
+            final Boolean isTestEventContextFlag = (Boolean) context.getAdditionalProperties().get(TEST_ACTION_CONTEXT_TEST_EVENT);
+
+            return isTestEventContextFlag &&
+                    event.getAction().getBundle().equals(TestEventHelper.TEST_ACTION_BUNDLE) &&
+                    event.getAction().getApplication().equals(TestEventHelper.TEST_ACTION_APPLICATION) &&
+                    event.getAction().getEventType().equals(TestEventHelper.TEST_ACTION_EVENT_TYPE);
+        } catch (final NullPointerException e) {
```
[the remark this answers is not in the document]
MikelAlejoBR: I guess an action should always have a bundle, application or event type, and that we simply reject any actions without this information. I wanted to double check: do I remove it then?

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
MikelAlejoBR: Done!

engine/pom.xml
MikelAlejoBR: No! Removed it!

engine/src/test/java/com/redhat/cloud/notifications/routers/EndpointTestResourceTest.java
MikelAlejoBR: Aaah, gotcha! Thanks!

engine/src/main/java/com/redhat/cloud/notifications/routers/EndpointTestResource.java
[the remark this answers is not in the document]
gwenneg: My bad, you can disregard my comment.

database/src/main/resources/db/migration/V1.71.0__RHCLOUD-22470_insert_test_action_events_bundle_application_event_type.sql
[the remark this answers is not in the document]
gwenneg: Sure, that's strictly equivalent from a Postgres query plan perspective.

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
MikelAlejoBR: Yeah, that's way better! Done!

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -658,6 +668,37 @@ public boolean unsubscribeEmail(
...
+    @Parameters({
+        @Parameter(
+                name = "uuid",
+                in = ParameterIn.PATH,
+                description = "The UUID of the endpoint to test",
+                schema = @Schema(type = SchemaType.STRING)
+            )
+    })
+    @Produces(APPLICATION_JSON)
+    @RolesAllowed(ConsoleIdentityProvider.RBAC_READ_INTEGRATIONS_ENDPOINTS)
```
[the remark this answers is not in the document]
gwenneg: Does it make sense for a user with read permission to test an integration that they didn't create and they can't modify? In case the integration does not work, they don't have any way to know who has the permission to fix it.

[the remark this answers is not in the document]
gwenneg: Does it make sense for a user with read permission to test an integration that they didn't create and they can't modify? In case the integration does not work, they don't have any way to know who has the permission to fix it. Based on that, I would change the permission:
```suggestion
    @RolesAllowed(ConsoleIdentityProvider.RBAC_WRITE_INTEGRATIONS_ENDPOINTS)
```

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
gwenneg: Actually, we don't even need to specify the version in this class. You can simply remove it.

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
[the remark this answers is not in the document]
gwenneg: The bundle, application and event type fields are mandatory in the incoming `Action` indeed. If any of them is missing, we will reject the Kafka message from the `ingress` topic. So there's no need to catch a NPE here.

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
```diff
@@ -0,0 +1,116 @@
...
+         */
+        Event testEvent = new Event();
+
+        Metadata metadata = new Metadata();
+        metadata.setAdditionalProperty(TEST_ACTION_METADATA_KEY, TEST_ACTION_METADATA_VALUE);
+
+        Payload payload = new Payload();
+        payload.setAdditionalProperty(TEST_ACTION_PAYLOAD_KEY, TEST_ACTION_PAYLOAD_VALUE);
+
+        testEvent.setMetadata(metadata);
```
gwenneg, line 66: I don't think we need metadata in the test action.
```suggestion
```

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
```diff
@@ -0,0 +1,116 @@
...
+         */
+        testAction.setApplication(TEST_ACTION_APPLICATION);
+        testAction.setBundle(TEST_ACTION_BUNDLE);
+        testAction.setEvents(List.of(testEvent));
+        testAction.setEventType(TEST_ACTION_EVENT_TYPE);
+        testAction.setId(UUID.randomUUID());
+        testAction.setRecipients(List.of(recipient));
+        testAction.setOrgId(orgId);
+        testAction.setTimestamp(LocalDateTime.now(Clock.systemUTC()));
+        testAction.setVersion(TEST_ACTION_VERSION);
```
gwenneg, line 83: We can omit the version. The latest version will be automatically used.
```suggestion
```

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
```diff
@@ -0,0 +1,116 @@
...
+ * The purpose of this class is to help creating test events for testing the
+ * integrations that the clients have set up. The used constants are also
+ * exposed for any required checks or tests that might need them. For more
+ * information, take a look at <a href="https://issues.redhat.com/browse/RHCLOUD-22470">RHCLOUD-22470</a>.
+ */
+public class TestEventHelper {
+    /**
+     * Regular action test data.
+     */
+    public static final String TEST_ACTION_CONTEXT_ENDPOINT_ID = "test-action-context-endpoint-id";
```
gwenneg, line 26: This will be seen by the org admin, let's use something a bit more user-friendly:
```suggestion
    public static final String TEST_ACTION_CONTEXT_ENDPOINT_ID = "integration-id";
```

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
```diff
@@ -0,0 +1,116 @@
...
+
+        /*
+         * Set the rest of the action's members.
+         */
+        testAction.setApplication(TEST_ACTION_APPLICATION);
+        testAction.setBundle(TEST_ACTION_BUNDLE);
+        testAction.setEvents(List.of(testEvent));
+        testAction.setEventType(TEST_ACTION_EVENT_TYPE);
+        testAction.setId(UUID.randomUUID());
+        testAction.setRecipients(List.of(recipient));
```
gwenneg, line 80: We don't need recipients in this action as it is not meant to test email integations.
```suggestion
```

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
```diff
@@ -0,0 +1,116 @@
...
+     */
+    public static final String TEST_ACTION_CONTEXT_ENDPOINT_ID = "test-action-context-endpoint-id";
+    public static final String TEST_ACTION_METADATA_KEY = "test-metadata-key";
+    public static final String TEST_ACTION_METADATA_VALUE = "test-metadata-value";
+    public static final String TEST_ACTION_PAYLOAD_KEY = "message";
+    public static final String TEST_ACTION_PAYLOAD_VALUE = "Congratulations! The integration you created on https://console.redhat.com was successfully tested!";
+    public static final String TEST_ACTION_RECIPIENT = "test-recipient-1";
+    public static final String TEST_ACTION_VERSION = "0.0.0";
+    /**
+     * The test action or event will be sent with a defined "console" bundle, the "notifications" application and
```
gwenneg, line 34:
```suggestion
     * The test action or event will be sent with a defined "console" bundle, the "integrations" application and
```

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
```diff
@@ -0,0 +1,116 @@
...
+     * a new event type that has been inserted in the V1.71.0 migration.
+     */
+    public static final String TEST_ACTION_BUNDLE = "console";
+    public static final String TEST_ACTION_APPLICATION = "integrations";
+    public static final String TEST_ACTION_EVENT_TYPE = "integration-test";
+
+    /**
+     * Creates a test action ready to be sent to the engine. It sets the endpoint's UUID and a flag in the context, so
+     * that it can be easily identified.
```
gwenneg, line 43:
```suggestion
     * Creates a test action ready to be sent to the engine. It sets the endpoint's UUID in the context.
```

engine/src/test/java/com/redhat/cloud/notifications/routers/EndpointTestResourceTest.java
```diff
@@ -0,0 +1,131 @@
...
+        );
+
+        final var actionsList = actionsOut.received();
+
+        // Only one test action should have been sent to Kafka.
+        final var expectedActionsCount = 1;
+        Assertions.assertEquals(expectedActionsCount, actionsList.size(), "unexpected number of actions sent to Kafka");
+
+        final String kafkaActionRaw = actionsList.get(0).getPayload();
+        final Action kafkaAction = Json.decodeValue(kafkaActionRaw, Action.class);
```
gwenneg, line 77:
```suggestion
        final Action kafkaAction = Parser.decode(kafkaActionRaw);
```

common/src/test/java/com/redhat/cloud/notifications/models/event/TestEventHelperTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+        Assertions.assertTrue(TestEventHelper.isIntegrationTestEvent(testEvent), "the test event was not identified as such");
+    }
+
+    /**
+     * Tests that when an event has no context, the function under test
+     * correctly identifies it as a normal event.
+     */
+    @Test
+    public void testIsIntegrationEventNoContext() {
```
gwenneg, line 99: The test will fail because the B/A/ET do not match the expectations from the tested method, not because the event has no context.

common/src/test/java/com/redhat/cloud/notifications/models/event/TestEventHelperTest.java
```diff
@@ -0,0 +1,197 @@
...
+        final Context context = new Context();
+        context.setAdditionalProperty("test-property", "one-two-three");
+
+        nonTestAction.setContext(context);
+
+        final var nonTestEvent = new com.redhat.cloud.notifications.models.Event();
+        nonTestEvent.setAction(nonTestAction);
+
+        Assertions.assertFalse(TestEventHelper.isIntegrationTestEvent(nonTestEvent), "the event should not have been identified as a test event");
+    }
```
gwenneg, line 125: This is probably no longer needed.

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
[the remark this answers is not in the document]
gwenneg: Some tests will cause a NPE, only because they do not test something that can happen in real life. There's an easy way to work around that without catching the NPE:
```java
            return TestEventHelper.TEST_ACTION_BUNDLE.equals(event.getAction().getBundle()) &&
                    TestEventHelper.TEST_ACTION_APPLICATION.equals(event.getAction().getApplication()) &&
                    TestEventHelper.TEST_ACTION_EVENT_TYPE.equals(event.getAction().getEventType());
```

common/src/test/java/com/redhat/cloud/notifications/models/event/TestEventHelperTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+        final Action testAction = TestEventHelper.createTestAction(endpointUuid, "random-org-id");
+
+        // Convert the action to JSON and back to simulate the event going
+        // through Kafka. If not, some additional properties of the context are
+        // not serialized as String, and won't match all the types and the way
+        // they get serialized when sent via Kafka and received via Kafka as
+        // well.
+        final String jsonAction = Json.encode(testAction);
+        final Action rawAction = Json.decodeValue(jsonAction, Action.class);
```
gwenneg, line 188:
```suggestion
        final String jsonAction = Parser.encode(testAction);
        final Action rawAction = Parser.decode(jsonAction);
```

engine/src/test/java/com/redhat/cloud/notifications/events/EndpointProcessorTest.java
```diff
@@ -0,0 +1,108 @@
...
+        // Create the action with the endpoint reference and then a corresponding event for the processor.
+        final Action testAction = TestEventHelper.createTestAction(endpointUuid, orgId);
+
+        // Convert the action to JSON and back to simulate the event going
+        // through Kafka. If not, some additional properties of the context are
+        // not serialized as String, and won't match all the types and the way
+        // they get serialized when sent via Kafka and received via Kafka as
+        // well.
+        final String jsonAction = Json.encode(testAction);
+        final Action rawAction = Json.decodeValue(jsonAction, Action.class);
```
gwenneg, line 65:
```suggestion
        final String jsonAction = Parser.encode(testAction);
        final Action rawAction = Parser.decode(jsonAction);
```

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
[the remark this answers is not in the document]
MikelAlejoBR: It makes a lot of sense! Changed!

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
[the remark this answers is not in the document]
MikelAlejoBR: Thank you so much! Changed!

engine/src/main/java/com/redhat/cloud/notifications/routers/EndpointTestResource.java
```diff
@@ -0,0 +1,45 @@
...
+    @Consumes(APPLICATION_JSON)
+    @Path("/test")
+    @POST
+    public void testEndpoint(@Valid final EndpointTestRequest endpointTestRequest) {
+        final Action testAction = TestEventHelper.createTestAction(
+            endpointTestRequest.endpointUuid,
+            endpointTestRequest.orgId
+        );
+
+        this.eventEmitter.send(Json.encode(testAction));
```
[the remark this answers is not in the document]
MikelAlejoBR: Done!

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -160,6 +160,41 @@ public boolean incrementEndpointServerErrors(UUID endpointId, int maxServerError
...
+            .getCurrentSession()
+            .createQuery(query, Endpoint.class)
+            .setParameter("uuid", endpointUuid)
+            .setParameter("orgId", orgId)
+            .uniqueResult();
+
+        // Throw a no result exception to avoid a null pointer exception from
+        // the "List.of".
+        if (endpoint == null) {
+            throw new NoResultException();
```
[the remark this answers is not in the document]
MikelAlejoBR: Done as well!

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
MikelAlejoBR: Done!

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
MikelAlejoBR: Removed!

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
MikelAlejoBR: Renamed! I changed it to `integration-uuid` to be more precise, if that's okay!

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
MikelAlejoBR: Removed!

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
MikelAlejoBR: Refactored!

common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java
MikelAlejoBR: This too!

engine/src/test/java/com/redhat/cloud/notifications/routers/EndpointTestResourceTest.java
MikelAlejoBR: Done!

common/src/test/java/com/redhat/cloud/notifications/models/event/TestEventHelperTest.java
MikelAlejoBR: You are right. I have removed the test, since it doesn't make any sense to have it around...

common/src/test/java/com/redhat/cloud/notifications/models/event/TestEventHelperTest.java
MikelAlejoBR: Removed as well!

common/src/test/java/com/redhat/cloud/notifications/models/event/TestEventHelperTest.java
MikelAlejoBR: Changed!

engine/src/test/java/com/redhat/cloud/notifications/events/EndpointProcessorTest.java
MikelAlejoBR: Done as well!

gwenneg (approved):
LGTM

PR #1603, opened 2022-11-15
[RHCLOUD-22065] fix: reference to the Sources endpoint in the properties

MikelAlejoBR:
We need to reference the ClowdApp name plus the deployment name for it to work.

## Links

[RHCLOUD-22065](https://issues.redhat.com/browse/RHCLOUD-22065)


gwenneg (approved):
LGTM

PR #1604, opened 2022-11-15
[RHCLOUD-22065] Feature sources processors

MikelAlejoBR:
Basically:

* Moves the `SecretUtils` and `SourcesService` classes to the common package, so that the engine can use them.
* The test `SecretUtilsTest` is kept in the `backend` module, because apparently it requires a datasource set in Quarkus.
* Fetches the secrets from Sources if the feature is enabled.

## Links


engine/src/main/resources/application.properties
```diff
@@ -131,3 +131,7 @@ quarkus.cache.caffeine.from-open-bridge-history-filler.expire-after-write=PT24H
 
 # Use this property to load the templates from the DB. Temp, to be removed soon.
 notifications.use-templates-from-db=false
+
+# Sources integration URLs and details. It is used to store the secrets' data for the camel and webhook endpoints.
+quarkus.rest-client.sources.read-timeout=1000
+quarkus.rest-client.sources.url=${clowder.endpoints.sources-api:http://localhost:8000}
```
gwenneg, line 137:
```suggestion
quarkus.rest-client.sources.url=${clowder.endpoints.sources-api-svc:http://localhost:8000}
```

common/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretsUtilsTestProfile.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Since the "common" module doesn't contain an "application.properties" file,
+ * we are feeding the {@link SecretUtilsTest} tests with some default
+ * configuration properties.
+ */
+public class SecretsUtilsTestProfile implements QuarkusTestProfile {
```
gwenneg, line 13: Why did you introduce H2 here? That's a great DBMS for testing but since we depend on Postgres on all envs, I'd stick with Postgres for testing as well, unless H2 provides something important Postgres doesn't.

engine/src/main/resources/application.properties

common/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretsUtilsTestProfile.java
MikelAlejoBR: Basically Quarkus required me to set up a database connection for that test, and since the test doesn't use any database, I thought of just using H2 for this. I've changed it so that it uses pgsql instead.

gwenneg: I think we'll need something similar to [TestLifecycleManager](https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/test/java/com/redhat/cloud/notifications/TestLifecycleManager.java) in this module.

gwenneg: Let's deal with that in two steps:
- First, could you please move `SecretUtilsTest` to notifications-backend or engine? We'll merge this PR after that.
- Then in a subsequent PR, let's try to move the class back to this module.

From a feature perspective, your dev is finished so let's not delay its deployment on stage/prod 😃

MikelAlejoBR: The problem with moving this stuff back to the `backend` module is that the `engine` would not be able to use the `SecretUtils`, and I'd have to duplicate the code for that. That's the reason why I decided to move the `SecretUtils` to the `common` package, so I could define everything there and use it both in the `backend` and the `engine`...

gwenneg: I'm only asking to move the test back to `backend`. Everything else can stay in `common` 😃

MikelAlejoBR: I created a `TestProfile` instead of using a `TestLifecycleManager` because no other test in the `common` package seems to need a database... And even the `SecretUtilsTest` class requests the database just because. So I thought that I would simply create that profile for that test.

gwenneg: You did that for a good reason which I understand completely.

However, testing `common` in a better way is a topic larger than this PR that has been at the bottom of my todo list for a while. There are other tests in `backend` or `engine` that should ideally be in `common`. I'd simply like to take care of that later (maybe during a tech debt sprint) instead of adding a profile in this PR that doesn't do anything except keeping Quarkus happy.

MikelAlejoBR: Makes total sense. I've moved back the file to the `backend` module!

common/pom.xml
```diff
@@ -69,6 +73,15 @@
...
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5-mockito</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jdbc-postgresql</artifactId>
+        </dependency>
```
gwenneg, line 84:
```suggestion
```

common/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretsUtilsTestProfile.java
gwenneg: Thanks!

common/pom.xml
MikelAlejoBR: Done!

gwenneg (approved):
LGTM

gwenneg:

PR #1607, opened 2022-11-16
RHCLOUD-22955 Add OpenTelemetry support for backend.

pilhuhn:


backend/pom.xml
```diff
@@ -85,6 +85,27 @@
...
+            <artifactId>opentelemetry-extension-trace-propagators</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry.instrumentation</groupId>
+            <artifactId>opentelemetry-jdbc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-extension-annotations</artifactId>
+        </dependency>
```
gwenneg, line 107: It looks like we'll need to revisit these dependencies with Quarkus `2.14.0.Final` as they moved a few things. For instance, `quarkus-opentelemetry-exporter-otlp` is no longer needed (and does not exist anymore).

.rhcicd/clowdapp-backend.yaml
```diff
@@ -318,6 +324,12 @@ parameters:
 - name: SENTRY_ENABLED
   description: Enable Sentry (or not)
   value: "false"
+- name: QUARKUS_OPENTELEMETRY_ENABLED
+  value: "false"
+- name: QUARKUS_OPENTELEMETRY_ENDPOINT
+  value: "localhost:4317"
+- name: QUARKUS_DATASOURCE_JDBC_DRIVER
+  value: jdbc:postgresql://127.0.0.1:5432/notifications
```
gwenneg, line 332: The param name mentions a driver while the value is a connection string. Which one do you want to configure here?

backend/src/main/resources/application.properties
```diff
@@ -80,3 +82,7 @@ quarkus.scheduler.metrics.enabled=true
 # Sources integration URLs and details. It is used to store the secrets' data for the camel and webhook endpoints.
 quarkus.rest-client.sources.read-timeout=1000
 quarkus.rest-client.sources.url=${clowder.endpoints.sources-api-svc:http://localhost:8000}
+
+# OpenTelemetry -- see also jdbc driver above.
+quarkus.opentelemetry.enabled=true
```
gwenneg, line 87:
```suggestion
quarkus.opentelemetry.enabled=false
```

.rhcicd/clowdapp-backend.yaml
pilhuhn: This is what I want to ask you. I even believe that the Driver could always be the otel one, even if otel is disabled. It will then just be a thin wrapper around the real code.

gwenneg: I agree that we can permanently use the otel driver and connection string (even if otel is disabled) as long as it looks stable on stage.

gwenneg:
`master` now depends on Quarkus 2.14.1.Final.

backend/src/main/resources/application.properties
```diff
@@ -8,7 +8,9 @@ quarkus.http.port=8085
 quarkus.datasource.db-kind=postgresql
 quarkus.datasource.username=postgres
 quarkus.datasource.password=postgres
-quarkus.datasource.jdbc.url=jdbc:postgresql://127.0.0.1:5432/notifications
+quarkus.datasource.jdbc.url=jdbc:otel:postgresql://127.0.0.1:5432/notifications
```
gwenneg, line 11: This won't work with [clowder-quarkus-config-source](https://github.com/RedHatInsights/clowder-quarkus-config-source) out of the box. I'll prepare a PR there to retrieve the DB config without introducing a breaking change.

backend/src/main/resources/application.properties
```diff
@@ -80,3 +82,7 @@ quarkus.scheduler.metrics.enabled=true
 # Sources integration URLs and details. It is used to store the secrets' data for the camel and webhook endpoints.
 quarkus.rest-client.sources.read-timeout=1000
 quarkus.rest-client.sources.url=${clowder.endpoints.sources-api-svc:http://localhost:8000}
+
+# OpenTelemetry -- see also jdbc driver above.
+quarkus.opentelemetry.enabled=false
+quarkus.opentelemetry.tracer.exporter.otlp.endpoint=http://localhost:4317
```
gwenneg, line 88: Will the OTLP host and port be provided by Clowder eventually?

backend/src/main/resources/application.properties
gwenneg: https://github.com/RedHatInsights/clowder-quarkus-config-source/pull/118

.rhcicd/clowdapp-backend.yaml
```diff
@@ -320,6 +324,10 @@ parameters:
 - name: SENTRY_ENABLED
   description: Enable Sentry (or not)
   value: "false"
+- name: QUARKUS_OPENTELEMETRY_ENABLED
+  value: "false"
+- name: QUARKUS_OPENTELEMETRY_ENDPOINT
+  value: "localhost:4317"
```
gwenneg, line 330: I'm not sure if that's mandatory:
```suggestion
  value: "http://localhost:4317"
```

backend/src/main/resources/application.properties
pilhuhn: Yes, eventually.

backend/pom.xml
```diff
@@ -86,6 +86,23 @@
...
```
gwenneg, line 104: This is already a transitive dependency coming from `quarkus-opentelemetry`.
```suggestion
```

gwenneg (approved):
LGTM

PR #1612, opened 2022-11-18
RHCLOUD-22474 Add resources to manage daily digest hour per org id

g-duval:


.rhcicd/clowdapp-aggregator.yaml
```diff
@@ -61,6 +61,8 @@ objects:
           value: ${ENV_NAME}
         - name: PROMETHEUS_PUSHGATEWAY_URL
           value: ${PROMETHEUS_PUSHGATEWAY}
+        - name: NOTIFICATIONS_AGGREGATOR_ACCORDING_ORG_PREF
```
gwenneg, line 64: Comment about the entire `cronjob` object in this file: I think we'll need to duplicate that `cronjob`. The old one will remain unchanged while the new one will have a different schedule and be disabled by default from the env vars / app-interface configuration. Then, when we'll be ready to switch from the old to the new cronjob, we'll do that from app-interface and we'll enable the feature flag at the same time.

aggregator/src/main/java/com/redhat/cloud/notifications/db/AggregationCronjobParameterRepository.java
```diff
@@ -0,0 +1,30 @@
...
+@ApplicationScoped
+public class AggregationCronjobParameterRepository {
+
+    @Inject
+    Session session;
+
+    @Inject
+    EntityManager entityManager;
+
+    public List<AggregationCronjobParameters> getOrdIdToProceed(LocalDateTime now) {
```
gwenneg, line 22:
```suggestion
    public List<AggregationCronjobParameters> getOrgIdToProceed(LocalDateTime now) {
```

aggregator/src/main/java/com/redhat/cloud/notifications/db/AggregationCronjobParameterRepository.java
```diff
@@ -0,0 +1,30 @@
...
+        LocalTime currentHour = LocalTime.of(now.getHour(), 0, 0);
+        String query = "SELECT acp FROM AggregationCronjobParameters acp WHERE acp.expectedRunningTime = :currentHour";
```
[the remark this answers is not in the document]
gwenneg: We'll need to talk with the team about outages: if `aggregator` is down when it's time for a specific org to be aggregated, then we should probably run the aggregation process when `aggregator` is back online even if the current hour doesn't match the settings anymore.

common-aggregator/src/main/java/com/redhat/cloud/notifications/models/AggregationCronjobParameters.java
```diff
@@ -0,0 +1,88 @@
...
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.util.Objects;
+
+import static com.fasterxml.jackson.annotation.JsonFormat.Shape.STRING;
+import static com.fasterxml.jackson.databind.PropertyNamingStrategies.SnakeCaseStrategy;
+
+@Entity
+@Table(name = "aggregation_cronjob_parameter")
+@JsonNaming(SnakeCaseStrategy.class)
```
gwenneg, line 20: You will find many JPA entities with Jackson annotations in the code because we chose a long time ago to use a single layer of classes for the DB and the REST API. However, I've never been a big fan of that decision because it makes everything more complicated when we have to implement something, especially when it involves multiple Maven modules. For this specific case, I'd prefer if we had the JPA entity separated from the POJO that will be used for the REST API. The entity would stay in `common-aggregator` because it is need in `backend` and `aggregator` while the POJO for the REST API would only be in `backend`.

aggregator/src/main/java/com/redhat/cloud/notifications/db/AggregationCronjobParameterRepository.java
```diff
@@ -0,0 +1,30 @@
...
+
```
gwenneg, line 25: What about the old behavior that we need to preserve? If the current hour is midnight UTC, all orgs that did not customize the aggregator execution time should also be included in the list of orgs to aggregate. Currently "midnight UTC" is configured from the `cronjob` scheduling in the YAML file but we could move that scheduling to the app and use a `@ConfigProperty` field to use it here.

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
```diff
@@ -15,29 +17,55 @@
...
+
+    private static final String DEFAULT_ORG_ID = "NA";
+
+    public List<EmailAggregationKey> getApplicationsWithPendingAggregation(String orgIdToAggregate, LocalDateTime start, LocalDateTime end) {
+        String orgIdCriteria = "ea.org_id = :orgId AND acj.org_id IS NOT NULL";
+        if (DEFAULT_ORG_ID.equals(orgIdToAggregate)) {
+            orgIdCriteria = "acj.org_id IS NULL";
+        }
+        String query = "SELECT DISTINCT ea.org_id, ea.bundle, ea.application FROM email_aggregation ea "
+                + "LEFT JOIN aggregation_cronjob_parameter acj ON ea.org_id = acj.org_id "
```
gwenneg, line 30: Is it really necessary to `JOIN` here? We already have the `org_id` value when the method is invoked and we do not select anything from `aggregation_cronjob_parameter` in the query.

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
```diff
@@ -15,29 +17,55 @@
...
+                + " WHERE " + orgIdCriteria + " AND ea.created > :start AND ea.created <= :end";
```
gwenneg, line 31: Wouldn't this work?
```suggestion
                + " WHERE ea.org_id = :orgId AND ea.created > :start AND ea.created <= :end";
```
and then below
```
.setParameter("orgId", orgIdToAggregate)
```

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
```diff
@@ -15,29 +17,55 @@
 public class EmailAggregationResources {
 
     @Inject
-    Session session;
+    EntityManager entityManager;
+
+    private static final String DEFAULT_ORG_ID = "NA";
```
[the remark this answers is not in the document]
gwenneg: Okay I just found the answer in the SQL script 😃

aggregator/src/main/java/com/redhat/cloud/notifications/db/AggregationCronjobParameterRepository.java
gwenneg: I just understood that the SQL scripts help preserving the old behavior.

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
gwenneg: Please ignore that comment for now.

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationResources.java
gwenneg: Please ignore that comment for now.

common-aggregator/src/main/java/com/redhat/cloud/notifications/models/AggregationCronjobParameters.java
g-duval: Ok

aggregator/src/main/java/com/redhat/cloud/notifications/db/AggregationCronjobParameterRepository.java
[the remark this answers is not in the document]
gwenneg: > Well I prefer full version including SELECT :).

👍

aggregator/src/main/java/com/redhat/cloud/notifications/db/AggregationCronjobParameterRepository.java
```diff
@@ -0,0 +1,30 @@
...
+public class AggregationCronjobParameterRepository {
+
+    @Inject
+    Session session;
+
+    @Inject
+    EntityManager entityManager;
+
+    public List<AggregationCronjobParameters> getOrdIdToProceed(LocalDateTime now) {
+        LocalTime currentHour = LocalTime.of(now.getHour(), 0, 0);
```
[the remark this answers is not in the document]
gwenneg: As discussed in a meeting, the current code is good enough for the first version. We may revisit that later.

backend/src/main/java/com/redhat/cloud/notifications/routers/OrgConfigResource.java
```diff
@@ -0,0 +1,72 @@
...
+    public Response getDailyDigestTimePreference(@Context SecurityContext sec) {
+        String orgId = getOrgId(sec);
+        Log.infof("Delete daily digest time preference for orgId %s", orgId);
+        AggregationCronjobParameters storedParameters = aggregationCronjobParametersRepository.findJobAggregationCronjobParameters(orgId);
+        if (null != storedParameters) {
+            DailyDigestTimeSettings dailyDigestTimePreference = new DailyDigestTimeSettings();
+            dailyDigestTimePreference.dailyDigestTimePreference = storedParameters.getExpectedRunningTime().format(DateTimeFormatter.ofPattern("HH:mm:ss"));
+            return Response.ok(dailyDigestTimePreference).build();
+        } else {
+            return Response.status(Response.Status.NOT_FOUND).build();
```
gwenneg, line 69: Instead of returning a 404, we should initialize the ACP value just like what you did in the cronjob code.

backend/src/main/java/com/redhat/cloud/notifications/routers/OrgConfigResource.java
```diff
@@ -0,0 +1,72 @@
...
+    @Produces(APPLICATION_JSON)
+    @RolesAllowed(ConsoleIdentityProvider.RBAC_READ_NOTIFICATIONS)
+    public Response getDailyDigestTimePreference(@Context SecurityContext sec) {
+        String orgId = getOrgId(sec);
+        Log.infof("Delete daily digest time preference for orgId %s", orgId);
+        AggregationCronjobParameters storedParameters = aggregationCronjobParametersRepository.findJobAggregationCronjobParameters(orgId);
+        if (null != storedParameters) {
+            DailyDigestTimeSettings dailyDigestTimePreference = new DailyDigestTimeSettings();
+            dailyDigestTimePreference.dailyDigestTimePreference = storedParameters.getExpectedRunningTime().format(DateTimeFormatter.ofPattern("HH:mm:ss"));
+            return Response.ok(dailyDigestTimePreference).build();
```
gwenneg, line 67: Maybe we could return a `LocalTime` object directly?

backend/src/main/java/com/redhat/cloud/notifications/routers/OrgConfigResource.java
```diff
@@ -0,0 +1,66 @@
...
+    @Inject
+    AggregationOrgConfigRepository aggregationOrgConfigRepository;
+
+    @PUT
+    @Path("/daily-digest/time-preference")
+    @Consumes(TEXT_PLAIN)
+    @Produces(TEXT_PLAIN)
+    @Transactional
+    @RolesAllowed(ConsoleIdentityProvider.RBAC_WRITE_NOTIFICATIONS)
+    public Response saveDailyDigestTimePreference(@Context SecurityContext sec, String dailyDigestTimePreference) {
```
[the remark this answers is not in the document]
g-duval: works better indeed

backend/src/main/java/com/redhat/cloud/notifications/routers/OrgConfigResource.java
```diff
@@ -0,0 +1,63 @@
...
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.SecurityContext;
+import java.time.LocalTime;
+
+import static com.redhat.cloud.notifications.routers.SecurityContextUtil.getOrgId;
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+@Path(Constants.API_NOTIFICATIONS_V_1_0 + "/org-config")
+public class OrgConfigResource {
```
gwenneg, line 28: @josejulio This class contains new REST endpoints that will be used to store/retrieve the aggregation execution time for each org. I'm not sure if we should include these endpoints in the OpenAPI spec or hide them. WDYT?

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationRepository.java
```diff
@@ -0,0 +1,85 @@
...
+                .executeUpdate();
+    }
+
+    @Transactional
+    public void updateLastCronJobRunAccordingOrgPref(LocalDateTime end) {
+
+        String hqlQuery = "UPDATE AggregationOrgConfig ac SET ac.lastRun=:end WHERE ac.orgId IN " +
+            "(SELECT DISTINCT ea.orgId FROM EmailAggregation ea, AggregationOrgConfig acp WHERE " +
+            "ea.orgId = acp.orgId AND ea.created > acp.lastRun AND ea.created <= :end " +
+            "AND hour(acp.scheduledExecutionTime) = :runningHour)";
```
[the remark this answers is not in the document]
gwenneg: We talked about this on Slack. The current query is fine for now.

backend/src/main/java/com/redhat/cloud/notifications/routers/OrgConfigResource.java
gwenneg: I'm merging before your answer. We'll add the annotations if needed afterwards.

gwenneg (approved):
LGTM

PR #1614, opened 2022-11-22
[RHCLOUD-22065] Sources logs tweaks

MikelAlejoBR:
Basically:

* Includes the created secret ID for all the log statements that make sense to have it.
* Puts the "the secret was not created due to null/blank objects" log statement in the "debug" level.
* Removes the endpoint's IDs when creating the secrets, because at that point the endpoints are not yet persisted, and their IDs are null.

## Links

### DIFF
--- a/common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
+++ b/common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
@@ -106,7 +106,7 @@ public class SecretUtils {
             if (basicAuthId != null) {
                 if (this.isBasicAuthNullOrBlank(basicAuth)) {
                     this.sourcesService.delete(basicAuthId);
-                    Log.infof("[endpoint_id: %s] Basic authentication secret deleted in Sources during an endpoint update operation", endpoint.getId());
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
```diff
@@ -74,7 +74,7 @@ public void createSecretsForEndpoint(Endpoint endpoint) {
             if (secretToken != null && !secretToken.isBlank()) {
                 final long id = this.createSecretTokenSecret(secretToken);
 
-                Log.infof("[endpoint_id: %s][secret_id: %s] Secret token secret created in Sources", endpoint.getId(), id);
+                Log.infof("[endpoint_id: %s] Secret token secret created in Sources", id);
```
gwenneg, line 77:
```suggestion
                Log.infof("[secret_id: %s] Secret token secret created in Sources", id);
```

MikelAlejoBR: Nice catch! Done!

gwenneg (approved):
LGTM

PR #1616, opened 2022-11-24
[RHCLOUD-22361] Feature: validate query parameters

MikelAlejoBR:
Validates that the "limit", "offset" and "pageNumber" parameters have valid values, and returns a 400 if they don't.

## Links

[[RHCLOUD-22361]](https://issues.redhat.com/browse/RHCLOUD-22361)

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/EventResource.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/EventResource.java
@@ -18,6 +18,7 @@ import org.jboss.resteasy.reactive.RestQuery;
 
 import javax.annotation.security.RolesAllowed;
 import javax.inject.Inject;
+import javax.validation.Valid;
 import javax.ws.rs.BadRequestException;
 import javax.ws.rs.BeanParam;
 import javax.ws.rs.GET;
@@ -59,12 +60,8 @@ public class EventResource {
                                          @RestQuery String eventTypeDisplayName, @RestQuery LocalDate startDate, @RestQuery LocalDate endDate,
                                          @RestQuery Set<String> endpointTypes, @RestQuery Set<Boolean> invocationResults,
                                          @RestQuery Set<EventLogEntryActionStatus> status,
-                                         @BeanParam Query query,
+                                         @BeanParam @Valid Query query,
                                          @RestQuery boolean includeDetails, @RestQuery boolean includePayload, @RestQuery boolean includeActions) {
-        if (query.getLimit().getLimit() < 1 || query.getLimit().getLimit() > 200) {
-            throw new BadRequestException("Invalid 'limit' query parameter, its value must be between 1 and 200");
-        }
-
         Set<EndpointType> basicTypes = Collections.emptySet();
         Set<CompositeEndpointType> compositeTypes = Collections.emptySet();
         Set<NotificationStatus> notificationStatusSet = status == null ? Set.of() : toNotificationStatus(status);
--- a/backend/src/test/java/com/redhat/cloud/notifications/db/QueryTest.java
+++ b/backend/src/test/java/com/redhat/cloud/notifications/db/QueryTest.java
@@ -1,19 +1,25 @@
 package com.redhat.cloud.notifications.db;
 
+import com.redhat.cloud.notifications.MockServerConfig;
+import com.redhat.cloud.notifications.TestHelpers;
 import io.quarkus.test.junit.QuarkusTest;
+import io.restassured.http.ContentType;
+import io.restassured.http.Header;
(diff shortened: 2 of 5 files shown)


backend/src/test/java/com/redhat/cloud/notifications/db/QueryTest.java
```diff
@@ -78,4 +105,295 @@ public void testSort() {
...
+
+        // Get the annotation's values from the "offset" field of the "Query" class.
+        final Field offsetField = Query.class.getDeclaredField("offset");
+        final Min offsetMinAnnotation = offsetField.getAnnotation(Min.class);
+        final long minQueryOffsetValue = offsetMinAnnotation.value();
+
+        // Get a random number that will be used for the validation. The number must be between the "min" and "max"
+        // values of these annotations, so make sure we generate one that satisfies that constraint.
+        final int minRand = (int) Math.min(minQueryOffsetValue, Math.min(minQueryPageNumberValue, minQueryPageSizeValue));
+        final int validValue = ThreadLocalRandom.current().nextInt(minRand, (int) maxQueryPageSizeValue);
```
josejulio, line 137: Could we use a fixed (or a set of) number instead of randomly generating one? Will make the test easier to read. You test the edges and some numbers in between

backend/src/test/java/com/redhat/cloud/notifications/db/QueryTest.java
```diff
@@ -78,4 +105,295 @@ public void testSort() {
...
+
+        final Query query = new Query();
+
+        // Set the fields as accessible to be able to set a value.
+        offsetField.setAccessible(true);
+        pageNumberField.setAccessible(true);
+        pageSizeField.setAccessible(true);
```
josejulio, line 144: You can instead remove the `private` modifier and use the `default` one as we do with `sort_by` https://github.com/RedHatInsights/notifications-backend/blob/5784bb04af4c9b4060b274ac1140e46643c1f775/backend/src/main/java/com/redhat/cloud/notifications/db/Query.java#L34

MikelAlejoBR: All right, I wasn't sure if that was okay or not... Modified! :smiley:

backend/src/test/java/com/redhat/cloud/notifications/db/QueryTest.java
MikelAlejoBR: Sure, I've added a fixed number instead of a randomly generated one. My idea was to not ever think about these tests again if we change the collection limits, but to be honest I don't think we will ever change them, so... Changed!

josejulio (approved):
lgtm

josejulio:
@MikelAlejoBR could you rebase? thanks!

MikelAlejoBR:
> @MikelAlejoBR could you rebase? thanks!

Done!

backend/src/main/java/com/redhat/cloud/notifications/routers/EventResource.java
```diff
@@ -59,12 +60,8 @@ public Page<EventLogEntry> getEvents(@Context SecurityContext securityContext, @
                                          @RestQuery String eventTypeDisplayName, @RestQuery LocalDate startDate, @RestQuery LocalDate endDate,
                                          @RestQuery Set<String> endpointTypes, @RestQuery Set<Boolean> invocationResults,
                                          @RestQuery Set<EventLogEntryActionStatus> status,
-                                         @BeanParam Query query,
+                                         @BeanParam @Valid Query query,
                                          @RestQuery boolean includeDetails, @RestQuery boolean includePayload, @RestQuery boolean includeActions) {
-        if (query.getLimit().getLimit() < 1 || query.getLimit().getLimit() > 200) {
-            throw new BadRequestException("Invalid 'limit' query parameter, its value must be between 1 and 200");
```
gwenneg, line 65: Please do not lower the max limit for this endpoint. It has to keep returning 200 entries because other ConsoleDot apps depend on it.

gwenneg (changes requested):

backend/src/main/java/com/redhat/cloud/notifications/routers/EventResource.java
MikelAlejoBR: Oh, sorry about that. I will revert the `pageSize` back to 200.

gwenneg (approved):
LGTM

josejulio:
Triggering a rebuild

josejulio:
> warning The PR check failure is unusual and must be investigated before this PR is merged.

Errors are unrelated

PR #1618, opened 2022-11-24
RHCLOUD-22344 Fail requests with unknown sort configuration

josejulio:
As confirmed by sentry - all the unknown columns used in the last 30 days come from the unknown sort tests (things like: `08EEXZP9vB`, `KSvIe2Qm5h`, etc).

There is however one using `endpoint_id` - which is used in  [EndpointResource.getEndpointHistory](https://github.com/RedHatInsights/notifications-backend/blob/f5ece5f791d33ebce78d5e59a4b12ef3399453a8/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java#L570-L575)

https://github.com/RedHatInsights/notifications-backend/blob/f5ece5f791d33ebce78d5e59a4b12ef3399453a8/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java#L570-L575

This is however useless, as all the contents belong to the same endpoint, so adding and marking as deprecated.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/Query.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/Query.java
@@ -1,9 +1,11 @@
 package com.redhat.cloud.notifications.db;
 
 import io.quarkus.logging.Log;
+import io.sentry.Sentry;
 
 import javax.ws.rs.BadRequestException;
 import javax.ws.rs.DefaultValue;
+import javax.ws.rs.InternalServerErrorException;
 import javax.ws.rs.QueryParam;
 import java.util.Map;
 import java.util.Optional;
@@ -158,8 +160,9 @@ public class Query {
         }
 
         if (sortFields == null) {
-            Log.warnf("NOTIF-674 SortFields not set.");
-            sortFields = Map.of();
+            InternalServerErrorException isee = new InternalServerErrorException("SortFields are not set");
+            Sentry.captureException(isee);
+            throw isee;
         }
 
         if (!SORT_BY_PATTERN.matcher(sortBy).matches()) {
@@ -169,7 +172,7 @@ public class Query {
         String[] sortSplit = sortBy.split(":");
         Sort sort = new Sort(sortSplit[0]);
         if (!sortFields.containsKey(sort.sortColumn.toLowerCase())) {
-            Log.warnf("NOTIF-674 Unknown sort field passed: %s", sort.sortColumn);
+            throw new BadRequestException("Unknown sort field specified: " + sort.sortColumn);
         } else {
             sort.sortColumn = sortFields.get(sort.sortColumn.toLowerCase());
         }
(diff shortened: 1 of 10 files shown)


backend/src/main/java/com/redhat/cloud/notifications/db/Query.java
```diff
@@ -158,8 +174,9 @@ public Optional<Sort> getSort() {
         }
 
         if (sortFields == null) {
-            Log.warnf("NOTIF-674 SortFields not set.");
-            sortFields = Map.of();
+            InternalServerErrorException isee = new InternalServerErrorException("SortFields are not set");
+            Sentry.captureException(isee);
```
gwenneg, line 178: Is this needed because RESTEasy prevents the exception from being written in the log? If so, is `Sentry.captureException` still needed if we throw a `RuntimeException` instead of an `InternalServerErrorException`?

josejulio: > Is this needed because RESTEasy prevents the exception from being written in the log?

Yes, that is my understanding.

>  If so, is `Sentry.captureException` still needed if we throw a `RuntimeException` instead of an `InternalServerErrorException`?

I guess that would work - is there any preference over `Sentry.captureException` ?

gwenneg: `Sentry.captureException` is less standard/obvious than throwing a simple `RuntimeException`. I'm okay with using `Sentry.captureException` as long as there's a comment explaining what it does and why we need it 😃

josejulio: Updated it - instead of using `Sentry.capture` I'm simply logging an error.

backend/src/main/java/com/redhat/cloud/notifications/db/Query.java
```diff
@@ -158,8 +174,8 @@ public Optional<Sort> getSort() {
         }
 
         if (sortFields == null) {
-            Log.warnf("NOTIF-674 SortFields not set.");
-            sortFields = Map.of();
+            Log.errorf("Sort fields are not set - this mean that an API is using sorting without specifying what sort values are allowed");
```
gwenneg, line 177: Nitpick:
```suggestion
            Log.error("Sort fields are not set - this means that an API is using sorting without specifying what sort values are allowed");
```

backend/src/main/java/com/redhat/cloud/notifications/db/Query.java
```diff
@@ -1,12 +1,15 @@
 package com.redhat.cloud.notifications.db;
 
 import io.quarkus.logging.Log;
+import io.sentry.Sentry;
```
gwenneg, line 4:
```suggestion
```

josejulio: Always forget to clean my imports :sweat_smile:

gwenneg (approved):
LGTM

gwenneg:

PR #1619, opened 2022-11-25
[RHCLOUD-22065] feature: include "sources-api" as a dependency on the engine

MikelAlejoBR:
I forgot to include "sources-api" as a dependency for the engine, and that is making it complain that it doesn't find the Sources endpoint in the configuration file.

## Links


gwenneg (approved):
LGTM

PR #1624, opened 2022-11-28
[RHCLOUD-22065] fix: NullPointerException when sending blank basic authentications

MikelAlejoBR:
When sending blank passwords or usernames inside the basic authentication objects, the "isBlank" calls triggered a NullPointerException.

## Links

### DIFF
--- a/common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
+++ b/common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
@@ -5,6 +5,7 @@ import com.redhat.cloud.notifications.models.Endpoint;
 import com.redhat.cloud.notifications.models.EndpointProperties;
 import com.redhat.cloud.notifications.models.SourcesSecretable;
 import io.quarkus.logging.Log;
+import org.apache.commons.lang3.StringUtils;
 import org.eclipse.microprofile.rest.client.inject.RestClient;
 
 import javax.enterprise.context.ApplicationScoped;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
```diff
@@ -225,11 +225,29 @@ private long createSecretTokenSecret(final String secretToken) {
      * @param basicAuthentication the object to check.
      * @return {@code true} if the object is null, or if the password and the username are blank.
      */
-    private boolean isBasicAuthNullOrBlank(final BasicAuthentication basicAuthentication) {
+    protected boolean isBasicAuthNullOrBlank(final BasicAuthentication basicAuthentication) {
         if (basicAuthentication == null) {
             return true;
         }
 
-        return basicAuthentication.getUsername().isBlank() && basicAuthentication.getPassword().isBlank();
```
josejulio, line 233: I think we have access to apache's [StringUtils.isBlank](https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/StringUtils.html#isBlank(java.lang.CharSequence))
We could use that one instead.

MikelAlejoBR: Thanks, yeah! I like it more since it makes the check much more simple.

josejulio (approved):
lgtm

PR #1626, opened 2022-11-29
Fix delete endpoint

josejulio:
Looks like RHOSE is not answering with the deleted type and instead returning a 404.
Check if we were deleting an endpoint and delete it if that's the case.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointReadyChecker.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointReadyChecker.java
@@ -92,10 +92,16 @@ public class EndpointReadyChecker {
                     em.remove(ep);
                 }
             } catch (WebApplicationException wae) {
-                String path = "GET " + BASE_PATH + "/{bridgeId}/processors/{processorId}";
-                rhoseErrorMetricsRecorder.record(path, wae);
-                Log.warn("Getting data from OB failed", wae);
-                ep.setStatus(EndpointStatus.FAILED);
+                // 404 means that the processor could not be found - but we were deleting an endpoint.
+                // This means is no longer there and we can finish with the delete process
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointReadyChecker.java
```diff
@@ -92,10 +92,16 @@ public void execute() {
                     em.remove(ep);
                 }
             } catch (WebApplicationException wae) {
-                String path = "GET " + BASE_PATH + "/{bridgeId}/processors/{processorId}";
-                rhoseErrorMetricsRecorder.record(path, wae);
-                Log.warn("Getting data from OB failed", wae);
-                ep.setStatus(EndpointStatus.FAILED);
+                // 404 means that the processor could not be found - but we were deleting an endpoint.
+                // This means is no longer there and we can finish with the delete process
+                if (wae.getResponse().getStatus() == 404 && ep.getStatus().equals(EndpointStatus.DELETING)) {
```
gwenneg, line 97: Did you confirm with the RHOSE team that the 404 is the expected behavior?

josejulio: Not yet - asked them but waiting for an answer.
This is what I'm currently getting on stage.

gwenneg: We should probably wait for their answer before merging this PR then.

josejulio: > if you get 404 it means it does not exist anymore - the resource was deleted successfully

josejulio:
Merging to try on stage

PR #1627, opened 2022-12-01
Fix event type sort field

gwenneg:
This should fix the following bug found by @Mishrasubha while calling `GET /eventTypes` with the `name` sort param:
> ERROR: column reference "name" is ambiguous


josejulio (approved):

gwenneg:

PR #1636, opened 2022-12-07
Initial write-up of Advisor daily digest notification aggregator

PaulWay:
Advisor will produce a daily digest including:

* Number of new recommendations
* Number of resolved recommendations
* Number of deactivated recommendations
* The list of new recommendations, with the count of systems newly affected
* The list of resolved recommendations, with the count of systems previously affected
* The list of deactivated recommendations

This aggregator attempts to gather this information from the events over the last 24 hours.

Signed-off-by: Paul Wayper <paulway@redhat.com>

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
@@ -9,7 +9,6 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
-
 /*
  * Events are of the form:
  * {
@@ -88,36 +87,31 @@ public class AdvisorEmailAggregator extends AbstractEmailPayloadAggregator {
     private static final String EVENT_TYPE_KEY = "event_type";
     private static final String EVENTS_KEY = "events";
     private static final String PAYLOAD_KEY = "payload";
-    private static final String INVENTORY_ID = "inventory_id";
-    private static final String PAYLOAD_RULE_ID = "rule_id";
-    private static final String PAYLOAD_RULE_DESCRIPTION = "rule_description";
-    private static final String PAYLOAD_RULE_TOTAL_RISK = "total_risk";
-    private static final String PAYLOAD_RULE_HAS_INCIDENT = "has_incident";
-    private static final String PAYLOAD_RULE_URL = "rule_url";
 
-    // Advisor event payloads
-    private static final String NEW_RECOMMENDATION = "new_recommendation";
-    private static final String RESOLVED_RECOMMENDATION = "resolved_recommendation";
-    private static final String DEACTIVATED_RECOMMENDATION = "deactivated_recommendation";
+    public static final String PAYLOAD_RULE_ID = "rule_id";
+    public static final String PAYLOAD_RULE_DESCRIPTION = "rule_description";
+    public static final String PAYLOAD_RULE_TOTAL_RISK = "total_risk";
+    public static final String PAYLOAD_RULE_HAS_INCIDENT = "has_incident";
+    public static final String PAYLOAD_RULE_URL = "rule_url";
 
     // Advisor events aggregator data contents
-    private static final String ADVISOR_KEY = "advisor";
-    private static final String NEW_RECOMMENDATIONS = "new_recommendations";
-    private static final String RESOLVED_RECOMMENDATIONS = "resolved_recommendations";
-    private static final String DEACTIVATED_RECOMMENDATIONS = "deactivated_recommendations";
-    private static final String CONTENT_RULE_DESCRIPTION = "description";
-    private static final String CONTENT_RULE_TOTAL_RISK = "total_risk";
-    private static final String CONTENT_RULE_HAS_INCIDENT = "has_incident";
-    private static final String CONTENT_RULE_URL = "url";
-    private static final String CONTENT_SYSTEM_COUNT = "systems";
+    public static final String ADVISOR_KEY = "advisor";
+    public static final String NEW_RECOMMENDATIONS = "new_recommendations";
+    public static final String RESOLVED_RECOMMENDATIONS = "resolved_recommendations";
+    public static final String DEACTIVATED_RECOMMENDATIONS = "deactivated_recommendations";
+    public static final String CONTENT_RULE_DESCRIPTION = "description";
+    public static final String CONTENT_RULE_TOTAL_RISK = "total_risk";
+    public static final String CONTENT_RULE_HAS_INCIDENT = "has_incident";
+    public static final String CONTENT_RULE_URL = "url";
+    public static final String CONTENT_SYSTEM_COUNT = "systems";
 
     // Advisor event types
-    private static final String NEW_RECOMMENDATION_EVENT = "new-recommendation";
-    private static final String RESOLVED_RECOMMENDATION_EVENT = "resolved-recommendation";
-    private static final String DEACTIVATED_RULE_EVENT = "deactivated-rule";
+    public static final String NEW_RECOMMENDATION = "new-recommendation";
+    public static final String RESOLVED_RECOMMENDATION = "resolved-recommendation";
+    public static final String DEACTIVATED_RECOMMENDATION = "deactivated-recommendation";
     private static final Set<String> EVENT_TYPES = new HashSet<>(Arrays.asList(
-        NEW_RECOMMENDATION_EVENT, RESOLVED_RECOMMENDATION_EVENT,
-        DEACTIVATED_RULE_EVENT
+            NEW_RECOMMENDATION, RESOLVED_RECOMMENDATION,
+            DEACTIVATED_RECOMMENDATION
     ));
 
     // Advisor final payload helpers
@@ -137,7 +131,6 @@ public class AdvisorEmailAggregator extends AbstractEmailPayloadAggregator {
     @Override
     void processEmailAggregation(EmailAggregation notification) {
         JsonObject notifPayload = notification.getPayload();
-        JsonObject advisor = context.getJsonObject(ADVISOR_KEY);
         String eventType = notifPayload.getString(EVENT_TYPE_KEY);
         /* Major assumption: that there is at most one event per day from each
          * rule,system pair - i.e. that the tuple (day, rule, system) is
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
@@ -61,7 +61,7 @@ public class AdvisorEmailAggregator extends AbstractEmailPayloadAggregator {
     // Advisor event payloads
     private static final String NEW_RECOMMENDATION = "new_recommendation";
     private static final String RESOLVED_RECOMMENDATION = "resolved_recommendation";
-    private static final String DEACTIVATED_RECOMMENDATION = "resolved_recommendation";
+    private static final String DEACTIVATED_RECOMMENDATION = "deactivated_recommendation";
 
     // Advisor events aggregator data contents
     private static final String ADVISOR_KEY = "advisor";
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
@@ -10,6 +10,41 @@ import java.util.Map;
 import java.util.Set;
 
 
+/*
+ * Events are of the form:
+ * {
+   "bundle":"rhel",
+   "application":"advisor",
+   "event_type":"new-recommendation",
+   "timestamp":"2022-11-30T17:30:20.773398",
+   "account_id":"5758117",
+   "org_id":"7806094",
+   "context":{
+      "inventory_id":"6ad30f3e-0497-4e74-99f1-b3f9a6120a6f",
+      "hostname":"my-computer-jmartine",
+      "display_name":"my-computer-jmartine",
+      "rhel_version":"8.4",
+      "host_url":"https://console.redhat.com/insights/inventory/6ad30f3e-0497-4e74-99f1-b3f9a6120a6f",
+      "tags":[]
+   },
+   "events":[
+      {
+         "metadata":{},
+         "payload":{
+            "rule_id":"insights_core_egg_not_up2date|INSIGHTS_CORE_EGG_NOT_UP2DATE",
+            "rule_description":"System is not able to get the latest recommendations and may miss bug fixes when the Insights Client Core egg file is outdated",
+            "total_risk":"2",
+            "publish_date":"2021-03-13T18:44:00+00:00",
+            "rule_url":"https://console.redhat.com/insights/advisor/recommendations/insights_core_egg_not_up2date|INSIGHTS_CORE_EGG_NOT_UP2DATE/",
+            "reboot_required":false,
+            "has_incident":false,
+            "report_url":"https://console.redhat.com/insights/advisor/recommendations/insights_core_egg_not_up2date|INSIGHTS_CORE_EGG_NOT_UP2DATE/6ad30f3e-0497-4e74-99f1-b3f9a6120a6f"
+         }
+      }
+   ]
+}
+ */
+
 public class AdvisorEmailPayloadAggregator extends AbstractEmailPayloadAggregator {
 
     // Notification common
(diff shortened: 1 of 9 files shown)


engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
```diff
@@ -0,0 +1,85 @@
...
+        if (!EVENT_TYPES.contains(eventType)) {
+            return;
+        }
+
+        notifPayload.getJsonArray(EVENTS_KEY).stream().forEach(eventObject -> {
+            JsonObject event = (JsonObject) eventObject;
+            JsonObject payload = event.getJsonObject(PAYLOAD_KEY);
+            String ruleId = payload.getString(PAYLOAD_RULE_ID);
+
+            if (eventType == NEW_RECOMMENDATION) {
```
gwenneg, line 67:
```suggestion
            if (eventType.equals(NEW_RECOMMENDATION)) {
```

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
```diff
@@ -0,0 +1,85 @@
...
+                if (! newRecommendations.contains(ruleId)) {
+                    newRecommendations.put(ruleId, new HashSet<String>());
+                }
+                newRecommendations.get(ruleId).add(inventoryId);
+            } else if (eventType == RESOLVED_RECOMMENDATION) {
```
gwenneg, line 72:
```suggestion
            } else if (eventType.equals(RESOLVED_RECOMMENDATION)) {
```

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
```diff
@@ -0,0 +1,85 @@
...
+                if (! resolvedRecommendations.contains(ruleId)) {
+                    resolvedRecommendations.put(ruleId, new HashSet<String>());
+                }
+                resolvedRecommendations.get(ruleId).add(inventoryId);
+            } else if (eventType == DEACTIVATED_RECOMMENDATION) {
```
gwenneg, line 77:
```suggestion
            } else if (eventType.equals(DEACTIVATED_RECOMMENDATION)) {
```

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
```diff
@@ -0,0 +1,85 @@
...
+                deactivatedRecommendations.add(ruleId);
+            }
+        });
+        advisor.put(NEW_RECOMMENDATIONS, newRecommendations);
+        advisor.put(RESOLVED_RECOMMENDATIONS, newRecommendations);
+        advisor.put(DEACTIVATED_RECOMMENDATIONS, newRecommendations);
```
gwenneg, line 83:
```suggestion
        advisor.put(RESOLVED_RECOMMENDATIONS, resolvedRecommendations);
        advisor.put(DEACTIVATED_RECOMMENDATIONS, deactivatedRecommendations);
```

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
```diff
@@ -0,0 +1,85 @@
...
+            } else if (eventType == RESOLVED_RECOMMENDATION) {
+                if (! resolvedRecommendations.contains(ruleId)) {
+                    resolvedRecommendations.put(ruleId, new HashSet<String>());
+                }
+                resolvedRecommendations.get(ruleId).add(inventoryId);
+            } else if (eventType == DEACTIVATED_RECOMMENDATION) {
+                deactivatedRecommendations.add(ruleId);
+            }
+        });
+        advisor.put(NEW_RECOMMENDATIONS, newRecommendations);
```
gwenneg, line 81: Each invocation of the `processEmailAggregation` method relies on the same `advisor` instance. This means you don't have to do this `put` and the others below on each invocation, but only once from the class constructor (`public AdvisorDailyDigestEmailPayloadAggregator()`).

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
```diff
@@ -0,0 +1,85 @@
...
+        notifPayload.getJsonArray(EVENTS_KEY).stream().forEach(eventObject -> {
+            JsonObject event = (JsonObject) eventObject;
+            JsonObject payload = event.getJsonObject(PAYLOAD_KEY);
+            String ruleId = payload.getString(PAYLOAD_RULE_ID);
+
+            if (eventType == NEW_RECOMMENDATION) {
+                if (! newRecommendations.contains(ruleId)) {
+                    newRecommendations.put(ruleId, new HashSet<String>());
+                }
+                newRecommendations.get(ruleId).add(inventoryId);
```
gwenneg, line 71: Here's a shorter way of doing that:
```suggestion
                newRecommendations.computeIfAbsent(ruleId, key -> new HashSet<>()).add(inventoryId);
```

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
```diff
@@ -0,0 +1,85 @@
...
+            } else if (eventType == RESOLVED_RECOMMENDATION) {
+                if (! resolvedRecommendations.contains(ruleId)) {
+                    resolvedRecommendations.put(ruleId, new HashSet<String>());
+                }
+                resolvedRecommendations.get(ruleId).add(inventoryId);
```
gwenneg, line 76: Here's a shorter way of doing that:
```suggestion
                resolvedRecommendations.computeIfAbsent(ruleId, key -> new HashSet<>()).add(inventoryId);
```

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
```diff
@@ -0,0 +1,85 @@
...
+    private static final String DEACTIVATED_RULE_EVENT = "deactivated-rule";
+    private static final Set<String> EVENT_TYPES = new HashSet<>(Arrays.asList(
+        NEW_RECOMMENDATION_EVENT, RESOLVED_RECOMMENDATION_EVENT,
+        DEACTIVATED_RULE_EVENT
+    ));
+
+    // Advisor final payload helpers
+    private final HashMap<String, HashSet<String>> newRecommendations = new HashMap<String, HashSet<>>();
+    private final HashMap<String, HashSet<String>> resolvedRecommendations = new HashSet<String, HashSet<>>();
+    private final HashSet<String> deactivatedRecommendations = new HashSet<>();
```
gwenneg, line 43: This part is causing compilation errors. Here's a fix:
```suggestion
    private final Map<String, Set<String>> newRecommendations = new HashMap<>();
    private final Map<String, Set<String>> resolvedRecommendations = new HashMap<>();
    private final Set<String> deactivatedRecommendations = new HashSet<>();
```
⚠️ You will need a few `import`s at the top of the class for this to work.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public AdvisorDailyDigestEmailPayloadAggregator() {
+        JsonObject advisor = new JsonObject();
+
+        context.put(ADVISOR_KEY, advisor);
```
gwenneg, line 48: Minor comment: this could be shorter:
```suggestion
        context.put(ADVISOR_KEY, new JsonObject());
```

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
PaulWay: Yeah, I was wondering about that.  So I need to basically get those in the class constructor and then put data into them in `processEmailAggregation`, yeah?

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
PaulWay: Ah, excellent, I wondered if there'd be some inbuilt facility for doing that.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
PaulWay: If I put the references to the `Map` and `Set` objects in there, won't I need the `advisor` variable?

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
PaulWay: Yep, oops, copy-and-paste error!

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
PaulWay: Got it - thanks!

PaulWay:
OK, I think I've fixed all those things @gwenneg - let's see how this version looks!

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
gwenneg: This suggestion hasn't been applied yet. I'm only writing that in case you missed it but it's not mandatory.

gwenneg:
Checkstyle (our tool used for linting the Java code) is unhappy because of a few issues:
```
Error:  /home/runner/work/notifications-backend/notifications-backend/engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java:8:1: Wrong order for 'java.util.HashMap' import. [ImportOrder]
Error:  /home/runner/work/notifications-backend/notifications-backend/engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java:77:21: '!' is followed by whitespace. [NoWhitespaceAfter]
```

gwenneg:
Since the aggregation logic looks good now, let's talk about what's missing in this PR:
- `AdvisorDailyDigestEmailPayloadAggregator` needs to be declared in [EmailPayloadAggregatorFactory](https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/EmailPayloadAggregatorFactory.java) under the right bundle/application couple
- You will need the HTML template for the daily digest. Here's [an example](https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/main/resources/templates/ResourceOptimization/dailyEmailBody.html). You'll find many other HTML templates in the parent package of that file.
- Testing the aggregation logic would be nice. Here's [an example](https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/test/java/com/redhat/cloud/notifications/processors/email/aggregators/ResourceOptimizationPayloadAggregatorTest.java). You'll find other similar tests in the same package.
- [Advisor](https://github.com/RedHatInsights/notifications-backend/blob/826aa4957ee319fe08dc2dff696fa3a60e363b93/engine/src/main/java/com/redhat/cloud/notifications/templates/Advisor.java#L46-L54) currently supports `EmailSubscriptionType.INSTANT` email only. This needs to be changed to include `EmailSubscriptionType.DAILY`.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
PaulWay: Yep, missed that - just fixed it.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
PaulWay: OK, good, I think I've fixed that then, yeah?

PaulWay:
OK, if I understand it then the imports should be in alphabetical order - this suits me very well :-)

PaulWay:
Let's start work on the rest of your helpful checklist :grinning:

PaulWay:
As a quick side discussion, we need to have a `Set` of systems impacted by each rule because a system may generate multiple notifications in a day and it is technically possible (although usually very unlikely) for a system to trigger the same rule anew twice in a day.  We haven't yet looked at the actual notifications of a big customer to see if this might actually never occur, or might happen more often than we think.

PaulWay:
I've gone with a big assumption, that the above is not true - that in general each system will never report a new rule or a resolved rule more than once per day.

PaulWay:
OK @gwenneg - how's this looking so far?

gwenneg:
It looks good so far 👍

josejulio:
The email template was added as part of https://github.com/RedHatInsights/notifications-backend/pull/1641 (thanks @mhuth !)

@PaulWay Could you rebase and add the following to complete the setup?
It can be done as part of this PR or as a separate one once this is merged.

It basically tells notifications that we want to support Advisor's daily email and to use the new template everytime we want to generate a DAILY email.

```
diff --git a/engine/src/main/java/com/redhat/cloud/notifications/templates/Advisor.java b/engine/src/main/java/com/redhat/cloud/notifications/templates/Advisor.java
index 63dc5cc7..89a35fc7 100644
--- a/engine/src/main/java/com/redhat/cloud/notifications/templates/Advisor.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/templates/Advisor.java
@@ -17,6 +17,8 @@ public class Advisor implements EmailTemplate {
             } else if (eventType.equals("deactivated-recommendation")) {
                 return Templates.deactivatedRecommendationInstantEmailTitle();
             }
+        } else if (type == EmailSubscriptionType.DAILY) {
+            return Templates.dailyEmailTitle();
         }
 
         throw new UnsupportedOperationException(String.format(
@@ -35,6 +37,8 @@ public class Advisor implements EmailTemplate {
             } else if (eventType.equals("deactivated-recommendation")) {
                 return Templates.deactivatedRecommendationInstantEmailBody();
             }
+        } else if (type == EmailSubscriptionType.DAILY) {
+            return Templates.dailyEmailBody();
         }
 
         throw new UnsupportedOperationException(String.format(
@@ -45,12 +49,12 @@ public class Advisor implements EmailTemplate {
 
     @Override
     public boolean isSupported(String eventType, EmailSubscriptionType type) {
-        return (eventType.equals("new-recommendation") || eventType.equals("resolved-recommendation") || eventType.equals("deactivated-recommendation")) && type == EmailSubscriptionType.INSTANT;
+        return type == EmailSubscriptionType.DAILY || (eventType.equals("new-recommendation") || eventType.equals("resolved-recommendation") || eventType.equals("deactivated-recommendation")) && type == EmailSubscriptionType.INSTANT;
     }
 
     @Override
     public boolean isEmailSubscriptionSupported(EmailSubscriptionType type) {
-        return type == EmailSubscriptionType.INSTANT;
+        return type == EmailSubscriptionType.INSTANT || type == EmailSubscriptionType.DAILY;
     }
 
     @CheckedTemplate(requireTypeSafeExpressions = false)
@@ -67,6 +71,10 @@ public class Advisor implements EmailTemplate {
         public static native TemplateInstance deactivatedRecommendationInstantEmailTitle();
 
         public static native TemplateInstance deactivatedRecommendationInstantEmailBody();
+
+        public static native TemplateInstance dailyEmailTitle();
+
+        public static native TemplateInstance dailyEmailBody();
     }
 
 }
```

gwenneg:
I'm fixing the remaining build errors.

gwenneg (approved):
LGTM

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
```diff
@@ -0,0 +1,197 @@
...
+    void processEmailAggregation(EmailAggregation notification) {
+        JsonObject notifPayload = notification.getPayload();
+        String eventType = notifPayload.getString(EVENT_TYPE_KEY);
+        /* Major assumption: that there is at most one event per day from each
+         * rule,system pair - i.e. that the tuple (day, rule, system) is
+         * unique for new, and resolved, events.  This is generally true,
+         * because a rule would have to be new twice or resolved twice - i.e.
+         * new, then resolved, then occur again - on the same system in order
+         * for this to not be wrong.  This means we don't need to remember the
+         * inventory ID of each system impacted by or resolving a rule. */
```
gwenneg, line 137: @PaulWay If you change your mind about this, please let me know and I'll help you change the aggregation logic.

PR #1646, opened 2022-12-13
Fix the flexible sections of Advisor daily email templates

mhuth:


mhuth:
Use the orEmpty resolver in case the referenced attributes don't exist in the JSON context object.  It will give a size of 0, as per the docs:
https://quarkus.io/guides/qute-reference#built-in-resolvers

PR #1647, opened 2022-12-13
Allow email template iframe to be resized to better view email template

mhuth:


mhuth:
Adds a resize widget to the email template iframe so its easier to see the marked-up email.  Seems to work ok in my testing in chrome.
Screenshot from 2022-12-13 14-32-58

josejulio (approved):
lgtm

PR #1650, opened 2022-12-14
Resize iframe inline style in JSX accepted way

mhuth:
The previous way I did it as a string broke the JSX rendering of the page.  This PR does the inline style in JSX friendly way.


mhuth:
@josejulio @gwenneg The tests are failing but it looks like its an environmental problem - test environment can't connect with kafka - rather than related to the code in the PR.

josejulio (approved):

PR #1655, opened 2022-12-15
RHCLOUD-23364 schedule daily digest every15min

g-duval:


aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationRepository.java
```diff
@@ -23,13 +24,15 @@ public class EmailAggregationRepository {
...
+        // it covers cases when aggregation job may be run with few minutes late (ie: 05:01, 07,32)
         String query = "SELECT DISTINCT ea.orgId, ea.bundleName, ea.applicationName, acp.lastRun FROM EmailAggregation ea, AggregationOrgConfig acp WHERE " +
             "ea.orgId = acp.orgId AND ea.created > acp.lastRun AND ea.created <= :end " +
-            "AND hour(acp.scheduledExecutionTime) = :runningHour";
-
+            "AND :endTime >= acp.scheduledExecutionTime AND (:endTime - acp.scheduledExecutionTime) < CAST(:cutoff as LocalTime)";
         Query hqlQuery = entityManager.createQuery(query)
-                .setParameter("runningHour", end.getHour())
-                .setParameter("end", end);
+                .setParameter("endTime", end.toLocalTime())
```
gwenneg, line 33: Since `end` is actually `now`, we could name everything like this:
```suggestion
                .setParameter("now", now.toLocalTime())
```

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationRepository.java
```diff
@@ -23,13 +24,15 @@ public class EmailAggregationRepository {
     EntityManager entityManager;
 
     public List<AggregationCommand> getApplicationsWithPendingAggregationAccordinfOrgPref(LocalDateTime end) {
+        // Must takes every EmailAggregation supposed to be processed on last 15 min
+        // it covers cases when aggregation job may be run with few minutes late (ie: 05:01, 07,32)
         String query = "SELECT DISTINCT ea.orgId, ea.bundleName, ea.applicationName, acp.lastRun FROM EmailAggregation ea, AggregationOrgConfig acp WHERE " +
             "ea.orgId = acp.orgId AND ea.created > acp.lastRun AND ea.created <= :end " +
-            "AND hour(acp.scheduledExecutionTime) = :runningHour";
-
+            "AND :endTime >= acp.scheduledExecutionTime AND (:endTime - acp.scheduledExecutionTime) < CAST(:cutoff as LocalTime)";
```
gwenneg, line 31: Is it possible to replace
```
(:endTime - acp.scheduledExecutionTime) < CAST(:cutoff as LocalTime)
```
with something based on Postgres's [extract(minute ...)](https://www.postgresql.org/docs/11/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT) function?

Something like this (untested):
```
EXTRACT(MINUTE FROM TIME :endTime) - EXTRACT(MINUTE FROM TIME acp.scheduledExecutionTime) < 15
```

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -47,8 +47,8 @@ public class DailyEmailAggregationJob {
     @ConfigProperty(name = "prometheus.pushgateway.url")
     String prometheusPushGatewayUrl;
 
-    @ConfigProperty(name = "notification.default.daily.digest.hour", defaultValue = "0")
-    int defaultDailyDigestHour;
+    @ConfigProperty(name = "notification.default.daily.digest.hour", defaultValue = "00:00")
+    String defaultDailyDigestHour;
```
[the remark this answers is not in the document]
gwenneg: Quarkus and its underlying [config dependency](https://github.com/smallrye/smallrye-config) include many converters that can deserialize config values into Java types. I thought there would be something for `LocalTime` that we could use here, but apparently that doesn't exist (yet)...

backend/src/main/java/com/redhat/cloud/notifications/routers/OrgConfigResource.java
```diff
@@ -41,6 +45,9 @@ public class OrgConfigResource {
     @RolesAllowed(ConsoleIdentityProvider.RBAC_WRITE_NOTIFICATIONS)
     public Response saveDailyDigestTimePreference(@Context SecurityContext sec, @NotNull LocalTime expectedTime) {
         String orgId = getOrgId(sec);
+        if (!ALLOWED_MINUTES.contains(expectedTime.getMinute())) {
+            return Response.status(Response.Status.BAD_REQUEST).build();
```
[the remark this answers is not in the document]
g-duval: Ok

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -47,8 +48,8 @@ public class DailyEmailAggregationJob {
     @ConfigProperty(name = "prometheus.pushgateway.url")
     String prometheusPushGatewayUrl;
 
-    @ConfigProperty(name = "notification.default.daily.digest.hour", defaultValue = "0")
-    int defaultDailyDigestHour;
+    @ConfigProperty(name = "notification.default.daily.digest.time", defaultValue = "00:00")
```
gwenneg, line 51:
```suggestion
    @ConfigProperty(name = "notifications.default.daily.digest.time", defaultValue = "00:00")
```

backend/src/main/java/com/redhat/cloud/notifications/routers/OrgConfigResource.java
```diff
@@ -20,27 +20,37 @@
...
 public class OrgConfigResource {
 
+    static final List<Integer> ALLOWED_MINUTES = Arrays.asList(0, 15, 30, 45);
+
     @Inject
     AggregationOrgConfigRepository aggregationOrgConfigRepository;
 
-    @ConfigProperty(name = "notification.default.daily.digest.hour", defaultValue = "0")
-    int defaultDailyDigestHour;
+    @ConfigProperty(name = "notification.default.daily.digest.time", defaultValue = "00:00")
```
gwenneg, line 38:
```suggestion
    @ConfigProperty(name = "notifications.default.daily.digest.time", defaultValue = "00:00")
```

backend/src/main/java/com/redhat/cloud/notifications/routers/OrgConfigResource.java
```diff
@@ -20,27 +20,37 @@
...
+    @ConfigProperty(name = "notifications.default.daily.digest.time", defaultValue = "00:00")
+    LocalTime defaultDailyDigestTime;
 
     @PUT
     @Path("/daily-digest/time-preference")
     @Consumes(APPLICATION_JSON)
     @Produces(APPLICATION_JSON)
     @Transactional
     @RolesAllowed(ConsoleIdentityProvider.RBAC_WRITE_NOTIFICATIONS)
+    @Operation(summary = "Save daily digest time preference as UTC format. To cover all time zone conversions to UTC format, minute value have to be 00, 15, 30 or 45.")
```
gwenneg, line 47:
```suggestion
    @Operation(summary = "Save the daily digest UTC time preference. To cover all time zones conversion to UTC, the accepted minute values are 00, 15, 30 and 45.")
```

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationRepository.java
gwenneg: My previous suggestion was indeed not so good :smile:

What about this alternative?
```java
    public List<AggregationCommand> getApplicationsWithPendingAggregationAccordinfOrgPref(LocalDateTime now) {
        // Must takes every EmailAggregation supposed to be processed on last 15 min
        // it covers cases when aggregation job may be run with few minutes late (ie: 05:01, 07,32)
        String query = "SELECT DISTINCT ea.org_id, ea.bundle, ea.application, acp.last_run FROM email_aggregation ea, aggregation_org_config acp WHERE " +
            "ea.org_id = acp.org_id AND ea.created > acp.last_run AND ea.created <= :now " +
            // BETWEEN is inclusive so we need to subtract one millisecond from the right boundary.
            "AND  CAST(:now AS TIME) BETWEEN acp.scheduled_execution_time AND acp.scheduled_execution_time + (15 * INTERVAL '1 minute') - (1 * INTERVAL '1 microsecond')";
        Query hqlQuery = entityManager.createNativeQuery(query)
                .setParameter("now", now);

        List<Object[]> records = hqlQuery.getResultList();
        return records.stream()
                .map(emailAggregationRecord -> new AggregationCommand(
                    new EmailAggregationKey((String) emailAggregationRecord[0], (String) emailAggregationRecord[1], (String) emailAggregationRecord[2]),
                    ((Timestamp) emailAggregationRecord[3]).toLocalDateTime(),
                    now,
                    DAILY
                ))
                .collect(toList());
    }
```

gwenneg: My only concern about the current query  is that it not easy to understand at first glance.

gwenneg: I think what you did is fine 😃

PR #1662, opened 2022-12-22
[RHCLOUD-22513] feature: create behavior groups specifying the bundle name

MikelAlejoBR:
The frontend team has requested us to implement a change to be able to create a behavior group by using the bundle's name, instead of its UUID. The reason behind this is that they seem to be able to provide the bundle name directly, but for the UUID, they need to fetch all the bundles to find the right one by name.

## Links

[[RHCLOUD-22513]](https://issues.redhat.com/browse/RHCLOUD-22513)

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BundleRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BundleRepository.java
@@ -82,4 +82,27 @@ public class BundleRepository {
                     .getResultList();
         }
     }
+
+    /**
+     * Finds the bundle by name.
+     * @param bundleName the name to find the bundle by.
+     * @return the found bundle from the database.
+     */
+    public Bundle findByName(final String bundleName) {
+        final String findByNameQuery =
+            "SELECT " +
+                    "b " +
+            "FROM " +
+                "Bundle AS b " +
+            "WHERE " +
+                "b.name = :name";
+
+        try {
+            return this.entityManager.createQuery(findByNameQuery, Bundle.class)
+                .setParameter("name", bundleName)
+                .getSingleResult();
+        } catch (final NoResultException e) {
+            throw new NotFoundException("the provided bundle name does not exist");
+        }
+    }
 }
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BundleRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BundleRepository.java
@@ -10,6 +10,7 @@ import javax.persistence.NoResultException;
 import javax.transaction.Transactional;
 import javax.ws.rs.NotFoundException;
 import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
 
 @ApplicationScoped
@@ -88,7 +89,7 @@ public class BundleRepository {
      * @param bundleName the name to find the bundle by.
      * @return the found bundle from the database.
      */
-    public Bundle findByName(final String bundleName) {
+    public Optional<Bundle> findByName(final String bundleName) {
         final String findByNameQuery =
             "SELECT " +
                     "b " +
@@ -98,11 +99,13 @@ public class BundleRepository {
                 "b.name = :name";
 
         try {
-            return this.entityManager.createQuery(findByNameQuery, Bundle.class)
-                .setParameter("name", bundleName)
-                .getSingleResult();
+            return Optional.of(
+                this.entityManager.createQuery(findByNameQuery, Bundle.class)
+                    .setParameter("name", bundleName)
+                    .getSingleResult()
+            );
         } catch (final NoResultException e) {
-            throw new NotFoundException("the provided bundle name does not exist");
+            return Optional.empty();
         }
     }
 }
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/NotificationResource.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/NotificationResource.java
@@ -19,6 +19,7 @@ import com.redhat.cloud.notifications.routers.models.PageLinksBuilder;
 import com.redhat.cloud.notifications.routers.models.behaviorgroup.CreateBehaviorGroupRequest;
 import com.redhat.cloud.notifications.routers.models.behaviorgroup.CreateBehaviorGroupResponse;
 import com.redhat.cloud.notifications.routers.models.behaviorgroup.UpdateBehaviorGroupRequest;
+import org.apache.commons.lang3.StringUtils;
 import org.eclipse.microprofile.openapi.annotations.Operation;
 import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
 import org.eclipse.microprofile.openapi.annotations.media.Content;
@@ -225,13 +226,26 @@ public class NotificationResource {
     })
     @RolesAllowed(ConsoleIdentityProvider.RBAC_WRITE_NOTIFICATIONS)
     @Transactional
-    public CreateBehaviorGroupResponse createBehaviorGroup(@Context SecurityContext sec,
-                              @RequestBody(required = true) @Valid @NotNull CreateBehaviorGroupRequest request) {
+    public CreateBehaviorGroupResponse createBehaviorGroup(
+        @Context final SecurityContext sec,
+        @RequestBody(required = true) @Valid @NotNull final CreateBehaviorGroupRequest request
+    ) {
         String accountId = getAccountId(sec);
         String orgId = getOrgId(sec);
 
+        // Set the bundle ID from the incoming request. If the user provided a
+        // bundle name instead, fetch the bundle ID from the database. This is
+        // to make the frontend's life easier, in the way that they won't need
+        // to fetch the bundles to get the exact bundle ID. More information in
+        // the request's class, or in RHCLOUD-22513.
+        UUID bundleId = request.bundleId;
+        if (bundleId == null && !StringUtils.isBlank(request.bundleName)) {
+            final Bundle bundle = this.bundleRepository.findByName(request.bundleName);
+            bundleId = bundle.getId();
+        }
+
         BehaviorGroup behaviorGroup = new BehaviorGroup();
-        behaviorGroup.setBundleId(request.bundleId);
+        behaviorGroup.setBundleId(bundleId);
         behaviorGroup.setDisplayName(request.displayName);
 
         behaviorGroup = behaviorGroupRepository.createFull(
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/NotificationResource.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/NotificationResource.java
@@ -19,7 +19,6 @@ import com.redhat.cloud.notifications.routers.models.PageLinksBuilder;
 import com.redhat.cloud.notifications.routers.models.behaviorgroup.CreateBehaviorGroupRequest;
 import com.redhat.cloud.notifications.routers.models.behaviorgroup.CreateBehaviorGroupResponse;
 import com.redhat.cloud.notifications.routers.models.behaviorgroup.UpdateBehaviorGroupRequest;
-import org.apache.commons.lang3.StringUtils;
 import org.eclipse.microprofile.openapi.annotations.Operation;
 import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
 import org.eclipse.microprofile.openapi.annotations.media.Content;
@@ -240,7 +239,7 @@ public class NotificationResource {
         // to fetch the bundles to get the exact bundle ID. More information in
         // the request's class, or in RHCLOUD-22513.
         UUID bundleId = request.bundleId;
-        if (bundleId == null && !StringUtils.isBlank(request.bundleName)) {
+        if (bundleId == null) {
             final Optional<Bundle> bundle = this.bundleRepository.findByName(request.bundleName);
             if (bundle.isEmpty()) {
                 throw new NotFoundException("the specified bundle was not found in the database");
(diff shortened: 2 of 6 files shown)


backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BundleRepository.java
```diff
@@ -82,4 +82,27 @@ public List<Application> getApplications(UUID id) {
...
+                "Bundle AS b " +
+            "WHERE " +
+                "b.name = :name";
+
+        try {
+            return this.entityManager.createQuery(findByNameQuery, Bundle.class)
+                .setParameter("name", bundleName)
+                .getSingleResult();
+        } catch (final NoResultException e) {
+            throw new NotFoundException("the provided bundle name does not exist");
```
gwenneg, line 105: Nitpick:

Throwing a JAX-RS exception here makes this method tightly-coupled with the REST layer while we could have a cleaner separation between the persistence and the REST layers.

This method could return an `Optional<Bundle>` and `Optional.empty()` in case of `NoResultException` to avoid that tight-coupling between layers. Then the calling method would throw the `NotFoundException` when the optional is empty.

backend/src/main/java/com/redhat/cloud/notifications/routers/models/behaviorgroup/CreateBehaviorGroupRequest.java
```diff
@@ -35,4 +50,13 @@ public class CreateBehaviorGroupRequest {
...
      */
     public Set<UUID> eventTypeIds;
+
+    /**
+     * Validates that the bundle is identifiable.
+     * @return true if the bundle ID is null and the bundle name is blank. False otherwise.
+     */
+    @AssertFalse(message = "either the bundle name or the bundle UUID are required")
+    private boolean isBundleUuidNullOrBundleNameBlank() {
+        return this.bundleId == null && StringUtils.isBlank(this.bundleName);
```
gwenneg, line 60: `StringUtils.noBlank` may not be available anymore with the latest version of `master`. `org.apache.commons.lang3.StringUtils` came from a transitive dependency which may no longer be in the classpath. If so, you can simply do an old-school check 😄

backend/src/main/java/com/redhat/cloud/notifications/routers/models/behaviorgroup/CreateBehaviorGroupRequest.java
```diff
@@ -3,21 +3,36 @@
...
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import org.apache.commons.lang3.StringUtils;
 
+import javax.validation.constraints.AssertFalse;
 import javax.validation.constraints.NotNull;
 import java.util.List;
 import java.util.Set;
 import java.util.UUID;
 
+/**
```
gwenneg, line 14: This is a Javadoc kind of comment. It feels like a simple comment would be enough. WDYT?
```suggestion
/*
```

backend/src/main/java/com/redhat/cloud/notifications/routers/NotificationResource.java
```diff
@@ -225,13 +226,26 @@ public List<Facet> getBundleFacets(@Context SecurityContext sec, @QueryParam("in
...
         String accountId = getAccountId(sec);
         String orgId = getOrgId(sec);
 
+        // Set the bundle ID from the incoming request. If the user provided a
+        // bundle name instead, fetch the bundle ID from the database. This is
+        // to make the frontend's life easier, in the way that they won't need
+        // to fetch the bundles to get the exact bundle ID. More information in
+        // the request's class, or in RHCLOUD-22513.
+        UUID bundleId = request.bundleId;
+        if (bundleId == null && !StringUtils.isBlank(request.bundleName)) {
```
gwenneg, line 242: We already know from the `@AssertFalse` constraint that `bundleName` is not blank if `bundleId` is null.
```suggestion
        if (bundleId == null) {
```

backend/src/test/java/com/redhat/cloud/notifications/db/repositories/BundleRepositoryTest.java
```diff
@@ -0,0 +1,45 @@
...
+import io.quarkus.test.junit.QuarkusTest;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import javax.inject.Inject;
+import javax.ws.rs.NotFoundException;
+
+@QuarkusTest
+@QuarkusTestResource(TestLifecycleManager.class)
+public class BundleRepositoryTest {
```
gwenneg, line 16:
```suggestion
public class BundleRepositoryTest extends DbIsolatedTest {
```

backend/src/test/java/com/redhat/cloud/notifications/routers/NotificationResourceTest.java
```diff
@@ -972,6 +972,51 @@ void testBehaviorGroupSameName() {
...
+        final Header identityHeader = initRbacMock("tenant", "sameBehaviorGroupName", "user", FULL_ACCESS);
+
+        given()
+            .header(identityHeader)
+            .when()
+            .contentType(JSON)
+            .body(Json.encode(createBehaviorGroupRequest))
+            .post("/notifications/behaviorGroups")
+            .then()
+            .statusCode(400);
```
gwenneg, line 1017: Could you please check the message that is returned with the HTTP error?

backend/src/main/java/com/redhat/cloud/notifications/routers/models/behaviorgroup/CreateBehaviorGroupRequest.java
josejulio: It's already used elsewhere, if master is not failing, then it should be fine.
I wonder if we should depend on it directly - unsure about maven semantics i.e. it's posible to tell maven to depend on the lib on at least version x.y.z ?

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BundleRepository.java
MikelAlejoBR: Yeah, makes a lot of sense. Dealt with!

backend/src/main/java/com/redhat/cloud/notifications/routers/NotificationResource.java
MikelAlejoBR: Great catch! Refactored.

backend/src/main/java/com/redhat/cloud/notifications/routers/models/behaviorgroup/CreateBehaviorGroupRequest.java
MikelAlejoBR: I moved the comment from the request's class to the handler's method, since I think it makes more sense to have it there, close to the logic of how the creation is dealt with.

backend/src/test/java/com/redhat/cloud/notifications/db/repositories/BundleRepositoryTest.java
MikelAlejoBR: Done!

backend/src/test/java/com/redhat/cloud/notifications/routers/NotificationResourceTest.java
MikelAlejoBR: Done as well!

backend/src/main/java/com/redhat/cloud/notifications/routers/models/behaviorgroup/CreateBehaviorGroupRequest.java
MikelAlejoBR: It seems like we could use the [`maven-enforcer-plugin`](https://maven.apache.org/enforcer/maven-enforcer-plugin/) to set the minimum version of `commons-lang` to `3.12.0`, which is the one we are currently using? Otherwise I don't mind doing it the old way.

Reference: https://stackoverflow.com/questions/32971660/enforce-a-minimum-version-of-maven

MikelAlejoBR: All right I just saw that the pipeline is failing because of the missing dependency

gwenneg: That specific dependency is declared in the [Quarkus BOM](https://github.com/quarkusio/quarkus/blob/main/bom/application/pom.xml) so if we add it to our POM, we should use the version from Quarkus (which means we won't declare any version as it will come from the BOM).

gwenneg: However, do we really need to add a direct dependency only to avoid writing an `if (x != null && !x.isBlank())` condition? This feels like we'll use a bazooka to kill a fly to me 😄

backend/src/test/java/com/redhat/cloud/notifications/db/repositories/BundleRepositoryTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+        // Call the function under test.
+        final Optional<Bundle> result = this.bundleRepository.findByName(bundleName);
+
+        Assertions.assertTrue(result.isPresent(), "no bundle was fetched from the database");
+        Assertions.assertEquals(bundle.getId(), result.get().getId(), "unexpected bundle fetched by name from the database");
+    }
+
+    /**
+     * Tests that a not found exception is thrown when the bundle cannot be found by its name.
```
gwenneg, line 41:
```suggestion
     * Tests that an empty Optional is returned when the bundle cannot be found by its name.
```

backend/src/test/java/com/redhat/cloud/notifications/routers/NotificationResourceTest.java
```diff
@@ -972,6 +973,66 @@ void testBehaviorGroupSameName() {
...
+        Assertions.assertNotNull(constraintViolations, "the constraint violations key is not present");
+        Assertions.assertEquals(1, constraintViolations.size(), "only one error message was expected, but more were found");
+
+        final JsonObject error = constraintViolations.getJsonObject(0);
+        final String errorMessage = error.getString("message");
+
+        Assertions.assertNotNull(errorMessage, "the error message is null");
+        Assertions.assertEquals("either the bundle name or the bundle UUID are required", errorMessage, "unexpected error message received");
+    }
+
```
gwenneg, line 1035: Could you please add a test about the 400 that is returned when the bundle name is passed in the request and the bundle is not found in the DB?

backend/src/main/java/com/redhat/cloud/notifications/routers/models/behaviorgroup/CreateBehaviorGroupRequest.java
MikelAlejoBR: All right, I've removed the `StringUtils` bit :+1:

backend/src/main/java/com/redhat/cloud/notifications/routers/models/behaviorgroup/CreateBehaviorGroupRequest.java
```diff
@@ -35,4 +40,15 @@ public class CreateBehaviorGroupRequest {
      * the behavior group).
      */
     public Set<UUID> eventTypeIds;
+
+    /**
+     * Validates that the bundle is identifiable.
+     * @return true if the bundle ID is null and the bundle name is blank. False otherwise.
```
gwenneg, line 46:
```suggestion
     * @return true if the bundle ID is not null or the bundle name is not blank. False otherwise.
```

backend/src/test/java/com/redhat/cloud/notifications/routers/NotificationResourceTest.java
MikelAlejoBR: Done! On this one there is a minor problem: the persistence layer raises a `NotFoundException` too. That is why in the test I validate both error messages: the one from the handler, and the other from the persistence layer.

I initially started refactoring the handler a bit, with one extra DB call to validate the bundle ID too, but the ID gets validated in the persistence layer as well, so I thought that I was just duplicating work here...

backend/src/main/java/com/redhat/cloud/notifications/routers/models/behaviorgroup/CreateBehaviorGroupRequest.java
MikelAlejoBR: Phew! On this one I've changed the name of the function so that it better reflects what it is supposed to do too, because with the change in the Javadocs the returning result wasn't being honored by what the function was doing...

backend/src/test/java/com/redhat/cloud/notifications/models/behaviorgroup/CreateBehaviorGroupRequestTest.java
```diff
@@ -0,0 +1,115 @@
...
+     * Tests that a valid request doesn't raise any constraint violations.
+     */
+    @Test
+    void testValidRequest() {
+        final var requestBundleId = new CreateBehaviorGroupRequest();
+        requestBundleId.bundleId = UUID.randomUUID();
+        requestBundleId.displayName = "valid display name";
+
+        final var requestBundleName = new CreateBehaviorGroupRequest();
+        requestBundleName.bundleId = UUID.randomUUID();
```
gwenneg, line 37: Shouldn't this one be
```java
        requestBundleName.bundleName= // something
```
?

MikelAlejoBR: You are right, yes! I've fixed the test.

gwenneg (approved):
LGTM

PR #1663, opened 2022-12-23
[RHCLOUD-21842] feature: limit the maximum allowed number of behavior group creations

MikelAlejoBR:
QE has spotted that after a few thousand behavior groups that get created, the frontend's performance degrades. However, having a customer with more than 15 behavior groups configured is going to be a rare case.

That is why we are limiting the number of behavior groups that can be created, since we assume that if more need to be created, there will highly likely be old ones that are not used, and therefore need to be deleted first.

## Links

[[RHCLOUD-21842]](https://issues.redhat.com/browse/RHCLOUD-21842)

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
@@ -81,7 +81,9 @@ public class BehaviorGroupRepository {
 
     @Transactional
     BehaviorGroup create(String accountId, String orgId, BehaviorGroup behaviorGroup, boolean isDefaultBehaviorGroup) {
-        if (!this.isAllowedToCreateMoreBehaviorGroups(orgId)) {
+        // The organization ID might be null if a system behavior group is being
+        // created, that is why the check is only forced for actual tenants.
+        if (orgId != null && !orgId.isBlank() && !this.isAllowedToCreateMoreBehaviorGroups(orgId)) {
             throw new BadRequestException("behavior group creation limit reached. Please consider deleting unused behavior groups before creating more.");
         }
 
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
@@ -29,6 +29,11 @@ import static org.hibernate.jpa.QueryHints.HINT_PASS_DISTINCT_THROUGH;
 @ApplicationScoped
 public class BehaviorGroupRepository {
 
+    /**
+     * Represents the maximum number of behavior groups that can be created for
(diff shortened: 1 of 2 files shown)


MikelAlejoBR:
I am going to convert it to Draft until I run a few queries in stage and production, to see how many accounts we have with more BGs that the allowed ones. I will update the Jira ticket.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
```diff
@@ -555,4 +564,30 @@ private void checkBehaviorGroup(UUID behaviorGroupId, boolean isDefaultBehaviorG
...
+    private boolean isAllowedToCreateMoreBehaviorGroups(final String accountNumber, final String orgId) {
+        final String countQuery =
+            "SELECT " +
+                "COUNT(bg) " +
+            "FROM " +
+                "BehaviorGroup AS bg " +
+            "WHERE " +
+                "(bg.accountId = :account_number OR bg.accountId IS NULL) " +
+            "AND " +
+                "bg.orgId = :org_id";
```
gwenneg, line 584: Let's skip the `accountId` conditions here. Checking the `orgId` is enough.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
```diff
@@ -76,6 +81,10 @@ public UUID getBundleId(String orgId, UUID behaviorGroupId) {
 
     @Transactional
     BehaviorGroup create(String accountId, String orgId, BehaviorGroup behaviorGroup, boolean isDefaultBehaviorGroup) {
+        if (!this.isAllowedToCreateMoreBehaviorGroups(accountId, orgId)) {
+            throw new BadRequestException("behavior group creation limit reached. Please consider deleting unused behavior groups before creating more.");
+        }
```
gwenneg, line 86: The `accountId`/`orgId` may be `null` if we are creating a system behavior group. Let's skip the check and always allow the BG creation in that case.

MikelAlejoBR: All right, done! I haven't written a test because the `testCreateAndUpdateAndDeleteDefaultBehaviorGroup` test already covers the new condition on the check.

MikelAlejoBR:
I think that the PR failures are unrelated. These two runs have yielded the same errors:

- Run [3650](https://ci.ext.devshift.net/job/RedHatInsights-notifications-backend-pr-check/3660/console) for this PR.
- Run [3650](https://ci.ext.devshift.net/job/RedHatInsights-notifications-backend-pr-check/3650/console) for [PR#1732](https://github.com/RedHatInsights/notifications-backend/pull/1732).

gwenneg (approved):
LGTM

PR #1670, opened 2023-01-03
RHCLOUD-22228 add info message text

aferd:


aferd:
adding text to BE to be able to implement UI side.

josejulio:
According to the JIRA ticket:
> I thought we decided you only get one at a time. When it's unchecked, you get the info, and when it's checked, you get the warning!

So we might need to use `isSubscribed` to decide that? 

Also, it looks like this only applies to notifications that are part of a default behavior group.

backend/src/main/java/com/redhat/cloud/notifications/routers/models/SettingsValueJsonForm.java
```diff
@@ -88,7 +88,7 @@ public static SettingsValueJsonForm fromSettingsValue(SettingsValues values) {
                             field.label = "Instant notification";
                             field.description = "Immediate email for each triggered application event. See notification settings for configuration.";
                             field.checkedWarning = "Opting into this notification may result in a large number of emails";
-                            field.infoMessage = "You may still receive forced notifications for this service";
```
josejulio: Ah - not what I meant with the `isSubscribed` .

There is a variable called `isSubscribed` above that tell us if the user is currently subscribed to the event type.
We could use that to determine if we want to fill the `checkedWarning` or the `infoMessage` - that is according to what UX said.

> I thought we decided you only get one at a time. When it's unchecked, you get the info, and when it's checked, you get the warning!

But on second though this could be dynamic thing that happens on the frontend - so maybe not required to really verify that on the backend.

josejulio (approved):

PR #1679, opened 2023-01-06
RHCLOUD-23540 Improve RHOSE and Camel errors handling

gwenneg:


engine/src/main/java/com/redhat/cloud/notifications/db/repositories/NotificationHistoryRepository.java
```diff
@@ -45,28 +45,21 @@ public void createNotificationHistory(NotificationHistory history) {
...
                 .setParameter("status", notificationHistory.getStatus())
                 .setParameter("id", notificationHistory.getId())
                 .setParameter("invocationTime", notificationHistory.getInvocationTime())
                 .executeUpdate();
-
-        if (count == 0) {
-            throw new NoResultException("Update returned no rows");
-        } else if (count > 1) {
-            throw new IllegalStateException("Update count was " + count);
-        }
```
gwenneg, line 67: This could never happen since we're retrieving the records from the primary key.

engine/src/main/java/com/redhat/cloud/notifications/events/FromOpenBridgeHistoryFiller.java
```diff
@@ -103,23 +100,27 @@ public void execute() {
         }
 
         for (ProcessingError pe : items) {
+            try {
```
gwenneg, line 103: Each error needs to be processed in its own try/catch block to make sure the loop is never broken by an exception.

engine/src/main/java/com/redhat/cloud/notifications/events/FromOpenBridgeHistoryFiller.java
```diff
@@ -103,23 +100,27 @@ public void execute() {
...
-            });
+
+                statelessSessionFactory.withSession(statelessSession -> {
+                    Map<String, Object> decodedPayload = decodeItem(pe);
+                    // TODO reinjectIf needed (?)
+                    boolean updated = camelHistoryFillerHelper.updateHistoryItem(decodedPayload);
+                    if (updated) {
+                        messagesWithError.increment();
+                    } else {
+                        Log.infof("RHOSE notification history update failed because no record was found with [id=%s]", historyId);
```
gwenneg, line 118: The old `Log.warn` instruction created noise in Sentry. `Log.info` should be enough.

josejulio (approved):
lgtm

PR #1683, opened 2023-01-10
RHCLOUD-23573 Require channel field for Slack integrations

gwenneg:


backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -229,6 +230,9 @@ public Endpoint createEndpoint(@Context SecurityContext sec,
         }
 
         if (isForOpenBridge(endpoint)) {
+            CamelProperties cp = endpoint.getProperties(CamelProperties.class);
+            checkSlackChannel(cp);
```
josejulio, line 234: We are only supporting slack for now. We will need to refactor this call (and probably there are other slack-specific stuff around in the update).

josejulio (approved):
lgtm

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
gwenneg: Yes, the entire RHOSE code in notifications is tightly-coupled with Slack. We'll refactor everything when another kind of sink will be available.

PR #1684, opened 2023-01-10
RHCLOUD-23576 Remove OB feature flags

gwenneg:
After this PR, `FromOpenBridgeHistoryFiller` will be disabled by default on dev machine but it will keep running every 10 seconds on ephemeral, stage and prod.


g-duval:
LGTM

PR #1685, opened 2023-01-10
[RHCLOUD-22281] feature: don't update a processor if it is not ready to be updated

MikelAlejoBR:
Up to now when a processor was not ready to be updated, we still sent the update request, which would result in an unnecessary call to RHOSE and an error report for us. This should spare RHOSE from receiving that request.

I chose to use an `InternalServerErrorException` because all of the web application exceptions are catched, so sending a `BadRequest` would still be wrapped in a `5xx` response to the client.

## Links

[[RHCLOUD-22281]](https://issues.redhat.com/browse/RHCLOUD-22281)

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
@@ -459,6 +459,26 @@ public class EndpointResource {
             CamelProperties requestProperties = endpoint.getProperties(CamelProperties.class);
             String slackChannel = checkSlackChannel(requestProperties);
 
+            /*
+             * Before the endpoint and its processor get updated, we need to
+             * check that the processor was in an actionable state before
+             * updating it in RHOSE.
+             *
+             * This will achieve two goals:
+             *
+             * 1. Avoid returning the RHOSE error directly to the
+             * user.
+             * 2. Avoid unnecessarily calling RHOSE.
+             *
+             * Check https://issues.redhat.com/browse/RHCLOUD-22281
+             * for more information.
+             */
+            if (dbEndpoint.getStatus() != EndpointStatus.READY && dbEndpoint.getStatus() != EndpointStatus.FAILED) {
+                throw new BadRequestException(
+                    "the processor associated with the endpoint is not ready to be modified"
+                );
+            }
+
             dbEndpoint.setStatus(EndpointStatus.PROVISIONING);
 
             CamelProperties dbProperties = dbEndpoint.getProperties(CamelProperties.class);
--- a/backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
+++ b/backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
@@ -36,6 +36,7 @@ import io.restassured.http.Header;
 import io.restassured.response.Response;
 import io.vertx.core.json.JsonObject;
 import org.eclipse.microprofile.rest.client.inject.RestClient;
+import org.hibernate.Session;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
@@ -47,7 +48,9 @@ import org.mockito.Mockito;
 
 import javax.inject.Inject;
 import javax.persistence.EntityManager;
+import javax.transaction.Transactional;
 import java.util.ArrayList;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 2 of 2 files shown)


backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -458,6 +467,24 @@ public Response updateEndpoint(@Context SecurityContext sec,
...
+                         * goals:
+                         *
+                         * 1. Avoid returning the RHOSE error directly to the
+                         * user.
+                         * 2. Avoid unnecessarily calling RHOSE.
+                         *
+                         * Check https://issues.redhat.com/browse/RHCLOUD-22281
+                         * for more information.
+                         */
+                        if (endpointStatus != EndpointStatus.READY) {
```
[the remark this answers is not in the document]
MikelAlejoBR: All right! I've added that one too :smile:

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -468,6 +476,24 @@ public Response updateEndpoint(@Context SecurityContext sec,
...
+             * 2. Avoid unnecessarily calling RHOSE.
+             *
+             * Check https://issues.redhat.com/browse/RHCLOUD-22281
+             * for more information.
+             */
+            if (previousEndpointStatus != EndpointStatus.READY && previousEndpointStatus != EndpointStatus.FAILED) {
+                throw new InternalServerErrorException(
+                        "the processor associated with the endpoint is not ready to be modified"
+                );
+            }
```
gwenneg, line 495: Can we move that check to line 462 and then avoid storing the status in `previousEndpointStatus`?

gwenneg: Regarding the HTTP error that will be returned, I'd return a 400 with the same message instead of a 500.

MikelAlejoBR: Sure thing! Done!

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
```diff
@@ -697,6 +703,20 @@ void addOpenBridgeEndpoint() {
...
+             * was ready to be updated. The transaction is required for
+             * Hibernate to run the update query.
+             */
+            final Transaction transaction = this.databaseSession.beginTransaction();
+            this.entityManager
+                .createQuery("UPDATE Endpoint SET status = :status WHERE id = :id")
+                .setParameter("status", EndpointStatus.READY)
+                .setParameter("id", UUID.fromString(id))
+                .executeUpdate();
+            transaction.commit();
```
gwenneg, line 718: It would be a bit simpler to run that query from a method (which must NOT be `private`) annotated with `@Transactional`. It would rely on `EntityManager` instead of requiring a `Session` here. What you did works though so I don't mind approving this version.

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
```diff
@@ -2110,6 +2130,101 @@ void testEndpointTestNotFound() {
...
+                 * was being deleted. The transaction is required for Hibernate to
+                 * run the update query.
+                 */
+                final Transaction transaction = this.databaseSession.beginTransaction();
+                this.entityManager
+                        .createQuery("UPDATE Endpoint SET status = :status WHERE id = :id")
+                        .setParameter("status", nonActionableStatus)
+                        .setParameter("id", UUID.fromString(id))
+                        .executeUpdate();
+                transaction.commit();
```
gwenneg, line 2208: Same here.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -459,6 +459,27 @@ public Response updateEndpoint(@Context SecurityContext sec,
             CamelProperties requestProperties = endpoint.getProperties(CamelProperties.class);
             String slackChannel = checkSlackChannel(requestProperties);
 
+            /*
+             * Before the endpoint and its processor get updated, we set the
+             * endpoint's status to "PROVISIONING" since the processor's update
+             * might take some time. That is why we need to check that the
+             * processor was in an actionable state before updating it in RHOSE.
```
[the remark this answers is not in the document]
MikelAlejoBR: Done, thank you!

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
MikelAlejoBR: I did it and it works perfectly, thank you!

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
MikelAlejoBR: Done too!

gwenneg (approved):
LGTM

PR #1691, opened 2023-01-12
[RHCLOUD-23551] update Ansible templates

g-duval:
image-2023-01-12-15-38-43-829


engine/src/main/resources/templates/Ansible/instantEmailBodyV2.html
```diff
@@ -0,0 +1,16 @@
...
+{#content-title-section1}
+    Report available
+{/content-title-section1}
+{#content-body-section1}
+    <p>Ansible report available for download.</p>
+{/content-body-section1}
+{#content-button-section1}
+    <a href="{environment.url}/ansible/insights/reports/{action.context.slug}">Download now</a>
+{/content-button-section1}
+{/include}
```
josejulio, line 16: I :heart:  that this one is so simple

engine/src/main/java/com/redhat/cloud/notifications/templates/TemplateService.java
```diff
@@ -25,6 +25,8 @@
 @ApplicationScoped
 public class TemplateService {
 
+    public static final String HCC_LOGO_TARGET = "Logo-Red_Hat-Hybrid_Cloud_Console-A-Reverse-RGB";
```
josejulio, line 28: This one is only used in the tests, could we move the variable there? (test code class if it's used by multiple tests)

josejulio (changes requested):

engine/src/main/java/com/redhat/cloud/notifications/templates/TemplateService.java
g-duval: You are right, I moved it into TestHelper class which is much more appropriate.

engine/src/main/resources/templates/Ansible/instantEmailBodyV2.html
g-duval: Thanks

josejulio (approved):
lgtm

engine/src/test/java/com/redhat/cloud/notifications/TestHelpers.java
```diff
@@ -327,4 +327,25 @@ public static Action createComplianceAction() {
 
         return emailActionMessage;
     }
+
+    public static Action createAnsibleAction(String slug) {
```
gwenneg, line 331: image

PR #1694, opened 2023-01-16
[RHCLOUD-23628] update compliance templates

g-duval:
image

image

image

### DIFF
--- a/common/src/main/java/com/redhat/cloud/notifications/config/FeatureFlipper.java
+++ b/common/src/main/java/com/redhat/cloud/notifications/config/FeatureFlipper.java
@@ -72,6 +72,9 @@ public class FeatureFlipper {
     @ConfigProperty(name = "notifications.use-policies-email-templates-v2.enabled", defaultValue = "false")
     boolean policiesEmailTemplatesV2Enabled;
 
+    @ConfigProperty(name = "compliance.use-policies-email-templates-v2.enabled", defaultValue = "false")
+    boolean complianceEmailTemplatesV2Enabled;
+
     void logFeaturesStatusAtStartup(@Observes StartupEvent event) {
         Log.infof("=== %s startup status ===", FeatureFlipper.class.getSimpleName());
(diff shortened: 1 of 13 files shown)


common/src/main/java/com/redhat/cloud/notifications/config/FeatureFlipper.java
```diff
@@ -72,6 +72,9 @@ public class FeatureFlipper {
     @ConfigProperty(name = "notifications.use-policies-email-templates-v2.enabled", defaultValue = "false")
     boolean policiesEmailTemplatesV2Enabled;
 
+    @ConfigProperty(name = "compliance.use-policies-email-templates-v2.enabled", defaultValue = "false")
```
josejulio, line 75: I suppose you mean:

```suggestion
    @ConfigProperty(name = "notifications.use-compliance-email-templates-v2.enabled", defaultValue = "false")
```

engine/src/main/java/com/redhat/cloud/notifications/templates/TemplateService.java
```diff
@@ -25,6 +25,8 @@
 @ApplicationScoped
 public class TemplateService {
 
+    public static final String HCC_LOGO_TARGET = "Logo-Red_Hat-Hybrid_Cloud_Console-A-Reverse-RGB";
```
josejulio, line 28: Same as the other PR - Lets move this variable to test/**

josejulio (changes requested):

common/src/main/java/com/redhat/cloud/notifications/config/FeatureFlipper.java
g-duval: oups :)

engine/src/main/java/com/redhat/cloud/notifications/templates/TemplateService.java
g-duval: Yep, I moved it

josejulio (approved):
lgtm

PR #1695, opened 2023-01-18
[RHCLOUD-23629] Update Cost Management templates

g-duval:
image

image

image

image

image

image

image


engine/src/main/java/com/redhat/cloud/notifications/templates/TemplateService.java
```diff
@@ -25,6 +25,8 @@
 @ApplicationScoped
 public class TemplateService {
 
+    public static final String HCC_LOGO_TARGET = "Logo-Red_Hat-Hybrid_Cloud_Console-A-Reverse-RGB";
```
josejulio, line 28: Same as the others - lets move this to tests.

josejulio (changes requested):
Great work!

I just have the nitpick about the variable being used only by tests

engine/src/main/java/com/redhat/cloud/notifications/templates/TemplateService.java
g-duval: Sure, I did it

josejulio (approved):
lgtm

PR #1696, opened 2023-01-18
[RHCLOUD-23722] fix: Sources not receiving the "x-rh-identity header" from the engine

MikelAlejoBR:
Sources was returning 401 responses when the engine was trying to fetch the related secrets to an endpoint. This was happening because the Sources REST Service class was configured to forward any "x-rh-identity" header received from the client, but in the case of the engine, since there was no header to forward, the request was being sent without any header.

By switching to an explicit header specification for the REST Service class, we force the caller to provide the proper "x-rh-identity" header's content. Also, the secret utilities have been updated to honor this interface.

Finally, the tests have been updated as well to make sure that the secret utilities are sending the headers in every single case.

## Links

[[RHCLOUD-23722]](https://issues.redhat.com/browse/RHCLOUD-23722)


common/src/main/java/com/redhat/cloud/notifications/XRhIdentityUtils.java
```diff
@@ -0,0 +1,30 @@
...
+     * @param accountId the EBS account number to place in the contents.
+     * @param orgId the organization id to place in the contents.
+     * @return a base64 encoded string.
+     */
+    public static String generateEncodedXRhIdentity(final String accountId, final String orgId) {
+        final JsonObject identity = new JsonObject();
+
+        if (accountId != null && !accountId.isBlank()) {
+            identity.put("account_number", accountId);
+        }
```
gwenneg, line 19: Is the `account_number` field needed in Sources? Otherwise, it'd be great if we could remove it here because that data may be removed from the `endpoints` table in the DB soon.

common/src/main/java/com/redhat/cloud/notifications/XRhIdentityUtils.java
```diff
@@ -0,0 +1,30 @@
...
+
+        if (orgId != null && !orgId.isBlank()) {
```
gwenneg, line 21: The `org_id` field is probably mandatory in Sources. If so, let's throw an exception if it is `null` or blank in this method.

common/src/test/java/com/redhat/cloud/notifications/XRhIdentityUtilsTest.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Tests that when generating an encoded x-rh-identity content with both
+     * an account number and an org id, the content is encoded as expected.
+     */
+    @Test
+    void generateEncodedXRhIdentity() {
+        final String got = XRhIdentityUtils.generateEncodedXRhIdentity("ebs-account-number", "organization-id");
+
+        final String expectedValue = "eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6ImVicy1hY2NvdW50LW51bWJlciIsIm9yZ19pZCI6Im9yZ2FuaXphdGlvbi1pZCJ9fQ==";
+        Assertions.assertEquals(expectedValue, got, "the x-rh-identity contents were not correctly base64 encoded");
```
gwenneg, line 17: For all tests in this class, you could also check that the result of `Base64Utils.decode(got)` contains the expected JSON.

common/src/main/java/com/redhat/cloud/notifications/XRhIdentityUtils.java
MikelAlejoBR: Sure, let's do that! Sources also prioritizes the organization ID, so I?m all for removing the account number.

common/src/main/java/com/redhat/cloud/notifications/XRhIdentityUtils.java
MikelAlejoBR: All right, done!

common/src/test/java/com/redhat/cloud/notifications/XRhIdentityUtilsTest.java
MikelAlejoBR: I've added the decoding testing as well in the test!

gwenneg (approved):
LGTM

PR #1699, opened 2023-01-19
[RHCLOUD-23752] refactor: specify the environment's URL via an environment variable

MikelAlejoBR:
The idea behind this change is to allow developers to customize the environment's URL that is used across notifications. By doing this all the notifications can use the specified URL instead of relying on the hard coded values.

## Links

[[RHCLOUD-23752]](https://issues.redhat.com/browse/RHCLOUD-23752)

### DIFF
--- a/.rhcicd/clowdapp-engine.yaml
+++ b/.rhcicd/clowdapp-engine.yaml
@@ -217,6 +217,9 @@ parameters:
 - name: CPU_REQUEST
   description: CPU request
   value: 500m
+- name: ENV_BASE_URL
+  default: https://console.redhat.com
+  description: The environment's base URL
 - name: ENV_NAME
   description: ClowdEnvironment name (ephemeral, stage, prod)
   required: true
--- a/engine/src/main/java/com/redhat/cloud/notifications/templates/models/Environment.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/templates/models/Environment.java
@@ -10,19 +10,14 @@ public class Environment {
     @ConfigProperty(name = "env.name", defaultValue = "local-dev")
     String environment;
 
+    @ConfigProperty(name = "env.base.url", defaultValue = "/")
+    String url;
+
     public String name() {
         return this.environment;
     }
 
     public String url() {
-
-        switch (environment) {
-            case "prod":
-                return "https://console.redhat.com";
-            case "stage":
-                return "https://console.stage.redhat.com";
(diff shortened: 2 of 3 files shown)


MikelAlejoBR:
> Looks good to me, but grogu aka baby yoda is sad :'( We will need to update stage environment's url

Let me create an MR...

josejulio:
> > Looks good to me, but grogu aka baby yoda is sad :'( We will need to update stage environment's url
> 
> Let me create an MR...

Just to confirm, I don't think the ci tests failed because of the missing stage's url - but it's something we will have to do.

.rhcicd/clowdapp-engine.yaml
```diff
@@ -217,6 +217,10 @@ parameters:
 - name: CPU_REQUEST
   description: CPU request
   value: 500m
+- name: ENV_BASE_URL
+  default: https://console.redhat.com
+  description: The environment's base URL
+  required: true
```
gwenneg, line 223: This is not needed since we have a default value in the Java code.
```suggestion
```

engine/src/main/java/com/redhat/cloud/notifications/templates/models/Environment.java
```diff
@@ -10,19 +10,14 @@ public class Environment {
     @ConfigProperty(name = "env.name", defaultValue = "local-dev")
     String environment;
 
+    @ConfigProperty(name = "env.base.url", defaultValue = "https://console.redhat.com")
```
gwenneg, line 13: Should we keep the previous default value which was `/` instead of using `https://console.redhat.com`?
```suggestion
    @ConfigProperty(name = "env.base.url", defaultValue = "/")
```

.rhcicd/clowdapp-engine.yaml
MikelAlejoBR: True! Removed!

engine/src/main/java/com/redhat/cloud/notifications/templates/models/Environment.java
MikelAlejoBR: Switched back to the previous default value!

gwenneg:

gwenneg (approved):
LGTM

gwenneg:
@josejulio Do you want to take another look before the merge?

josejulio (approved):

PR #1700, opened 2023-01-20
[RHCLOUD-23853] Update default email template

g-duval:
image


josejulio (approved):
lgtm

gwenneg:
No feature flag for that one?

g-duval:
> 

No because as I undersood, this default template usage is already behind a feature flag enabled on stage only.
It should never be sent as it is to a real enduser.

gwenneg:
Makes sense 👍

PR #1702, opened 2023-01-20
RHCLOUD-23553 Add function to check if behavior group is forcing emai…

josejulio:
…l to users

PR #1703, opened 2023-01-20
Display warning on apps forcing emails

josejulio:
Follow up to https://github.com/RedHatInsights/notifications-backend/pull/1702 to only show the message about "forced emails" in applications that have forced emails


aferd (approved):
lgtm!

PR #1707, opened 2023-01-23
RHCLOUD-23885 Simplify RHOSE processors create/update/delete operations

gwenneg:
This should make the migration to V2 of the RHOSE REST API easier.


MikelAlejoBR (approved):
Nice refactor! It is way easier to understand now! :smile:

PR #1708, opened 2023-01-23
feature: debug what "x-rh-identity" header is being sent to Sources

MikelAlejoBR:
Let's see if this helps understanding why Sources is returning 401s for Notifications.

## Links

[[HCLOUD-23722]](https://issues.redhat.com/browse/RHCLOUD-23722)

### DIFF
--- a/common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
+++ b/common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
@@ -10,6 +10,7 @@ import org.eclipse.microprofile.rest.client.inject.RestClient;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
+import java.util.UUID;
 
 @ApplicationScoped
 public class SecretUtils {
@@ -33,8 +34,13 @@ public class SecretUtils {
 
             final Long basicAuthSourcesId = props.getBasicAuthenticationSourcesId();
             if (basicAuthSourcesId != null) {
+                final String xRhIdentityContents = XRhIdentityUtils.generateEncodedXRhIdentity(endpoint.getOrgId());
+
+                // Log the org ID and the contents of the x-rh-identity header we are about to send.
+                Log.debugf("[endpoint_id: %s][org_id: %s] the encoded x-rh-identity contents are '%s'", endpoint.getId(), endpoint.getOrgId(), xRhIdentityContents);
+
                 final Secret secret = this.sourcesService.getById(
-                    XRhIdentityUtils.generateEncodedXRhIdentity(endpoint.getOrgId()),
+                    xRhIdentityContents,
                     basicAuthSourcesId
                 );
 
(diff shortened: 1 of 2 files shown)


gwenneg (approved):
LGTM

common/src/main/java/com/redhat/cloud/notifications/XRhIdentityUtils.java
```diff
@@ -21,6 +22,10 @@ public static String generateEncodedXRhIdentity(final String orgId) {
         final JsonObject json = new JsonObject();
         json.put("identity", identity);
 
-        return Base64Utils.encode(json.toString());
+        // Log the org ID and the contents of the x-rh-identity header we are about to send..
+        final String generatedXRhIdentityContents = Base64Utils.encode(json.toString());
+        Log.debugf("the encoded org ID is '%s', and the encoded x-rh-identity contents are '%s'", orgId, generatedXRhIdentityContents);
```
gwenneg, line 27: Quarkus doesn't seem happy with `Log` in this class. Maybe that's because the method is `static`.

You can replace `Log` with a `org.jboss.logging.Logger`:
```java
private static final Logger LOGGER = Logger.getLogger(XRhIdentityUtils.class);
```

gwenneg: My bad, the failure comes from the test: it is not a `@QuarkusTest` so `Log` doesn't work. Building the `Logger` as suggested above should work.

MikelAlejoBR: Arg, I'm sorry. I thought that it would be interesting to log the endpoint's ID too and I ended up adding the logs to the secrets utils... I don't know if you're okay with it or you prefer the other version

MikelAlejoBR: I'm facing the same error, let me fix it

engine/src/main/resources/application.properties
```diff
@@ -140,3 +140,6 @@ quarkus.rest-client.sources.url=${clowder.endpoints.sources-api-svc:http://local
 
 quarkus.opentelemetry.enabled=false
 quarkus.opentelemetry.tracer.exporter.otlp.endpoint=http://localhost:4317
+
+# Set the debug level of the x-rh-identity contents generator to debug potential integration issues with Sources.
+quarkus.log.category."com.redhat.cloud.notifications.XRhIdentityUtils".level=DEBUG
```
gwenneg, line 145: The class needs to be changed here.

common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
```diff
@@ -235,16 +282,22 @@ private long createBasicAuthentication(final String orgId, final BasicAuthentica
...
         secret.authenticationType = Secret.TYPE_SECRET_TOKEN;
         secret.password = secretToken;
 
+        final String xRhIdentityContents = XRhIdentityUtils.generateEncodedXRhIdentity(orgId);
+
+        // Log the org ID and the contents of the x-rh-identity header we are about to send.
+        Log.debugf("[endpoint_id: %s][org_id: %s] the encoded x-rh-identity contents are '%s'", endpointId, orgId, xRhIdentityContents);
+
         final Secret createdSecret = this.sourcesService.create(
             XRhIdentityUtils.generateEncodedXRhIdentity(orgId),
```
gwenneg, line 302:
```suggestion
            xRhIdentityContents,
```

engine/src/main/resources/application.properties
MikelAlejoBR: You're right. Done!

common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
MikelAlejoBR: Nice catch!

gwenneg (approved):
LGTM

PR #1710, opened 2023-01-24
RHCLOUD-23546 Add temporary internal endpoint to delete behavior groups

gwenneg:
This PR introduces an internal endpoint that will be used to delete all behavior groups from a few internal accounts that will not be disclosed here.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/internal/BehaviorGroupCleanupResource.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/internal/BehaviorGroupCleanupResource.java
@@ -1,0 +1,67 @@
+package com.redhat.cloud.notifications.routers.internal;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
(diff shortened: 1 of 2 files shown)


backend/src/main/java/com/redhat/cloud/notifications/routers/internal/BehaviorGroupCleanupResource.java
```diff
@@ -0,0 +1,67 @@
...
+    @Inject
+    EntityManager entityManager;
+
+    @PUT
+    @Path("/{orgId}/initiate")
+    public void initiate(@RestPath String orgId) {
+        Log.warnf("If you are not sure of what you are doing, STOP NOW! " +
+                "Behavior groups cleanup is enabled for 1 minute for orgId=%s. " +
+                "Nothing has been deleted yet from the DB. " +
+                "Call the confirmation endpoint to delete all behavior groups from orgId=%s.", orgId, orgId);
```
josejulio, line 46: It's probably worth to return this string to the browser/client too

josejulio (approved):
lgtm

PR #1712, opened 2023-01-25
RHCLOUD-23647 Introduce emails only mode

gwenneg:
From now on, if `NOTIFICATIONS_EMAILS_ONLY_MODE_ENABLED` is set to `true` in the environment variables, Notifications will:
- not process any endpoint type other than `EMAIL_SUBSCRIPTION`
- not allow to create, update, delete, enable or disable an endpoint if its type is not `EMAIL_SUBSCRIPTION`


gwenneg:

backend/src/test/java/com/redhat/cloud/notifications/routers/EmailsOnlyModeTest.java
```diff
@@ -0,0 +1,170 @@
...
+                .basePath(API_INTEGRATIONS_V_1_0)
+                .header(identityHeader)
+                .when()
+                .contentType(JSON)
+                .body(Json.encode(endpoint))
+                .post("/endpoints")
+                .then()
+                .statusCode(400)
+                .extract().asString();
+        assertEquals(UNSUPPORTED_ENDPOINT_TYPE, responseBody);
```
josejulio, line 65: if this assert fails, it won't process the `featureFlipper.setEmailsOnlyMode(false);` which might break more tests, making it harder to track it down to the real error.
Could you wrap this into a `try {} finally { featureFlipper.setEmailsOnlyMode(false); }`  or have a `afterEach` setting the mode to `false`?

This applies for all the other tests.

josejulio (changes requested):
Is there any way to disable the `TOCAMEL_CHANNEL` and `FROMCAMEL_CHANNEL` ? 
Else we would need to create them even if we are not using them. Or maybe that's something that can be done already from the ENVs?

backend/src/test/java/com/redhat/cloud/notifications/routers/EmailsOnlyModeTest.java
gwenneg: Good catch, fixed!

gwenneg:
> Is there any way to disable the `TOCAMEL_CHANNEL` and `FROMCAMEL_CHANNEL` ? Else we would need to create them even if we are not using them. Or maybe that's something that can be done already from the ENVs?

Very good question! I added the required param to the ClowdApp template.

josejulio (approved):

gwenneg:
I renamed the `fromCamel` and `toCamel` channels in the configuration because otherwise the env vars wouldn't work as expected. Now everything works, I tested it locally.

PR #1715, opened 2023-01-25
[RHCLOUD-23939] Update Rbac label to User Access on email templates

g-duval:
Rbac label have to be updated to Access User on email templates


engine/src/main/resources/templates/Rbac/customGroupCreatedEmailBody.html
```diff
@@ -1,5 +1,6 @@
+
```
josejulio, line 1: Is the new line intentional? 

```suggestion

```

g-duval: Nop :)

josejulio (approved):
lgtm

PR #1718, opened 2023-01-26
Make EndpointReadyChecker resilient to all kinds of exceptions

gwenneg:
When a NPE is thrown during the processing of one of our endpoints, the scheduled job is interrupted immediately while it should process the other endpoints.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointReadyChecker.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointReadyChecker.java
@@ -74,10 +74,11 @@ public class EndpointReadyChecker {
                 .setHint("javax.persistence.lock.timeout", SKIP_LOCKED)
                 .getResultList();
 
+        String processorId = null;
         for (Endpoint ep : endpoints) {
             try {
                 CamelProperties cp = em.find(CamelProperties.class, ep.getId()); // TODO Fetch in one go
-                String processorId = cp.getExtras().get("processorId");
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointReadyChecker.java
```diff
@@ -75,33 +75,37 @@ public void execute() {
...
+                        em.remove(ep);
+                    } else {
+                        String path = "GET " + BASE_PATH + "/{bridgeId}/processors/{processorId}";
+                        rhoseErrorMetricsRecorder.record(path, wae);
+                        Log.warn("Getting data from OB failed", wae);
+                        ep.setStatus(EndpointStatus.FAILED);
+                    }
                 }
+            } catch (Exception e) {
+                Log.error("Could not retrieve the status of a RHOSE processor", e);
```
MikelAlejoBR, line 108: Would it be possible to maybe get the processor ID outside the `try {/* ... */}` block to then log it along with the error message? Maybe it is not required, it is just a suggestion :+1:

MikelAlejoBR (approved):
LGTM!

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointReadyChecker.java
gwenneg: That's a good suggestion, thanks! I've updated the code a bit.

PR #1721, opened 2023-01-26
[RHCLOUD-23966] feature: implement PSK authentication with Sources

MikelAlejoBR:
The initial implementation simply forwarded the received x-rh-identity header to Sources. This worked for the backend, but not the engine, since the engine doesn't have a client that sends such header.

The second implementation generated a minimal x-rh-identity header with an org-id inside of it, but it also didn't work because RBAC expects an username to identify the user. This also had the drawback of the user needing the Sources administrator role.

This implementation uses the PSK to authenticate to Sources, and then a second header so that Sources knows to which tenants the operations tie to.

## Links

[[RHCLOUD-23966]](https://issues.redhat.com/browse/RHCLOUD-23966)


.rhcicd/clowdapp-backend.yaml
```diff
@@ -207,6 +207,11 @@ objects:
           value: ${QUARKUS_REST_CLIENT_KC_URL}
         - name: NOTIFICATIONS_USE_SOURCES_SECRETS_BACKEND
           value: ${SOURCES_ENABLED}
+        - name: SOURCES_PSK
+          valueFrom:
+            secretKeyRef:
+              name: sources-psk
+              key: psk
```
MikelAlejoBR, line 214: I have named the secret like this because this might be useful in ephemeral, because Sources creates a Secret of the same name:

https://github.com/RedHatInsights/sources-api-go/blob/main/deploy/clowdapp.yaml#L5-L15

By naming the secret like this, Notifications can simply pick it up from the one that Sources creates. Although we can also use another name and key combination for the secret and simply write the PSK that Sources defined in their ClowdApp.

What do you think?

backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTestTestProfile.java
```diff
@@ -0,0 +1,20 @@
+package com.redhat.cloud.notifications.routers.sources;
+
+import io.quarkus.test.junit.QuarkusTestProfile;
+
+import java.util.Map;
+
+public class SecretUtilsTestTestProfile implements QuarkusTestProfile {
```
gwenneg, line 7: There's a drawback with Quarkus test profiles: they trigger one (sometimes two) full Quarkus restart (including the whole Flyway migration) during the tests execution, which leads to longer tests and more resources consumed.

The `sources.psk` config value could be set with a simple `System.setProperty("sources.psk", SOURCES_PSK_TEST);` at the beginning of the tests execution. Quarkus should be happy with that.

backend/src/main/resources/application.properties
```diff
@@ -91,3 +91,8 @@ quarkus.rest-client.sources.url=${clowder.endpoints.sources-api-svc:http://local
 # OpenTelemetry -- see also jdbc driver above.
 quarkus.opentelemetry.enabled=false
 quarkus.opentelemetry.tracer.exporter.otlp.endpoint=http://localhost:4317
+
+# Sources' development PSK value. Specified here to avoid Quarkus from complaining that the configuration parameter
+# is missing. In the case that you are using a real Sources application to test the integration, you will need to
+# set this PSK in the SOURCES_PSKS environment variable on that end.
+sources.psk=development-value-123
```
gwenneg, line 98: When does Quarkus complain exactly? Is it during the tests execution, in dev mode, in prod mode?

engine/src/main/resources/application.properties
```diff
@@ -141,5 +141,7 @@ quarkus.rest-client.sources.url=${clowder.endpoints.sources-api-svc:http://local
 quarkus.opentelemetry.enabled=false
 quarkus.opentelemetry.tracer.exporter.otlp.endpoint=http://localhost:4317
 
-# Set the debug level of secret utils to debug potential issues with Sources.
-quarkus.log.category."com.redhat.cloud.notifications.routers.sources.SecretUtils".level=DEBUG
+# Sources' development PSK value. Specified here to avoid Quarkus from complaining that the configuration parameter
+# is missing. In the case that you are using a real Sources application to test the integration, you will need to
+# set this PSK in the SOURCES_PSKS environment variable on that end.
+sources.psk=development-value-123
```
gwenneg, line 147: Same question.

.rhcicd/clowdapp-backend.yaml
gwenneg: Your link shows a slightly different name: `internal-psk`. Did I miss something or do we need to rename our secret?

backend/src/main/resources/application.properties
MikelAlejoBR: When running with `quarkus:dev` Quarkus doesn't start unless this value is specified, since I didn't provide a default value for the `@ConfigProperty`. The same happens when running tests, the dependency injection starts going a bit crazy, and it happens when I comment the property and recompile the project.

backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTestTestProfile.java
MikelAlejoBR: All right, I'm switching to that...

.rhcicd/clowdapp-backend.yaml
MikelAlejoBR: I am really surprised that I didn't even manage to copy and paste properly. You're absolutely right, its name is `internal-psk`.

.rhcicd/clowdapp-backend.yaml
```diff
@@ -207,6 +207,11 @@ objects:
           value: ${QUARKUS_REST_CLIENT_KC_URL}
         - name: NOTIFICATIONS_USE_SOURCES_SECRETS_BACKEND
           value: ${SOURCES_ENABLED}
+        - name: SOURCES_PSK
+          valueFrom:
+            secretKeyRef:
+              name: internal-psk
```
gwenneg, line 213: Thanks for fixing the name. Now it raises another question: is `internal-psk` good enough to be put in the Notifications vault in stage and prod? From the vault alone, it will be impossible to understand that this secret is related to Sources. Of course the answer can be found in this ClowdApp template, but for a maintainability perspective this is not ideal... WDYT?

gwenneg: The new name sounds better to me 👍

Ephemeral can rely on secrets from a different environment but I won't put any more details here as this is not the right place to discuss that 😄 Let's continue elsewhere.

MikelAlejoBR: Perfect! After talking about it, we can consider this thread dealt with, then! Thanks :)

gwenneg (approved):
LGTM

gwenneg:
From the pod in ephemeral:
> Error: secret "sources-api-psk" not found

MikelAlejoBR:
It is reporting that the `sources-api-psk` secret is not present. I'm looking into it.

gwenneg (approved):
LGTM

PR #1724, opened 2023-01-30
[RHCLOUD-23993] feature: turn Sources ClowdApp into an optional dependency

MikelAlejoBR:
Another team needs to run Notifications without Sources, and they can't do that if Sources is listed as a mandatory dependency.

## Links

[[RHCLOUD-23993]](https://issues.redhat.com/browse/RHCLOUD-23993)

### DIFF
--- a/.rhcicd/clowdapp-backend.yaml
+++ b/.rhcicd/clowdapp-backend.yaml
@@ -15,7 +15,8 @@ objects:
     dependencies:
     - notifications-engine
     - rbac
-    - sources-api
+    optionalDependencies:
+    - sources-api # https://issues.redhat.com/browse/RHCLOUD-23993
     database:
       name: notifications-backend
       version: 13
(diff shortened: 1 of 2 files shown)


.rhcicd/clowdapp-backend.yaml
```diff
@@ -16,6 +16,8 @@ objects:
     - notifications-engine
     - rbac
     - sources-api
```
g-duval, line 18: Sources-api is now prensent on optional dependencues, but it is still present on required dependency as well.
Is it a simple typo ?

MikelAlejoBR: Oops! Nice catch :fishing_pole_and_fish: ! It should only be listed in the optional dependencies. Thank you!

gwenneg (approved):
LGTM

gwenneg:

PR #1726, opened 2023-01-31
RHCLOUD-23956 Fix RHOSE processor update

gwenneg:


backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -484,9 +484,11 @@ public Response updateEndpoint(@Context SecurityContext sec,
                 Log.error("RHOSE processor update failed", e);
                 return Response.serverError().build();
             }
-        }
 
-        endpointRepository.updateEndpoint(endpoint);
+            endpointRepository.updateEndpoint(dbEndpoint);
+        } else {
+            endpointRepository.updateEndpoint(endpoint);
```
gwenneg, line 490: This isn't pretty but at least this code does what we need. We may refactor the entire endpoint update logic and replace it with something cleaner later...

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointReadyChecker.java
```diff
@@ -79,6 +84,13 @@ public void execute() {
             try {
                 CamelProperties cp = em.find(CamelProperties.class, ep.getId()); // TODO Fetch in one go
                 processorId = cp.getExtras().get("processorId");
+
+                // TODO Temporary code meant to clean the DB from stage only. Remove it as soon as it's been executed once.
+                if ("stage".equals(environment) && processorId == null) {
+                    em.remove(ep);
+                    continue;
+                }
```
gwenneg, line 92: Because of my tests on stage, a few Slack endpoints lost their `processorid` field in the DB. They can't be updated anymore and are stuck forever in this job. This will remove them from the DB.

MikelAlejoBR (approved):
LGTM!

PR #1732, opened 2023-02-01
[RHCLOUD-23848] feature: limit and order the returned rules for Advisor aggregations

MikelAlejoBR:
The Advisor team requested that the aggregation emails should only return the 5 most critical rules received that day, instead of all of the rules that might have been triggered.

## Links
[[RHCLOUD-23848]](https://issues.redhat.com/browse/RHCLOUD-23848)

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
@@ -3,9 +3,12 @@ package com.redhat.cloud.notifications.processors.email.aggregators;
 import com.redhat.cloud.notifications.models.EmailAggregation;
 import io.vertx.core.json.JsonObject;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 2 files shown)


gwenneg:
@MikelAlejoBR The aggregation logic was updated in a bit in another PR, this PR will need a rebase.

MikelAlejoBR:
> @MikelAlejoBR The aggregation logic was updated in a bit in another PR, this PR will need a rebase.

All right, done! I was thinking about refactoring this, though, and using a list instead of the maps we are using right now for the different recommendations. I don't know if I should maybe wait and suggest a refactor in a different PR. That way we could fulfill the client's request on this one, and improve and simplify the code in the following one.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
```diff
@@ -220,4 +243,42 @@ private void countRecommendationsAndIncidents(String ruleIncident) {
...
+        });
+
+        // Create the resulting map we are about to return. Make it a linked
+        // hash map to preserve the insertion order.
+        final LinkedHashMap<String, Map<String, Object>> resultMap = new LinkedHashMap<>(MAXIMUM_NUMBER_RETURNED_EVENTS);
+
+        // Populate the new map.
+        for (final Map.Entry<String, Map<String, Object>> entry : sortedEntries) {
+            if (resultMap.size() > MAXIMUM_NUMBER_RETURNED_EVENTS - 1) {
+                break;
```
gwenneg, line 274: Is the Advisor team aware that some data, which may be important (high risk), will be dropped?

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
```diff
@@ -220,4 +243,42 @@ private void countRecommendationsAndIncidents(String ruleIncident) {
...
+
+            // The order is reversed because we want the list to be in
+            // descending order, since we want the elements sorted from the
+            // highest risk to the lowest one.
+            return o2TotalRisk.compareTo(o1TotalRisk);
+        });
+
+        // Create the resulting map we are about to return. Make it a linked
+        // hash map to preserve the insertion order.
+        final LinkedHashMap<String, Map<String, Object>> resultMap = new LinkedHashMap<>(MAXIMUM_NUMBER_RETURNED_EVENTS);
```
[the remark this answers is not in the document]
MikelAlejoBR: Very interesting! I have modified the code as suggested.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregator.java
MikelAlejoBR: Let me forward this question to them, I will answer as soon as they reply.

MikelAlejoBR: They just confirmed that this behavior is the desired one :man_dancing:

gwenneg: Thanks for asking!

engine/src/test/java/com/redhat/cloud/notifications/processors/email/aggregators/AdvisorEmailAggregatorTest.java
```diff
@@ -175,4 +185,206 @@ void testAggregate() {
...
+        testRules.add(testRule4);
+        testRules.add(testRule2);
+        testRules.add(testRule1);
+        testRules.add(testRule3);
+        testRules.add(testRule5);
+
+        // Test that the new recommendations are equivalent to the test rules
+        // and that they are in the correct order.
+        final List<Map.Entry<String, Map<String, Object>>> newRecommendationsCollection = new ArrayList<>(newRecommendations.entrySet());
+        this.assertRulesAreEqualAndCorrectOrder(testRules, newRecommendationsCollection);
```
gwenneg, line 292: This works because `testRule` 1 to 5 all have a different risk value. If two rules have the same risk value, their order will be unpredictable. Similarly, if there are 6 rules with the same highest risk value, we can't predict which one will be excluded from the aggregation result. I'm not sure this is a problem though. The current test looks good enough to me.

gwenneg (approved):
LGTM

PR #1737, opened 2023-02-03
[RHCLOUD-23717] feature: limit behavior groups name to 150 characters

MikelAlejoBR:
The front end already has that limit established, so this needs to be done in the back end as well so that direct requests are also limited in this sense.

## Links
[[RHCLOUD-23717]](https://issues.redhat.com/browse/RHCLOUD-23717)

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/models/behaviorgroup/UpdateBehaviorGroupRequest.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/models/behaviorgroup/UpdateBehaviorGroupRequest.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 
+import javax.validation.constraints.AssertFalse;
 import javax.validation.constraints.Size;
 import java.util.List;
 import java.util.Set;
@@ -31,4 +32,16 @@ public class UpdateBehaviorGroupRequest {
      * This will effectively set the linked event types..
      */
     public Set<UUID> eventTypeIds;
+
+    /**
+     * Validates that the display name is not blank when it is provided. Since
+     * this is a potential payload for the "update" operation, clients might
+     * not even send the display name if they don't want to update it. That is
+     * why we let "null" display names go through.
+     * @return true if the display name is not null, but it is blank.
+     */
+    @AssertFalse(message = "the display name cannot be empty")
+    private boolean isDisplayNameNotNullAndBlank() {
+        return this.displayName != null && this.displayName.isBlank();
+    }
 }
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/models/behaviorgroup/UpdateBehaviorGroupRequest.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/models/behaviorgroup/UpdateBehaviorGroupRequest.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 
+import javax.validation.constraints.Size;
 import java.util.List;
 import java.util.Set;
 import java.util.UUID;
@@ -15,6 +16,7 @@ public class UpdateBehaviorGroupRequest {
     /**
      * If not null, changes the display name of the behavior group to this value.
      */
+    @Size(max = 150, message = "the display name cannot exceed {max} characters")
     public String displayName;
 
     /**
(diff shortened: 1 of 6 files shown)


backend/src/main/java/com/redhat/cloud/notifications/routers/models/behaviorgroup/UpdateBehaviorGroupRequest.java
```diff
@@ -15,6 +16,7 @@ public class UpdateBehaviorGroupRequest {
     /**
      * If not null, changes the display name of the behavior group to this value.
      */
+    @Size(max = 150, message = "the display name cannot exceed {max} characters")
```
gwenneg, line 19: We should probably make the display name `@NotBlank` here to be consistent with `CreateBehaviorGroupRequest`.

backend/src/test/java/com/redhat/cloud/notifications/routers/NotificationResourceTest.java
```diff
@@ -1081,6 +1083,106 @@ void testNotFoundBehaviorGroupNotExists() {
...
+
+        // Create the fixtures in the database.
+        final Bundle bundle = this.helpers.createBundle(TEST_BUNDLE_NAME, "Bundle-display-name");
+        final BehaviorGroup behaviorGroup = this.helpers.createBehaviorGroup(accountId, orgId, "valid display name", bundle.getId());
+
+        // Get the value of the "max" property for the "Size" annotation.
+        final Field classField = UpdateBehaviorGroupRequest.class.getDeclaredField("displayName");
+        final Size sizeClassAnnotation = classField.getAnnotation(Size.class);
+
+        final UpdateBehaviorGroupRequest createBehaviorGroupRequest = new UpdateBehaviorGroupRequest();
```
gwenneg, line 1153:
```suggestion
        final UpdateBehaviorGroupRequest updateBehaviorGroupRequest = new UpdateBehaviorGroupRequest();
```

backend/src/test/java/com/redhat/cloud/notifications/routers/NotificationResourceTest.java
```diff
@@ -1081,6 +1083,106 @@ void testNotFoundBehaviorGroupNotExists() {
...
+                .header(identityHeader)
+                .when()
+                .contentType(JSON)
+                .body(Json.encode(createBehaviorGroupRequest))
+                .post("/notifications/behaviorGroups")
+                .then()
+                .statusCode(400)
+                .extract()
+                .body()
+                .asString();
```
gwenneg, line 1116: FYI, `extract().asString()` returns the same thing.

backend/src/main/java/com/redhat/cloud/notifications/routers/models/behaviorgroup/UpdateBehaviorGroupRequest.java
MikelAlejoBR: I wasn't sure whether to do it on this one or to open a new PR, but since you commented this, I have included that check too :+1:

backend/src/test/java/com/redhat/cloud/notifications/routers/NotificationResourceTest.java
MikelAlejoBR: Whoops! Nice catch :fishing_pole_and_fish:

backend/src/test/java/com/redhat/cloud/notifications/routers/NotificationResourceTest.java
MikelAlejoBR: Aaaah, TIL! Refactored!

backend/src/main/java/com/redhat/cloud/notifications/routers/models/behaviorgroup/UpdateBehaviorGroupRequest.java
```diff
@@ -29,4 +32,16 @@ public class UpdateBehaviorGroupRequest {
...
+     * Validates that the display name is not blank when it is provided. Since
+     * this is a potential payload for the "update" operation, clients might
+     * not even send the display name if they don't want to update it. That is
+     * why we let "null" display names go through.
+     * @return true if the display name is not null, but it is blank.
+     */
+    @AssertFalse(message = "the display name cannot be empty")
+    private boolean isDisplayNameNotNullAndBlank() {
+        return this.displayName != null && this.displayName.isBlank();
+    }
```
gwenneg, line 46: Is there any reason to do this instead of using the `@NotBlank` annotation on the field?

MikelAlejoBR: By setting the `@NotBlank` annotation directly I would force the clients to always send tie `displayName`. For whatever reason I thought we wanted to let them not specify the `displayName` if they don't want to? :thinking: 

Either way, if you're okay with making them send the `displayName` every time I don't have any problem in simplifying this to a `@NotBlank` annotation.

gwenneg: My bad, I just realized that the display name is optional in `UpdateBehaviorGroupRequest` and is only updated in the DB if the field is not `null`. Your check is perfect 👍

gwenneg: I should have read the comment you added first 😄

MikelAlejoBR: Hahaha no worries at all :smile:

gwenneg (approved):
LGTM

PR #1741, opened 2023-02-07
RHCLOUD-23850 Add QUARKUS_LOG_CLOUDWATCH_API_CALL_TIMEOUT env var

gwenneg:
See the description of https://github.com/quarkiverse/quarkus-logging-cloudwatch/pull/375 for more information about the new env var.

After this PR, if we send log entries to CloudWatch and the remote server does not respond within 30 seconds, our request will be canceled with a timeout to prevent the accumulation of log entries in the app memory. The cancelation will be logged as an error and reported in Sentry.


MikelAlejoBR (approved):
LGTM!

PR #1742, opened 2023-02-07
[RHCLOUD-23515] Update email template in db from file system

g-duval:
Be able to create/update email templates from file system, when engine starts.
Restrict write access to Template API to internal admin only for the moment.

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/RunOnEngineStartup.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/RunOnEngineStartup.java
@@ -27,8 +27,10 @@ public class RunOnEngineStartup {
         startupUtils.logExternalServiceUrl("quarkus.rest-client.rbac-s2s.url");
         if (featureFlipper.isInjectEmailTemplateToDbOnStartupEnabled()) {
             List<String> warnings = emailTemplateMigrationService.migrate();
-            Log.warn("Email template migration ended with warnings, please check the logs for more details");
-            warnings.stream().forEach(t -> Log.info(t));
+            if (!warnings.isEmpty()) {
+                Log.warn("Email template migration ended with warnings, please check the logs for more details");
+                warnings.stream().forEach(t -> Log.info(t));
+            }
         }
     }
 }
--- a/engine/src/main/java/com/redhat/cloud/notifications/RunOnEngineStartup.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/RunOnEngineStartup.java
@@ -25,8 +25,9 @@ public class RunOnEngineStartup {
         startupUtils.initAccessLogFilter();
         startupUtils.logGitProperties();
         startupUtils.logExternalServiceUrl("quarkus.rest-client.rbac-s2s.url");
-        if (featureFlipper.isUseTemplatesFromDb()) {
+        if (featureFlipper.isInjectEmailTemplateToDbOnStartupEnabled()) {
             List<String> warnings = emailTemplateMigrationService.migrate();
+            Log.warn("Email template migration ended with warnings, please check the logs for more details");
             warnings.stream().forEach(t -> Log.info(t));
         }
     }
--- a/engine/src/main/java/com/redhat/cloud/notifications/RunOnEngineStartup.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/RunOnEngineStartup.java
@@ -1,6 +1,5 @@
 package com.redhat.cloud.notifications;
 
-import com.redhat.cloud.notifications.config.FeatureFlipper;
 import com.redhat.cloud.notifications.templates.EmailTemplateMigrationService;
 import io.quarkus.logging.Log;
 import io.quarkus.runtime.Startup;
@@ -16,9 +15,6 @@ public class RunOnEngineStartup {
 
     @Inject
     EmailTemplateMigrationService emailTemplateMigrationService;
-
-    @Inject
-    FeatureFlipper featureFlipper;
 
     @PostConstruct
     void postConstruct() {
--- a/engine/src/main/java/com/redhat/cloud/notifications/RunOnEngineStartup.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/RunOnEngineStartup.java
@@ -1,9 +1,12 @@
 package com.redhat.cloud.notifications;
 
+import com.redhat.cloud.notifications.config.FeatureFlipper;
+import com.redhat.cloud.notifications.templates.EmailTemplateMigrationService;
+import io.quarkus.logging.Log;
 import io.quarkus.runtime.Startup;
-
 import javax.annotation.PostConstruct;
 import javax.inject.Inject;
+import java.util.List;
 
 @Startup
 public class RunOnEngineStartup {
@@ -11,10 +14,20 @@ public class RunOnEngineStartup {
     @Inject
     StartupUtils startupUtils;
 
+    @Inject
+    EmailTemplateMigrationService emailTemplateMigrationService;
+
+    @Inject
+    FeatureFlipper featureFlipper;
+
     @PostConstruct
     void postConstruct() {
         startupUtils.initAccessLogFilter();
         startupUtils.logGitProperties();
         startupUtils.logExternalServiceUrl("quarkus.rest-client.rbac-s2s.url");
+        if (featureFlipper.isUseTemplatesFromDb()) {
+            List<String> warnings = emailTemplateMigrationService.migrate();
+            warnings.stream().forEach(t -> Log.info(t));
+        }
     }
 }
--- a/engine/src/main/java/com/redhat/cloud/notifications/RunOnEngineStartup.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/RunOnEngineStartup.java
@@ -25,12 +25,11 @@ public class RunOnEngineStartup {
         startupUtils.initAccessLogFilter();
         startupUtils.logGitProperties();
         startupUtils.logExternalServiceUrl("quarkus.rest-client.rbac-s2s.url");
-        if (featureFlipper.isInjectEmailTemplateToDbOnStartupEnabled()) {
-            List<String> warnings = emailTemplateMigrationService.migrate();
-            if (!warnings.isEmpty()) {
-                Log.warn("Email template migration ended with warnings, please check the logs for more details");
-                warnings.stream().forEach(t -> Log.info(t));
-            }
+
+        List<String> warnings = emailTemplateMigrationService.migrate();
+        if (!warnings.isEmpty()) {
+            Log.warn("Email template migration ended with warnings, please check the logs for more details");
+            warnings.stream().forEach(t -> Log.info(t));
         }
     }
 }
--- a/engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationService.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationService.java
@@ -402,37 +402,37 @@ public class EmailTemplateMigrationService {
                 warnings, "application-services", "rhosak", List.of("disruption"),
                 "Rhosak/serviceDisruptionTitle", "txt", "Rhosak service disruption email title",
                 "Rhosak/serviceDisruptionBody", "html", "Rhosak service disruption email body",
-                featureFlipper.isRosakEmailTemplatesV2Enabled()
+                featureFlipper.isRhosakEmailTemplatesV2Enabled()
         );
         createInstantEmailTemplate(
                 warnings, "application-services", "rhosak", List.of("instance-created"),
                 "Rhosak/instanceCreatedTitle", "txt", "Rhosak instance created email title",
                 "Rhosak/instanceCreatedBody", "html", "Rhosak instance created email body",
-                featureFlipper.isRosakEmailTemplatesV2Enabled()
+                featureFlipper.isRhosakEmailTemplatesV2Enabled()
         );
         createInstantEmailTemplate(
                 warnings, "application-services", "rhosak", List.of("instance-deleted"),
                 "Rhosak/instanceDeletedTitle", "txt", "Rhosak instance deleted email title",
                 "Rhosak/instanceDeletedBody", "html", "Rhosak instance deleted email body",
-                featureFlipper.isRosakEmailTemplatesV2Enabled()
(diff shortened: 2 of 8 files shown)


engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationService.java
```diff
@@ -469,8 +538,12 @@ private String loadResourceTemplate(String name, String extension) {
      */
     private void createInstantEmailTemplate(List<String> warnings, String bundleName, String appName, List<String> eventTypeNames,
             String subjectTemplateName, String subjectTemplateExtension, String subjectTemplateDescription,
-            String bodyTemplateName, String bodyTemplateExtension, String bodyTemplateDescription) {
+            String bodyTemplateName, String bodyTemplateExtension, String bodyTemplateDescription, boolean useTemplateV2) {
 
+        if (useTemplateV2) {
+            subjectTemplateName += "V2";
+            bodyTemplateName += "V2";
+        }
```
gwenneg, line 546: Is there something preventing this?
```suggestion
        if (featureFlipper.isRhosakEmailTemplatesV2Enabled()) {
            subjectTemplateName += "V2";
            bodyTemplateName += "V2";
        }
```
If not, I would really like that change to be applied as it would make the PR much lighter.

engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationService.java
```diff
@@ -527,8 +600,12 @@ private boolean instantEmailTemplateExists(EventType eventType) {
      */
     private void createDailyEmailTemplate(List<String> warnings, String bundleName, String appName,
                                           String subjectTemplateName, String subjectTemplateExtension, String subjectTemplateDescription,
-                                          String bodyTemplateName, String bodyTemplateExtension, String bodyTemplateDescription) {
+                                          String bodyTemplateName, String bodyTemplateExtension, String bodyTemplateDescription, boolean useTemplateV2) {
 
+        if (useTemplateV2) {
+            subjectTemplateName += "V2";
+            bodyTemplateName += "V2";
+        }
```
gwenneg, line 608: Same kind of question than above.

engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationService.java
g-duval: This method is always called with the activation flag like `featureFlipper.isRhosakEmailTemplatesV2Enabled()` at last parameter ie : 
`        createInstantEmailTemplate(
                warnings, "application-services", "rhosak", List.of("scheduled-upgrade"),
                "Rhosak/scheduledUpgradeTitle", "txt", "Rhosak scheduled upgrade email title",
                "Rhosak/scheduledUpgradeBody", "html", "Rhosak scheduled upgrade email body",
                featureFlipper.isRhosakEmailTemplatesV2Enabled()
        );`
This is why we don't need to check this flag one more time here. WDYT ?

gwenneg: I asked a stupid question, please ignore it :smile:

engine/src/main/java/com/redhat/cloud/notifications/RunOnEngineStartup.java
```diff
@@ -1,20 +1,33 @@
...
+
+    @Inject
+    FeatureFlipper featureFlipper;
+
     @PostConstruct
     void postConstruct() {
         startupUtils.initAccessLogFilter();
         startupUtils.logGitProperties();
         startupUtils.logExternalServiceUrl("quarkus.rest-client.rbac-s2s.url");
+        if (featureFlipper.isUseTemplatesFromDb()) {
```
gwenneg, line 28: I would feel a bit safer if we triggered the migration with a dedicated feature flag. Migrating the templates and immediately using them from the DB seems a bit risky if something goes wrong.

engine/src/main/java/com/redhat/cloud/notifications/RunOnEngineStartup.java
```diff
@@ -1,20 +1,33 @@
...
+            List<String> warnings = emailTemplateMigrationService.migrate();
+            warnings.stream().forEach(t -> Log.info(t));
```
gwenneg, line 30: Maybe we should log one entry if `!warnings.isEmpty()` to trigger a Sentry alert and make sure an actual issue is not ignored by the team?
```java
            List<String> warnings = emailTemplateMigrationService.migrate();
            if (!warnings.isEmpty()) {
                Log.warn("Email template migration ended with warnings, please check the logs for more details");
                warnings.stream().forEach(t -> Log.info(t));
            }
```

engine/src/test/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationServiceTest.java
```diff
@@ -86,6 +86,9 @@ void testMigration() {
         Application edgeManagement = resourceHelpers.createApp(rhel.getId(), "edge-management");
         EventType imageCreation = resourceHelpers.createEventType(edgeManagement.getId(), "image-creation");
         EventType updateDevices = resourceHelpers.createEventType(edgeManagement.getId(), "update-devices");
+        // App: inventory
+        Application inventoryManagement = resourceHelpers.createApp(rhel.getId(), "inventory");
+        EventType inventoryValidationError = resourceHelpers.createEventType(inventoryManagement.getId(), "validation-error");
```
gwenneg, line 91:
```suggestion
        Application inventory = resourceHelpers.createApp(rhel.getId(), "inventory");
        EventType inventoryValidationError = resourceHelpers.createEventType(inventory.getId(), "validation-error");
```

engine/src/test/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationServiceTest.java
```diff
@@ -198,6 +201,8 @@ void testMigration() {
             // App: edge-management
             findAndCompileInstantEmailTemplate(imageCreation.getId());
             findAndCompileInstantEmailTemplate(updateDevices.getId());
+            // App: Inventory
+            findAndCompileInstantEmailTemplate(inventoryValidationError.getId());
```
gwenneg, line 205:
```suggestion
            findAndCompileInstantEmailTemplate(inventoryValidationError.getId());
            findAndCompileAggregationEmailTemplate(rhel.getName(), inventory.getName(), DAILY);
```

engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationService.java
```diff
@@ -93,7 +102,8 @@ public List<String> migrate() {
         createInstantEmailTemplate(
                 warnings, "openshift", "advisor", List.of("new-recommendation"),
                 "AdvisorOpenshift/newRecommendationInstantEmailTitle", "txt", "AdvisorOpenshift new recommendation email title",
-                "AdvisorOpenshift/newRecommendationInstantEmailBody", "html", "AdvisorOpenshift new recommendation email body"
+                "AdvisorOpenshift/newRecommendationInstantEmailBody", "html", "AdvisorOpenshift new recommendation email body",
+                featureFlipper.isAdvisorEmailTemplatesV2Enabled()
```
gwenneg, line 106:
```suggestion
                featureFlipper.isAdvisorOpenShiftEmailTemplatesV2Enabled()
```

engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationService.java
```diff
@@ -218,12 +228,31 @@ public List<String> migrate() {
         createInstantEmailTemplate(
                 warnings, "rhel", "inventory", List.of("validation-error"),
                 "Inventory/validationErrorEmailTitle", "txt", "Inventory instant email title",
-                "Inventory/validationErrorEmailBody", "html", "Inventory instant email body"
+                "Inventory/validationErrorEmailBody", "html", "Inventory instant email body",
+                featureFlipper.isInventoryEmailTemplatesV2Enabled()
         );
         createDailyEmailTemplate(
-                warnings, "rhel", "Inventory",
+                warnings, "rhel", "inventory",
```
gwenneg, line 235: Good catch :smiley:

engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationService.java
```diff
@@ -398,29 +453,36 @@ public List<String> migrate() {
...
         );
         createDailyEmailTemplate(
                 warnings, "rhel", "vulnerability",
                 "Vulnerability/dailyEmailTitle", "txt", "Vulnerability daily email title",
-                "Vulnerability/dailyEmailBody", "html", "Vulnerability daily email body"
+                "Vulnerability/dailyEmailBody", "html", "Vulnerability daily email body",
+                featureFlipper.isVulnerabilityEmailTemplatesV2Enabled()
         );
 
+        getOrCreateTemplate(warnings, "Common/insightsEmailBody", "html", "Advisor Insights email body");
```
gwenneg, line 484:
```suggestion
        getOrCreateTemplate(warnings, "Common/insightsEmailBody", "html", "Common Insights email body");
```

engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationService.java
```diff
@@ -431,11 +493,18 @@ public List<String> migrate() {
      * Existing templates are never updated by this migration service.
      */
     Template getOrCreateTemplate(List<String> warnings, String name, String extension, String description) {
+        String templateFromFS = loadResourceTemplate(name, extension);
```
gwenneg, line 496: This could be reused at line 514:
```java
            template.setData(templateFromFS);
```

engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationService.java
```diff
@@ -431,11 +493,18 @@ public List<String> migrate() {
...
         try {
+            boolean hasBeenUpdated = false;
             Template template = entityManager.createQuery("FROM Template WHERE name = :name", Template.class)
                     .setParameter("name", name)
                     .getSingleResult();
-            warnings.add(String.format("Template found in DB: %s", name));
+            if (!template.getData().equals(templateFromFS)) {
+                template.setData(templateFromFS);
+                template = entityManager.merge(template);
```
gwenneg, line 504: Could you please check if that line can be removed?
```suggestion
```
This method is run within the boundaries of a JTA transaction so any change to `template` should be reflected in the DB (because of Hibernate's dirty check on managed entities) at the end of the transaction scope.

engine/src/test/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessorWithMigratedTemplateTest.java
```diff
@@ -0,0 +1,232 @@
...
+import static com.redhat.cloud.notifications.TestConstants.DEFAULT_ORG_ID;
+import static com.redhat.cloud.notifications.models.EmailSubscriptionType.INSTANT;
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.ArgumentMatchers.*;
+import static org.mockito.Mockito.*;
+
+@QuarkusTest
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@QuarkusTestResource(TestLifecycleManager.class)
+@TestProfile(EmailSubscriptionTypeProcessorWithMigratedTemplateTest.EnvVarTestSetup.class)
```
gwenneg, line 50: Test profiles from Quarkus come with a high time/resources cost because the entire app has to be restarted when the profile changes (which can happen twice for a single profile), which includes a rerun of all the Flyway scripts among other things.

Changing the feature flags values from `featureFlipper.setXXX` is more efficient and should be enough here.
```suggestion
@QuarkusTest
@QuarkusTestResource(TestLifecycleManager.class)
```

engine/src/test/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessorWithMigratedTemplateTest.java
```diff
@@ -0,0 +1,232 @@
...
+    WebhookTypeProcessor webhookSender;
+
+    @InjectSpy
+    TemplateRepository templateRepository;
+
+    @Inject
+    FeatureFlipper featureFlipper;
+
+    String commonTest() {
+        mockGetUsers(8);
```
gwenneg, line 79: Is there any point mocking 8 users while only `username-1` seems to be used in the following code?

engine/src/test/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessorWithMigratedTemplateTest.java
```diff
@@ -0,0 +1,232 @@
...
+            // Display name
+            assertTrue(bodyRequest.contains("My test machine"), "Body should contain the display_name");
+
+            assertTrue(bodyRequest.contains(TestHelpers.HCC_LOGO_TARGET));
+
+            return bodyRequest;
+        } catch (Exception e) {
+            e.printStackTrace();
+            fail(e);
+        }
```
gwenneg, line 138: Is that necessary? If the method `throws Exception`, wouldn't the test fail in a similar way?

engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationService.java
g-duval: Right I tested it, it works as you said. I remove this line.

engine/src/test/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessorWithMigratedTemplateTest.java
gwenneg: I think there's a way to avoid relying on the Quarkus profile. However, I don't want to block this PR so we'll revisit this later.

gwenneg (approved):
LGTM

PR #1746, opened 2023-02-08
[RHCLOUD-23992] Remove sensible data on email templates

g-duval:
Only Inventory, Rhosak and Vulnerability have some differences between secured and commercial versions.
 
image

image

image

Other application templates are identical.

image

image

image

image

image

image

### DIFF
--- a/engine/src/test/java/com/redhat/cloud/notifications/SecuredEmailTemplatesInDbHelper.java
+++ b/engine/src/test/java/com/redhat/cloud/notifications/SecuredEmailTemplatesInDbHelper.java
@@ -41,7 +41,7 @@ public abstract class SecuredEmailTemplatesInDbHelper {
     protected TemplateService templateService;
 
     @Inject
-    private FeatureFlipper featureFlipper;
+    FeatureFlipper featureFlipper;
 
     @Inject
     EmailTemplateMigrationService emailTemplateMigrationService;


.rhcicd/clowdapp-engine.yaml
```diff
@@ -425,3 +427,5 @@ parameters:
   value: "false"
 - name: NOTIFICATIONS_USE_EVENT_TYPE_FOR_SUBSCRIPTION_ENABLED
   value: "false"
+- name: NOTIFICATIONS_USE_SECURED_EMAIL_TEMPLATES_ENABLED
+  value: "false"
```
josejulio, line 431: We should default to `true`  - the secure version. 
Upon merging we should add a note to set this value to false in clowdapp in the next deployment

g-duval: I undestood it's the purpose of RHCLOUD-24563. Because we already have similar flags for other features developped for secured version (like NOTIFICATIONS_EMAILS_ONLY_MODE_ENABLED), shouldn't we switch all thoses default values at once to be consistent ? WDYT ?

g-duval: for sure, RHCLOUD-24563 have to be done on next days :)

josejulio: We are already unconsistent :) some are set to secured and some are not- so I guess is better to include this one using the secured approach and migrate the rest as part of RHCLOUD-24563

gwenneg: The flag is now set to `true` in the ClowdApp template. I'm resolving this conversation.

engine/src/test/java/com/redhat/cloud/notifications/SecuredEmailTemplatesInDbHelper.java
```diff
@@ -0,0 +1,101 @@
...
+    protected StatelessSessionFactory statelessSessionFactory;
+
+    @InjectSpy
+    protected TemplateRepository templateRepository;
+
+    @Inject
+    protected TemplateService templateService;
+
+    @Inject
+    private FeatureFlipper featureFlipper;
```
gwenneg, line 44: Quarkus doesn't like when something `private` is injected. It works but Quarkus complains in the logs:
```
2023-03-10T14:21:18.6616524Z 2023-03-10 14:21:18,647 INFO  [io.qua.arc.pro.BeanProcessor] (build-13) Found unrecommended usage of private members (use package-private instead) in application beans:
2023-03-10T14:21:18.6617591Z 	- @Inject field com.redhat.cloud.notifications.SecuredEmailTemplatesInDbHelper#featureFlipper,
2023-03-10T14:21:18.6618466Z 	- @Inject field com.redhat.cloud.notifications.SecuredEmailTemplatesInDbHelper#featureFlipper,
2023-03-10T14:21:18.6619324Z 	- @Inject field com.redhat.cloud.notifications.SecuredEmailTemplatesInDbHelper#featureFlipper
2023-03-10T14:21:18.6620065Z 	- and 6 more - please enable debug logging to see the full list
```
```suggestion
    FeatureFlipper featureFlipper;
```

g-duval: Hum ok, good to know :)

josejulio: FWIW this seems to be the reason why it doesn't like it: https://quarkus.io/guides/cdi-reference#native-executables-and-private-members

PR #1749, opened 2023-02-09
[RHCLOUD-23484] feature: trigger daily digests in stage

MikelAlejoBR:
Implements a couple of endpoints, one in the backend and another one in the engine, to be able to trigger daily digest emails with a custom range and organization ID. The goal is to help test and verify the templates that are sent.

## Links

[[RHCLOUD-23484]](https://issues.redhat.com/browse/RHCLOUD-23484)


engine/src/test/java/com/redhat/cloud/notifications/routers/DailyDigestResourceTest.java
```diff
@@ -0,0 +1,143 @@
...
+            .contentType(JSON)
+            .body(Json.encode(triggerDailyDigestRequestDto))
+            .post("/daily-digest")
+            .then()
+            .statusCode(400)
+            .extract()
+            .asString();
+
+        Assertions.assertEquals("the daily digests can only be triggered in the stage environment", response, "unexpected error message received");
+    }
```
MikelAlejoBR, line 76: For some reason in this test the `response` is empty and the assertion fails, but even though I've debugged the process and dig deep on this I am not able to see why it comes empty :thinking: :confused:

josejulio: Only this one is failing?

MikelAlejoBR: Yes! I have tried launching the engine and hitting the endpoint and it also returns an empty body :confused:

josejulio: Perhaps a Produces(TEXT_PLAIN) ? 
If still not returning anything instead of returning void, it could return a `Response` and you can manually build the return value like its done [here](https://github.com/RedHatInsights/notifications-backend/blob/9fe178aa5bb690abe085a6dcf52a10c291a9457f/backend/src/main/java/com/redhat/cloud/notifications/routers/UserConfigResource.java#L99-L103).

MikelAlejoBR: Perfect, I'll try that! Thank you very much :smile:

MikelAlejoBR: Using `@Produces(TEXT_PLAIN)` had the same result... Although your suggestion to build the `Response` fixed the issue! Thank you :man_dancing:

common/src/main/java/com/redhat/cloud/notifications/Constants.java
```diff
@@ -8,4 +8,5 @@ public interface Constants {
     String API_NOTIFICATIONS_V_1_0 = "/api/notifications/v1.0";
     String API_INTERNAL = "/internal";
     String X_RH_IDENTITY_HEADER = "x-rh-identity";
+    String X_RH_NOTIFICATIONS_ENGINE_ORG_ID = "x-rh-notifications-engine-org-id";
```
josejulio, line 11: Is this used anywhere?

common/src/main/java/com/redhat/cloud/notifications/routers/dailydigest/TriggerDailyDigestRequestDto.java
```diff
@@ -0,0 +1,88 @@
...
+import java.time.LocalTime;
+import java.time.ZoneOffset;
+import java.util.Objects;
+
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public class TriggerDailyDigestRequestDto {
+
+    @JsonProperty("application_name")
+    @NotBlank
+    private final String applicationName;
```
josejulio, line 20: I guess you also need to mark them as `@Required`  - i.e. they can be null but not blank (IIRC)

engine/src/main/java/com/redhat/cloud/notifications/templates/models/Environment.java
```diff
@@ -20,4 +20,20 @@ public String name() {
...
+        return "local-dev".equals(this.environment);
+    }
+
+    /**
+     * Checks if the "ENV_NAME" environment variable is set to "stage".
+     * @return true if the "ENV_NAME" environment variable is set to "stage".
+     */
+    public boolean isEnvironmentStage() {
+        return "stage".equals(this.environment);
+    }
```
josejulio, line 38: This could probably be moved to `common`

josejulio (changes requested):

common/src/main/java/com/redhat/cloud/notifications/routers/dailydigest/TriggerDailyDigestRequestDto.java
MikelAlejoBR: This is not needed! The `@NotBlank` annotation makes sure that the target `String` isn't null, empty or blank!

engine/src/main/java/com/redhat/cloud/notifications/templates/models/Environment.java
MikelAlejoBR: Good idea, I've moved it and updated the references! Thanks!

josejulio (approved):
lgtm

common/src/main/java/com/redhat/cloud/notifications/Constants.java
MikelAlejoBR: No! Good catch, I removed it!

MikelAlejoBR:
Looking into the failure...-

engine/src/main/java/com/redhat/cloud/notifications/routers/DailyDigestResource.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Triggers a daily digest by sending a command to the Kafka "aggregation"
+     * queue.
+     * @param triggerDailyDigestRequestDto the settings of the aggregation
+     *                                     command.
+     */
+    @Consumes(APPLICATION_JSON)
+    @Produces(TEXT_PLAIN)
+    @POST
+    public Response triggerDailyDigest(final TriggerDailyDigestRequestDto triggerDailyDigestRequestDto) {
```
gwenneg, line 46: Nitpick: This method could return `void`.

engine/src/main/java/com/redhat/cloud/notifications/routers/DailyDigestResource.java
```diff
@@ -0,0 +1,74 @@
...
+        );
+
+        final AggregationCommand aggregationCommand = new AggregationCommand(
+            emailAggregationKey,
+            triggerDailyDigestRequestDto.getStart(),
+            triggerDailyDigestRequestDto.getEnd(),
+            EmailSubscriptionType.DAILY
+        );
+
+        final Message<String> message = KafkaMessageWithIdBuilder.build(Json.encode(aggregationCommand));
```
gwenneg, line 68: I would use an `ObjectMapper` here to serialize the command instead of `Json.encode` to make sure this class behaves consistently with the `aggregator` module.

gwenneg: `KafkaMessageWithIdBuilder` isn't really needed as we don't consume the ID on the other side of the Kafka topic, but it doesn't hurt so it can stay.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/InternalResource.java
```diff
@@ -469,4 +481,29 @@ public Response getDailyDigestTimePreference(@PathParam("orgId") String orgId) {
...
+    /**
+     * Sends a daily digest command to the engine via a REST request.
+     * @param triggerDailyDigestRequestDto the settings of the digest.
+     */
+    @Consumes(APPLICATION_JSON)
+    @POST
+    @Path("/daily-digest")
+    @RolesAllowed(ConsoleIdentityProvider.RBAC_INTERNAL_USER)
+    public void triggerDailyDigest(@NotNull @Valid final TriggerDailyDigestRequestDto triggerDailyDigestRequestDto) {
+        if (!this.environment.isEnvironmentLocal() && !this.environment.isEnvironmentStage()) {
```
gwenneg, line 494: This class allows a run in a local env (which makes sense) while `DailyDigestResource`  does not. They should be consistent.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/InternalResource.java
```diff
@@ -469,4 +481,29 @@ public Response getDailyDigestTimePreference(@PathParam("orgId") String orgId) {
...
+            throw new BadRequestException("the daily digests can only be triggered in the stage environment");
+        }
+
+        if (
+            !this.applicationRepository.applicationBundleExists(
+                triggerDailyDigestRequestDto.getApplicationName(),
+                triggerDailyDigestRequestDto.getBundleName()
+            )
+        ) {
+            throw new BadRequestException("unable to find the specified application — bundle combination");
```
gwenneg, line 504: `—` that's a huge dash😄

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/InternalResource.java
```diff
@@ -469,4 +481,29 @@ public Response getDailyDigestTimePreference(@PathParam("orgId") String orgId) {
...
         }
     }
+
+    /**
+     * Sends a daily digest command to the engine via a REST request.
+     * @param triggerDailyDigestRequestDto the settings of the digest.
+     */
+    @Consumes(APPLICATION_JSON)
+    @POST
+    @Path("/daily-digest")
```
gwenneg, line 491: I wonder if we should prevent collisions with future internal endpoints related to the daily digest:
```suggestion
    @Path("/daily-digest/trigger")
```

engine/src/test/java/com/redhat/cloud/notifications/routers/DailyDigestResourceTest.java
```diff
@@ -0,0 +1,143 @@
...
+        // Make sure that the message was received before continuing.
+        Awaitility.await().until(
+            () -> aggregationsOut.received().size() == 1
+        );
+
+        final var aggregationCommands = aggregationsOut.received();
+
+        // Only one test aggregation command should have been sent to Kafka.
+        final var expectedAggregationsCommandCount = 1;
+        Assertions.assertEquals(expectedAggregationsCommandCount, aggregationCommands.size(), "unexpected number of aggregation commands sent to Kafka");
```
gwenneg, line 125: That's already asserted by line 118.

common/src/main/java/com/redhat/cloud/notifications/routers/dailydigest/TriggerDailyDigestRequestDto.java
```diff
@@ -0,0 +1,88 @@
...
+import javax.validation.constraints.NotBlank;
+import java.beans.ConstructorProperties;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.ZoneOffset;
+import java.util.Objects;
+
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public class TriggerDailyDigestRequestDto {
```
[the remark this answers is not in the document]
MikelAlejoBR: Agree! Removed :)

common/src/main/java/com/redhat/cloud/notifications/routers/dailydigest/TriggerDailyDigestRequestDto.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @JsonProperty("application_name")
```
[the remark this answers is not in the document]
MikelAlejoBR: All right, removed the redundant annotations!

common/src/main/java/com/redhat/cloud/notifications/models/Environment.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public String url() {
+        return this.url;
+    }
+
+    /**
+     * Checks if the "ENV_NAME" environment variable is set to "local-dev".
+     * @return true if the "ENV_NAME" environment variable is set to "local-dev".
+     */
+    public boolean isEnvironmentLocal() {
```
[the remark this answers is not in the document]
MikelAlejoBR: I like your suggestions better too, changed!

common/src/main/java/com/redhat/cloud/notifications/models/Environment.java
```diff
@@ -0,0 +1,39 @@
...
+        return "local-dev".equals(this.environment);
+    }
+
+    /**
+     * Checks if the "ENV_NAME" environment variable is set to "stage".
+     * @return true if the "ENV_NAME" environment variable is set to "stage".
+     */
+    public boolean isEnvironmentStage() {
```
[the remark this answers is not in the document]
MikelAlejoBR: Done as well!

common/src/main/java/com/redhat/cloud/notifications/models/Environment.java
```diff
@@ -0,0 +1,39 @@
...
+    public String url() {
+        return this.url;
+    }
+
+    /**
+     * Checks if the "ENV_NAME" environment variable is set to "local-dev".
+     * @return true if the "ENV_NAME" environment variable is set to "local-dev".
+     */
+    public boolean isEnvironmentLocal() {
+        return "local-dev".equals(this.environment);
```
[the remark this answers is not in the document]
MikelAlejoBR: I've put the two values in constants!

common/src/main/java/com/redhat/cloud/notifications/routers/dailydigest/TriggerDailyDigestRequestDto.java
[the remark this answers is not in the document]
MikelAlejoBR: I am going to unresolve this one, since now the tests are telling me the following thing:

```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `com.redhat.cloud.notifications.models.AggregationCommand` (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
 at [Source: (String)"{"aggregationKey":{"orgId":"org-id","bundle":"bundle-name","application":"application-name"},"start":"2023-03-08T10:43:33.55235369","end":"2023-03-13T10:43:33.55235369","subscriptionType":"DAILY"}"; line: 1, column: 2]
```

So I am guessing that that might be why I annotated everything...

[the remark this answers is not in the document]
gwenneg: Ok then feel free to put back the annotation.

engine/src/main/java/com/redhat/cloud/notifications/routers/DailyDigestResource.java
MikelAlejoBR: All right, applied the suggestion!

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/InternalResource.java
MikelAlejoBR: You're right! I think I forgot about this one. Done!

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/InternalResource.java
MikelAlejoBR: I tend to use the em dash everywhere! :joy:

engine/src/test/java/com/redhat/cloud/notifications/routers/DailyDigestResourceTest.java
MikelAlejoBR: All right! I removed the redundant assertion!

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/InternalResource.java
MikelAlejoBR: Changed!

gwenneg (approved):
LGTM

PR #1762, opened 2023-02-22
change doc version to 2023

ryandeussing:


g-duval:
LGTM ! As you know, because we are on migration phase bettween old and new email templates, I will replicate this update on new templates as well.

g-duval:
This PR is still tagged as Draft, don't forget to update it as ready for review to trigger integration pipelines ;)

PR #1767, opened 2023-02-27
[RHCLOUD-24419] Limit loaded EmailAggregation

g-duval:
Limit loaded EmailAggregation elements to prevent heap overflow

### DIFF
--- a/engine/src/test/java/com/redhat/cloud/notifications/processors/email/EmailAggregatorTest.java
+++ b/engine/src/test/java/com/redhat/cloud/notifications/processors/email/EmailAggregatorTest.java
@@ -36,7 +36,12 @@ import java.util.stream.Collectors;
 import static com.redhat.cloud.notifications.models.EmailSubscriptionType.DAILY;
 import static org.junit.jupiter.api.Assertions.*;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.reset;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 @QuarkusTest
@@ -45,7 +50,7 @@ class EmailAggregatorTest {
     @InjectMock
     EmailTemplateFactory emailTemplateFactory;
 
-    @Inject
+    @InjectSpy
     EmailAggregationRepository emailAggregationRepository;
 
     @InjectMock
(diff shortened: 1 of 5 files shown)


engine/src/test/java/com/redhat/cloud/notifications/processors/email/EmailAggregatorTest.java
```diff
@@ -63,13 +66,21 @@ class EmailAggregatorTest {
...
+    final EmailAggregationKey aggregationKey = new EmailAggregationKey("org-1", "rhel", "policies");
+
     @AfterEach
     void afterEach() {
         featureFlipper.setUseEventTypeForSubscriptionEnabled(false);
     }
 
+    @BeforeEach
+    void beforeEach() {
+        emailAggregator.aggregationMaxPageSize = 5;
```
gwenneg, line 78: It doesn't look like the test in this class triggers multiple iterations of the new `do while` loop that loads the aggregation from the DB. Did I read it wrong?

g-duval: Well it does but it's not obvious. I added a comment and some Mockito verify methods to clarify it.

engine/src/test/java/com/redhat/cloud/notifications/processors/email/EmailAggregatorTest.java
```diff
@@ -98,31 +114,49 @@ void shouldTestRecipientsFromSubscription() {
...
+        verify(emailAggregationRepository, times(1)).getEmailAggregation(any(EmailAggregationKey.class), any(LocalDateTime.class), any(LocalDateTime.class), anyInt(), anyInt());
+        verify(emailAggregationRepository, times(1)).getEmailAggregation(any(EmailAggregationKey.class), any(LocalDateTime.class), any(LocalDateTime.class), eq(0), eq(emailAggregator.aggregationMaxPageSize));
+        reset(emailAggregationRepository); // just reset mockito counter
 
         // nobody subscribed to the right event type yet
-        assertTrue(result.size() == 0);
+        assertEquals(0, result.size());
 
         resourceHelpers.createEventTypeEmailSubscription("org-1", "user-2", eventType1, DAILY);
+        // because on previous step, nobody subscribed to the rignt event type, we already have 4 records on database
```
gwenneg, line 137: Is that really the reason? From what I can see in the code, `EmailAggregator#getAggregated` does not purge the DB data so even if someone subscribed to the event type, we would still have the 4 records in the DB here.

gwenneg: This would sound more accurate to me:
```suggestion
        // because after the previous aggregate() call the email_aggregation DB table was not purged, we already have 4 records on database
```

g-duval: Right

gwenneg (approved):
LGTM

PR #1783, opened 2023-03-09
[RHCLOUD-24563] Set restrictive env config as default value

g-duval:
Use flags in a consistent way by making the restrictive mode the default

### DIFF
--- a/.rhcicd/clowdapp-engine.yaml
+++ b/.rhcicd/clowdapp-engine.yaml
@@ -261,6 +261,8 @@ objects:
           value: ${NOTIFICATIONS_USE_EVENT_TYPE_FOR_SUBSCRIPTION_ENABLED}
         - name: NOTIFICATIONS_USE_SECURED_EMAIL_TEMPLATES_ENABLED
           value: ${NOTIFICATIONS_USE_SECURED_EMAIL_TEMPLATES_ENABLED}
+        - name: NOTIFICATIONS_USE_TEMPLATES_FROM_DB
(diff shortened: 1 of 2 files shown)


MikelAlejoBR (approved):
LGTM!

gwenneg (changes requested):
Do not merge (yet), I'll take a look later.

.rhcicd/clowdapp-engine.yaml
```diff
@@ -327,7 +329,7 @@ parameters:
   value: "false"
 - name: NOTIFICATIONS_EMAILS_ONLY_MODE_ENABLED
```
gwenneg, line 330: `clowdapp-backend.yaml` also contains that flag. Both files should have the same flag value.

.rhcicd/clowdapp-engine.yaml
```diff
@@ -438,3 +440,5 @@ parameters:
   value: ""
 - name: NOTIFICATIONS_USE_SECURED_EMAIL_TEMPLATES_ENABLED
   value: "true"
+- name: NOTIFICATIONS_USE_TEMPLATES_FROM_DB
+  value: "true"
```
gwenneg, line 444: Is that really needed in this PR?

gwenneg: It is, I just remembered why 😄

.rhcicd/clowdapp-engine.yaml
g-duval: Right, I forgot this :)

gwenneg (approved):
LGTM

PR #1784, opened 2023-03-09
Making IT service optional

josejulio:
I'm removing the it keystore from `QUARKUS_HTTP_SSL_CERTIFICATE_KEY_STORE_FILE` (global keystore for http) because we might have to use a new one from clowdapp.

I'm setting it specifically for the ITservice.


g-duval (approved):
LGTM

gwenneg (approved):

PR #1789, opened 2023-03-13
Fix Slack notifications data deserialization

gwenneg:
`JsonObject#getMap` doesn't work well with Qute when fields from `context` need to be used in a template, so I replaced it with `ObjectMapper#readValue(String.class, Map.class)`.


MikelAlejoBR (approved):
LGTM!

PR #1796, opened 2023-03-15
RHCLOUD-24697 Revert Flyway scripts execution from engine in ephemeral

gwenneg:
We tested it and it didn't work as we hoped because of Bonfire limitations...


g-duval:
LGTM

g-duval (approved):

PR #1798, opened 2023-03-16
[RHCLOUD-24734] feature: migrate database secrets to Sources

MikelAlejoBR:
Implements the migration of the database secrets present in the camel and webhook properties to Sources.

## Links

[[RHCLOUD-24734]](https://issues.redhat.com/browse/RHCLOUD-24734)


backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -233,18 +256,22 @@ public boolean updateEndpoint(Endpoint endpoint) {
                 case WEBHOOK:
                     WebhookProperties properties = endpoint.getProperties(WebhookProperties.class);
                     return entityManager.createQuery(webhookQuery)
+                            .setParameter("basicAuthenticationSourcesId", properties.getBasicAuthenticationSourcesId())
```
gwenneg, line 259: I don't think we should migrate the credentials to Sources by modifying the public REST API, especially when the changes here will never be used with a public REST request (the sources IDs are marked as `@JsonIgnore`).

I'd prefer a temporary REST internal endpoint dedicated to the migration. It would rely on DB queries very similar to what you already wrote, but with a cleaner separation between public and internal matters.

gwenneg (changes requested):

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
gwenneg: Here are two examples of internal endpoints dedicated to migrations:
- https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationService.java
- https://github.com/RedHatInsights/notifications-backend/blob/master/backend/src/main/java/com/redhat/cloud/notifications/routers/internal/SubscriptionToEventTypeMigrationService.java

MikelAlejoBR: All right! I have created everything ad-hoc so that we can simply wipe it out easily once we're done with the migrations, as you suggested.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -352,4 +357,121 @@ public boolean existsByUuidAndOrgId(final UUID endpointUuid, final String orgId)
...
+    /**
+     * Returns a list of endpoints with its properties loaded, which have been
+     * identified as having "basic authentication" and "secret token" secrets
+     * stored in the database, but no references to those secrets in sources.
+     * This signals that the endpoints' secrets need to be migrated to Sources.
+     *
+     * @return a list of endpoints with their properties loaded.
+     */
+    @Deprecated(forRemoval = true)
+    public List<Endpoint> findEndpointWithPropertiesWithStoredSecrets() {
```
gwenneg, line 370: This method could rely on a single **non-native** query:
```sql
SELECT e FROM Endpoint e
WHERE EXISTS (
  SELECT cp FROM CamelProperties cp
  WHERE cp.id = e.id AND cp.basicAuthenticationSourcesId IS NULL AND cp.secretTokenSourcesId IS NULL
  AND (cp.basicAuthentication IS NOT NULL OR cp.secretToken IS NOT NULL)
) OR EXISTS (
  SELECT wp FROM WebhookProperties wp
  WHERE wp.id = e.id AND wp.basicAuthenticationSourcesId IS NULL AND wp.secretTokenSourcesId IS NULL
  AND (wp.basicAuthentication IS NOT NULL OR wp.secretToken IS NOT NULL)
)
```
If the query is not native, then Hibernate will perform a dirty check and update the sources ID automatically, which means that the next method `updateEndpointSecrets` should not be needed. `SourcesSecretsMigrationService#migrateEndpointSecretsSources` will need to be annotated with `@Transactional`.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -352,4 +356,87 @@ public boolean existsByUuidAndOrgId(final UUID endpointUuid, final String orgId)
...
+    }
+
+    /**
+     * Updates the given endpoint's secrets' references in the database.
+     * @param genericProperties the generic {@link EndpointProperties} of the
+     *                          endpoint.
+     */
+    @Deprecated(forRemoval = true)
+    @Transactional
+    public void updateEndpointSecrets(final EndpointProperties genericProperties) {
```
gwenneg, line 398: This method shouldn't be needed anymore.

MikelAlejoBR: Removed!

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/SourcesSecretsMigrationService.java
```diff
@@ -0,0 +1,68 @@
...
+
+        int processedCounter = 0;
+        int errorCounter = 0;
+        for (final Endpoint endpoint : endpoints) {
+            // On failure, don't stop processing the endpoints and keep going,
+            // since this is not a destructive operation.
+            try {
+                this.secretUtils.createSecretsForEndpoint(endpoint);
+            } catch (final WebApplicationException e) {
+                Log.errorf("[endpoint_id: %s] error when migrating the endpoint secrets: unable to create the secrets for the endpoint: %s", e.getResponse().getEntity());
```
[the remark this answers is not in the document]
MikelAlejoBR: Nice catch! Done!

backend/src/test/java/com/redhat/cloud/notifications/db/ResourceHelpers.java
```diff
@@ -285,4 +291,220 @@ public void setupTransformationTemplate() {
...
+        // Create a few endpoints that shouldn't be picked by the function
+        // under test, because they contain references to Sources secrets.
+        for (int i = 0; i < 12; i++) {
+            final Endpoint endpoint = new Endpoint();
+            endpoint.setDescription(UUID.randomUUID().toString());
+            endpoint.setName(UUID.randomUUID().toString());
+            endpoint.setOrgId(DEFAULT_ORG_ID);
+
+            // Should it be a camel, or a webhook endpoint?
+            if (random.nextBoolean()) {
```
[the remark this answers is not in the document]
MikelAlejoBR: Yeah, I didn't consider that possibility, to be honest. I've refactored it to follow your suggestion. That way we can ensure we have properties of all types!

Thanks!A

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/SourcesSecretsMigrationService.java
```diff
@@ -0,0 +1,68 @@
...
+        final List<Endpoint> endpoints = this.endpointRepository.findEndpointWithPropertiesWithStoredSecrets();
+
+        int processedCounter = 0;
+        int errorCounter = 0;
+        for (final Endpoint endpoint : endpoints) {
+            // On failure, don't stop processing the endpoints and keep going,
+            // since this is not a destructive operation.
+            try {
+                this.secretUtils.createSecretsForEndpoint(endpoint);
+            } catch (final WebApplicationException e) {
```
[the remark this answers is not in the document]
MikelAlejoBR: Done too!

PR #1802, opened 2023-03-17
 [RHCLOUD-24704] feature: measure the time it takes to call to Sources

MikelAlejoBR:
These metrics will help us understand if we need caching the incoming secrets for a better performance in Notifications.

## Links

[[RHCLOUD-24704]](https://issues.redhat.com/browse/RHCLOUD-24704)


gwenneg (approved):
LGTM

PR #1808, opened 2023-03-21
RHCLOUD-24624 Polish the Slack integration details

gwenneg:


gwenneg:
> Thank you. I had the opportunity to dig a bit deeper in Camel. It made me wonder if I could have "easily" done the "test an endpoint" feature in the engine with a Camel Route as well?
> 
> Similar to what you did building a REST endpoint but having a `.to` to the Kafka channel.

Yes, that could have worked. However, I have a strong preference for REST endpoints declared with JAX-RS annotations which are, in my opinion, much easier to implement or maintain for advanced/complex use cases. I like Camel very much for the amazing help it provides when it comes to integrating with an external system. Not so much to expose a REST endpoint unless it has to be part of a Camel route to the outside world 😄

MikelAlejoBR:
Interesting! Thanks!

PR #1809, opened 2023-03-22
RHCLOUD-24813 Send notifications to Microsoft Teams

gwenneg:
Based on #1808. I'll rebase this PR when the other one is merged.


gwenneg:
Don't worry about the PR check failure, it's unrelated.

engine/src/main/java/com/redhat/cloud/notifications/processors/teams/TeamsRouteBuilder.java
```diff
@@ -0,0 +1,50 @@
...
+        /*
+         * This route transforms an incoming REST payload into a message that is eventually sent to Teams.
+         */
+        from(TEAMS_DIRECT_ENDPOINT)
+                .routeId(TEAMS_OUTGOING_ROUTE)
+                .process(teamsNotificationProcessor)
+                .removeHeaders("CamelHttp*")
+                .setHeader(HTTP_METHOD, constant("POST"))
+                .setHeader(CONTENT_TYPE, constant("application/json"))
+                .toD("${exchangeProperty.webhookUrl}", maxEndpointCacheSize);
```
MikelAlejoBR, line 48: I was wondering: does Camel first accept the request and return a 200 to the caller, and then call MS Teams? I don't see it clear because:

* The `TeamsProcessor` has a `try-catch` block surrounding sending the notification to Camel. But does that `try-catch` block just catch if, for example, the Camel context is misbehaving or even down?
* I don't think that Camel, after accepting the Notification from the `TeamsProcessor`, will hold that connection until MS Teams responds to then forward back the response, right?

Thank you!

MikelAlejoBR (commented):
Looking good!

engine/src/main/java/com/redhat/cloud/notifications/processors/teams/TeamsRouteBuilder.java
gwenneg: This is all very similar to exposing a REST endpoint with `resteasy` and calling an external service with `rest-client` from the method that serves the REST request.

engine/src/main/resources/application.properties
```diff
@@ -142,3 +142,4 @@ quarkus.opentelemetry.tracer.exporter.otlp.endpoint=http://localhost:4317
 sources.psk=development-value-123
 
 quarkus.rest-client.internal-slack.url=${clowder.endpoints.notifications-engine-service:http://localhost:8087}
+quarkus.rest-client.internal-teams.url=${clowder.endpoints.notifications-engine-service:http://localhost:8087}
```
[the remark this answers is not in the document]
gwenneg: I agree in the long run. However, this is temporary and should be replaced with Kafka shortly.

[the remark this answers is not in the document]
MikelAlejoBR: Aaaah perfect, thanks!

engine/src/main/java/com/redhat/cloud/notifications/processors/teams/TeamsRouteBuilder.java
MikelAlejoBR: Aaah I see! Thank you! It makes more sense, yes...

PR #1812, opened 2023-03-22
Fix daily digest trigger endpoint path

gwenneg:
I tried to trigger a daily digest in stage and it failed with a 404.

The REST client interface didn't match this path:
https://github.com/RedHatInsights/notifications-backend/blob/f1b415363fd28c79ddf4437685ba642b2587d502/engine/src/main/java/com/redhat/cloud/notifications/routers/DailyDigestResource.java#L49


MikelAlejoBR (approved):
LGTM!

PR #1825, opened 2023-03-27
[RHCLOUD-24889] Minor fixes on email templates V2

g-duval:


engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/PatchEmailPayloadAggregator.java
```diff
@@ -30,10 +30,10 @@ public class PatchEmailPayloadAggregator extends AbstractEmailPayloadAggregator
...
     private static final List<String> EVENT_TYPES = Arrays.asList(NEW_ADVISORIES_EVENT);
 
-    private static final String ENHANCEMENT_TYPE = "enhancement";
-    private static final String BUGFIX_TYPE = "bugfix";
-    private static final String SECURITY_TYPE = "security";
-    private static final String UNSPECIFIED_TYPE = "unspecified";
+    private static final String ENHANCEMENT_TYPE = "Enhancement";
+    private static final String BUGFIX_TYPE = "Bugfix";
+    private static final String SECURITY_TYPE = "Security";
+    private static final String UNSPECIFIED_TYPE = "Unspecified";
```
josejulio, line 36: The comparation below is done using a lower case type - and I personally like the pattern - whatever you send is lower case and we compare with lower case string versions.

```
String advisoryType = payload.getString(ADVISORY_TYPE).toLowerCase();
// Drop on unknown type
 if (!ADVISORY_TYPES.contains(advisoryType)) {
    return;
}
```

Was this failing?

g-duval: You are right, thank you for the warning :).
Changing it on Patch aggregator is little bit overkill, let's do that using simple css because it's a rendering issue.

engine/src/main/resources/templates/Patch/dailyEmailBodyV2.html
```diff
@@ -21,7 +21,7 @@
                 <table class="rh-data-table rh-m-bordered">
                     <thead>
                         <tr>
-                          <th>{key} advisories</th>
+                          <th style="text-transform: capitalize">{key} advisories</th>
```
josejulio, line 24: Ah yes, this is even better!

PR #1831, opened 2023-03-28
Adds Cache for CodeQL analysis

josejulio:
I noticed CodeQL download all dependencies on every run.
Adding a cache for java and javascript (depending what is being run)


gwenneg (approved):
Good catch! LGTM

PR #1832, opened 2023-03-28
Rhcloud 23516 generate one email fro multiple recipients

g-duval:


engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
```diff
@@ -77,6 +78,38 @@ public void init() {
...
+            //      also add metrics for these failures
+
+            webhookSender.doHttpRequest(
+                event, endpoint,
+                bopRequest,
+                getPayload(user, action, subject, body), "POST", bopUrl, persistHistory);
+
+            processedTimer.stop(registry.timer("processor.email.processed", "bundle", action.getBundle(), "application", action.getApplication()));
+
+            processTime.record(Duration.between(start, LocalDateTime.now(UTC)));
```
MikelAlejoBR, line 106: I guess these would stop if an exception occurs, but I am not sure what's the behavior here. What do you think about moving these two statements to a `finally` block?

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
```diff
@@ -77,6 +78,38 @@ public void init() {
...
+        } catch (Exception e) {
+            Log.info("Email sending failed", e);
```
MikelAlejoBR, line 108: I think this one should be a `Log.error` instead

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
```diff
@@ -77,6 +78,38 @@ public void init() {
         bopApiToken = LineBreakCleaner.clean(bopApiToken);
     }
 
+    public void sendEmail(Set<User> user, Event event, TemplateInstance subject, TemplateInstance body, boolean persistHistory) {
```
MikelAlejoBR, line 81: Nitpicky, but I'd vote for renaming `user` to `users`

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
```diff
@@ -77,6 +78,38 @@ public void init() {
...
+            // TODO Add recipients processing from policies-notifications processing (failed recipients)
+            //      by checking the NotificationHistory's details section (if missing payload - fix in WebhookTypeProcessor)
+
+            // TODO If the call fails - we should probably rollback Kafka topic (if BOP is down for example)
+            //      also add metrics for these failures
+
+            webhookSender.doHttpRequest(
+                event, endpoint,
+                bopRequest,
+                getPayload(user, action, subject, body), "POST", bopUrl, persistHistory);
```
MikelAlejoBR, line 102: Can you please put all the arguments in new lines?

common/src/main/java/com/redhat/cloud/notifications/config/FeatureFlipper.java
```diff
@@ -132,6 +132,9 @@ public class FeatureFlipper {
     @ConfigProperty(name = "notifications.instant-emails.enabled", defaultValue = "false")
     boolean instantEmailsEnabled;
 
+    @ConfigProperty(name = "notifications.remove-user-from-email-template.enabled", defaultValue = "false")
+    boolean removeUserDataFromEmailTemplateEnabled;
+
```
MikelAlejoBR, line 137: This is just out of curiosity. In this PR you're adding support for multi-user single-emails, right? Why is the feature flag called `remove-user-form-email-template`, though :thinking: ?

MikelAlejoBR (commented):
Just a few minor comments! LGTM!

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
g-duval: Ok I updated it as error level

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
g-duval: I just duplicated the original method by adding a set of user instead of a single user, but I think you are right, it should be on a finally block.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
g-duval: Sure :)

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
g-duval: Sure

common/src/main/java/com/redhat/cloud/notifications/config/FeatureFlipper.java
g-duval: Because is because we can do that because thanks to V2, we can remove user data from email templates.
Bur you are right, this name doesn't reflect the real purpose of this feature, I will update it.

g-duval:
> Should we mark this one as draft until we update and use all V2 email templates?

Beacuse template V2 were activated on production few hours ago, and because this feature is behind a feature flag, I think it can be reviewed ?
Of course if you think if too soon, I can mark it as draft to wait a litte bit more ? There is no real rush here ;)

josejulio:
> Beacuse template V2 were activated on production few hours ago, and because this feature is behind a feature flag, I think it can be reviewed ?

I guess it's good to go, we just need to make sure we don't enable this with any old template

g-duval:
> @g-duval This PR needs a rebase. I'll take a look after it's done.

Done @gwenneg

g-duval:
> > @g-duval This PR needs a rebase. I'll take a look after it's done.
> 
> Done @gwenneg 

I will rerebase it tomorrow 😂

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -263,29 +268,59 @@ private void processAggregateEmailsByAggregationKey(EmailAggregationKey aggregat
         }
 
         if (subject != null && body != null) {
-            for (Map.Entry<User, Map<String, Object>> aggregation :
-                    emailAggregator.getAggregated(aggregationKey, emailSubscriptionType, startTime, endTime).entrySet()) {
+            Map<User, Map<String, Object>> aggregationsByUsers = emailAggregator.getAggregated(aggregationKey, emailSubscriptionType, startTime, endTime);
+
+            if (featureFlipper.isSendSingleEmailForMultipleRecipientsEnabled()) {
+                Map<Map<String, Object>, Set<User>> aggregationsEmailContext = aggregationsByUsers.keySet().stream()
+                    .collect(Collectors.groupingBy(aggregationsByUsers::get, Collectors.toSet()));
```
gwenneg, line 275: I'm okay with that as a temporary solution but we'll need to make the whole aggregation process a lot simpler than it is now. My comment is not related to your PR, I already didn't like that process before 😅

engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
```diff
@@ -202,6 +202,16 @@ public void doHttpRequest(Event event, Endpoint endpoint, HttpRequest<Buffer> re
...
+                if (isEmailEndpoint) {
+                    JsonObject details = new JsonObject();
+                    try {
+                        int totalRecipients = payload.getJsonArray("emails").getJsonObject(0).getJsonArray("bccList").size();
+                        details.put("total_recipients", totalRecipients);
+                        history.setDetails(details.getMap());
+                    } catch (Exception ex) {
+                        Log.info("This email payload doesn't use bccList");
+                    }
+                }
```
gwenneg, line 214: `history#details` will be overwritten in case of exception or `INTERNAL_FAILURE` status. I think this block should move a few lines lower, inside the `if (resp.statusCode() >= 200 && resp.statusCode() < 300)` block.

engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
```diff
@@ -202,6 +202,16 @@ public void doHttpRequest(Event event, Endpoint endpoint, HttpRequest<Buffer> re
 
                 boolean serverError = false;
                 boolean shouldResetEndpointServerErrors = false;
+                if (isEmailEndpoint) {
+                    JsonObject details = new JsonObject();
```
gwenneg, line 206:
```suggestion
                    Map<String, Object> details = new HashMap<>();
```

engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
```diff
@@ -202,6 +202,16 @@ public void doHttpRequest(Event event, Endpoint endpoint, HttpRequest<Buffer> re
...
+                    try {
+                        int totalRecipients = payload.getJsonArray("emails").getJsonObject(0).getJsonArray("bccList").size();
+                        details.put("total_recipients", totalRecipients);
+                        history.setDetails(details.getMap());
+                    } catch (Exception ex) {
+                        Log.info("This email payload doesn't use bccList");
```
gwenneg, line 212: This would be an issue the dev team has to investigate, let's raise the log level. Since the `catch` block could be executed for other reasons than the lack of `bccList` field, we should also include the exception in the log entry.
```suggestion
                        Log.error("Could not set the total_recipients field in the history details", e);
```

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
g-duval: I fully agree, the full aggregation process should be simpler

engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
g-duval: You are right, I used a slite different approach to preserve it in case of INTERNAL_FAILURE

engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
g-duval: Ok

engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
g-duval: Ok

engine/src/test/java/com/redhat/cloud/notifications/TestHelpers.java
```diff
@@ -73,7 +78,13 @@ public static EmailAggregation createEmailAggregation(String orgId, String bundl
...
         emailActionMessage.setOrgId(orgId);
 
-        aggregation.setPayload(TestHelpers.wrapActionToJsonObject(emailActionMessage));
+        JsonObject payload = TestHelpers.wrapActionToJsonObject(emailActionMessage);
+        if (null != extraRecipient) {
+            Recipient recipient = new Recipient();
+            recipient.setUsers(List.of(extraRecipient));
+            payload.put("recipients", List.of(recipient));
+        }
+        aggregation.setPayload(payload);
```
gwenneg, line 87:
```suggestion
        if (null != extraRecipient) {
            JsonObject payload = TestHelpers.wrapActionToJsonObject(emailActionMessage);
            Recipient recipient = new Recipient();
            recipient.setUsers(List.of(extraRecipient));
            payload.put("recipients", List.of(recipient));
            aggregation.setPayload(payload);
        }
```

gwenneg (approved):
LGTM

gwenneg:
Did I break something with my commit? 🤔 Checking...

gwenneg:
I did 😢 Fixing the build.

gwenneg (approved):

PR #1834, opened 2023-03-28
refactor: allow specifying the full URL for the BOP host

MikelAlejoBR:
This will help us adapt the service to environments that don't have the BOP implemented with the HTTPS protocol.

### DIFF
--- a/.rhcicd/clowdapp-engine.yaml
+++ b/.rhcicd/clowdapp-engine.yaml
@@ -149,7 +149,7 @@ objects:
         - name: PROCESSOR_EMAIL_BOP_ENV
           value: ${BACKOFFICE_CLIENT_ENV}
         - name: PROCESSOR_EMAIL_BOP_URL
-          value: ${BACKOFFICE_SCHEME}://${BACKOFFICE_HOST}/v1/sendEmails
+          value: ${BACKOFFICE_SCHEME}://${BACKOFFICE_HOST}:${BACKOFFICE_PORT}/v1/sendEmails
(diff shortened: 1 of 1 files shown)


.rhcicd/clowdapp-engine.yaml
```diff
@@ -254,6 +254,9 @@ parameters:
 - name: BACKOFFICE_HOST
   description: backoffice URL
   value: backoffice-proxy.apps.ext.spoke.preprod.us-west-2.aws.paas.redhat.com
+- name: BACKOFFICE_SCHEME
+  description: The scheme for the backoffice host's URL
+  value: https
```
josejulio, line 259: Probably a more disruptive change, but wouldn't it be better to  use something like BACKOFFICE_URL ?

Also - i find it curious that we are using an external url to talk with BOP - thought it was within the cluster. 

WDYT @gwenneg  ?

MikelAlejoBR: Done!

josejulio (approved):
lgtm

We will need to update the references to BACKOFFICE_HOST in stage/prod.

josejulio (approved):
lgtm

We are still using BACKOFFICE_SCHEME as BACKOFFICE_HOST is not managed by us

gwenneg:
Is there a reason why we're not simply using one env var instead of three?

josejulio:
> I just remembered `BACKOFFICE_HOST` is defined outside of our configuration in app-interface

exactly

josejulio:
Merging - this was tested on the target environment

PR #1841, opened 2023-04-03
feature: include more details when requests to Sources fail

MikelAlejoBR:
Includes the endpoint's UUID and the invoked method in order to clarify what might have gone wrong.

## Links
[[RHCLOUD-24435]](https://issues.redhat.com/browse/RHCLOUD-24435)

### DIFF
--- a/common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
+++ b/common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
@@ -12,6 +12,8 @@ import org.eclipse.microprofile.rest.client.inject.RestClient;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
+import javax.ws.rs.WebApplicationException;
+import java.util.UUID;
 
 @ApplicationScoped
 public class SecretUtils {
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
```diff
@@ -300,4 +345,18 @@ protected boolean isBasicAuthNullOrBlank(final BasicAuthentication basicAuthenti
...
+     * problem might have been. Then, it rethrows the exception.
+     * @param endpointUuid the endpoint's UUID to prepend.
+     * @param e the exception that was thrown by the {@link SourcesService}.
+     */
+    protected void prependEndpointUuidToExceptionAndRethrow(final UUID endpointUuid, final WebApplicationException e) {
+        throw new WebApplicationException(
+            String.format("[endpoint_uuid: %s]%s", endpointUuid, e.getMessage()),
+            e.getResponse()
+        );
+    }
```
MikelAlejoBR, line 361: I was thinking that maybe an alternative could be the following one:

```
final UUID debugUuid = UUID.random();

Log.errorf(String.format("[debug_uuid: %s][endpoint_uuid: %s]%s", debugUuid, endpointUuid, e.getMessage());

throw new WebApplicationException(String.format("an error occurred when reaching out Sources. Debug UUID: %s", debugUuid));
```
Since the consumers are not interested in this much detail? Although I don't know if it is the best moment to get exquisite about this...

PR #1843, opened 2023-04-04
[RHCLOUD-25078] feature: checked exceptions for the "SecretUtils" service

MikelAlejoBR:
The checked exceptions will force the caller to explicitly deal with any errors that may occur during a Sources request call. The goal of this is to force the developer to think about which actions should be taken when the external service is reporting errors.

The reason for this change is that the "get endpoints" endpoint was reporting 404 errors when at least one endpoint had references to non-existing Sources secrets. In theory this should not happen, because we already took good care of making sure that every operation leaves the Sources references as they should in the [RHCLOUD-24435](https://issues.redhat.com/browse/RHCLOUD-24435) ticket.

However, there are a couple of "unlikely" scenarios in which this may happen:

* The secrets dissapear in Sources for any unexpected reason. One of them could be, for example, a bad data migration on their side.
* QE tests may spawn two Sources instances, and this could make some tests fail if they end up creating secrets with cross references to those instances.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
@@ -431,6 +431,8 @@ public class EndpointResource {
                     this.secretUtils.updateSecretsForEndpoint(dbEndpoint);
                 } catch (final SourcesException e) {
                     Log.error(e.getMessage(), e);
+
+                    throw new InternalServerErrorException("Unable to update the integration due to an internal error");
                 }
             }
         }
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
@@ -52,6 +52,7 @@ import javax.ws.rs.BeanParam;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
+import javax.ws.rs.InternalServerErrorException;
 import javax.ws.rs.NotFoundException;
 import javax.ws.rs.POST;
 import javax.ws.rs.PUT;
@@ -230,6 +231,8 @@ public class EndpointResource {
                 this.secretUtils.createSecretsForEndpoint(endpoint);
             } catch (final SourcesException e) {
                 Log.error(e.getMessage(), e);
+
+                throw new InternalServerErrorException("Unable to create the integration due to an internal error");
             }
         }
 
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
@@ -26,6 +26,8 @@ import com.redhat.cloud.notifications.routers.models.EndpointPage;
 import com.redhat.cloud.notifications.routers.models.Meta;
 import com.redhat.cloud.notifications.routers.models.RequestSystemSubscriptionProperties;
 import com.redhat.cloud.notifications.routers.sources.SecretUtils;
+import com.redhat.cloud.notifications.routers.sources.SourcesException;
+import io.quarkus.logging.Log;
 import io.vertx.core.json.JsonObject;
 import org.eclipse.microprofile.openapi.annotations.Operation;
 import org.eclipse.microprofile.openapi.annotations.enums.ParameterIn;
@@ -171,7 +173,11 @@ public class EndpointResource {
         for (Endpoint endpoint: endpoints) {
             // Fetch the secrets from Sources.
             if (this.featureFlipper.isSourcesUsedAsSecretsBackend()) {
-                this.secretUtils.loadSecretsForEndpoint(endpoint);
+                try {
+                    this.secretUtils.loadSecretsForEndpoint(endpoint);
+                } catch (final SourcesException e) {
+                    Log.error(e.getMessage(), e);
+                }
             }
         }
 
@@ -220,7 +226,11 @@ public class EndpointResource {
          * for more details.
          */
         if (this.featureFlipper.isSourcesUsedAsSecretsBackend()) {
-            this.secretUtils.createSecretsForEndpoint(endpoint);
+            try {
+                this.secretUtils.createSecretsForEndpoint(endpoint);
+            } catch (final SourcesException e) {
+                Log.error(e.getMessage(), e);
+            }
         }
 
         return endpointRepository.createEndpoint(endpoint);
@@ -296,7 +306,11 @@ public class EndpointResource {
         } else {
             // Fetch the secrets from Sources.
             if (this.featureFlipper.isSourcesUsedAsSecretsBackend()) {
-                this.secretUtils.loadSecretsForEndpoint(endpoint);
+                try {
+                    this.secretUtils.loadSecretsForEndpoint(endpoint);
+                } catch (final SourcesException e) {
+                    Log.error(e.getMessage(), e);
+                }
             }
 
             return endpoint;
@@ -319,7 +333,11 @@ public class EndpointResource {
         // Clean up the secrets in Sources.
         if (this.featureFlipper.isSourcesUsedAsSecretsBackend()) {
             Endpoint ep = endpointRepository.getEndpoint(orgId, id);
-            this.secretUtils.deleteSecretsForEndpoint(ep);
+            try {
+                this.secretUtils.deleteSecretsForEndpoint(ep);
+            } catch (final SourcesException e) {
+                Log.error(e.getMessage(), e);
+            }
         }
 
         endpointRepository.deleteEndpoint(orgId, id);
@@ -406,7 +424,11 @@ public class EndpointResource {
                 databaseEndpointProps.setBasicAuthentication(incomingEndpointProps.getBasicAuthentication());
                 databaseEndpointProps.setSecretToken(incomingEndpointProps.getSecretToken());
 
-                this.secretUtils.updateSecretsForEndpoint(dbEndpoint);
+                try {
+                    this.secretUtils.updateSecretsForEndpoint(dbEndpoint);
+                } catch (final SourcesException e) {
+                    Log.error(e.getMessage(), e);
+                }
             }
         }
 
--- a/backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
+++ b/backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
@@ -27,6 +27,7 @@ import com.redhat.cloud.notifications.routers.engine.EndpointTestService;
 import com.redhat.cloud.notifications.routers.models.EndpointPage;
 import com.redhat.cloud.notifications.routers.models.RequestSystemSubscriptionProperties;
 import com.redhat.cloud.notifications.routers.sources.Secret;
+import com.redhat.cloud.notifications.routers.sources.SourcesRuntimeException;
 import com.redhat.cloud.notifications.routers.sources.SourcesService;
 import io.quarkus.runtime.LaunchMode;
 import io.quarkus.runtime.configuration.ProfileManager;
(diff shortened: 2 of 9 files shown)


backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -162,7 +164,11 @@ public EndpointPage getEndpoints(
         for (Endpoint endpoint: endpoints) {
             // Fetch the secrets from Sources.
             if (this.featureFlipper.isSourcesUsedAsSecretsBackend()) {
-                this.secretUtils.loadSecretsForEndpoint(endpoint);
+                try {
+                    this.secretUtils.loadSecretsForEndpoint(endpoint);
+                } catch (final SourcesException e) {
+                    Log.error(e.getMessage());
```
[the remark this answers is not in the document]
MikelAlejoBR: True! I have modified it to be `Log.error(e.getMessage(), e)`, since the error message is already prepared to be logged, with all the details we could require. Thanks!

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -162,7 +164,11 @@ public EndpointPage getEndpoints(
         for (Endpoint endpoint: endpoints) {
             // Fetch the secrets from Sources.
             if (this.featureFlipper.isSourcesUsedAsSecretsBackend()) {
-                this.secretUtils.loadSecretsForEndpoint(endpoint);
+                try {
+                    this.secretUtils.loadSecretsForEndpoint(endpoint);
+                } catch (final SourcesException e) {
+                    Log.error(e.getMessage(), e);
+                }
```
gwenneg, line 171: I'll add comments to each case but it doesn't necessarily mean I'm requesting changes. I just want to make sure I understand each of them.

gwenneg: In this case, we're loading multiple endpoints of various types and a Sources exception with one of these endpoints could prevent the entire list from being loaded and returned to the caller. Sources exceptions can be thrown for many reasons, one of them being the absence of data on the Sources side while the Notifications DB contains an ID that points to a Sources DB entry.

It makes complete sense to me to make sure a failure with one endpoint does not prevent the other endpoints from being loaded. Now, what do we do with the endpoint affected by the failure?
- Return inconsistent data? (current code in this PR)
- Exclude the endpoint from the returned list? (weird from a customer perspective)
- Return something that shows the endpoint exists but could not be loaded properly?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -211,7 +217,11 @@ public Endpoint createEndpoint(@Context SecurityContext sec,
          * for more details.
          */
         if (this.featureFlipper.isSourcesUsedAsSecretsBackend()) {
-            this.secretUtils.createSecretsForEndpoint(endpoint);
+            try {
+                this.secretUtils.createSecretsForEndpoint(endpoint);
+            } catch (final SourcesException e) {
+                Log.error(e.getMessage(), e);
+            }
```
gwenneg, line 224: Here, we are unable to store a secret in Sources during an endpoint creation. I think the entire endpoint creation should fail in that case, resulting in a 500 status for the caller. We would still need the error log entry.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -268,7 +278,11 @@ public Endpoint getEndpoint(@Context SecurityContext sec, @PathParam("id") UUID
         } else {
             // Fetch the secrets from Sources.
             if (this.featureFlipper.isSourcesUsedAsSecretsBackend()) {
-                this.secretUtils.loadSecretsForEndpoint(endpoint);
+                try {
+                    this.secretUtils.loadSecretsForEndpoint(endpoint);
+                } catch (final SourcesException e) {
+                    Log.error(e.getMessage(), e);
+                }
```
gwenneg, line 285: Here, we're loading a single endpoint and an exception is thrown while retrieving its secret from Sources. Is it wise returning inconsistent data or should we show the caller that something went wrong?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -291,7 +305,11 @@ public Response deleteEndpoint(@Context SecurityContext sec, @PathParam("id") UU
         // Clean up the secrets in Sources.
         if (this.featureFlipper.isSourcesUsedAsSecretsBackend()) {
             Endpoint ep = endpointRepository.getEndpoint(orgId, id);
-            this.secretUtils.deleteSecretsForEndpoint(ep);
+            try {
+                this.secretUtils.deleteSecretsForEndpoint(ep);
+            } catch (final SourcesException e) {
+                Log.error(e.getMessage(), e);
+            }
```
gwenneg, line 312: Here, we're deleting an endpoint and the Sources data deletion throws an exception. It looks fine to me to keep deleting the data on our side and log the exception. There is however a risk of leaving orphan records in the Sources DB, depending on the cause of the exception and which actions are taken by the dev team.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -378,7 +396,11 @@ public Response updateEndpoint(@Context SecurityContext sec,
                 databaseEndpointProps.setBasicAuthentication(incomingEndpointProps.getBasicAuthentication());
                 databaseEndpointProps.setSecretToken(incomingEndpointProps.getSecretToken());
 
-                this.secretUtils.updateSecretsForEndpoint(dbEndpoint);
+                try {
+                    this.secretUtils.updateSecretsForEndpoint(dbEndpoint);
+                } catch (final SourcesException e) {
+                    Log.error(e.getMessage(), e);
+                }
```
gwenneg, line 403: Like for the endpoint creation, I would be in favor of a failure (500) returned to the caller if an endpoint update fails because we were unable to update a secret in Sources. The error log entry is still needed.

gwenneg (commented):
@josejulio @g-duval Please do not hesitate to share your opinion about my comments below.

engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
```diff
@@ -133,7 +134,11 @@ private JsonObject buildPayload(Event event, Endpoint endpoint, String originalE
 
         if (featureFlipper.isSourcesUsedAsSecretsBackend()) {
             // Get the basic authentication and secret token secrets from Sources.
-            secretUtils.loadSecretsForEndpoint(endpoint);
+            try {
+                this.secretUtils.loadSecretsForEndpoint(endpoint);
+            } catch (final SourcesException e) {
+                Log.error(e.getMessage(), e);
+            }
```
gwenneg, line 141: I forgot to comment this case.

We're processing a Camel endpoint and the secret cannot be loaded from Sources for some reason. I think the exception thrown here should bubble up to the `EventConsumer` try/catch block. It does really make sense to continue processing the endpoint if the secret is missing.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
MikelAlejoBR: Good call, I agree, yes. I was thinking on returning a generic error message like "Unable to create the integration due to an internal error" and that's it, since the user might not be interested in knowing anything about the secrets back end. Same for the apps.

What do you think?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
MikelAlejoBR: I +1 to that, yes. I was thinking to add another `Info` line there, to say something like "[endpoint_uuid: %s] Integration removed with Sources errors. Associated secrets might not have been deleted" to easily see that "aha, this error message is tied to the successful deletion".

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
MikelAlejoBR: Agree, yes. I will return a similar message to the one above.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
MikelAlejoBR: I am not sure... in this case I wouldn't? Because the user might want to edit or view what the integration has, and if we're not able to show it fully... it might be better to simply throw an error in here?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
josejulio: We should return something to let the user knows that there was a problem loading the endpoint - but I'm not sure how to handle that - maybe a different http code and a custom header with the error? or  a different response that informs the error and wraps the endpoint list ? i.e. `{ errors: [ {} ]}, endpoints: [...]}`

But i agree that we should not return inconsistent data.

As a side note - Is it OK to continue retrieving all the passwords? (I'm guessing there could be security concerns here).

Wondering if secrets has (or has plans to) a UI to securely view the secrets - if that's the case we could simply link to secrets UI or similar.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
josejulio: Sounds good to me

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
josejulio: Could we  have more information about the `SourcesException`? If it is a timeout exception or something that could succed after a retry, then it's worth to fail the request.

IF it's something like "secret not found" it might be worth to let the user know that we couldn't retrieve the secret and allow the user to update to something new.

engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
josejulio: Agree - and maybe have some sort of retry mechanism

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
MikelAlejoBR: Done!

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
MikelAlejoBR: Done!

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
MikelAlejoBR: Done!

engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelTypeProcessor.java
MikelAlejoBR: Done!

MikelAlejoBR:
Regarding the following things:

- Returning a custom structure when listing the endpoints.
- Returning different errors when returning an endpoint.

I wanted to bring up a team discussion on whether we should keep returning credentials to clients or not. I would like to discuss this matter in the bi weekly meeting.

gwenneg:
@MikelAlejoBR What's the status of this dev? Did we answer all pending questions?

MikelAlejoBR:
Yes @gwenneg. Basically since we are going to stop returning the secrets to the users this PR is going to change a bit. I am going to turn it into a draft.

MikelAlejoBR:
Since we are making so many changes to the Sources service integration we decide it to close it.

PR #1851, opened 2023-04-06
[RHCLOUD-25096] expose more metrics

g-duval:


engine/src/main/java/com/redhat/cloud/notifications/processors/slack/RetryCounterProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public void process(Exchange exchange) throws Exception {
+        final String routeId = exchange.getFromRouteId();
+        switch (routeId) {
+            case SLACK_INCOMING_ROUTE :
+                slackRetryCounter.increment();
+                break;
+            default:
+                Log.warnf("Unsupported route Id: %s", routeId);
```
gwenneg, line 35: Is it really necessary to log something? We could simply do nothing in the `default` block.

gwenneg:
```suggestion
                Log.warnf("Unsupported route Id: %s", routeId);
                break;
```

engine/src/main/java/com/redhat/cloud/notifications/processors/slack/RetryCounterProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+import javax.inject.Inject;
+
+import static com.redhat.cloud.notifications.processors.slack.SlackRouteBuilder.SLACK_INCOMING_ROUTE;
+
+@ApplicationScoped
+public class RetryCounterProcessor implements Processor {
+
+    public static final String CAMEL_SLACK_RETRY_COUNTER = "camel.slack.retry.counter";
+    @Inject
+    MeterRegistry registry;
```
gwenneg, line 18: We only need this injection or the one in the constructor, but not both 😃

engine/src/main/java/com/redhat/cloud/notifications/processors/slack/RetryCounterProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+    public RetryCounterProcessor(MeterRegistry registry) {
+        this.registry = registry;
+        slackRetryCounter = registry.counter(CAMEL_SLACK_RETRY_COUNTER);
+    }
+
+    @Override
+    public void process(Exchange exchange) throws Exception {
+        final String routeId = exchange.getFromRouteId();
+        switch (routeId) {
+            case SLACK_INCOMING_ROUTE :
```
gwenneg, line 31: Shouldn't this be `SLACK_OUTGOING_ROUTE`? Retries will most likely happen when the call to Slack fails.

engine/src/main/resources/application.properties
```diff
@@ -132,6 +135,8 @@ reinject.enabled=false
 # Use this property to load the templates from the DB. Temp, to be removed soon.
 notifications.use-templates-from-db=false
 
+%test.processor.webhook.retry.back-off.initial-value=0.001S
```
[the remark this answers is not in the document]
gwenneg: Should we make the same change for `back-off.max-value`?
https://github.com/RedHatInsights/notifications-backend/blob/4038eae8664351c8790301182f53055ab7614a65/engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java#L59-L63

engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
```diff
@@ -254,8 +272,10 @@ public void doHttpRequest(Event event, Endpoint endpoint, HttpRequest<Buffer> re
                 }
 
                 if (serverError) {
+                    updateRetryMetrics(isEmailEndpoint);
```
gwenneg, line 275: This only counts 1 of the 4 exceptions that can trigger a retry.

engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
```diff
@@ -276,12 +297,39 @@ public void doHttpRequest(Event event, Endpoint endpoint, HttpRequest<Buffer> re
                 details.put("error_message", e.getMessage()); // TODO This message isn't always the most descriptive..
                 history.setDetails(details.getMap());
             }
+            updateMetrics(history.getStatus(), isEmailEndpoint);
```
gwenneg, line 300: We talked about counting successes (instead of processed events) and failures from a `finally` block depending on the notification status. Is that no longer part of the plan?

engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
gwenneg: FailSafe has an `onRetry` clause that could be used from the builder at the beginning of the class. The only problem is that we can't determine if we're retrying an email or a webhook call with the current retry policy. We could have two retry policies though and use the correct one depending on the endpoint type.

engine/src/main/resources/application.properties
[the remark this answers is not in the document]
g-duval: FYI max-value must be higher than `initial-value` * nb retries

g-duval:
> I don't feel comfortable with our current Slack exception management (the one I wrote) because it affects all Camel routes while it should be focused on Slack only. I'll push a commit in this PR to change a few things, then we'll continue the review.

An idea could be using `doCatch` block with a specific exception for Slack like:
`        from(SLACK_DIRECT_ENDPOINT)
                .routeId(SLACK_OUTGOING_ROUTE)
                .process(slackNotificationProcessor)
                .doTry()
                    .toD("slack:${exchangeProperty.channel}?webhookUrl=${exchangeProperty.webhookUrl}", maxEndpointCacheSize)
                .doCatch(IOException.class)
                    .throwException(new SlackRetryException())
                .doFinally();`
And 
`        onException(SlackRetryException.class)
                .maximumRedeliveries(maxRedeliveries)
                .redeliveryDelay(redeliveryDelay)
                .onRedelivery(retryCounterProcessor)
                .retryAttemptedLogLevel(INFO)
                .to(EXCEPTION_DIRECT_ENDPOINT)
                .handled(true); `
I tested it, it works almost the same way, but outgoing route metrics are different.
I guess I will investigate that tomorrow.

engine/src/test/java/com/redhat/cloud/notifications/processors/webhook/WebhookTest.java
```diff
@@ -85,7 +94,14 @@ public class WebhookTest {
...
     void beforeEach() {
-        micrometerAssertionHelper.saveCounterValuesBeforeTest(PROCESSED_COUNTER_NAME);
+        micrometerAssertionHelper.saveCounterValuesBeforeTest(PROCESSED_WEBHOOK_COUNTER);
+        micrometerAssertionHelper.saveCounterValuesBeforeTest(PROCESSED_EMAIL_COUNTER);
+        micrometerAssertionHelper.saveCounterValuesBeforeTest(FAILED_WEBHOOK_COUNTER);
+        micrometerAssertionHelper.saveCounterValuesBeforeTest(FAILED_EMAIL_COUNTER);
+        micrometerAssertionHelper.saveCounterValuesBeforeTest(RETRIED_WEBHOOK_COUNTER);
+        micrometerAssertionHelper.saveCounterValuesBeforeTest(RETRIED_EMAIL_COUNTER);
+        micrometerAssertionHelper.saveCounterValuesBeforeTest(SUCCESSFUL_WEBHOOK_COUNTER);
+        micrometerAssertionHelper.saveCounterValuesBeforeTest(SUCCESSFUL_EMAIL_COUNTER);
```
gwenneg, line 104:
```suggestion
        micrometerAssertionHelper.saveCounterValuesBeforeTest(PROCESSED_WEBHOOK_COUNTER, PROCESSED_EMAIL_COUNTER, FAILED_WEBHOOK_COUNTER, FAILED_EMAIL_COUNTER, RETRIED_WEBHOOK_COUNTER, RETRIED_EMAIL_COUNTER, SUCCESSFUL_WEBHOOK_COUNTER, SUCCESSFUL_EMAIL_COUNTER);
```

gwenneg (approved):
LGTM

PR #1855, opened 2023-04-11
[RHCLOUD-25182] Add Google Chat connector

g-duval:


engine/src/main/java/com/redhat/cloud/notifications/processors/google/chat/NotificationErrorProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+
+        String responseBody = httpException.getResponseBody();
+        Map<String, String> responseHeaders = httpException.getResponseHeaders();
+
+        StringBuilder builder = new StringBuilder();
+        for (String key : responseHeaders.keySet()) {
+            builder.append(key).append(":").append(responseHeaders.get(key)).append(",");
+        }
+
+        Log.infof("Received error response from endpoint: %s, body: %s, headers: %s",
```
MikelAlejoBR, line 33: I think an `errorf` should be used here

engine/src/main/java/com/redhat/cloud/notifications/processors/google/chat/NotificationErrorProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+            new String[]{(exchange.getFromEndpoint() != null ? exchange.getFromEndpoint().getEndpointUri() :
+                    "unknown"), responseBody, builder.toString()});
```
MikelAlejoBR, line 35: Also, I was thinking that since we are storing the `orgId`, `historyId` and `webhookUrl` properties in the exchange, we could get them here and maybe use them in the logs to know exactly which notification failed.

MikelAlejoBR (approved):
LGTM! Just a minor optional comment.

I was thinking that maybe in the future we can start using producer templates in order to send payloads to the `direct` endpoints, instead of having to use `REST` endpoints with a dummy REST Client as we are doing now.

It comes with the downside of coupling the processor directly with the Camel endpoint, but I am unsure if that would be an inconvenience. This way we could save ourselves from having dummy REST clients just to "integrate" with Camel.

Just a thought for opening a discussion in the future :)

engine/src/main/java/com/redhat/cloud/notifications/processors/google/chat/NotificationErrorProcessor.java
g-duval: Sure good Idea !

g-duval:
> LGTM! Just a minor optional comment.
> 
> I was thinking that maybe in the future we can start using producer templates in order to send payloads to the `direct` endpoints, instead of having to use `REST` endpoints with a dummy REST Client as we are doing now.
> 
> It comes with the downside of coupling the processor directly with the Camel endpoint, but I am unsure if that would be an inconvenience. This way we could save ourselves from having dummy REST clients just to "integrate" with Camel.
> 
> Just a thought for opening a discussion in the future :)

Thanks, that's very intresting, anyway we will have to refactor a bit Slack, Teams and Google integrations to avoid code and test dupplicates. Could be an opportunity to review routes according your recommandations.

PR #1856, opened 2023-04-13
Update event-schema to 1.3.0

josejulio:
- Using ObjectMapper from lib
 - Using Notifications generated object instead of custom made


g-duval (approved):
lgtm

PR #1862, opened 2023-04-18
[RHCLOUD-25386] fix error on aggregated emails

g-duval:


josejulio:
I think the JIRA id is missing a digit

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
```diff
@@ -83,9 +84,17 @@ public void sendEmail(User user, Event event, TemplateInstance subject, Template
...
         Timer.Sample processedTimer = Timer.start(registry);
+
         EventType eventType = event.getEventType();
-        String bundleName = eventType.getApplication().getBundle().getName();
-        String applicationName = eventType.getApplication().getName();
+        String bundleName = "NA";
+        String applicationName = "NA";
+        if (eventType != null) {
+            bundleName = eventType.getApplication().getBundle().getName();
+            applicationName = eventType.getApplication().getName();
```
josejulio, line 93: Is this going to work ? This is a stateless session, from what I remember it throws an exception - but I could be wrong :sweat:

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
```diff
@@ -83,9 +84,17 @@ public void sendEmail(User user, Event event, TemplateInstance subject, Template
...
+        } else if (event.getEventWrapper().getEvent() instanceof Action action) {
+            bundleName = action.getBundle();
+            applicationName = action.getApplication();
+        }
```
josejulio, line 97: This should always be true, we are not aggregating with cloud events yet - could you move this to the first option?

josejulio (approved):

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
josejulio: nevermind, it's equal to the previous code

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
josejulio: disregard my previous comment.

g-duval: The first option shouldn't be trying to get App and Bundle from eventType ? because we have much more instant notifications than aggregated notifications ?

PR #1886, opened 2023-04-27
feat: create new Ansible endpoint type

vkrizan:
Adds Event-Driven Ansible type that aliases WebHook processor.

RHCLOUD-25405

### DIFF
--- a/common/src/main/java/com/redhat/cloud/notifications/models/EndpointType.java
+++ b/common/src/main/java/com/redhat/cloud/notifications/models/EndpointType.java
@@ -10,7 +10,9 @@ public enum EndpointType {
     @JsonProperty("email_subscription")
     EMAIL_SUBSCRIPTION(false),
     @JsonProperty("camel")
-    CAMEL(true);
+    CAMEL(true),
+    @JsonProperty("ansible")
+    ANSIBLE(false);
 
     public final boolean requiresSubType;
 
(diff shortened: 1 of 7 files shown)


common/src/main/java/com/redhat/cloud/notifications/models/EndpointType.java
```diff
@@ -10,7 +10,9 @@ public enum EndpointType {
     @JsonProperty("email_subscription")
     EMAIL_SUBSCRIPTION(false),
     @JsonProperty("camel")
-    CAMEL(true);
+    CAMEL(true),
+    @JsonProperty("eda")
```
vkrizan, line 14: @jeromemarc internally the type is gonna be "eda". Do we want to use a longer name?

josejulio:
We also need to update this file https://github.com/RedHatInsights/notifications-backend/blob/f439ed0704386525c15fbb12d2fd365ac17b8f1e/common/src/main/java/com/redhat/cloud/notifications/models/Endpoint.java#L85

something like:

```
 @JsonSubTypes.Type(value = WebhookProperties.class, name = "eda"), 
```

common/src/main/java/com/redhat/cloud/notifications/models/EndpointType.java
vkrizan: I've decided to go with "ansible". It's better than an abbreviated EDA, and the abbreviation can likely change.

gwenneg:
@vkrizan When we met last week, my mind was focused on the Kafka messages processing and I forgot that we needed to add the `ANSIBLE` endpoint type at several places to properly create/update/load the properties for that endpoint type. Everything should be good now.

gwenneg (approved):
LGTM

gwenneg:

PR #1890, opened 2023-04-28
[RHCLOUD-25186] feature allow overriding default message test endpoint

MikelAlejoBR:
The second step on sending "test events" for integrations was to allow
sending custom messages on those events. This implements it in a way
that the client can decide whether to send a message or not. But if they
do, the message must not be blank.

## Links

[[RHCLOUD-25186]](https://issues.redhat.com/browse/RHCLOUD-25186)

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
@@ -20,7 +20,7 @@ import com.redhat.cloud.notifications.models.EndpointType;
 import com.redhat.cloud.notifications.models.NotificationHistory;
 import com.redhat.cloud.notifications.models.SourcesSecretable;
 import com.redhat.cloud.notifications.models.SystemSubscriptionProperties;
-import com.redhat.cloud.notifications.routers.endpoints.EndpointTestRequest;
+import com.redhat.cloud.notifications.routers.endpoints.InternalEndpointTestRequest;
 import com.redhat.cloud.notifications.routers.engine.EndpointTestService;
 import com.redhat.cloud.notifications.routers.models.EndpointPage;
 import com.redhat.cloud.notifications.routers.models.Meta;
@@ -591,7 +591,7 @@ public class EndpointResource {
             throw new NotFoundException("integration not found");
         }
 
-        final var endpointTestRequest = new EndpointTestRequest(uuid, orgId);
+        final var endpointTestRequest = new InternalEndpointTestRequest(uuid, orgId);
 
         this.endpointTestService.testEndpoint(endpointTestRequest);
     }
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
@@ -20,6 +20,7 @@ import com.redhat.cloud.notifications.models.EndpointType;
 import com.redhat.cloud.notifications.models.NotificationHistory;
 import com.redhat.cloud.notifications.models.SourcesSecretable;
 import com.redhat.cloud.notifications.models.SystemSubscriptionProperties;
+import com.redhat.cloud.notifications.routers.endpoints.EndpointTestRequest;
 import com.redhat.cloud.notifications.routers.endpoints.InternalEndpointTestRequest;
 import com.redhat.cloud.notifications.routers.engine.EndpointTestService;
 import com.redhat.cloud.notifications.routers.models.EndpointPage;
@@ -573,6 +574,7 @@ public class EndpointResource {
      * @return a "no content" response on success.
      */
     @APIResponse(responseCode = "204", description = "No Content")
+    @Consumes(APPLICATION_JSON)
     @POST
     @Path("/{uuid}/test")
     @Parameters({
@@ -584,16 +586,26 @@ public class EndpointResource {
             )
     })
     @RolesAllowed(ConsoleIdentityProvider.RBAC_WRITE_INTEGRATIONS_ENDPOINTS)
-    public void testEndpoint(@Context SecurityContext sec, @RestPath UUID uuid) {
+    public void testEndpoint(@Context SecurityContext sec, @RestPath UUID uuid, @RequestBody EndpointTestRequest requestBody) {
         final String orgId = SecurityContextUtil.getOrgId(sec);
 
         if (!this.endpointRepository.existsByUuidAndOrgId(uuid, orgId)) {
             throw new NotFoundException("integration not found");
         }
 
-        final var endpointTestRequest = new InternalEndpointTestRequest(uuid, orgId);
+        final InternalEndpointTestRequest internalEndpointTestRequest;
+        if (requestBody == null) {
+            internalEndpointTestRequest = new InternalEndpointTestRequest(uuid, null, orgId);
+        } else {
+            if (requestBody.isMessageBlank()) {
+                throw new BadRequestException("the custom message cannot be empty");
+            }
 
-        this.endpointTestService.testEndpoint(endpointTestRequest);
+            internalEndpointTestRequest = new InternalEndpointTestRequest(uuid, requestBody.message, orgId);
+
+        }
+
+        this.endpointTestService.testEndpoint(internalEndpointTestRequest);
     }
 
     private static void checkSystemEndpoint(EndpointType endpointType) {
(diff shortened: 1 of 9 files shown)


backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
```diff
@@ -2055,18 +2056,20 @@ void testEndpointTest() {
         given()
             .header(identityHeader)
             .when()
+            .contentType(JSON)
```
MikelAlejoBR, line 2059: I was forced to specify the content type as `application/json` even if those requests contained no body, as otherwise I would get a "415 Unsupported Media Type" error back from Quarkus. Is there a way to avoid this?

engine/src/test/java/com/redhat/cloud/notifications/routers/EndpointTestResourceTest.java
```diff
@@ -106,4 +117,82 @@ void testEndpoint() {
...
+        final Payload payload = event.getPayload();
+        final Map<String, Object> payloadAdditionalProperties = payload.getAdditionalProperties();
+
+        final int expectedPayloadAdditionalPropertiesCount = 1;
+        Assertions.assertEquals(expectedPayloadAdditionalPropertiesCount, payloadAdditionalProperties.size(), "unexpected number of payload additional properties");
+
+        final String payloadValue = (String) payload.getAdditionalProperties().get(TestEventHelper.TEST_ACTION_PAYLOAD_KEY);
+
+        Assertions.assertEquals(customMessage, payloadValue, "unexpected event payload value");
+    }
```
MikelAlejoBR, line 197: For some odd reason this test times out waiting for messages. I am guessing I might need to use a different memory sink for this test?

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
gwenneg: We depend on `quarkus-resteasy-reactive-jackson` which makes `application/json` the expected content type for all inputs and outputs. I don't think there's an issue here though.

gwenneg:
We could provide one default body per endpoint/integration type with a REST endpoint that takes the endpoint type as the input and returns that default body. WDYT @MikelAlejoBR?

MikelAlejoBR:
> We could provide one default body per endpoint/integration type with a REST endpoint that takes the endpoint type as the input and returns that default body. WDYT @MikelAlejoBR? It would be easier for users to test the endpoint with a custom body based on the default one.

If I understood correctly, you are suggesting to generate what an event would look like for each integration type and then send that body. If that is what you're thinking, that was the phase 3 of the development of this feature:

1. Simply send a generic event for the integration.
2. Allow sending a custom message.
3. Allow the user to select the integration type in order to receive an event similar to what they would receive if the real application had sent it.

Is this so? We could skip step two completely, but we'll have to talk to Alexandra too to cancel the frontend modification for this PR.

josejulio:
> We could provide one default body per endpoint/integration type with a REST endpoint that takes the endpoint type as the input and returns that default body. WDYT @MikelAlejoBR? It would be easier for users to test the endpoint with a custom body based on the default one.

I had a similar idea for the email templates (the part about having default values for these)

https://issues.redhat.com/browse/RHCLOUD-25625

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -515,16 +517,26 @@ public boolean unsubscribeEmail(
             )
     })
     @RolesAllowed(ConsoleIdentityProvider.RBAC_WRITE_INTEGRATIONS_ENDPOINTS)
-    public void testEndpoint(@Context SecurityContext sec, @RestPath UUID uuid) {
+    public void testEndpoint(@Context SecurityContext sec, @RestPath UUID uuid, @RequestBody EndpointTestRequest requestBody) {
```
gwenneg, line 520: `requestBody` could be a `@NotBlank String` param. Is there any reason why we need the `EndpointTestRequest` class?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -515,16 +517,26 @@ public boolean unsubscribeEmail(
...
         }
 
-        final var endpointTestRequest = new EndpointTestRequest(uuid, orgId);
+        final InternalEndpointTestRequest internalEndpointTestRequest;
+        if (requestBody == null) {
+            internalEndpointTestRequest = new InternalEndpointTestRequest(uuid, null, orgId);
+        } else {
+            if (requestBody.isMessageBlank()) {
+                throw new BadRequestException("the custom message cannot be empty");
+            }
```
gwenneg, line 533: If `EndpointTestRequest` is needed, this validation could be done with a `@NotBlank` annotation on `EndpointTestRequest#message` and the `@Valid` annotation on `requestBody`.

common/src/main/java/com/redhat/cloud/notifications/routers/endpoints/EndpointTestRequest.java
```diff
@@ -1,23 +1,27 @@
...
-public class EndpointTestRequest {
-
-    @NotNull
-    public UUID endpointUuid;
+public final class EndpointTestRequest {
+    public String message;
 
-    @NotBlank
-    public String orgId;
+    public EndpointTestRequest(@JsonProperty("message") final String message) {
```
gwenneg, line 15: Is that `@JsonProperty` annotation needed?

common/src/main/java/com/redhat/cloud/notifications/routers/endpoints/InternalEndpointTestRequest.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public InternalEndpointTestRequest(final UUID endpointUuid, final String orgId) {
+        this.endpointUuid = endpointUuid;
+        this.orgId = orgId;
+    }
+
+    public InternalEndpointTestRequest(
+        @JsonProperty("endpointUuid") final UUID endpointUuid,
+        @JsonProperty("message") final String message,
+        @JsonProperty("orgId") final String orgId
```
gwenneg, line 34: I don't think we need these `@JsonProperty` annotations.

engine/src/test/java/com/redhat/cloud/notifications/routers/EndpointTestResourceTest.java
gwenneg: FYI I'm still reviewing the tests. I may answer this question soon.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
MikelAlejoBR: I did it so that the following structure gets deserialized automatically:

```json
{
    "message": "Hello, World!"
}
```

Since otherwise I had to be using a JSON Object and what not...

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
MikelAlejoBR: True! It's way better and cleaner in the way you suggest it. Done!

common/src/main/java/com/redhat/cloud/notifications/routers/endpoints/EndpointTestRequest.java
MikelAlejoBR: I think so, because if I remove it, then I start getting the following errors:

```com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot construct instance of `com.redhat.cloud.notifications.routers.endpoints.EndpointTestRequest` (although at least one Creator exists): cannot deserialize from Object value (no delegate- or property-based Creator)
```

MikelAlejoBR: Apparently I can [add an empty constructor as well](https://stackoverflow.com/a/53192674/6319771), as an alternative. But I think it is just cleaner to have a single constructor, even if it is annotated... What do you think, though?

common/src/main/java/com/redhat/cloud/notifications/routers/endpoints/InternalEndpointTestRequest.java
MikelAlejoBR: In this case I had the same problem as above, since I only defined that constructor...

gwenneg:
@MikelAlejoBR Could you please take a look at the build failure?

gwenneg:
Rebased onto `master`.

gwenneg (approved):
LGTM

PR #1891, opened 2023-04-29
[SPM-2047] refactor: update patch daily digest to new format

michalslomczynski:
Changes:
- Added synopsis field to aggregator
- Order of advisory types in emails
- Included synopsis column in emails
- Group `unspecified` advisories under `other` group


engine/src/main/resources/templates/Patch/dailyEmailBodyV2.html
michalslomczynski, line 1: @g-duval is it fine if I modify V2?

g-duval: Yes it is :+1:

g-duval (approved):
/lgtm

PR #1902, opened 2023-05-04
Adding date to security compliance tag

xxlhacker:
Adding the build date to the security compliance tag in the build script


casey-williams-rh (approved):
👍

PR #1903, opened 2023-05-05
[RHCLOUD-25597] Export service integration — Entity transformations

MikelAlejoBR:
The transformers will take care of transforming the data in the precise
way we want it to be exported.

## Links

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/exports/TransformationException.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/exports/TransformationException.java
@@ -1,0 +1,10 @@
+package com.redhat.cloud.notifications.exports;
+
+/**
+ * A wrapper exception class which wraps any transformation errors.
+ */
+public class TransformationException extends Exception {
+    public TransformationException(final Throwable e) {
+        super(e);
+    }
+}
--- a/engine/src/main/java/com/redhat/cloud/notifications/exports/transformers/event/CSVEventTransformer.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/exports/transformers/event/CSVEventTransformer.java
@@ -1,0 +1,50 @@
+package com.redhat.cloud.notifications.exports.transformers.event;
+
+import com.redhat.cloud.notifications.exports.ResultsTransformer;
+import com.redhat.cloud.notifications.exports.TransformationException;
+import com.redhat.cloud.notifications.models.Event;
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVPrinter;
+
+import java.io.IOException;
+import java.io.StringWriter;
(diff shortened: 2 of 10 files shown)


engine/src/main/java/com/redhat/cloud/notifications/exports/TransformationException.java
```diff
@@ -0,0 +1,10 @@
+package com.redhat.cloud.notifications.exports;
+
+/**
+ * A wrapper exception class which wraps any transformation errors.
+ */
+public class TransformationException extends Exception {
```
gwenneg, line 6: This is not used in this PR. Will it be used later?

engine/src/main/java/com/redhat/cloud/notifications/exports/transformers/event/CSVEventTransformer.java
```diff
@@ -0,0 +1,38 @@
...
+        builder.append(System.lineSeparator());
+
+        for (final Event event : events) {
+            builder.append(
+                String.format(
+                    "%s,%s,%s,%s,%s%s",
+                    event.getId(),
+                    event.getBundleDisplayName(),
+                    event.getApplicationDisplayName(),
+                    event.getEventTypeDisplayName(),
```
gwenneg, line 29: What if a comma is used in a display name? Commas are supported in CSV values as long as these values are wrapped into double quotes. We should either support that or forbid commas in display names.

josejulio: We should search for csv library that support these features, it can get tricky if we want to support other data (such as descriptions) that can already have commas and double quotes.

engine/src/main/java/com/redhat/cloud/notifications/exports/TransformationException.java
MikelAlejoBR: I was thinking to standardize what the transformers should look like. I also modified it in order to show that it might throw an exception, which is basically a generic wrapper to the multiple exceptions the transformers might throw. That way we have to handle just a single exception... However I am not very much convinced about it.

I also thought that this might be the way to end up building a factory for the transformers, but I don't know...

engine/src/main/java/com/redhat/cloud/notifications/exports/transformers/event/CSVEventTransformer.java
MikelAlejoBR: Yeah, good idea. I thought we could do it without having to deal with exceptions here but of course there are many cases I didn't think about... So I've used Apache's CSV library, which implements [RFC 4180](https://datatracker.ietf.org/doc/html/rfc4180) for CSVs, among others.

gwenneg (approved):
LGTM

PR #823, opened 2021-12-03
Adding an first iteration of email template for report download

h-kataria:
This is an email template for Insights for AAP to send out a link to customers when Report is ready for download

https://issues.redhat.com/browse/AA-825

PR #1908, opened 2023-05-10
RHCLOUD-25835 Offboard RHOSAK

gwenneg:
Areas not covered by this PR:
- [x] The B/A/ET hierarchy needs to be removed from the DB
- [ ] The email templates need to be removed from the DB
- [ ] The Slack/Teams/Chat templates need to be updated to remove the logic dedicated to RHOSAK
- [ ] The frontend needs an update to stop showing the Application Services bundle


engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/EmailPayloadAggregatorFactory.java
```diff
@@ -30,8 +29,6 @@ public static AbstractEmailPayloadAggregator by(EmailAggregationKey aggregationK
                 switch (application.toLowerCase()) { // TODO Remove toLowerCase if possible
                     case ANSIBLE:
                         return new AnsibleEmailAggregator();
-                    case RHOSAK:
```
josejulio, line 33: Unrelated, are we still onboarding `ansible` ? It looks like it's part of the `application-services`  bundle aswell.

engine/src/test/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationServiceTest.java
```diff
@@ -112,18 +112,6 @@ void testMigration() {
...
-        // App: rhosak
-        Application rhosak = resourceHelpers.createApp(applicationServices.getId(), "rhosak");
-        EventType scheduledUpgrade = resourceHelpers.createEventType(rhosak.getId(), "scheduled-upgrade");
-        EventType disruption = resourceHelpers.createEventType(rhosak.getId(), "disruption");
-        EventType instanceCreated = resourceHelpers.createEventType(rhosak.getId(), "instance-created");
-        EventType instanceDeleted = resourceHelpers.createEventType(rhosak.getId(), "instance-deleted");
-        EventType actionRequired = resourceHelpers.createEventType(rhosak.getId(), "action-required");
-
         /*
          * Bundle: ansible
```
josejulio, line 116: Bundle `ansible` ? On the other comment I saw that `ansible` was an application - maybe we need to check if there is something wrong?

josejulio (approved):
lgtm - some unrelated questions

engine/src/test/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationServiceTest.java
gwenneg: Good questions.

This looks like unfinished business and we should contact whoever added the `ansible` instant template to check if that still makes sense.

gwenneg: https://github.com/RedHatInsights/notifications-backend/pull/823

gwenneg:
```
* Suggested changes for ansible:
 - Remove code from aggregator
 - Use ansible as the bundle and reports as the app +
 - Add instant body and title
```

PR #1909, opened 2023-05-10
RHCLOUD-21865 Stop supporting deprecated sort fields

josejulio:


backend/src/main/java/com/redhat/cloud/notifications/db/Query.java
```diff
@@ -193,11 +178,6 @@ public Optional<Sort> getSort() {
         if (!sortFields.containsKey(lowerCaseSortColumn)) {
             throw new BadRequestException("Unknown sort field specified: " + sort.sortColumn);
         } else {
-            // NOTIF-674 Delete after confirming all deprecated fields are no longer used
-            if (deprecatedSortFields != null && deprecatedSortFields.contains(lowerCaseSortColumn)) {
-                Log.warnf("NOTIF-674 Using deprecated sort field: %s", lowerCaseSortColumn);
```
josejulio, line 198: Checked the logs and the sentry channel, the last occurence is April 4th

gwenneg: I started writing a question exactly about that and got distracted by something else. You answered before I could post it 😄

common/src/main/java/com/redhat/cloud/notifications/models/NotificationHistory.java
```diff
@@ -32,17 +31,7 @@ public class NotificationHistory extends CreationTimestamped {
             "created", "nh.created",
             "invocation_time", "nh.invocationTime",
             "invocation_result", "nh.invocationResult",
-            "status", "nh.status",
-            // NOTIF-674 Delete after the frontend has been updated
-            "nh.created", "nh.created",
-            "invocationtime", "nh.invocationTime",
-            "invocationresult", "nh.invocationResult",
-            "endpoint_id", "endpoint_id"
```
gwenneg, line 40: Is it normal that we're not allowing a sort on the endpoint anymore?

josejulio: > Is it normal that we're not allowing a sort on the endpoint anymore?

The only reason I can see is that sorting by `endpoint_id` doesn't make lots of sense to me :thinking: .

But the format is good and we could include it on the supported sorting. 
What do you think?

gwenneg: I just checked the code. It doesn't make sense to sort on the `endpoint_id` indeed.

gwenneg (approved):
LGTM

PR #1913, opened 2023-05-11
Deploy the same image tag for backend and engine in ephemeral

gwenneg:
This PR:
- makes the PR check two times faster by only triggering one ephemeral deployment while testing both `backend` and `engine` images
- makes the IQE tests run against the same image tag of both `backend` and `engine`, solving the frequent compatibility issues we had when a breaking change in a new component image was not compatible with the old image of the other component


MikelAlejoBR (approved):
Very nice improvement! LGTM!

PR #1914, opened 2023-05-11
[RHCLOUD-25597] Export service integration — Request handler

MikelAlejoBR:
Handles the incoming export requests accordingly. The handler makes sure
that the export requests are requested for Notifications for the
supported resources and formats. It also validates that the incoming
inputs are supported and valid, before querying the database and
transforming the results.

## Links


engine/src/main/resources/application.properties
```diff
@@ -53,6 +53,13 @@ mp.messaging.incoming.fromcamel.group.id=integrations
 mp.messaging.incoming.fromcamel.key.deserializer=org.apache.kafka.common.serialization.StringDeserializer
 mp.messaging.incoming.fromcamel.value.deserializer=org.apache.kafka.common.serialization.StringDeserializer
 
+# Input queue for the "export requests" coming from the export service.
+mp.messaging.incoming.export-requests.connector=smallrye-kafka
+mp.messaging.incoming.export-requests.topic=platform.export.requests
+mp.messaging.incoming.export-requests.group.id=export-requests
```
gwenneg, line 59: Let's stick with the same group ID we use for all other topics.
```suggestion
mp.messaging.incoming.export-requests.group.id=integrations
```

engine/src/main/resources/application.properties
```diff
@@ -176,3 +183,15 @@ quarkus.rest-client.internal-google-chat.trust-store-type=${clowder.endpoints.no
...
+# point to the private internal server.
+# Export service's development PSK value. Specified here to avoid Quarkus from
+# complaining that the configuration parameter is missing. In the case that you
+# are using a real Export Service application to test the integration, you will
+# need to set this PSK in the PSKS environment variable on that end.
+export-service.psk=development-value-123
+quarkus.rest-client.export-service.url=${clowder.endpoints.export-service.url:http://localhost:10010}
+quarkus.rest-client.export-service.trust-store=${clowder.endpoints.notifications-engine-service.trust-store-path}
+quarkus.rest-client.export-service.trust-store-password=${clowder.endpoints.notifications-engine-service.trust-store-password}
+quarkus.rest-client.export-service.trust-store-type=${clowder.endpoints.notifications-engine-service.trust-store-type}
```
gwenneg, line 197:
```suggestion
quarkus.rest-client.export-service.trust-store=${clowder.endpoints.export-service.trust-store-path}
quarkus.rest-client.export-service.trust-store-password=${clowder.endpoints.export-service.trust-store-password}
quarkus.rest-client.export-service.trust-store-type=${clowder.endpoints.export-service.trust-store-type}
```

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsService.java
```diff
@@ -0,0 +1,91 @@
...
+
+import javax.ws.rs.Consumes;
+import javax.ws.rs.HeaderParam;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.MediaType;
+import java.util.UUID;
+
+@RegisterRestClient(configKey = "export-service")
+public interface ExportsService {
```
gwenneg, line 16: Nitpick:
```suggestion
@Path("/app/export/v1")
public interface ExportsService {
```

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsEventListener.java
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    /**
+     * Extracts the export request's {@link UUID}, which comes in the Cloud
+     * Event's subject. Beware that the Cloud Event's {@link UUID} is not the
+     * same as the export request's {@link UUID}, or the resource's
+     * {@link UUID}.
+     * @param subject the received subject of the Cloud Event.
+     * @return the extracted {@link UUID} from the subject.
+     */
```
gwenneg, line 324: Is there any reason why the request UUID is not used as the cloud event UUID?

Why is a resource UUID needed in addition to the request UUID?

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsEventListener.java
```diff
@@ -0,0 +1,335 @@
...
+     * Listens to the exports channel, extracts the request's data, and if
+     * Notifications is the target application, and the requested resource and
+     * formats are supported, the corresponding payload is sent to the export
+     * service.
+     * @param message the incoming message from the channel.
+     * @return a read commit on the received message.
+     */
+    @Blocking
+    @Incoming(EXPORT_CHANNEL)
+    @Transactional
```
gwenneg, line 83: This should have no effect with the stateless session.

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsService.java
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * Sends a JSON payload to the export service.
+     * @param xRhExportServicePsk the export service PSK required for the authorization.
+     * @param exportRequestUuid the {@link UUID} of the export request.
+     * @param application the application the export request got requested to.
+     * @param resourceUuid the {@link UUID} of the requested resource.
+     * @param exportContents the payload of the request.
+     */
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Path("/app/export/v1/{exportRequestUuid}/{application}/{resourceUuid}/upload")
```
gwenneg, line 27: How does the export service deal with several apps with the same name but from different bundles? (like `advisor`)

Maybe it's only available for the `rhel` bundle.

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsService.java
```diff
@@ -0,0 +1,91 @@
...
+     * debugging.
+     * @param response the received response from the Export Service.
+     * @return the {@link RuntimeException} to be thrown.
+     *
+    @ClientExceptionMapper
+    static RuntimeException toException(final Response response) {
+        final String errMessage = String.format("The export service responded with a %s status: %s", response.getStatus(), response.readEntity(String.class));
+
+        throw new WebApplicationException(errMessage, response);
+    }*/
```
gwenneg, line 90: Should this be uncommented before the merge?

gwenneg (commented):
I'm not done reviewing but here's a first round of comments/questions.

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsEventListener.java
```diff
@@ -0,0 +1,335 @@
...
+    public CompletionStage<Void> eventListener(final Message<String> message) {
```
gwenneg, line 84: If you don't need to access the headers of the Kafka message, then this signature should be enough:
```suggestion
    public void eventListener(final String payload) {
```

gwenneg: Then you can replace `return message.ack();` with `return;` everywhere.

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsEventListener.java
```diff
@@ -0,0 +1,335 @@
...
+        }
+
+        // Attempt deserializing the received message as a Cloud Event.
+        final ConsoleCloudEvent receivedEvent;
+        try {
+            receivedEvent = this.consoleCloudEventParser.fromJsonString(message.getPayload());
+        } catch (final ConsoleCloudEventParsingException e) {
+            Log.errorf("the received payload from the 'export-requests' topic is not a parseable Cloud Event: %s", e.getMessage());
+
+            this.meterRegistry.counter(EXPORTS_SERVICE_FAILURES_COUNTER).increment();
```
gwenneg, line 97: WDYT of adding a tag to differentiate each failure case in the metrics?

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsService.java
gwenneg: I just understood who `application` is and that's us. Please ignore this comment 😄

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsEventListener.java
```diff
@@ -0,0 +1,335 @@
...
+        final String resource = exportRequest.getResource();
+
+        // Check that we support the requested resource type to export.
+        if (!this.isValidResourceType(resource)) {
+            Log.errorf("[export_request_uuid: %s][resource_uuid: %s] export request could not be fulfilled: the requested resource type '%s' is not handled. Original cloud event: %s", exportRequestUuid, resourceUuid, resource, message.getPayload());
+
+            this.meterRegistry.counter(EXPORTS_SERVICE_FAILURES_COUNTER).increment();
+
+            final ExportError exportError = new ExportError(HttpStatus.SC_BAD_REQUEST, "the specified resource type is unsupported by this application");
+            this.exportsService.notifyErrorExport(this.exportServicePsk, exportRequestUuid, application, resourceUuid, exportError);
```
gwenneg, line 152: We could make it obvious in all calls that `exportsService` that `application` is actually a hardcoded value:
```suggestion
            this.exportsService.notifyErrorExport(this.exportServicePsk, exportRequestUuid, APPLICATION_NAME, resourceUuid, exportError);
```

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsService.java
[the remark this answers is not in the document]
MikelAlejoBR: It seems to be the [export service](https://github.com/RedHatInsights/export-service-go) but I got confused at some point and I was convinced that it was in plural... I've reverted it to the singular version.

engine/src/main/resources/application.properties

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsService.java
MikelAlejoBR: Extracted that `@Path`!

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsEventListener.java
MikelAlejoBR: There are three UUIDs:

- The Cloud Event message's UUID, which represents that unique message itself.
- The subject's UUID, which represents an export request.
- The resource's UUID, which is a specific data request from that export request.

The export request could be seen as the parent of the resource: one export request may require multiple resource requests from multiple applications. So, maybe that export request's UUID can come in the payload too to make it easier to parse? We can ask for that to the export team, I think.

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsEventListener.java
MikelAlejoBR: Removed!

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsEventListener.java
gwenneg: It looks like the Cloud Event UUID and the subject (or request) UUID are two identifiers for the same thing. If so, I'd use the same UUID for both. But that's not a decision to make on our side anyway 😃

Thanks for clarifying the resource UUID, it makes sense to me now.

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsEventListener.java
MikelAlejoBR: We'll simplify to that, for sure. Thanks!

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsEventListener.java
```diff
@@ -0,0 +1,335 @@
...
+
+            return message.ack();
+        }
+
+        // Extract the export request's UUID from the subject.
+        final UUID exportRequestUuid;
+        try {
+            exportRequestUuid = this.extractExportUuidFromSubject(receivedEvent.getSubject());
+        } catch (final IllegalArgumentException | IllegalStateException e) {
+            Log.errorf("unable to extract the export request's UUID from the subject '%s': %s. Original Cloud Event: %s", receivedEvent.getSubject(), e.getMessage(), message.getPayload());
```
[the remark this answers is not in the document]
MikelAlejoBR: This too!

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsEventListener.java
MikelAlejoBR: Oh crap. You said a tag. I've implemented and pushed different counters for the errors :man_facepalming: . Let me fix that...

MikelAlejoBR: I am having an issue with this one. I have pushed the latest code to the PR. Essentially, I am unable to make the proper counting with the Micrometer Assertion Helper... I think I am passing the right tags with the correct values to assert, but I might be doing something wrong here?

It's the latest commit the one that has these changes.

engine/src/test/java/com/redhat/cloud/notifications/exports/ExportEventListenerTest.java
```diff
@@ -0,0 +1,640 @@
...
+    }
+
+    /**
+     * Tests that when an export request is received with an invalid resource
+     * type, then an error is sent to the export service.
+     */
+    @Test
+    void testInvalidResourceTypeRaisesError() {
+        // Save the counter values to assert the "errors count" change later.
+        this.micrometerAssertionHelper.saveCounterValuesBeforeTest(ExportEventListener.EXPORTS_SERVICE_FAILURES_COUNTER);
```
gwenneg, line 108: Let's save and assert both `EXPORTS_SERVICE_FAILURES_COUNTER` and `EXPORTS_SERVICE_SUCCESSES_COUNTER` in all tests to make sure that they are only incremented when they should be.

engine/src/test/java/com/redhat/cloud/notifications/exports/ExportEventListenerTest.java
```diff
@@ -0,0 +1,640 @@
...
+    void setUpMockServerRoutes() {
+        final ClientAndServer mockServer = MockServerLifecycleManager.getClient();
+
+        mockServer
+            .when(request().withPath(".*/error"))
+            .respond(response().withStatusCode(200));
+
+        mockServer
+            .when(request().withPath(".*/upload"))
+            .respond(response().withStatusCode(200));
```
gwenneg, line 90: Testing with MockServer works, but it requires more work than tests based on `@InjectMock @RestClient` and Mockito:
https://github.com/RedHatInsights/notifications-backend/blob/35a969b12e921e38260427f35d8f749a6d5258d3/backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java#L2078-L2085

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsEventListener.java
```diff
@@ -0,0 +1,335 @@
...
+                    final ExportError exportError = new ExportError(HttpStatus.SC_INTERNAL_SERVER_ERROR, "unable to serialize message in the correct format");
+                    this.exportsService.notifyErrorExport(this.exportServicePsk, exportRequestUuid, application, resourceUuid, exportError);
+                }
+            });
+        }
+
+        this.meterRegistry.counter(EXPORTS_SERVICE_SUCCESSES_COUNTER).increment();
+
+        return message.ack();
+    }
```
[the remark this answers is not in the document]
MikelAlejoBR: I'm going to reopen this one too, because I am thinking on moving the transformation logic to yet another class too, and maybe wrap those two in a service:

- `ExportEventsService#extractFiltersAndTransform` which would make use of the filters extractor and the transformer.

That way the handler would only call that and deal with potential exceptions, and then send the proper response to the service.

.rhcicd/clowdapp-engine.yaml
```diff
@@ -319,7 +321,7 @@ parameters:
   description: When this is true, all integration types except emails are disabled
   value: "true"
 - name: NOTIFICATIONS_LOG_LEVEL
-  description: Log level for com.redhat.cloud.notifications
+  description: Log level for com.redhat.cloud.notificationsq
```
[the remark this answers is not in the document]
MikelAlejoBR: Done, nice catch!

engine/src/main/resources/application.properties
```diff
@@ -176,3 +187,15 @@ quarkus.rest-client.internal-google-chat.trust-store-type=${clowder.endpoints.no
...
 %test.quarkus.mailer.mock=false
+
+# Rest client settings for the export service. When used locally, it should
+# point to the private internal server.
+# Export service's development PSK value. Specified here to avoid Quarkus from
+# complaining that the configuration parameter is missing. In the case that you
+# are using a real Export Service application to test the integration, you will
+# need to set this PSK in the PSKS environment variable on that end.
+export-service.psk=development-value-123
+quarkus.rest-client.export-service.url=${clowder.endpoints.export-service.url:http://localhost:10010}
```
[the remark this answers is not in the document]
MikelAlejoBR: I have just checked AppInterface and I've seen that they've got an stage deployment ready. I've added the service as an optional dependency, as suggested.

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportEventListener.java
```diff
@@ -0,0 +1,319 @@
...
+            Log.debugf("[export_request_uuid: %s][resource_uuid: %s] bad \"from\" date format. Notifying the export service with a 400 error. Received date: %s", exportRequestUuid, resourceUuid, filters.get(FILTER_DATE_FROM));
+
+            this.meterRegistry.counter(EXPORTS_SERVICE_FAILURES_COUNTER).increment();
+
+            final ExportError exportError = new ExportError(HttpStatus.SC_BAD_REQUEST, "unable to parse the 'from' date filter with the 'yyyy-mm-dd' format");
+            this.exportService.notifyErrorExport(this.exportServicePsk, exportRequestUuid, APPLICATION_NAME, resourceUuid, exportError);
+
+            return;
+        } catch (final IllegalStateException e) {
+            Log.debugf("[export_request_uuid: %][resource_uuid: %s] well formatted but invalid \"from\" date received: %s. Error cause: %s", exportRequestUuid, resourceUuid, filters.get(FILTER_DATE_FROM), e.getMessage());
```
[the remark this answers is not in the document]
MikelAlejoBR: Done as well!

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportEventListener.java
```diff
@@ -0,0 +1,319 @@
...
+            this.meterRegistry.counter(EXPORTS_SERVICE_FAILURES_COUNTER).increment();
+
+            final ExportError exportError = new ExportError(HttpStatus.SC_BAD_REQUEST, "the 'to' date cannot be lower than the 'from' date");
+            this.exportService.notifyErrorExport(this.exportServicePsk, exportRequestUuid, APPLICATION_NAME, resourceUuid, exportError);
+
+            return;
+        }
+
+        // Handle exporting the requested resource type.
+        if (RESOURCE_TYPE_EVENTS.equals(resource)) {
```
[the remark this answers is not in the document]
MikelAlejoBR: True! I am refactoring this bit so I will take this into account to not have it duplicated.

engine/src/test/java/com/redhat/cloud/notifications/exports/ExportEventListenerTest.java
MikelAlejoBR: Done as well!

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportEventListener.java
```diff
@@ -0,0 +1,319 @@
...
+                            resultsTransformer = new JSONEventTransformer();
+                            contents = resultsTransformer.transform(events);
+                            this.exportService.uploadJSONExport(this.exportServicePsk, exportRequestUuid, APPLICATION_NAME, resourceUuid, contents);
+                        }
+                        default -> {
+                            final ExportError exportError = new ExportError(
+                                HttpStatus.SC_BAD_REQUEST,
+                                String.format("the specified format '%s' is unsupported for the request", format)
+                            );
+                            this.exportService.notifyErrorExport(this.exportServicePsk, exportRequestUuid, APPLICATION_NAME, resourceUuid, exportError);
```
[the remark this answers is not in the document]
MikelAlejoBR: Right, I moved things around a bit to a service, so now it should be easier to track these ones. I think I've got them all!

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsEventListener.java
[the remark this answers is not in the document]
MikelAlejoBR: Done! I moved the filter extraction bit to a class, and then created a service which calls the different transformators. That should make the handler easier to follow.

engine/src/test/java/com/redhat/cloud/notifications/exports/ExportEventListenerTest.java
gwenneg: Do you plan on keeping the MockServer approach here? I'm not requesting a change, I only want to make sure this class is in its final state before reviewing it in details.

MikelAlejoBR:
> This method needs a try {} catch (Exception e} { // Log the exception } block that includes everything to protect the app against Kafka poison pills.

Good one, because I was also wondering how we would catch the `WebApplicationExceptions` if the export service is sending non 200 responses.

> Ping :smile:

Yeah! I am waiting because I'll probably take care of that one last, since it's going to break a few tests :sweat: 

> Do you plan on keeping the MockServer approach here? I'm not requesting a change, I only want to make sure this class is in its final state before reviewing it in details.

No! It is true that it adds complexity and I'll revert to using mocks.

---

I was trying to bump the schemas' version because we've updated the `export` service's CE to make it easier and clearer to get the different UUIDs, instead of having to parse the subject. The thing is that I've realized that bumping that one makes tests fail due to a parsing error in the timestamps.

I will skip this until we sort it out —we're already working on it—, and I'll start with replacing the Mockserver and the other feedback :smile:

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportEventListener.java
```diff
@@ -0,0 +1,233 @@
...
+                final ExportError exportError = new ExportError(
+                    HttpStatus.SC_BAD_REQUEST,
+                    String.format("the specified format '%s' is unsupported for the request", format)
+                );
+
+                this.exportService.notifyErrorExport(this.exportServicePsk, exportRequestUuid, APPLICATION_NAME, resourceUuid, exportError);
+            }
+        }
+
+        this.meterRegistry.counter(EXPORTS_SERVICE_SUCCESSES_COUNTER).increment();
```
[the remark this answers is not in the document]
MikelAlejoBR: Done!

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportEventListener.java
```diff
@@ -0,0 +1,233 @@
...
+            case JSON -> this.exportService.uploadJSONExport(this.exportServicePsk, exportRequestUuid, APPLICATION_NAME, resourceUuid, exportedContents);
+            default -> {
+                Log.debugf("[export_request_uuid: %s][resource_uuid: %s][requested_format: %s] unsupported format", exportRequestUuid, resourceUuid, format);
+
+                final ExportError exportError = new ExportError(
+                    HttpStatus.SC_BAD_REQUEST,
+                    String.format("the specified format '%s' is unsupported for the request", format)
+                );
+
+                this.exportService.notifyErrorExport(this.exportServicePsk, exportRequestUuid, APPLICATION_NAME, resourceUuid, exportError);
```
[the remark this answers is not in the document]
MikelAlejoBR: Done as well!

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportEventListener.java
```diff
@@ -0,0 +1,233 @@
...
+    /**
+     * Listens to the exports channel, extracts the request's data, and if
+     * Notifications is the target application, and the requested resource and
+     * formats are supported, the corresponding payload is sent to the export
+     * service.
+     * @param payload the incoming payload from the channel.
+     */
+    @Blocking
+    @Incoming(EXPORT_CHANNEL)
+    public void eventListener(final String payload) {
```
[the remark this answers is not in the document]
MikelAlejoBR: Done!

engine/src/test/java/com/redhat/cloud/notifications/exports/ExportEventListenerTest.java
MikelAlejoBR: Yes, I'm going to start working on this one now

gwenneg:
⚠️ There's one thing we'll need to be careful about once the export service is enabled: it could lead to `engine` restarts if we rely on the default Kafka settings.

Kafka expects a consumer to poll a batch of messages every once in a while. There's a max poll interval in the settings that defines the maximum time a consumer can take between two polls. If the consumer fails to poll before that max delay is exceeded, then the consumer is kicked out of the consumer group. The poll interval is configurable. If our pod is kicked out of the Kafka consumer group, its health check will be marked as DOWN and the pod will be restarted in OpenShift.

This could happen with the export service integration because of the amount of data we'll need to upload via the REST calls. Lots of data will mean long running REST requests and a throughput that might be too low to satisfy the default max poll interval.

There are two ways to prevent that from happening with the configuration:
- by increasing the max poll delay
- and/or by lowering the number of messages polled from Kafka in each batch

engine/src/test/java/com/redhat/cloud/notifications/exports/ExportEventListenerTest.java
MikelAlejoBR: Done! I've left the Mock Server just to test that the "Content-Type" headers are correct for each method.

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportsService.java
MikelAlejoBR: Yes! There's a couple of things I need to finish in this PR:

1. This.
2. Update the schema.

MikelAlejoBR:
And is it possible to use a different consumer group for this integration?

gwenneg:
> And is it possible to use a different consumer group for this integration?

Possible, yes. But it wouldn't help with the poll delay topic. We would simply be kicked out from a different consumer group and the pods would still restart.

gwenneg:
Here's how the Kafka consumer config can be changed to prevent the issue: https://github.com/RedHatInsights/notifications-backend/blob/132e9fb8bdd5dfe717c79f1e32fc90dfac326992/.rhcicd/clowdapp-engine.yaml#L99-L102

gwenneg:
@MikelAlejoBR Could you please rebase the PR?

MikelAlejoBR:
Sure thing! Done!

gwenneg (approved):
LGTM

engine/src/main/java/com/redhat/cloud/notifications/exports/ExportEventListener.java
```diff
@@ -0,0 +1,225 @@
...
+                final int statusCode = wae.getResponse().getStatus();
+                final Response.Status.Family responseFamily = Response.Status.Family.familyOf(statusCode);
+
+                if (responseFamily == Response.Status.Family.CLIENT_ERROR) {
+                    this.failuresCounter.increment();
+                }
+
+                if (responseFamily == Response.Status.Family.SERVER_ERROR) {
+                    this.failuresCounter.increment();
+                }
```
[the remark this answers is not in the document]
MikelAlejoBR: I was preparing for the "tags metrics" change. The last step was to add tags to the counters to make it easier to differentiate which errors we're suffering from. However I was having some issues and we thought of postponing it to the last step.

Were you thinking on doing it in a separate PR?

engine/src/test/java/com/redhat/cloud/notifications/exports/ExportEventListenerTest.java
```diff
@@ -0,0 +1,436 @@
...
+@QuarkusTestResource(TestLifecycleManager.class)
+public class ExportEventListenerTest {
+    @ConfigProperty(name = "export-service.psk")
+    String exportServicePsk;
+
+    @InjectMock
+    EventRepository eventRepository;
+
+    @Inject
+    ExportEventListener exportEventListener;
```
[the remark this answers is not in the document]
MikelAlejoBR: Done!

gwenneg (approved):

PR #1921, opened 2023-05-17
[RHCLOUD-25667] Disallow loopback addresses in stage and production

g-duval:


common/src/main/java/com/redhat/cloud/notifications/models/validation/ValidNonPrivateUrlValidator.java
```diff
@@ -79,6 +82,12 @@ public boolean isValid(final String rawUrl, ConstraintValidatorContext constrain
             return false;
         }
 
+        if (ProfileManager.getLaunchMode() == NORMAL && address.isLoopbackAddress()) {
+            this.replaceDefaultMessage(constraintValidatorContext, LOOPBACK_ADDRESS);
+
+            return false;
+        }
+
```
MikelAlejoBR, line 90: Very interesting! When I thought about this I was wondering how would I be able to tell that the environment was not a `dev` one. TIL!

MikelAlejoBR (approved):
Nice one! LGTM!

PR #1922, opened 2023-05-18
[RHCLOUD-24734] fix: avoid exceptions blocking the migrations

MikelAlejoBR:
During the migration, we saw that constraint violation exceptions on endpoints with invalid URLs were rolling back the whole operation. The goal of this change is to handle those exceptions in order to be able to continue with the process even when a failure happens, and in order to handle the deletion of any created secrets when the error occurred.

## Links

[[RHCLOUD-24734]](https://issues.redhat.com/browse/RHCLOUD-24734)

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/internal/SourcesSecretsMigrationService.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/internal/SourcesSecretsMigrationService.java
@@ -65,6 +65,20 @@ public class SourcesSecretsMigrationService {
             } catch (final Exception e) {
                 if (this.extractConstraintViolationException(e) instanceof ConstraintViolationException cve) {
                     Log.errorf("[endpoint_id: %s] error when migrating the endpoint secrets: a constraint violation exception was raised: %s. Constraints list: %s", eligibleEndpoint.getKey(), cve.getMessage(), cve.getConstraintViolations());
+
+                    // Delete the secrets for the processed endpoint, since an
+                    // error occurred.
+                    final Endpoint endpoint = this.processedEndpoints.get(eligibleEndpoint.getKey());
+                    final EndpointProperties endpointProperties = endpoint.getProperties();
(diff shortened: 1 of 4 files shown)


backend/src/main/java/com/redhat/cloud/notifications/routers/internal/SourcesSecretsMigrationService.java
```diff
@@ -40,34 +52,105 @@ public class SourcesSecretsMigrationService {
...
+                    continue;
+                } else if (e instanceof WebApplicationException wae) {
+                    Log.errorf("[endpoint_id: %s] error when migrating the endpoint secrets: unable to create the secrets for the endpoint: %s", eligibleEndpoint.getKey(), wae.getResponse().getEntity());
+                } else {
+                    Log.errorf("[endpoint_id: %s] error when migrating the endpoint secrets: unable to create the secrets for the endpoint: %s", eligibleEndpoint.getKey(), e.getMessage());
+                }
+
+                // Delete the secrets for the processed endpoint, since an
+                // error occurred.
+                final Endpoint endpoint = this.processedEndpoints.get(eligibleEndpoint.getKey());
```
gwenneg, line 81: Will this ever return an endpoint? Line 123 (the `put` call) will only be executed if no exception is thrown when `migrateEndpoint` is run.

MikelAlejoBR: Mmmmm... You are right, I think the only time this will ever return an endpoint is if we hit a `ConstraintViolationException`, since the `put` call will be executed and then, when attempting to save the endpoint, the exception will be thrown... So I guess I can move the code up there, since if there is a `NotFoundException` then the secrets won't be created, and if we have a `WebApplicationException`, that simply means that Sources didn't like the payload and didn't create the secrets...

MikelAlejoBR:
I think this one is ready to be re-reviewed :smiley:

gwenneg (approved):
LGTM

PR #1926, opened 2023-05-22
Remove camel-sender-demo module

gwenneg:
We don't need this module anymore. The [eventing-integrations](https://github.com/RedHatInsights/eventing-integrations) repo can be used as an example, until it is merged into this repo.


g-duval (approved):
Nice, I was about to ask if we still need this project :)

PR #1927, opened 2023-05-22
[RHCLOUD-25998] Drawer notification DB table

g-duval:


database/src/main/resources/db/migration/V1.77.0__RHCLOUD-25998_add_drawer_notification_table.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE public.drawer_notification (
+    user_id varchar(50) NOT NULL,
+    org_id text NOT NULL,
+    event_id uuid NOT NULL,
+    created timestamp NOT NULL,
+    read boolean NOT NULL default false,
```
gwenneg, line 6: Sorry, I forgot something in the Jira ticket 😅 Since this field will be updated from the REST API, I think the `drawer_notification` table needs an `id uuid` column which would be the PK.

database/src/main/resources/db/migration/V1.77.0__RHCLOUD-25998_add_drawer_notification_table.sql
```diff
@@ -0,0 +1,11 @@
...
+    CONSTRAINT pk_drawer_notification PRIMARY KEY (user_id, event_id)
```
gwenneg, line 7: This can be changed to a `unique` constraint. I would also include `org_id` in that constraint because all queries will include an `org_id` and Postgres will perform faster if that field is part of the `unique` constraint.

database/src/main/resources/db/migration/V1.77.0__RHCLOUD-25998_add_drawer_notification_table.sql
g-duval: Ok

database/src/main/resources/db/migration/V1.77.0__RHCLOUD-25998_add_drawer_notification_table.sql
g-duval: Ok

engine/src/main/java/com/redhat/cloud/notifications/models/DrawerNotification.java
```diff
@@ -0,0 +1,119 @@
...
+import java.sql.Timestamp;
+import java.time.LocalDateTime;
+import java.util.UUID;
+
+import static com.fasterxml.jackson.annotation.JsonProperty.Access.READ_ONLY;
+import static java.time.ZoneOffset.UTC;
+
+@Entity
+@Table(name = "drawer_notification")
+public class DrawerNotification {
```
gwenneg, line 24:
```suggestion
public class DrawerNotification extends CreationTimestamped {
```

engine/src/main/java/com/redhat/cloud/notifications/models/DrawerNotification.java
```diff
@@ -0,0 +1,119 @@
...
+        this.event = event;
+    }
+
+    public LocalDateTime getCreated() {
+        return created.toLocalDateTime();
+    }
+
+    public void setCreated(LocalDateTime created) {
+        this.created = Timestamp.valueOf(created);
+    }
```
gwenneg, line 94:
```suggestion
```

engine/src/main/java/com/redhat/cloud/notifications/models/DrawerNotification.java
```diff
@@ -0,0 +1,119 @@
...
+        this.eventId = eventId;
+    }
+
+    @PrePersist
+    public void prePersist() {
+        // The 'created' field value can be set in tests.
+        if (created == null) {
+            created = Timestamp.valueOf(LocalDateTime.now(UTC));
+        }
+    }
```
gwenneg, line 118:
```suggestion
```

engine/src/main/java/com/redhat/cloud/notifications/models/DrawerNotification.java
gwenneg, line 24:
```suggestion
@JsonNaming(SnakeCaseStrategy.class)
public class DrawerNotification {
```

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
```diff
@@ -0,0 +1,26 @@
...
+        statelessSessionFactory.getCurrentSession().insert(drawerNotification);
+        return drawerNotification;
+    }
+
+    public List<DrawerNotification> getDrawerNotificationsByUserId(String userId) {
+        String query = "SELECT dn FROM DrawerNotification dn WHERE userId = :userId";
+        return statelessSessionFactory.getCurrentSession().createQuery(query, DrawerNotification.class)
+            .setParameter("userId", userId)
+            .getResultList();
+    }
```
gwenneg, line 25: The drawer notifications will be retrieved from the `backend` REST API which means we'll use a different repository in the `backend` module for that. You can move this method to the test class to keep testing the DB records creation.
```suggestion
```

gwenneg: Note for later: we'll always include the `orgId` in the queries to retrieve drawer notifications because two orgs may contain the same `userId`.

MikelAlejoBR (approved):
LGTM!

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
g-duval: Ho I see, I wasn't aware that was possible !
Do you know how a user can longin is such case ? Because on console dot login page the org id is not required ?

engine/src/main/java/com/redhat/cloud/notifications/models/DrawerNotification.java
```diff
@@ -0,0 +1,119 @@
+package com.redhat.cloud.notifications.models;
```
josejulio, line 1: Should this live in the common module?

engine/src/main/java/com/redhat/cloud/notifications/models/DrawerNotification.java
```diff
@@ -0,0 +1,119 @@
...
+    @NotNull
+    @Transient
+    private UUID eventId;
+
+    @ManyToOne(fetch = FetchType.LAZY, optional = false)
+    @JoinColumn(name = "event_id")
+    private Event event;
+
+    @NotNull
+    private boolean read;
```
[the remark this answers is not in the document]
gwenneg: The exact endpoint paths have yet to be determined, but the `read` field will probably be updated by calling something like:
```
PUT /drawer/{notificationId}/read
```
and
```
PUT /drawer/{notificationId}/unread
```
We won't pass the whole `DrawerNotification` object in the request, so I think a primitive type is fine for the `read` field.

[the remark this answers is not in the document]
MikelAlejoBR: Aaaah perfect! Thank you both for the clarification!

PR #1930, opened 2023-05-24
[RHCLOUD-25884] subscription to drawer

g-duval:


backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EmailSubscriptionRepository.java
```diff
@@ -87,10 +95,20 @@ protected int replicateUnsubscribeToEventTypeLevel(String orgId, String username
             .executeUpdate();
     }
 
-    @Transactional
     public int subscribeEventType(String orgId, String username, UUID eventTypeId, EmailSubscriptionType subscriptionType) {
-        String query = "INSERT INTO email_subscriptions(org_id, user_id, event_type_id, subscription_type) " +
-            "VALUES (:orgId, :userId, :eventTypeId, :subscriptionType ) " +
+        // Because Drawer subscription are opt-out by default
```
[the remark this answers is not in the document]
g-duval: Thanks, good point !

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EmailSubscriptionRepository.java
```diff
@@ -87,10 +95,20 @@ protected int replicateUnsubscribeToEventTypeLevel(String orgId, String username
             .executeUpdate();
     }
 
-    @Transactional
     public int subscribeEventType(String orgId, String username, UUID eventTypeId, EmailSubscriptionType subscriptionType) {
-        String query = "INSERT INTO email_subscriptions(org_id, user_id, event_type_id, subscription_type) " +
-            "VALUES (:orgId, :userId, :eventTypeId, :subscriptionType ) " +
+        // Opt-in: only subscriptions are stored on database
+        // Opt-on: only un-subscriptions are stored on database
```
gwenneg, line 100: I'm not sure about this so let's talk about it 😃

Storing everything in the DB wouldn't really slow down the queries if we have a good indexing strategy, so let's ignore the impact on performances.

From a maintenance perspective, I feel like it would be easier to understand the data/code, make changes or query the DB through Gabi if the table contained explicit data rather than having to compare the DB data with the code logic to figure out what it means.

g-duval: Thank you for your comment.
I understand the drawback by introducing a different logic on this table for support perspective.
Nevertheless, from my perspective, this extra effort will come anyway by adding a different subscription logic on notification.

In favor on this solution, it is able to handle a new bundle or application or event type without doing anything at subscription level, like it is already the case on email subscription.
Additionnaly, adding the subscribed boolean column on this table should help the support guy to remember the logic right ?

But you are right, lets talk about it with the team :)

josejulio: I think that @gwenneg said makes sense to me. We still have to query the database to verify if the value is there or not.
One thing that we need to add is some mechanism to set the default value when we're querying and the value is not already in the database (to have more consistency).

Talking about storage, I don't think storing would increase the values a lot.

Just to confirm, I'm not against this new logic (opt-in vs opt-out) being at this level. IIUC the concern is about storing or not storing the default value.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EmailSubscriptionRepository.java
```diff
@@ -70,6 +78,7 @@ protected int replicateSubscribeToEventTypeLevel(String orgId, String username,
             .setParameter("applicationName", applicationName)
             .setParameter("bundleName", bundleName)
             .setParameter("subscriptionType", subscriptionType.name())
+            .setParameter("subscribed", true) // only email subscription replication is supported
```
gwenneg, line 81: I think we should wrap the following code with an `if (emailSubscriptionType != DRAWER) { /* old code */ }`:

https://github.com/RedHatInsights/notifications-backend/blob/dd9ccdaf7ccd487e7bc30087f44ffcfe5d2e5d20/backend/src/main/java/com/redhat/cloud/notifications/routers/UserConfigResource.java#L104-L114

backend/src/test/java/com/redhat/cloud/notifications/routers/UserConfigResourceTest.java
```diff
@@ -483,16 +515,30 @@ void testSettingsByEventType() {
...
+
+            // Daily to true
+            updateAndCheckUserPreference(path, identityHeader, bundle, application, eventType, true, false, false);
+
+            // Instant to true
+            updateAndCheckUserPreference(path, identityHeader, bundle, application, eventType, false, true, false);
+
+            // Both to true
+            updateAndCheckUserPreference(path, identityHeader, bundle, application, eventType, true, true, false);
+        }
```
gwenneg, line 541: Why is the last arg `false` when the drawer is enabled and `true` in the first 4 `updateAndCheck...`? Shouldn't it be the opposite?

g-duval: Actually a part of this test was missing.

gwenneg (approved):
LGTM

PR #1931, opened 2023-05-24
RHCLOUD-26020 Enable Kafka topic name translation using Clowder config

gwenneg:
The topic name from `application.properties` is not directly used in stage or prod. It is used as a key to retrieve the actual topic name from the Clowder configuration. This only works if the config key starts with `mp.messaging` and ends with `topic`.

See [this](https://github.com/RedHatInsights/clowder-quarkus-config-source/blob/56502b08308acec1248d4feef4b5bbc5ae335625/src/main/java/com/redhat/cloud/common/clowder/configsource/ClowderConfigSource.java#L176-L189) for more details.

### DIFF
--- a/engine/src/main/resources/application.properties
+++ b/engine/src/main/resources/application.properties
@@ -180,7 +180,7 @@ quarkus.rest-client.internal-google-chat.trust-store-type=${clowder.endpoints.no
 %test.quarkus.mailer.mock=false
 
 # This will move out of the engine after the split
-notifications.camel.kafka-return-topic=platform.notifications.fromcamel
+mp.messaging.fromcamel.topic=platform.notifications.fromcamel
 
 # camel-quarkus-kafka configuration
 camel.component.kafka.brokers=localhost:9092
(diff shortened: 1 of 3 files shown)


engine/src/main/resources/application.properties
```diff
@@ -180,7 +180,7 @@ quarkus.rest-client.internal-google-chat.trust-store-type=${clowder.endpoints.no
 %test.quarkus.mailer.mock=false
 
 # This will move out of the engine after the split
-notifications.camel.kafka-return-topic=platform.notifications.fromcamel
+mp.messaging.fromcamel.topic=platform.notifications.fromcamel
```
josejulio, line 183: Why this one doesn't have the `incoming` like `mp.messaging.incoming.ingress.topic` ?

gwenneg: This config key is a "hack" of the config source lib. The prefix has to be `mp.messaging` and the suffix has to be `topic`, but if `incoming` or `outgoing` is present in the middle of the key, Quarkus complains because other config keys should be provided and the app doesn't start. If `incoming` and `outgoing` are missing, the config key is ignored by Quarkus and the app can start.

gwenneg: That "hack" is currently used in `eventing-integrations`. It's not pretty but transforming it to something cleaner isn't my immediate priority.

josejulio (approved):
lgtm

PR #1932, opened 2023-05-24
RHCLOUD-26020 Persist the notification history before calling Camel

gwenneg:
The notification history update fails in stage because the history record hasn't been created yet when the engine receives the return message. The reason behind that is pretty simple: we're calling Camel with a blocking call (REST) and that call includes the response that is sent through Kafka. The history is created after the Kafka response has been sent and processed, so it can't be updated.

This PR moves the history record creation before the call to Camel and guarantees that the return message will update the history.


MikelAlejoBR (approved):
LGTM!

PR #1934, opened 2023-05-25
RHCLOUD-26040 Remove Kafka headers from EventingProcessor

gwenneg:
These headers are not used on the consumer side of the Kafka topic.


josejulio (approved):
lgtm

PR #1952, opened 2023-05-30
RHCLOUD-26040 Introduce ConnectorSender

gwenneg:
This PR replaces REST with Kafka when `engine` dispatches a notification to Slack/Teams/GChat.


engine/src/test/java/com/redhat/cloud/notifications/processors/camel/google/chat/GoogleChatTestProfile.java
```diff
@@ -0,0 +1,6 @@
+package com.redhat.cloud.notifications.processors.camel.google.chat;
+
+import io.quarkus.test.junit.QuarkusTestProfile;
+
+public class GoogleChatTestProfile implements QuarkusTestProfile {
```
g-duval, line 5: Why do we need a specific test profile for each integration ?
As far I remember, we had an issue on unit tests because we were using identical mock urls for each tests.
If it's the only reason why a test profile is required, I think we should use a dedicated target url like we had before to save running test time ?

gwenneg: I added back the test profiles because the tests no longer worked after mocking the source Kafka endpoint. This is not related to the mock URL change. The same URL can be used for everything as long as `getClient().reset()` is called between each test.

gwenneg: The test profiles will be removed very soon when each connector will be moved to its own Maven module.

g-duval: Ok then if it's temporary

g-duval: Ok then if it's temporary

engine/src/main/java/com/redhat/cloud/notifications/processors/ConnectorSender.java
```diff
@@ -0,0 +1,51 @@
...
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+import java.util.UUID;
+
+@ApplicationScoped
+public class ConnectorSender {
+
+    public static final String TOCAMEL_CHANNEL = "tocamel";
+    // TODO notification should end with a s but eventing-integrations does not expect it...
+    public static final String CLOUD_EVENT_TYPE_PREFIX = "com.redhat.console.notification.toCamel.";
```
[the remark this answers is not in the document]
g-duval: Talked with Gwenneg, fine for now, to be fixed later.

g-duval (approved):

PR #1953, opened 2023-05-30
[RHCLOUD-26000] add Drawer minimal template

g-duval:


MikelAlejoBR (approved):
LGTM!

g-duval:
My idea is to use event type display name as default title value.
Then allow applications to customise it later thanks to RHCLOUD-26001.
Does it seems fair to you @gwenneg ?

gwenneg:
Sounds good to me. The event type display name should be enough even in the long run. We'll need to make sure it makes sense for a normal user because until now only org admins could see it.

engine/src/test/java/com/redhat/cloud/notifications/templates/TestDrawerDefaultTemplate.java
```diff
@@ -0,0 +1,57 @@
...
+        statelessSessionFactory.withSession(statelessSession -> {
+            String result = getDefaultDrawerTemplate()
+                .data("data", dataAsMap)
+                .render();
+            assertEquals("FooMachine triggered 2 events", result);
+        });
+    }
+
+    private TemplateInstance getDefaultDrawerTemplate() {
+        IntegrationTemplate integrationTemplate = templateRepository.findIntegrationTemplate(null, null, ORG, "drawer")
```
gwenneg, line 52: The `ORG` param is so confusing for a test focused on a `DEFAULT` template and yet it works... 😅 The whole logic around integration template feels counterintuitive to me, I hope we'll make it simpler someday. This doesn't have anything to do with your PR 😄

gwenneg (approved):
LGTM

PR #1957, opened 2023-05-31
Trigger fresh deployment in stage

gwenneg:


engine/src/main/resources/application.properties
```diff
@@ -4,7 +4,6 @@
 quarkus.kafka.devservices.port=9092
 
 quarkus.http.port=8087
-
```
josejulio, line 7: Empty lines be like
image

PR #1965, opened 2023-05-31
RHCLOUD-22207 | fix sonarqube pipeline

MikelAlejoBR:
The scanner was complaining that the ".java" files were not compiled,
and that should be provided via a property. It was working locally
because I always have the compiled project there, so by copying all the
contents of the project's directory it also copied the compiled classes.
However, in a pipeline, where everything is fresh, that didn't happen,
which I suppose is causing the scanner to complain.

## Links
[[RHCLOUD-22207]](https://issues.redhat.com/browse/RHCLOUD-22207)


gwenneg (approved):
LGTM

PR #1966, opened 2023-05-31
Revert changes introduced to investigate the Kafka outage

gwenneg:
https://github.com/RedHatInsights/notifications-backend/compare/132e9fb8bdd5dfe717c79f1e32fc90dfac326992...dc3eb724b4e37e485c1346fad991b8ce610f5254

This PR does NOT revert:
- the new ClowdApp params used to change the log level of Hibernate, Kafka client and SmallRye Reactive Messaging
- the removal of `@Acknowledgment(PRE_PROCESSING)` in `EventConsumer`
- the addition of `cloud-events=false` in the Reactive Messaging channels configuration

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
@@ -18,7 +18,6 @@ import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Timer;
 import io.quarkus.logging.Log;
 import io.smallrye.reactive.messaging.annotations.Blocking;
-import org.eclipse.microprofile.faulttolerance.Timeout;
 import org.eclipse.microprofile.reactive.messaging.Incoming;
 import org.eclipse.microprofile.reactive.messaging.Message;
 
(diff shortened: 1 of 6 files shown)


engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
```diff
@@ -226,12 +219,11 @@ public void process(String payload) {
                     TAG_KEY_APPLICATION, tags.getOrDefault(TAG_KEY_APPLICATION, ""),
                     TAG_KEY_EVENT_TYPE_FQN, tags.getOrDefault(TAG_KEY_EVENT_TYPE_FQN, "")
             ));
-            Log.debug("Processing Kafka message - end");
         }
+        return message.ack();
```
josejulio, line 223: Won't this lead to exceptions/disconnects if some endpoints take long to process?

engine/src/main/resources/application.properties
```diff
@@ -4,6 +4,7 @@
 quarkus.kafka.devservices.port=9092
 
 quarkus.http.port=8087
+
```
josejulio, line 7: welcome back!

MikelAlejoBR (approved):

engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
gwenneg: If it does, we will simply put the annotation back.

josejulio: I recall we had issues without the annotation `@Acknowledgment(PRE_PROCESSING)` - we should be extra careful with QA tests that might have tons of endpoints - I suppose that once we do the split it will be less trouble in that regard.

josejulio (approved):

PR #1971, opened 2023-06-02
[RHCLOUD-25883] add drawer processor step1

g-duval:


engine/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
```diff
@@ -16,4 +22,19 @@ public DrawerNotification create(DrawerNotification drawerNotification) {
         return drawerNotification;
     }
 
+    public List<DrawerNotification> create(Event event, String users) {
```
gwenneg, line 25: Do we still need the other `create` method?

database/src/main/resources/db/migration/V1.80.0__RHCLOUD-25883_add_drawer_notification_on_event_table.sql
```diff
@@ -0,0 +1,20 @@
+ALTER TABLE event add column rendered_drawer_notification TEXT;
+
+CREATE OR REPLACE FUNCTION insert_drawer_notifications(pUserIdList text, pOrgId text, pEventId uuid, pCreated timestamp)
+    RETURNS SETOF drawer_notification AS
+$BODY$
+DECLARE
+    createdNotifications REFCURSOR;
```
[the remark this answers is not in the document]
g-duval: That's right it's no longer necessary

engine/src/main/java/com/redhat/cloud/notifications/processors/drawer/DrawerProcessor.java
```diff
@@ -0,0 +1,138 @@
...
+            Log.infof(e, "Processing drawer notification failed [eventId=%s, historyId=%s]", event.getId(), historyId);
+        } finally {
+            long invocationTime = System.currentTimeMillis() - startTime;
+            history.setInvocationTime(invocationTime);
+            persistNotificationHistory(history);
+        }
+    }
+
+    public String buildNotificationMessage(Event event) {
+        JsonObject data = baseTransformer.toJsonObject(event);
```
gwenneg, line 115: We'll probably need to add the `Environment#url` to `data`. We can take care of that later.

engine/src/main/java/com/redhat/cloud/notifications/processors/drawer/DrawerProcessor.java
```diff
@@ -0,0 +1,138 @@
...
+            history = getHistoryStub(endpoint, event, 0L, historyId);
+            history.setStatus(NotificationStatus.SUCCESS);
+        } catch (Exception e) {
+            history = getHistoryStub(endpoint, event, 0L, historyId);
+            history.setStatus(NotificationStatus.FAILED_INTERNAL);
+            history.setDetails(Map.of("failure", e.getMessage()));
+            Log.infof(e, "Processing drawer notification failed [eventId=%s, historyId=%s]", event.getId(), historyId);
+        } finally {
+            long invocationTime = System.currentTimeMillis() - startTime;
+            history.setInvocationTime(invocationTime);
```
gwenneg, line 109: I'm not sure the invocation time makes sense for the Drawer since we will only push a message to Kafka which is always instant.

engine/src/main/java/com/redhat/cloud/notifications/processors/drawer/DrawerProcessor.java
```diff
@@ -0,0 +1,138 @@
...
+        try {
+            String userNameListAsStr = userList.stream().map(usr -> usr.getUsername()).collect(Collectors.joining(","));
+            List<DrawerNotification> drawerNotifications = drawerNotificationRepository.create(event, userNameListAsStr);
+
+            // build event thought qute template
+            String renderedData = buildNotificationMessage(event);
+            // store it on event table
+            event.setRenderedDrawerNotification(renderedData);
+            eventRepository.updateDrawerNotification(event);
+            endpoint = endpointRepository.getOrCreateDefaultSystemSubscription(event.getAccountId(), event.getOrgId(), EndpointType.DRAWER);
```
gwenneg, line 96: Why is a second `getOrCreateDefaultSystemSubscription` needed?

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EventRepository.java
```diff
@@ -85,4 +85,8 @@ public List<Event> findEventsToExport(final String orgId, final LocalDate from,
 
         return findEventsRanged.getResultList();
     }
+
+    public void updateDrawerNotification(Event event) {
+        statelessSessionFactory.getCurrentSession().update(event);
```
gwenneg, line 90: Not mandatory: I prefer when code performing DB operations is obvious and can be understood immediately, without having to read several methods to understand what will be modified in the DB. It can make the maintenance easier. Feel free to disagree 😄 
```suggestion
        String hql = "UPDATE Event SET renderedDrawerNotification = :renderedDrawerNotification WHERE id = :id";
        statelessSessionFactory.getCurrentSession().createQuery(hql)
                .setParameter("id", event.getId())
                .setParameter("renderedDrawerNotification", event.getRenderedDrawerNotification())
                .executeUpdate();
```

engine/src/test/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepositoryTest.java
```diff
@@ -63,32 +70,67 @@ void removeFixtures() {
         entityManager.createQuery("DELETE FROM DrawerNotification").executeUpdate();
     }
 
+    private void createDrawerNotification(DrawerNotification drawerNotification) {
+        entityManager.persist(drawerNotification);
+    }
+
     @Test
     @Transactional
     void testSimpleCreateGetCascadeDelete() {
```
gwenneg, line 79: Now that the old `DrawerNotificationRepository#create` method has been removed, this test only tests test code so we don't really need it anymore. That's a lot of "test" in the same sentence 😅.

engine/src/main/java/com/redhat/cloud/notifications/processors/drawer/DrawerProcessor.java
gwenneg: You're right, it could be useful even though the name would be a bit misleading since we're not invoking anything 😃 Let's keep it!

engine/src/test/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepositoryTest.java
gwenneg: Okay let's keep it for now.

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EventRepository.java
gwenneg: That's because we're mixing transactional queries with non-transactional ones. Can you try:
- adding `@Transactional` to `DrawerNotificationRepository#create`
- adding `@Transactional` to `EventRepository#updateDrawerNotification`
- my suggestion above with HQL instead of update
?

gwenneg: That's because we're mixing transactional queries with non-transactional ones. Let's try to improve that later. I don't want to delay the merge of this PR because of it.

gwenneg (approved):
LGTM

PR #1973, opened 2023-06-02
 RHCLOUD-25597 | fix: export service's endpoint name

MikelAlejoBR:
The application was complaining that the "export-service" was not being found in Clowder's endpoints section, and the reason was that the naming was wrong.

## Links


josejulio (approved):
lgtm

PR #1980, opened 2023-06-06
[RHCLOUD-26223] Fix single email for multiple recipients

g-duval:


engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -164,8 +167,10 @@ private void sendEmail(Event event, Set<Endpoint> endpoints) {
                 .getEmailSubscribersUserId(event.getOrgId(), bundleName, applicationName, eventTypeName, emailSubscriptionType));
 
         Set<User> userList = recipientResolver.recipientUsers(event.getOrgId(), requests, subscribers);
-        if (featureFlipper.isSendSingleEmailForMultipleRecipientsEnabled()) {
-            emailSender.sendEmail(userList, event, subject, body, true);
+        if (isSendSingleEmailForMultipleRecipientsEnabled(userList)) {
+            if (!userList.isEmpty()) {
```
josejulio, line 171: good catch!

josejulio: I'm wondering if this check should be inside the `sendEmail` method

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -279,4 +284,12 @@ private void processAggregateEmailsByAggregationKey(EmailAggregationKey aggregat
             emailAggregationRepository.purgeOldAggregation(aggregationKey, endTime);
         }
     }
+
+    private boolean isSendSingleEmailForMultipleRecipientsEnabled(Set<User> users) {
+        if (ProfileManager.getLaunchMode() == NORMAL && featureFlipper.isSendSingleEmailForMultipleRecipientsEnabled()) {
+            Set<String> strUsers = users.stream().map(User::getUsername).collect(Collectors.toSet());
+            return (strUsers.contains("gduval-prod") || strUsers.contains("gduval-stage"));
```
josejulio, line 291: I wonder if we should extend our feature flipper to be able to conditionally enable features on specific accounts. Just wanting to start a discussion and I don't think it should affect this PR

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSender.java
```diff
@@ -55,6 +55,9 @@ public class EmailSender {
     @ConfigProperty(name = "processor.email.bop_env")
     String bopEnv;
 
+    @ConfigProperty(name = "processor.email.no_reply")
```
josejulio, line 58: This is already part of the clowder file: https://github.com/RedHatInsights/notifications-backend/blob/master/.rhcicd/clowdapp-engine.yaml#L369C9-L371

josejulio (approved):
lgtm

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
gwenneg: That's one of the many ideas I wish I had time to implement in the quarkus-feature-flag extension (which doesn't exist yet) 😄

PR #2008, opened 2023-06-19
[RHCLOUD-25999] Update drawer topic name

g-duval:


gwenneg:
This won't be needed.

PR #2018, opened 2023-06-20
Prevent eventing history update failure

gwenneg:
On rare occasions, the following issue can happen in any environment:
```
2023-06-20 17:12:04,813 INFO  [com.red.clo.not.eve.ConnectorReceiver] (executor-thread-0) Camel notification history update failed because no record was found with [id=4ef1a736-ec9e-4b96-bf49-a385e40ebf94]
```
This PR prevents that by persisting the history in the DB before the engine receives the return Kafka message from the eventing apps.


MikelAlejoBR (approved):
LGTM!

PR #2032, opened 2023-06-26
RHCLOUD-25597 | feature: Grafana dashboards for the exports service integration

MikelAlejoBR:
These Grafana dashboards will help us identify whether the service is working as expected or if, on the other hand, we should investigate and take action because of a misbehavior.

## Links

### DIFF
--- a/.rhcicd/grafana/grafana-dashboard-notifications-general.configmap.yaml
+++ b/.rhcicd/grafana/grafana-dashboard-notifications-general.configmap.yaml
@@ -4522,7 +4522,7 @@ data:
         {
           "datasource": {
             "type": "prometheus",
-            "uid": "PDD8BE47D10408F45"
+            "uid": "$datasource"
           },
           "description": "Information about the health of the integration.",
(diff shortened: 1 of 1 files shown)


.rhcicd/grafana/grafana-dashboard-notifications-general.configmap.yaml
```diff
@@ -4518,6 +4519,202 @@ data:
             "align": false
           }
         },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "PDD8BE47D10408F45"
```
g-duval, line 4525: `"datasource": { "uid": ...}` should be `"uid": "$datasource"` like others.
It will allow to change targeted env easily.
(same comment for 9 others occurences of this section :) )

MikelAlejoBR: Good catch! Thanks!

g-duval (approved):
LGTM

PR #2034, opened 2023-06-26
[RHCLOUD-26001] add support of application/eventType template for drawer

g-duval:


engine/src/main/java/com/redhat/cloud/notifications/templates/DrawerTemplateMigrationService.java
```diff
@@ -0,0 +1,362 @@
...
+        Log.debug("Migration starting");
+
+        // Advisor
+        createDrawerIntegrationTemplate(
+            warnings, "rhel", "advisor", List.of(DEACTIVATED_RECOMMENDATION),
+            "Advisor/deactivatedRecommendationBody", "html", "Advisor deactivated recommendation drawer body"
+        );
+        createDrawerIntegrationTemplate(
+            warnings, "rhel", "advisor", List.of(NEW_RECOMMENDATION),
+            "Advisor/newRecommendationBody", "html", "Advisor new recommendation drawer body"
```
gwenneg, line 45: I wonder if we should have a clearer naming strategy for all DB templates (not only the drawer ones).

For example, instead of having:
- `Advisor/newRecommendationInstantEmailBody`
- `Advisor/newRecommendationBody`

we could use:
- `Advisor/email/newRecommendationBody`
- `Advisor/drawer/newRecommendationBody`

Changing the templates names wouldn't require to move the files. They would stay where they are right now.

WDYT?

engine/src/main/resources/templates/drawer/Advisor/deactivatedRecommendationBody.html
```diff
@@ -0,0 +1,7 @@
+{#let recs=data.events.size()}
+{#if recs > 1}
+    {recs} recommendations have recently been deactivated by Red Hat Insights and are no longer affecting your systems.
+{#else}
+    {recs} recommendation has recently been deactivated by Red Hat Insights and is no longer affecting your systems.
```
gwenneg, line 5:
```suggestion
    1 recommendation has recently been deactivated by Red Hat Insights and is no longer affecting your systems.
```

engine/src/main/resources/templates/drawer/AdvisorOpenShift/newRecommendationBody.html
```diff
@@ -0,0 +1 @@
+<b>{action.context.display_name}</b> has {action.events.size()} new {#if action.events.size() is 1}recommendation{#else}recommendations{/if}.
```
gwenneg, line 1:
```suggestion
<b>{data.context.display_name}</b> has {data.events.size()} new {#if data.events.size() is 1}recommendation{#else}recommendations{/if}.
```

engine/src/main/resources/templates/drawer/Inventory/validationErrorBody.html
```diff
@@ -0,0 +1 @@
+If no hosts were created by this change, the error will not appear in the service.
```
gwenneg, line 1: This template doesn't look very helpful.

gwenneg (commented):
I'm still reviewing. Here's a first round of comments.

engine/src/main/resources/templates/drawer/AdvisorOpenShift/newRecommendationBody.html
g-duval: Good catch :)

engine/src/main/resources/templates/drawer/Inventory/validationErrorBody.html
g-duval: It's part of mocks provided by UX team, so I gess it could be helpful for the user ?

gwenneg: There is useful information (host name and error message) in the email template this was extracted from.

This version is kind of empty and doesn't look like it'll help the user in any way.

engine/src/test/java/com/redhat/cloud/notifications/IntegrationTemplatesInDbHelper.java
```diff
@@ -0,0 +1,129 @@
...
+                eventType = resourceHelpers.createEventType(application.getId(), eventTypeToCreate);
+            }
+            eventTypes.put(eventTypeToCreate, eventType.getId());
+        }
+        migrate();
+    }
+
+    protected void migrate() {
+        drawerTemplateMigrationService.migrate();
+    }
```
gwenneg, line 76:
```suggestion
        drawerTemplateMigrationService.migrate();
    }
```

engine/src/test/java/com/redhat/cloud/notifications/IntegrationTemplatesInDbHelper.java
```diff
@@ -0,0 +1,129 @@
...
+
+    protected String generateDrawerTemplate(String eventTypeStr, Action action) {
+        return generateIntegrationTemplate(getIntegrationType(), eventTypeStr, action);
+    }
+
+    protected String generateDrawerTemplate(String eventTypeStr, Map<String, Object> context) {
+        return generateIntegrationTemplate("drawer", eventTypeStr, context);
+    }
+
```
gwenneg, line 85: This is never used:
```suggestion
```

engine/src/test/java/com/redhat/cloud/notifications/IntegrationTemplatesInDbHelper.java
```diff
@@ -0,0 +1,129 @@
...
+        TemplateInstance templateInstance = templateService.compileTemplate(integrationTemplate.getTheTemplate().getData(), integrationTemplate.getTheTemplate().getName());
+        return renderTemplate(templateInstance, action);
+    }
+
+    protected String generateIntegrationTemplate(String integrationType, String eventTypeStr, Map<String, Object> context) {
+        IntegrationTemplate integrationTemplate = templateRepository.findIntegrationTemplate(application.getId(), eventTypes.get(eventTypeStr), null, IntegrationTemplate.TemplateKind.ALL, integrationType).get();
+        TemplateInstance templateInstance = templateService.compileTemplate(integrationTemplate.getTheTemplate().getData(), integrationTemplate.getTheTemplate().getName());
+        return renderTemplate(templateInstance, context);
+    }
+
```
gwenneg, line 97: This is never used:
```suggestion
```

engine/src/test/java/com/redhat/cloud/notifications/IntegrationTemplatesInDbHelper.java
```diff
@@ -0,0 +1,129 @@
...
+    protected String renderTemplate(TemplateInstance templateInstance, Map<String, Object> context) {
+        String result = templateInstance
+            .data("action", Map.of("context", context, "bundle", getBundle(), "timestamp", LocalDateTime.now()))
+            .data("environment", environment)
+            .render();
+
+        return result.trim();
+    }
+
```
gwenneg, line 115: This is never used:
```suggestion
```

engine/src/test/java/com/redhat/cloud/notifications/templates/drawer/TestAdvisorTemplate.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.redhat.cloud.notifications.IntegrationTemplatesInDbHelper;
+import com.redhat.cloud.notifications.TestHelpers;
+import com.redhat.cloud.notifications.ingress.Action;
+import io.quarkus.test.junit.QuarkusTest;
+import org.junit.jupiter.api.Test;
+import java.util.List;
+
+import static com.redhat.cloud.notifications.processors.email.aggregators.AdvisorEmailAggregator.*;
+import static org.junit.jupiter.api.Assertions.*;
```
gwenneg, line 11:
```suggestion
import static com.redhat.cloud.notifications.processors.email.aggregators.AdvisorEmailAggregator.DEACTIVATED_RECOMMENDATION;
import static com.redhat.cloud.notifications.processors.email.aggregators.AdvisorEmailAggregator.NEW_RECOMMENDATION;
import static com.redhat.cloud.notifications.processors.email.aggregators.AdvisorEmailAggregator.RESOLVED_RECOMMENDATION;
import static org.junit.jupiter.api.Assertions.assertEquals;
```

engine/src/test/java/com/redhat/cloud/notifications/templates/drawer/TestEdgeManagementTemplate.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Test
+    void testRenderedTemplateImageCreation() {
+        long nbMonths = ChronoUnit.MONTHS.between(
+            LocalDate.from(ACTION.getTimestamp()),
+            LocalDate.now(UTC));
+
+        String result = generateDrawerTemplate(IMAGE_CREATION, ACTION);
+        assertEquals(String.format("A new image named <b>Test name</b> was created %d months ago.", nbMonths), result);
```
gwenneg, line 39: Won't this test fail when the age becomes greater than one year?

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
```diff
@@ -124,33 +124,56 @@ private InstantEmailTemplate loadDefaultEmailTemplate() throws Exception {
...
-     * @param appName Application name the template applies to. Can be null.
+     * @param appId Application name for the template applies to. Can be null.
+     * @param eventTypeId Event type Id for the template applies to. Can be null.
      * @param orgId The organization id for templates that are organization specific. Need to be of Kind ORG
      * @param templateKind Kind of template requested. If it does not exist, Kind DEFAULT is returned or Optional.empty()
      * @param integrationType Type of integration requested. E.g. 'slack', 'teams' or 'splunk'
      * @return IntegrationTemplate with potential fallback or Optional.empty() if there is not even a default template.
      */
-    public Optional<IntegrationTemplate> findIntegrationTemplate(String appName,
+    public Optional<IntegrationTemplate> findIntegrationTemplate(UUID appId,
```
gwenneg, line 134: This is not your responsibility but I really really don't like this method (old or new one, doesn't matter). This is such a mess from a maintenance perspective...

engine/src/main/java/com/redhat/cloud/notifications/templates/DrawerTemplateMigrationService.java
```diff
@@ -0,0 +1,360 @@
...
+
+            Template bodyTemplate = getOrCreateTemplate(bodyTemplateName, bodyTemplateExtension, bodyTemplateDescription, integrationType);
+            if (eventType.isPresent() && !integrationTemplateExists(eventType.get(), integrationType)) {
+                Log.infof("Creating integration template for event type: %s/%s/%s", bundleName, appName, eventTypeName);
+
+                IntegrationTemplate integrationTemplate = new IntegrationTemplate();
+                integrationTemplate.setTemplateKind(IntegrationTemplate.TemplateKind.EVENT_TYPE);
+                integrationTemplate.setIntegrationType(integrationType);
+                integrationTemplate.setCreated(LocalDateTime.now());
+                integrationTemplate.setApplication(eventType.get().getApplication());
```
gwenneg, line 286: This shouldn't be needed.
```suggestion
```

engine/src/main/java/com/redhat/cloud/notifications/processors/drawer/DrawerProcessor.java
```diff
@@ -178,16 +178,16 @@ public String buildNotificationMessage(Event event) {
...
+            .render().trim();
 
         return message;
     }
 
     @CacheResult(cacheName = "drawer-template")
-    TemplateInstance getTemplate() {
-        IntegrationTemplate integrationTemplate = templateRepository.findIntegrationTemplate(null, null, DEFAULT, "drawer")
+    TemplateInstance getTemplate(UUID applicationId, UUID eventTypeId, String orgId) {
+        IntegrationTemplate integrationTemplate = templateRepository.findIntegrationTemplate(applicationId, eventTypeId, orgId, ALL, "drawer")
```
gwenneg, line 190: Can we get rid of `applicationId`?
```suggestion
    TemplateInstance getTemplate(UUID eventTypeId, String orgId) {
        IntegrationTemplate integrationTemplate = templateRepository.findIntegrationTemplate(null, eventTypeId, orgId, ALL, "drawer")
```

common/src/main/java/com/redhat/cloud/notifications/models/IntegrationTemplate.java
```diff
@@ -103,6 +123,7 @@ public enum TemplateKind {
         DEFAULT, // A fallback if nothing more specific is defined
         APPLICATION, // specific for one (sending) application
         EVENT_TYPE, // specific for an event_type of an application
-        ORG  // defined for a single (customer) organisation
+        ORG,  // defined for a single (customer) organisation
+        ALL // all kind of template can be acceptable
```
gwenneg, line 127: Why is `ALL` needed?

database/src/main/resources/db/migration/V1.81.0__RHCLOUD-26001_add_custom_templates_for_drawer_notifs.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE integration_template
+    ADD COLUMN event_type_id UUID,
+    ADD constraint fk_generic_template_event_id FOREIGN KEY (event_type_id) REFERENCES event_type (id) ON DELETE CASCADE;
```
gwenneg, line 3:
```suggestion
    ADD constraint fk_integration_template_event_type_id FOREIGN KEY (event_type_id) REFERENCES event_type (id) ON DELETE CASCADE;
```

engine/src/main/java/com/redhat/cloud/notifications/processors/drawer/DrawerProcessor.java
gwenneg: Wouldn't this code work if `ALL` was replaced with `EVENT_TYPE`?

engine/src/main/java/com/redhat/cloud/notifications/processors/drawer/DrawerProcessor.java
```diff
@@ -178,16 +178,16 @@ public String buildNotificationMessage(Event event) {
...
             .orElseThrow(() -> new IllegalStateException("No default template defined for drawer"));
```
gwenneg, line 191: The exception message could be updated. If the exception is thrown, it means we didn't find:
- an integration template for the drawer matching the event type
- a default drawer integration template

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
```diff
@@ -124,33 +124,56 @@ private InstantEmailTemplate loadDefaultEmailTemplate() throws Exception {
...
+            hql += "AND (it.eventType IS NULL OR it.eventType.id = :eventTypeId) ";
+        } else {
+            hql += "AND (it.eventType IS NULL) ";
+        }
+
+        if (appId != null) {
+            hql += "AND (it.application IS NULL OR it.application.id = :appId) ";
+        } else {
+            hql += "AND (it.application IS NULL) ";
+        }
```
gwenneg, line 159: The old behavior was a bit different. Any reason to change that?
```suggestion
        if (appId != null) {
            hql += "AND it.application.id = :appId ";
        }
```

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
```diff
@@ -124,33 +124,56 @@ private InstantEmailTemplate loadDefaultEmailTemplate() throws Exception {
...
+            hql += "AND (it.orgId IS NULL OR it.orgId = :orgId) ";
+        } else {
+            hql += "AND (it.orgId IS NULL) ";
+        }
+
+        if (eventTypeId != null) {
+            hql += "AND (it.eventType IS NULL OR it.eventType.id = :eventTypeId) ";
+        } else {
+            hql += "AND (it.eventType IS NULL) ";
+        }
```
gwenneg, line 153: Wouldn't this be enough?
```suggestion
        if (eventTypeId != null) {
            hql += "AND it.eventType.id = :eventTypeId ";
        }
```

common/src/main/java/com/redhat/cloud/notifications/models/IntegrationTemplate.java
```diff
@@ -36,10 +36,14 @@ public class IntegrationTemplate extends CreationUpdateTimestamped {
     @JsonProperty(access = READ_ONLY)
     private UUID id;
 
-    @OneToOne(fetch = FetchType.LAZY)
+    @ManyToOne(fetch = FetchType.LAZY)
```
gwenneg, line 39: If `applicationId` is not used while dealing with event type template (as I suggested in other comments), then I wonder if that change is needed... :thinking:

common/src/main/java/com/redhat/cloud/notifications/models/IntegrationTemplate.java
g-duval: It's to be able to request any kind of template at once, no matter if it's an Org, event_type, application or default template lever.

engine/src/main/java/com/redhat/cloud/notifications/processors/drawer/DrawerProcessor.java
g-duval: Because the `findIntegrationTemplate` supports fallbacks between all kind of templates and the lowest level is le `default` type, this exception will be raised only if default template is missing.
It should never happends unless someone delete it from BD by mistake.

engine/src/test/java/com/redhat/cloud/notifications/templates/drawer/TestEdgeManagementTemplate.java
g-duval: Right

engine/src/main/java/com/redhat/cloud/notifications/processors/drawer/DrawerProcessor.java
g-duval: Not if a template is define at an org id level

common/src/main/java/com/redhat/cloud/notifications/models/IntegrationTemplate.java
g-duval: Because we can have one template per event type, and we can have multiple event types for a single application, this is why it seems natural for me to update it to have a consistent data on this table.

engine/src/main/resources/templates/drawer/Inventory/validationErrorBody.html
g-duval: I checked with Jess, the idea is to show to tenants what drawer notifications could looks like.
Then they will have to adjust and validate them based on stage data.

gwenneg:
Rebased onto `master` to increase the deployment timeout in ephemeral.

engine/src/main/java/com/redhat/cloud/notifications/templates/DrawerTemplateMigrationService.java
```diff
@@ -0,0 +1,360 @@
...
+        );
+
+        createDrawerIntegrationTemplate(
+            warnings, "console", "rbac", List.of("custom-role-created"),
+            "Rbac/customRoleCreatedBody", "html", "Rbac custom role created drawer body"
+        );
+
+        createDrawerIntegrationTemplate(
+            warnings, "console", "rbac", List.of("custom-role-updated"),
+            "Rbac/customRoleUpdatedBody", "html", "Rbac custom role created drawer body"
```
gwenneg, line 185:
```suggestion
            "Rbac/customRoleUpdatedBody", "html", "Rbac custom role updated drawer body"
```

gwenneg (approved):
LGTM

PR #2036, opened 2023-06-27
Render templates in tests using TemplateService

josejulio:


MikelAlejoBR (approved):
LGTM!

engine/src/test/java/com/redhat/cloud/notifications/EmailTemplatesInDbHelper.java
```diff
@@ -138,25 +135,20 @@ protected String generateAggregatedEmailBody(Action action) {
...
+    @Deprecated
+    protected String generateEmail(TemplateInstance template, Object actionOrEvent) {
 
+        String result = templateService.renderTemplate(createUser(), actionOrEvent, template);
         writeOrSendEmailTemplate(result, template.getTemplate().getId() + ".html");
 
         return result;
     }
 
+    @Deprecated
```
g-duval, line 147: Both `generateEmail` methods are now tagged as `Deprecated`.
The new equivalent method isn't it missing ?

josejulio: I guess that's wrong. These shouldn't be deprecated, thanks!

g-duval (approved):
LGTM

PR #2037, opened 2023-06-27
RHCLOUD-26685 Fix default email templates for cloud events

josejulio:
Depends on #2036 

@g-duval I think we will need to do something similar for the default templates when using the drawer.
If the bundle/app/event-type are needed, we could implement something to actually have that information.

### DIFF
--- a/engine/src/main/resources/templates/Common/insightsEmailBody.html
+++ b/engine/src/main/resources/templates/Common/insightsEmailBody.html
@@ -581,7 +581,7 @@
                             <table role="presentation" border="0" align="center" cellpadding="0" cellspacing="0" width="100%">
                                 <tr>
                                     <td class="rh-footer__text">
-                                        This email was sent by Hybrid Cloud Console | <a href="{environment.url}/user-preferences/notifications/{action.bundle}" target="_blank">Manage email preferences</a>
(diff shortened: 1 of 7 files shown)


engine/src/main/resources/templates/Common/insightsEmailBody.html
```diff
@@ -581,7 +581,7 @@ <h1>{#insert content-title}{/}</h1>
                             <table role="presentation" border="0" align="center" cellpadding="0" cellspacing="0" width="100%">
                                 <tr>
                                     <td class="rh-footer__text">
-                                        This email was sent by Hybrid Cloud Console | <a href="{environment.url}/user-preferences/notifications/{action.bundle}" target="_blank">Manage email preferences</a>
+                                        This email was sent by Hybrid Cloud Console | <a href="{environment.url}/user-preferences/notifications/{action.bundle.or("")}" target="_blank">Manage email preferences</a>
```
josejulio, line 584: We don't have bundles in cloud events. With the upcoming change about merging all bundles into a single screen I don't think we need to work on this. Cloud events are still in-progress and it might happen that when we finalize them the bundles are merged anyway.

g-duval (approved):
LGTM

g-duval:
> Depends on #2036
> 
> @g-duval I think we will need to do something similar for the default templates when using the drawer. If the bundle/app/event-type are needed, we could implement something to actually have that information.

Yes I think we have to, because those baet data are already used on Slack, Teams & Gchat templates.
The drawer default template doen't use it for the moment, but it could in the future

PR #2040, opened 2023-06-28
[RHCLOUD-26179] Remove email templates V1

g-duval:


engine/src/main/java/com/redhat/cloud/notifications/templates/EmailTemplateMigrationService.java
```diff
@@ -546,12 +479,13 @@ private String loadResourceTemplate(String name, String extension) {
      */
     private void createInstantEmailTemplate(List<String> warnings, String bundleName, String appName, List<String> eventTypeNames,
             String subjectTemplateName, String subjectTemplateExtension, String subjectTemplateDescription,
-            String bodyTemplateName, String bodyTemplateExtension, String bodyTemplateDescription, boolean useTemplateV2) {
+            String bodyTemplateName, String bodyTemplateExtension, String bodyTemplateDescription) {
 
-        if (useTemplateV2) {
+        if (!featureFlipper.isUseSecuredEmailTemplates()) {
```
g-duval, line 484: Today only Secured env are using templates without V2 extension.
They are V2 indeed but because it was created after V2 migration, we didn't add this extention on them.

V2 extension will have to be removed on RHCLOUD-26726. I didn't did it on this PR because it has already lot of file modified :)

josejulio: thanks!

engine/src/test/java/com/redhat/cloud/notifications/EmailTemplatesInDbHelper.java
```diff
@@ -154,7 +154,6 @@ protected String generateEmail(TemplateInstance templateInstance, Map<String, Ob
         String result = templateInstance
             .data("action", Map.of("context", context, "bundle", getBundle(), "timestamp", LocalDateTime.now()))
             .data("environment", environment)
-            .data("user", Map.of("firstName", "John", "lastName", "Doe"))
```
josejulio, line 157: We are not using the `user` in v2, but we still have code sending this variable to the templates. We should remove the user in a separate PR.

josejulio (changes requested):

engine/src/test/java/com/redhat/cloud/notifications/EmailTemplatesInDbHelper.java
josejulio: Coindentally, there is a conflict on this file.

g-duval: Thats right, thanks

josejulio:
This one had a conflict but upon a quick "fix conflicts" in github i broke something. :sweat_smile:

josejulio (approved):
lgtm, just need to fix the conflicts - my last commit didn't help.

PR #2043, opened 2023-06-29
RHCLOUD-26568 | feature: new Floorist queries

MikelAlejoBR:
The idea is to gather more accurate and useful data in order to see how Notifications is being used by the users.

## Links

[[RHCLOUD-26568]](https://issues.redhat.com/browse/RHCLOUD-26568)


.rhcicd/clowdapp-backend.yaml
```diff
@@ -183,25 +183,44 @@ objects:
...
         SELECT
-        COUNT(distinct account_id)
-        FROM endpoints
-        WHERE endpoint_type_v2 = 'CAMEL' AND endpoint_sub_type = 'splunk' AND enabled IS FALSE;
+          e.endpoint_type_v2 AS "Endpoint type",
+          count(e.endpoint_type_v2) AS "Count"
+        FROM
+          endpoints AS e
+        INNER JOIN
+          behavior_group_action AS bga ON bga.endpoint_id = e.id
```
josejulio, line 219: We should probably check if the behavior is also linked to an event type.

.rhcicd/clowdapp-backend.yaml
```diff
@@ -183,25 +183,44 @@ objects:
       secretName: ${FLOORIST_BUCKET_SECRET_NAME}
     suspend: ${{FLOORIST_SUSPEND}}
     queries:
-    - prefix: insights/notifications/splunk
+    # Count the number of created endpoints per endpoint type. Raw total, which
+    # includes the disabled ones or the ones not assigned in a behavior group.
```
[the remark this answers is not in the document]
MikelAlejoBR: Good question. The only way I can think of is by manually adding a `WHERE NOT IN ()` condition, which doesn't seem like a good solution... I wish org ids were pseudo-random, where internal accounts simply get a `rh-org-id` prefix or something like that...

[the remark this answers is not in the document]
josejulio: Maybe floorist could provide us a list of org-ids that are internal? Unsure how floorist works :)

.rhcicd/clowdapp-backend.yaml
MikelAlejoBR: Done!

.rhcicd/clowdapp-backend.yaml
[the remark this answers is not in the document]
MikelAlejoBR: Asking! I'll update the thread once I receive a response

josejulio (approved):
lgtm

.rhcicd/clowdapp-backend.yaml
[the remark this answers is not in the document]
MikelAlejoBR: Apparently Floorist doesn't have a way of filtering stuff... One of the ways I can think of is by seeing which accounts are internal accounts, grabbing those org ids, and then passing them via a Clowder environment variable.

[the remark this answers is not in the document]
josejulio: I wonder if it's something floorist can provide

.rhcicd/clowdapp-backend.yaml
```diff
@@ -183,25 +183,64 @@ objects:
...
-          DISTINCT "account_id"
-          FROM "endpoints"
-          WHERE "endpoint_type_v2"='CAMEL' AND "endpoint_sub_type"='splunk';
-    - prefix: insights/notifications/accounts
+          endpoint_type_v2 AS "Endpoint type",
+          count(endpoint_type_v2) AS "Count"
+        FROM
+          endpoints
+        GROUP BY
+        endpoint_type_v2;
```
gwenneg, line 196: I think we should include the endpoints subtype somehow in the queries. Each subtype of the `CAMEL` type should have its own record in the query result. And because the whole CAMEL / not CAMEL logic won't make any sense for someone outside of our team, it'd be better if we simply exported one "type" column showing directly the kind of endpoint.

gwenneg: Here's what I meant:

| Endpoint type | Count |
| - | - |
| email | 3 |
| slack | 12 |
| webhook | 7 |
| splunk | 42 |

.rhcicd/clowdapp-backend.yaml
```diff
@@ -183,25 +183,64 @@ objects:
...
+    # but in this case just counting the enabled ones.
+    - prefix: insights/notifications/endpoint_types
       query: >-
         SELECT
-        "applications"."id", "applications"."name", COUNT("applications"."id") AS "applications_count"
-        FROM "applications"
-        INNER JOIN "endpoint_email_subscriptions" ON "applications"."id"="endpoint_email_subscriptions"."application_id"
-        GROUP BY "applications"."id" HAVING COUNT("applications"."id") > 1;
-    - prefix: insights/notifications/accounts
+          endpoint_type_v2 AS "Enabled endpoint type",
```
gwenneg, line 202: Same comment for the endpoints type/subtype thing.

.rhcicd/clowdapp-backend.yaml
```diff
@@ -183,25 +183,64 @@ objects:
...
+        GROUP BY endpoint_type_v2;
+    # Count the number of enabled endpoints per type which are assigned to a
+    # behavior group, and therefore considered to be in active use.
+    - prefix: insights/notifications/endpoint_types
       query: >-
         SELECT
-        COUNT(distinct account_id)
-        FROM endpoints
-        WHERE endpoint_type_v2 = 'CAMEL' AND endpoint_sub_type = 'splunk' AND enabled IS FALSE;
+          e.endpoint_type_v2 AS "Endpoint type",
```
gwenneg, line 214: Same comment for the endpoints type/subtype thing.

.rhcicd/clowdapp-backend.yaml
```diff
@@ -183,25 +183,64 @@ objects:
...
+            COUNT(es.subscription_type) AS "Count"
+        FROM
+            email_subscriptions AS es
+        INNER JOIN
+            event_type AS et ON et.id = es.event_type_id
+        INNER JOIN
+            applications AS a ON a.id = et.application_id
+        GROUP BY
+            es.subscription_type,
+            a.display_name
```
gwenneg, line 242: From a business intelligence perspective, it would be useful to group the data by:
- subscription type
- bundle display name
- app display name
- event type display name

The BI team will certainly have the tools to analyze the data with a focus on any level of the B/A/ET hierarchy.

.rhcicd/clowdapp-backend.yaml
gwenneg: Untested:
```sql
        SELECT
          CASE
            WHEN endpoint_type_v2 = 'CAMEL' THEN endpoint_sub_type
            ELSE endpoint_type_v2
          END AS "Endpoint type",
          count(endpoint_type_v2) AS "Count"
        FROM
          endpoints
        GROUP BY
        endpoint_type_v2;
```

MikelAlejoBR: Updated!

.rhcicd/clowdapp-backend.yaml
MikelAlejoBR: Updated!

.rhcicd/clowdapp-backend.yaml
MikelAlejoBR: Updated!

.rhcicd/clowdapp-backend.yaml
MikelAlejoBR: I've done it in a single query. Were you thinking on maybe splitting the query in four?

gwenneg: One query is perfect 👍

.rhcicd/clowdapp-backend.yaml
```diff
@@ -183,25 +183,91 @@ objects:
...
+        FROM
+          endpoints AS e
+        INNER JOIN
+          behavior_group_action AS bga ON bga.endpoint_id = e.id
+        INNER JOIN
+            event_type_behavior AS etb ON etb.behavior_group_id = bga.behavior_group_id
+        WHERE
+          enabled=true
+        GROUP BY
+          e.endpoint_type_v2
```
gwenneg, line 247:
```suggestion
          e.endpoint_type_v2, e.endpoint_sub_type;
```

gwenneg: I'm not sure if semicolons are important here. If they are, then it was missing in this query.

.rhcicd/clowdapp-backend.yaml
```diff
@@ -183,25 +183,91 @@ objects:
...
+          event_type AS et ON et.id = es.event_type_id
+        INNER JOIN
+          applications AS a ON a.id = et.application_id
+        INNER JOIN
+          bundles AS b ON b.id = a.bundle_id
+        GROUP BY
+          b.display_name,
+          a.display_name,
+          et.display_name,
+          es.subscription_type
```
gwenneg, line 270: Semicolon needed?

.rhcicd/clowdapp-backend.yaml
```diff
@@ -183,25 +183,91 @@ objects:
...
+            ELSE
+              LOWER(e.endpoint_type_v2)
+          END AS "Endpoint type",
+          count(e.endpoint_type_v2) AS "Count"
+        FROM
+          endpoints AS e
+        INNER JOIN
+          behavior_group_action AS bga ON bga.endpoint_id = e.id
+        INNER JOIN
+            event_type_behavior AS etb ON etb.behavior_group_id = bga.behavior_group_id
```
gwenneg, line 243: Instead of running 3 queries for endpoints, we could run a single one that would return all the information:
| Endpoint type | Count | Enabled | Actively used |
| - | - | - | - |
| Slack | 18 | true | true |
| Slack | 7 | true | false |
[...]

That kind of data structure can easily by aggregated and/or filtered on any column using a pivot table, which is one of the basic tools used in business intelligence.

.rhcicd/clowdapp-backend.yaml
```diff
@@ -183,25 +183,91 @@ objects:
...
-    - prefix: insights/notifications/accounts
+          CASE
+            WHEN
+              endpoint_type_v2 = 'CAMEL'
+            THEN
+              LOWER(endpoint_sub_type)
+            ELSE
+              LOWER(endpoint_type_v2)
+          END AS "Endpoint type",
+          count(endpoint_type_v2) AS "Count"
```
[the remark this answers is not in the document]
MikelAlejoBR: Sure, changed!

.rhcicd/clowdapp-backend.yaml
```diff
@@ -183,25 +183,91 @@ objects:
...
+              LOWER(endpoint_sub_type)
+            ELSE
+              LOWER(endpoint_type_v2)
+          END AS "Enabled endpoint type",
+          count(endpoint_type_v2) AS "Count"
+        FROM
+          endpoints
+        WHERE
+          enabled=true
+        GROUP BY endpoint_type_v2;
```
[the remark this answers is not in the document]
MikelAlejoBR: Oops! Updated!

.rhcicd/clowdapp-backend.yaml
MikelAlejoBR: Sure, changed!

.rhcicd/clowdapp-backend.yaml
MikelAlejoBR: As well, yes!

.rhcicd/clowdapp-backend.yaml
MikelAlejoBR: I have come up with the following query, but grouping by the endpoint's ID —which is required to be used in the subquery, since otherwise SQL complains— makes lots of duplicated results —obviously—. I am trying to figure out a better way, but I'll leave the query here just in case you spot something:

```sql
SELECT
  -- When the endpoint is a Camel endpoint, get its subtype instead.
  CASE
    WHEN
      e.endpoint_type_v2 = 'CAMEL'
    THEN
      LOWER(e.endpoint_sub_type)
    ELSE
      LOWER(e.endpoint_type_v2)
  END AS "Endpoint type",
  COUNT(*) AS "Count",
  e.enabled AS "Enabled",
  -- If the endpoint has an associated event type and a behavior group, then
  -- that means that it is actively being used.
  EXISTS (
    SELECT
        1
    FROM
        behavior_group_action AS bga
    INNER JOIN
        event_type_behavior AS etb ON etb.behavior_group_id = bga.behavior_group_id
    WHERE
        bga.endpoint_id = e.id 
) AS "Actively used"
FROM
  endpoints AS e
WHERE
  e.enabled=TRUE
GROUP BY
  e.endpoint_type_v2,
  e.endpoint_sub_type,
  e.enabled,
  e.id
```

MikelAlejoBR: I have left the results of this query after running it in stage in the ticket —restricted to Red Hat employee for security purposes—.

gwenneg (approved):
LGTM

PR #2048, opened 2023-07-04
[RHCLOUD-25503] Add a timer metric on email aggregation service

g-duval:


MikelAlejoBR (approved):

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -171,6 +176,7 @@ public void consumeEmailAggregations(String aggregationCommandJson) {
 
         Log.infof("Processing received aggregation command: %s", aggregationCommand);
         processedAggregationCommandCount.increment();
+        Timer.Sample consumedTimer = Timer.start(registry);
```
gwenneg, line 179: I think we should count the time spent deserialization the Kafka payload as well.

g-duval: Ok

gwenneg (approved):
LGTM

PR #2051, opened 2023-07-05
[RHCLOUD-25503] Add aggregations on event log

g-duval:
Daily digest emails on Event log will look like:
image


.rhcicd/clowdapp-aggregator.yaml
```diff
@@ -61,8 +61,8 @@ objects:
           value: ${ENV_NAME}
         - name: PROMETHEUS_PUSHGATEWAY_URL
           value: ${PROMETHEUS_PUSHGATEWAY}
-        - name: NOTIFICATIONS_AGGREGATOR_ORG_PREF_ENABLED
```
gwenneg, line 64: Did you mean to remove this old feature in this PR? It looks like the branch/PR contains two unrelated topics.

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -96,12 +112,8 @@ public void processDailyEmail() {
                 }
             }
```
gwenneg, line 113: I just noticed how wrong the old code is: we're checking if all `futures` are completed on each loop iteration. It should only be done once after the loop is done iterating.

gwenneg: Fixed in https://github.com/RedHatInsights/notifications-backend/pull/2120

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -137,37 +151,29 @@ List<AggregationCommand> processAggregateEmailsWithOrgPref(LocalDateTime endTime
...
+    private void sendIt(AggregationCommand aggregationCommand) {
+
+        Payload.PayloadBuilder payloadBuilder = new Payload.PayloadBuilder();
+        Map<String, Object> payload = JsonObject.mapFrom(aggregationCommand).getMap();
+        payload.forEach(payloadBuilder::withAdditionalProperty);
+
+        Action action = new Action.ActionBuilder()
+            .withBundle(BUNDLE_NAME)
+            .withApplication(APP_NAME)
+            .withEventType(String.format("%s-%s-%s", EVENT_TYPE_NAME, aggregationCommand.getAggregationKey().getBundle(), aggregationCommand.getAggregationKey().getApplication()))
```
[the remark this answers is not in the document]
gwenneg: cc @josejulio @MikelAlejoBR

aggregator/src/main/resources/application.properties
```diff
@@ -9,12 +9,20 @@ quarkus.datasource.jdbc.url=jdbc:postgresql://127.0.0.1:5432/notifications
...
 mp.messaging.outgoing.aggregation.connector=smallrye-kafka
 mp.messaging.outgoing.aggregation.topic=platform.notifications.aggregation
 mp.messaging.outgoing.aggregation.group.id=integrations
 mp.messaging.outgoing.aggregation.key.serializer=org.apache.kafka.common.serialization.StringSerializer
 mp.messaging.outgoing.aggregation.value.serializer=org.apache.kafka.common.serialization.StringSerializer
 
+# Output queue
+mp.messaging.outgoing.egress.connector=smallrye-kafka
+mp.messaging.outgoing.egress.topic=platform.notifications.ingress
+mp.messaging.outgoing.egress.group.id=integrations
```
gwenneg, line 22: This is a common "mistake" in our configuration files but `group.id` only makes sense for `incoming` channels. There is no such thing as groups for Kafka producers.
```suggestion
```

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -72,18 +87,19 @@ public void processDailyEmail() {
...
             List<CompletableFuture<Void>> futures = new ArrayList<>();
             for (AggregationCommand aggregationCommand : aggregationCommands) {
                 try {
                     final String payload = objectMapper.writeValueAsString(aggregationCommand);
-                    futures.add(emitter.send(payload).toCompletableFuture());
+                    if (featureFlipper.isAggregatorSendOnIngress()) {
+                        sendIt(aggregationCommand);
+                    } else {
+                        futures.add(emitter.send(payload).toCompletableFuture());
+                    }
```
gwenneg, line 102: I wonder why we initially implemented the aggregation job using `CompletableFutures` rather than relying on `Emitter#send(Message)` as you did here. Maybe there was a reason, maybe not... Anyway it would be better to confirm that calling `Emitter#send(Message)` will actually lead to messages being sent without any interference from the Quarkus shutdown at the end of the CronJob execution.

josejulio: IIRC It was something about quarkus finishing before the messages were sent.

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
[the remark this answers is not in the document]
josejulio: It's still unclear to me why we can't only write to the EventLog table that has everything decoupled (i.e. no FKs) and use only the displayName of the EventLog's event-type :thinking:

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
gwenneg: > IIRC It was something about quarkus finishing before the messages were sent.

That makes sense when `Emitter#send(String)` is called because that's an async operation (as implied by the `CompletableFuture` return type). On the other hand, `Emitter#send(Message)` is a sync operation and I'm not sure we ever tried using it.

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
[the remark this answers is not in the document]
gwenneg: We could alter `Event#eventTypeDisplayName` value and put the additional information without adding the `details` field I mentioned, if that's what your suggesting. That would require event less effort in the backend and zero change in the frontend, so I like this even more than using `details` :smile:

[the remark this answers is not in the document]
g-duval: If I understand properly you are suggesting:
1. Adding only one application + event_type on bundle console to be able to receive aggregation events
2. Adding on propery on `notification_history#details` to be able to override the `eventTypeDisplayName`
3. Patching the event api to be able to return the custom `eventTypeDisplayName` in case of aggregation event

Form my point of view, I've to think and find a solution regarding following impacts: 
1.   The main aggregation event_type will be considered as regular event type, that involve the capacity to subscribe, unsubscribe to it, linking some templates etc...
2. Event log filter on event name have to be updated to look on regular event display name, but also on 'patched' display name, which can have an impact on api response time

@josejulio @gwenneg Could you please confirm my understanding ?

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -138,37 +152,29 @@ List<AggregationCommand> processAggregateEmailsWithOrgPref(LocalDateTime endTime
...
+                    new Event.EventBuilder()
+                        .withMetadata(new Metadata.MetadataBuilder().build())
+                        .withPayload(payloadBuilder.build())
+                        .build()))
+            .build();
+
+        String encodedAction = Parser.encode(action);
+        Log.infof("Encoded Payload: %s", encodedAction);
+        Message<String> message = Message.of(encodedAction);
+        emitterIngress.send(message);
```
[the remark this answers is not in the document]
g-duval: Ok :)

gwenneg:
Rebased onto `master` to fix conflicts.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailSubscriptionTypeProcessor.java
```diff
@@ -227,53 +318,62 @@ private void processAggregateEmailsByAggregationKey(EmailAggregationKey aggregat
...
                     event.setEventWrapper(new EventWrapperAction(action));
 
-                    emailSender.sendEmail(aggregation.getValue(), event, subject, body, false);
+                    NotificationHistory history = emailSender.sendEmail(aggregation.getValue(), event, subject, body, false, endpoint);
+                    Integer totalRecipients = (Integer) history.getDetails().get(TOTAL_RECIPIENTS_KEY);
+                    if (NotificationStatus.SUCCESS == history.getStatus()) {
+                        nbRecipientsSuccessfullySent += totalRecipients;
+                    } else {
+                        nbRecipientsFailureSent += totalRecipients;
+                    }
```
gwenneg, line 330: `history` can be `null`.
```suggestion
                    if (history != null) {
                        Integer totalRecipients = (Integer) history.getDetails().get(TOTAL_RECIPIENTS_KEY);
                        if (NotificationStatus.SUCCESS == history.getStatus()) {
                            nbRecipientsSuccessfullySent += totalRecipients;
                        } else {
                            nbRecipientsFailureSent += totalRecipients;
                        }
                    }
```

gwenneg (changes requested):

gwenneg (approved):
LGTM

PR #2053, opened 2023-07-06
RHCLOUD-23645 | feature: fetch users from MBOP instead of the IT service

MikelAlejoBR:
Notifications shouldn't be directly calling the IT service for users, since MBOP has been designed and developed to act as the middleman, thus saving us from having to deal with things like certificates in our application. Therefore, this is the first step to replace that integration.

## Links

[[RHCLOUD-23645]](https://issues.redhat.com/browse/RHCLOUD-23645)

### DIFF
--- a/engine/src/main/resources/application.properties
+++ b/engine/src/main/resources/application.properties
@@ -203,8 +203,7 @@ quarkus.rest-client.export-service.trust-store-type=${clowder.private-endpoints.
 # Quarkus caches
 quarkus.cache.caffeine.drawer-template.expire-after-write=PT5M
 
-# REST client settings for the back office proxy, BOP or MBOP.
+# REST client settings for the back office proxy, BOP or MBOP. The proxy isn't
+# a Clowder application, so the parameter must be overriden in stage and
+# production.
 quarkus.rest-client.mbop.url=http://localhost:8090
-quarkus.rest-client.mbop.trust-store=${clowder.private-endpoints.export-service-service.trust-store-path}
-quarkus.rest-client.mbop.trust-store-password=${clowder.private-endpoints.export-service-service.trust-store-password}
-quarkus.rest-client.mbop.trust-store-type=${clowder.private-endpoints.export-service-service.trust-store-type}
--- a/engine/src/main/resources/application.properties
+++ b/engine/src/main/resources/application.properties
@@ -130,6 +130,7 @@ quarkus.log.sentry.in-app-packages=*
 quarkus.log.sentry.dsn=FILL_ME
 
 # BOP properties
+quarkus.rest-client.mbop.url=https://backoffice-proxy.apps.ext.spoke.preprod.us-west-2.aws.paas.redhat.com
 processor.email.bop_url=https://backoffice-proxy.apps.ext.spoke.preprod.us-west-2.aws.paas.redhat.com
 processor.email.bop_apitoken=addme
 processor.email.bop_client_id=policies
@@ -202,8 +203,3 @@ quarkus.rest-client.export-service.trust-store-type=${clowder.private-endpoints.
 
 # Quarkus caches
 quarkus.cache.caffeine.drawer-template.expire-after-write=PT5M
-
-# REST client settings for the back office proxy, BOP or MBOP. The proxy isn't
-# a Clowder application, so the parameter must be overriden in stage and
-# production.
-quarkus.rest-client.mbop.url=http://localhost:8090
--- a/engine/src/main/resources/application.properties
+++ b/engine/src/main/resources/application.properties
@@ -202,3 +202,9 @@ quarkus.rest-client.export-service.trust-store-type=${clowder.private-endpoints.
 
 # Quarkus caches
 quarkus.cache.caffeine.drawer-template.expire-after-write=PT5M
+
+# REST client settings for the back office proxy, BOP or MBOP.
+quarkus.rest-client.mbop.url=http://localhost:8090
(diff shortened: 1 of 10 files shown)


engine/src/main/resources/application.properties
```diff
@@ -221,3 +221,9 @@ quarkus.cache.caffeine.drawer-template.expire-after-write=PT5M
 # Temp - For testing purposes - Revert ASAP
 # When set to 'false', the Camel routes discovery will be disabled in the engine, preventing any Google Chat, Teams or Slack notification from being sent
 quarkus.camel.routes-discovery.enabled=false
+
+# REST client settings for the back office proxy, BOP or MBOP.
+quarkus.rest-client.mbop.url=http://localhost:8090
+quarkus.rest-client.mbop.trust-store=${clowder.private-endpoints.export-service-service.trust-store-path}
+quarkus.rest-client.mbop.trust-store-password=${clowder.private-endpoints.export-service-service.trust-store-password}
+quarkus.rest-client.mbop.trust-store-type=${clowder.private-endpoints.export-service-service.trust-store-type}
```
josejulio, line 229: Clowder refeference needs updating. I think we should refer to `bop` and not `mbop`

engine/src/test/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProviderTest.java
```diff
@@ -275,6 +286,42 @@ public void getAllGroupsCache() {
...
+
+            // Return the mocked list first and then an empty one afterwards,
+            // since otherwise Mockito will always return the mocked list,
+            // which makes the MBOP loop go forever and hit heat space errors.
+            // In reality, an organization that has more than a 1000 users is
+            // going to be a rare case, and even if it is, the method can
+            // handle a few thousand elements.
+            Mockito
+                .when(this.mbopService.getUsersByOrgId(Mockito.eq(TestConstants.DEFAULT_ORG_ID), Mockito.eq(adminsOnly), Mockito.eq(RbacRecipientUsersProvider.MBOP_SORT_ORDER), Mockito.eq(this.MBOPMaxResultsPerPage), Mockito.anyInt()))
+                .thenReturn(MBOPUsers, new ArrayList<>());
```
josejulio, line 310: Could you test when the second "page" return some users?

josejulio (changes requested):

engine/src/test/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProviderTest.java
MikelAlejoBR: Yes! I've changed it so that we simulate two full pages plus and a third one with half a page, to test that the loop works correctly.

engine/src/main/resources/application.properties
MikelAlejoBR: I thought I had written a comment here, but it seems that I left it as a pending review. I was saying that I didn't see the MBOP service definition in Clowder, so I wasn't sure what to put here... I will double check with the access management team, just to be safe.

engine/src/test/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProviderTest.java
josejulio: Thanks!

engine/src/main/resources/application.properties
MikelAlejoBR: After checking with them they told us that BOP/MBOP are not Clowderized applications, and therefore the parameters must be manually set in AppInterface. That is why I gave the REST Client a default value locally, but didn't set any value in Clowder, so that we are forced to set a value in stage and production.

engine/src/main/resources/application.properties
```diff
@@ -201,3 +201,8 @@ quarkus.rest-client.export-service.trust-store-type=${clowder.private-endpoints.
 
 # Quarkus caches
 quarkus.cache.caffeine.drawer-template.expire-after-write=PT5M
+
+# REST client settings for the back office proxy, BOP or MBOP. The proxy isn't
+# a Clowder application, so the parameter must be overriden in stage and
+# production.
+quarkus.rest-client.mbop.url=http://localhost:8090
```
josejulio, line 208: I think we need a similar set of properties:

https://github.com/RedHatInsights/notifications-backend/blob/38e429815ffb216c8851504832000381ff813132/engine/src/main/resources/application.properties#L132-L135

Actually we can use the very same properties or even extend the other BOP bean.

MikelAlejoBR: I guess the answer is yes, but do we need the token and client IDs? In the REST client I have not implemented those bits...

josejulio: 2 weeks later... sorry! 
Yes, I believe we will need the token and client IDs to talk with BOP.

gwenneg (approved):
My comments above are completely minor. I forgot the most important part: LGTM

gwenneg:
@josejulio Are you happy with the latest code?

engine/src/main/resources/application.properties
```diff
@@ -129,6 +129,7 @@ quarkus.log.sentry.in-app-packages=*
 quarkus.log.sentry.dsn=FILL_ME
 
 # BOP properties
+quarkus.rest-client.mbop.url=https://backoffice-proxy.apps.ext.spoke.preprod.us-west-2.aws.paas.redhat.com
```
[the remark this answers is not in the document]

PR #2060, opened 2023-07-11
Make export-service a hard dependency

ezr-ondrej:
To make sure we always deploy the export-service along the engine :)


MikelAlejoBR (approved):
LGTM!

PR #2066, opened 2023-07-13
RHCLOUD-27092 Prevent duplicate Kafka processing for Splunk

gwenneg:
Related to https://github.com/RedHatInsights/eventing-integrations/pull/210

When `notifications-connector-splunk` will be deployed in production, it will belong to a different Kafka group (`notifications-connector-splunk`) compared to the old eventing app (`eventing-splunk`). Because of that change, the connector will automatically consume the Kafka topic from the smallest Kafka offset available rather than from the offset of the latest message consumed by the `eventing-splunk` Kafka group.

This PR introduces the `NOTIFICATIONS_SPLUNK_CONNECTOR_KAFKA_PROCESSING_ENABLED` feature flag in `engine` which will be used to prevent duplicate processing during the switch from eventing to the connector. When that flag is `true` and a message is produced, the connector will consume that message and eventing will ignore it. When the flag is `false` or missing, eventing will consume the message and the connector will ignore it.


g-duval (approved):
LGTM

PR #2074, opened 2023-07-20
RHCLOUD-27205 Make Kafka polling strategy easier to configure

gwenneg:
This PR makes changes to the Kafka polling strategy easier for connectors.

Changing the max poll records or delay is often needed when long running requests (such as calls to failing external hosts) are performed during the processing of a Kafka message.


MikelAlejoBR (approved):
LGTM!

PR #2077, opened 2023-07-21
[RHCLOUD-25882] Improve drawer Api response time

g-duval:


MikelAlejoBR (approved):
LGTM

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
```diff
@@ -40,10 +40,9 @@ public List<DrawerEntryPayload> getNotifications(String orgId, String username,
         query.setSortFields(DrawerNotification.SORT_FIELDS);
         query.setDefaultSortBy("created:DESC");
         Optional<Query.Sort> sort = query.getSort();
-
         String hql = "SELECT dn.id, dn.read, " +
-            "dn.event.bundleDisplayName, dn.event.applicationDisplayName, dn.event.eventTypeDisplayName, dn.created, dn.event.renderedDrawerNotification "
-            + "FROM DrawerNotification dn where dn.orgId = :orgId and dn.userId = :userid";
+            "ev.bundleDisplayName, ev.applicationDisplayName, ev.eventTypeDisplayName, dn.created, ev.renderedDrawerNotification "
+            + "FROM DrawerNotification dn inner join dn.event as ev where dn.orgId = :orgId and dn.userId = :userid";
```
gwenneg, line 45: Does that `inner join` have any impact on the response time? In the old code, Hibernate took care of that join automatically, maybe in a slightly different way.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
```diff
@@ -64,9 +63,8 @@ public List<DrawerEntryPayload> getNotifications(String orgId, String username,
 
     public Long count(String orgId, String username, Set<UUID> bundleIds, Set<UUID> appIds, Set<UUID> eventTypeIds,
                       LocalDateTime startDate, LocalDateTime endDate, Boolean readStatus) {
-
-        String hql = "SELECT count(dn.id) "
-            + "FROM DrawerNotification dn where dn.orgId = :orgId and dn.userId = :userid";
+        String hql = "SELECT COUNT(*) FROM DrawerNotification dn "
+                        + "inner join dn.event as ev where dn.orgId = :orgId and dn.userId = :userid";
```
gwenneg, line 67: Same question for this `inner join`.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
gwenneg: I'm very surprised to hear that Hibernate would do a cross join because it makes no sense based on the HQL query and entities mapping. Maybe there was a bug in Hibernate 5...

I just checked with Hibernate 6 and the SQL contains a default `join` (which is inner).

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
gwenneg: We don't need the inner join with `dn.event` here because we're not accessing the event table during the query execution.

gwenneg: I'd remove the `inner join` here because it will not be needed in for all queries and I think we can trust Hibernate 6 to make the right decision about the join type.

database/src/main/resources/db/migration/V1.84.0__RHCLOUD-25882_add_indexes_on_notification_drawer.sql
gwenneg, line 1: `V1.84` needs to be replaced with `V1.85`.

database/src/main/resources/db/migration/V1.84.0__RHCLOUD-25882_add_indexes_on_notification_drawer.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX ix_drawer_notification_org_id_user_id_created_read_event_id ON drawer_notification (org_id, user_id, created, read, event_id);
```
gwenneg, line 1: Will drawer entries always be sorted in descending order? If so, it could help to add DESC to the index declaration:
```suggestion
CREATE INDEX ix_drawer_notification_org_id_user_id_created_read_event_id ON drawer_notification (org_id, user_id, created DESC, read, event_id);
```

backend/src/main/java/com/redhat/cloud/notifications/routers/DrawerResource.java
```diff
@@ -55,13 +57,17 @@ public Page<DrawerEntryPayload> getDrawerEntries(@Context SecurityContext securi
...
         List<DrawerEntryPayload> drawerEntries = new ArrayList<>();
         if (count > 0) {
             drawerEntries = drawerRepository.getNotifications(orgId, username, bundleIds, appIds, eventTypeIds, startDate, endDate, readStatus, query);
         }
-
+        LocalDateTime now = LocalDateTime.now();
+        Log.infof("Drawer request duration %s for orgId: %s, userId: %s",
+            ChronoUnit.MILLIS.between(start, now),
+            orgId,
+            username);
```
gwenneg, line 70: Is that still needed with the connector?

database/src/main/resources/db/migration/V1.84.0__RHCLOUD-25882_add_indexes_on_notification_drawer.sql
g-duval: Good point

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
g-duval: Indeed generated SQL request seems have changed :)

gwenneg (approved):
LGTM

PR #2082, opened 2023-07-22
Change base image to ubi9/openjdk-17-runtime

gwenneg:
This is a test.

I'm discussing with the team responsible for `ubi9/openjdk-17-runtime` to determine if we can switch to that image.


g-duval:
lgtm, but webhook connector updates are missing :)

g-duval (approved):

PR #2087, opened 2023-07-24
[RHCLOUD-27283] Fix Sonar Jacoco config for test coverage report

g-duval:
Using sonar.coverage.jacoco.xmlReportPaths property will allow all sub modules to append the same report test file


MikelAlejoBR (approved):
LGTM

PR #2088, opened 2023-07-24
RHCLOUD-27292 Splunk/SNow don't need extras and _originalId

gwenneg:
`extras` and `_originalId` were never used in the Splunk or ServiceNow integrations, even in the old eventing apps.


gwenneg:

PR #2092, opened 2023-07-27
Configure Splunk HTTPS timeouts programmatically

gwenneg:
I suspect that the HTTP timeout settings we had until now for Splunk were ignored by the Camel `https` component.

This PR changes the configuration of these timeouts from `application.properties` to a programmatic approach with some logging at startup and the possibility to override the config values from the deployment params.

### DIFF
--- a/connector-common/src/main/java/com/redhat/cloud/notifications/connector/ConnectorConfig.java
+++ b/connector-common/src/main/java/com/redhat/cloud/notifications/connector/ConnectorConfig.java
@@ -1,14 +1,13 @@
 package com.redhat.cloud.notifications.connector;
 
+import io.quarkus.arc.DefaultBean;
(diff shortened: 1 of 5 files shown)


connector-common/src/main/java/com/redhat/cloud/notifications/connector/ConnectorConfig.java
```diff
@@ -77,109 +76,43 @@ public int getEndpointCacheMaxSize() {
         return endpointCacheMaxSize;
     }
 
-    public void setEndpointCacheMaxSize(int endpointCacheMaxSize) {
```
gwenneg, line 80: The setter were not used so I removed them all.

MikelAlejoBR (approved):
LGTM!

PR #2108, opened 2023-08-10
fix(HMS-2321): fix image builder instant email templates

amirfefer:
This PR removes timestamps from instant emails and adds a `n/a` value to the DNS column when the value is empty (can be empty in some cloud providers).

Screen Shot 2023-08-10 at 19 35 17

### DIFF
--- a/engine/src/main/resources/templates/ImageBuilder/launchFailedEmailBodyV2.html
+++ b/engine/src/main/resources/templates/ImageBuilder/launchFailedEmailBodyV2.html
@@ -16,7 +16,7 @@ Google Cloud Platform
 {/switch}
 {/content-title-section1}
 {#content-subtitle-section1}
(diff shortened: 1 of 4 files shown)


engine/src/main/resources/templates/ImageBuilder/launchFailedEmailBodyV2.html
```diff
@@ -16,7 +16,7 @@
 {/switch}
 {/content-title-section1}
 {#content-subtitle-section1}
-{action.timestamp.toStringFormat()} | launch id #{action.context.launch_id}
+launch id #{action.context.launch_id}
```
josejulio, line 19:
```suggestion
Launch id #{action.context.launch_id}
```

josejulio (commented):
Minor change to use `Launch` instad of `launch` on a subtitle

PR #2109, opened 2023-08-11
[RHCLOUD-27618] add bearer authentication support on webhooks

g-duval:


common/src/main/java/com/redhat/cloud/notifications/models/CamelProperties.java
```diff
@@ -100,6 +100,22 @@ public void setBasicAuthenticationSourcesId(Long basicAuthenticationSourcesId) {
         this.basicAuthenticationSourcesId = basicAuthenticationSourcesId;
     }
 
+    public Long getBearerAuthenticationSourcesId() {
+        return null;
+    }
+
+    public void setBearerAuthenticationSourcesId(Long bearerAuthenticationSourcesId) {
+        // do nothing here
```
josejulio, line 108: Are these placeholders for later in this PR or do plan to create a separate PR for this one?

common/src/main/java/com/redhat/cloud/notifications/models/SourcesSecretable.java
```diff
@@ -53,4 +53,12 @@ public interface SourcesSecretable {
...
      */
     void setBasicAuthenticationSourcesId(Long basicAuthenticationSourcesId);
+
+    Long getBearerAuthenticationSourcesId();
+
+    void setBearerAuthenticationSourcesId(Long bearerAuthenticationSourcesId);
+
+    String getBearerToken();
+
+    void setBearerToken(String bearerToken);
```
josejulio, line 63: Not for this PR, but I'm wondering if we should have an annotation (or similar) that takes care of adding the ID to the actual data.
```java
class MyProperties {
    @SourcesSecret
    String myTopSecretStuff;
}
```

josejulio: cc @MikelAlejoBR

MikelAlejoBR: The `SecretUtils` class contains a few checks with the `SourcesSecretable` interface, in order to be able to store/update/delete those secrets from the service: https://github.com/RedHatInsights/notifications-backend/blob/master/common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java#L92

I was thinking that we could totally use annotations for generating the getters/setters and storing the ID, but wouldn't that "break" the SecretUtils class? Or once generated, we should not have any problem?

josejulio: What do you mean?

MikelAlejoBR: Maybe I didn't understand you right. I thought that with the `@SourcesSecret` annotation getters, setters and a property would be automatically generated for that entity.

The "problem" I see is that I don't know if it is possible to generate code that adheres to an interface, and make that entity conform to it?

I am thinking that I might have understood the whole thing wrong.

josejulio: One way would be that the annotation only marks an attribute as "This is secretable - expect a `theAttributeId` property with setter/getter.

`SecretUtils` would look for those annotations and continue as usual (I found the annotation, I know that there is a `setTheStuffId` and `getTheStuffId`) and i'll use them accordingly. 

The enhancement is that we wouldn't need a fixed interface and it's going to be more flexible with more secrets across different types.

More information, but I don't think this provide any value other than being somewhat interesting:

Talking about code generation (one alternative) is that we create a processor that fetches the `SourcesSecret` annotation and generates a new class `MyOriginalNameSecretEnhanced` with the `ids` (getter/setters). We would still need to do as above.

We should be able to use that class within the same code as if was something we wrote directly.

[This](https://github.com/josejulio/versioned-path) is something I build during the past hackathon that mimics that for something else. Ultimately I think this approach will be less complicated as less handling of the super class is required.

common/src/main/java/com/redhat/cloud/notifications/models/CamelProperties.java
g-duval: That's it

gwenneg:
@josejulio @MikelAlejoBR I see you already took a look at this PR. Could one of you please take care of the review?

common/src/main/java/com/redhat/cloud/notifications/models/SourcesSecretable.java
MikelAlejoBR: All right, I have created a ticket so that we can apply this new approach: https://issues.redhat.com/browse/RHCLOUD-28488

common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
```diff
@@ -269,10 +285,10 @@ private long createBasicAuthentication(final BasicAuthentication basicAuthentica
      * @param orgId the organization id related to this operation for the tenant identification.
      * @return the id of the created secret.
      */
-    private long createSecretTokenSecret(final String secretToken, final String orgId) {
+    private long createSecretTokenSecret(final String secretToken, final String tokenType, final String orgId) {
         Secret secret = new Secret();
 
-        secret.authenticationType = Secret.TYPE_SECRET_TOKEN;
+        secret.authenticationType = tokenType;
         secret.password = secretToken;
```
MikelAlejoBR, line 292: Hhmmm... I was thinking that maybe it makes more sense to remove both the `SecretUtils#createBasicAuthentication(final BasicAuthentication basicAuthentication, final String orgId)` and the `createSecretTokenSecret(final String secretToken, final String tokenType, final String orgId)` functions and simply create the following one:

```java
    /**
     * Creates a secret in Sources.
     * @param authenticationType the type of secret we are storing.
     * @param username the username of the secret.
     * @param password the password of the secret.
     * @param orgId the organization ID which this secret belongs to.
     * @return the created secret's ID.
     */
    private long createSecret(final String authenticationType, final String username, final String password, final String orgId) {
        final Secret secret = new Secret();

        secret.authenticationType = authenticationType;
        secret.password = password;
        secret.username = username;

        final Secret createdSecret = this.sourcesService.create(
            orgId,
            this.sourcesPsk,
            secret
        );

        return createdSecret.id;
    }
```

What do you think?

common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
```diff
@@ -221,26 +233,30 @@ public void deleteSecretsForEndpoint(Endpoint endpoint) {
...
 
+    private void deleteSecret(Endpoint endpoint, Long secretTokenId, String format) {
+        this.sourcesService.delete(
+            endpoint.getOrgId(),
+            this.sourcesPsk,
+            secretTokenId
+        );
+        Log.infof(format, endpoint.getId(), secretTokenId);
+    }
+
```
MikelAlejoBR, line 259: Just naming nitpicks :see_no_evil: 

```diff
-   private void deleteSecret(Endpoint endpoint, Long secretTokenId, String format) {
+   private void deleteSecret(Endpoint endpoint, Long secretId, String logMessageFormat) {
        this.sourcesService.delete(
            endpoint.getOrgId(),
            this.sourcesPsk,
-           secretTokenId
+           secretId
        );
+
-       Log.infof(format, endpoint.getId(), secretTokenId);
+       Log.infof(logMessageFormat, endpoint.getId(), secretTokenId);
    }
```

common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
```diff
@@ -170,42 +177,47 @@ public void updateSecretsForEndpoint(Endpoint endpoint) {
...
+            if (secretToken == null || secretToken.isBlank()) {
+                Log.debugf("[endpoint_id: %s] %s not created in Sources: the secret token object is null or blank", endpoint.getId(), logDisplaySecretType);
+            } else {
+                final long id = this.createSecretTokenSecret(secretToken, secretType, endpoint.getOrgId());
+                Log.infof("[endpoint_id: %s][secret_id: %s] %s created in Sources during an endpoint update operation", endpoint.getId(), id, logDisplaySecretType);
+                return id;
             }
         }
+        return secretTokenId;
     }
```
MikelAlejoBR, line 221: Again, naming nitpicks:

```diff
-    private Long updateSecretToken(Endpoint endpoint, String secretToken, Long secretTokenId, String secretType, String logDisplaySecretType) {
+    private Long updateSecretToken(Endpoint endpoint, Long secretId, String secretType, String password, String logDisplaySecretType) {
-        if (secretTokenId != null) {
+        if (secretId != null) {
-            if (secretToken == null || secretToken.isBlank()) {
+            if (password == null || password.isBlank()) {
                this.sourcesService.delete(
                    endpoint.getOrgId(),
                    this.sourcesPsk,
-                   secretTokenId
+                   secretId
                );
-               Log.infof("[endpoint_id: %s][secret_id: %s] %s deleted in Sources during an endpoint update operation", endpoint.getId(), secretTokenId, logDisplaySecretType);
+               Log.infof("[endpoint_id: %s][secret_id: %s] %s deleted in Sources during an endpoint update operation", endpoint.getId(), secretId, logDisplaySecretType);
                return null;
            } else {
                Secret secret = new Secret();
-               secret.password = secretToken;
+               secret.password = password;

                this.sourcesService.update(
                    endpoint.getOrgId(),
                    this.sourcesPsk,
-                   secretTokenId,
+                   secretId,
                    secret
                );
                Log.infof("[endpoint_id: %s][secret_id: %s] %s updated in Sources", endpoint.getId(), secretTokenId, logDisplaySecretType);
                return secretTokenId;
            }
        } else {
-            if (secretToken == null || secretToken.isBlank()) {
+            if (password == null || password.isBlank()) {
                Log.debugf("[endpoint_id: %s] %s not created in Sources: the secret token object is null or blank", endpoint.getId(), logDisplaySecretType);
            } else {
-               final long id = this.createSecretTokenSecret(secretToken, secretType, endpoint.getOrgId());
+               final long id = this.createSecretTokenSecret(secretType, null, password, endpoint.getOrgId());
                Log.infof("[endpoint_id: %s][secret_id: %s] %s created in Sources during an endpoint update operation", endpoint.getId(), id, logDisplaySecretType);
                return id;
            }
        }
        return secretId;
    }
```

MikelAlejoBR (commented):
LGTM! Just a very few minor comments...

MikelAlejoBR:
Here:

https://github.com/RedHatInsights/notifications-backend/blob/ce0e492a649267e88950e7eb9152a979fca7ebf4/backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTest.java#L42-L48

I would change the description like this:

```diff
    /**
-     * Tests that the underlying "get by id" function gets called two times: one for the basic authentication and
-     * another one for the secret token. It also tests that the endpoint's properties hold the secrets from the
-     * returned payload from Sources.
+     * Tests that the underlying "get by id" function gets called three times:
+     * one for the basic authentication, one for the secret token, and one for
+     * the bearer secret. It also tests that the endpoint's properties hold the
+     * secrets from the returned payload from Sources.
     */
    @Test
    void getSecretsForEndpointTest() {
```

MikelAlejoBR:
Here:

https://github.com/RedHatInsights/notifications-backend/blob/ce0e492a649267e88950e7eb9152a979fca7ebf4/backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTest.java#L125-L131

I would change the comment to:

```diff
    /**
-     * Tests that when the endpoint's properties has the "basic authentication" and the "secret token" properties set,
-     * the function under test calls "create" upon Sources two times, and that the returned IDs are properly set on the
-     * endpoint's properties.
+     * Tests that when the endpoint's properties has the "basic authentication",
+     * "secret token" and "bearer secret" properties set, the function under
+     * test calls "create" upon Sources three times, and that the returned IDs
+     * are properly set on the endpoint's properties.
     */
    @Test
    void createSecretsForEndpointTest() {
```

MikelAlejoBR:
Here:

https://github.com/RedHatInsights/notifications-backend/blob/ce0e492a649267e88950e7eb9152a979fca7ebf4/backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTest.java#L382-L388

I would change the comment to:

```diff
    /**
-     * Tests that the function under test calls "update" upon Sources two times, one for the "basic authentication" and
-     * another one for the "secret token". It also checks that the responses are properly marshalled into the right
-     * endpoint properties.
+     * Tests that the function under test calls "update" upon Sources three
+     * times, one for the "basic authentication",one for the "secret token" and
+     * another one for the "bearer secret". It also checks that the responses
+     * are properly marshalled into the right endpoint properties.
     */
    @Test
    void updateSecretsForEndpointTest() {
```

MikelAlejoBR:
Here:

https://github.com/RedHatInsights/notifications-backend/blob/ce0e492a649267e88950e7eb9152a979fca7ebf4/backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTest.java#L463-L468

I would change the comment to:

```diff
    /**
-     * Tests that when the user provides a {@code null} "basic authentication" and "secret token" secret for an
-     * endpoint which has secrets stored in Sources, the deletion process for those secrets is triggered.
+     * Tests that when the user provides a {@code null} "basic authentication",
+     * "secret token" and "bearer secret" secrets for an endpoint which has
+     * secrets stored in Sources, the deletion process for those secrets is
+     * triggered.
     */
    @Test
    void updateSecretsForEndpointDeleteTest() {
```

MikelAlejoBR:
In the following snippet it seems that I forgot to add a comment to the test. If you may please add it I'd appreciate it :pray: 

https://github.com/RedHatInsights/notifications-backend/blob/ce0e492a649267e88950e7eb9152a979fca7ebf4/backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTest.java#L528-L530

```diff
+    /**
+     * Tests that when the endpoint didn't have previously stored secrets, but
+     * when the user provides the "basic authentication", "secret token" and
+     * "bearer authentication" secrets, the underlying "create" function gets
+     * called three times.
+     */
    @Test
    void updateSecretsForEndpointCreateTest() {
```

backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTest.java
```diff
@@ -528,10 +573,11 @@ void updateSecretsForEndpointCreateTest() {
         // Assert the results.
         Assertions.assertEquals(BASIC_AUTH_SOURCES_ID, props.getBasicAuthenticationSourcesId(), "the ID of the basic authentication secret from Sources doesn't match");
         Assertions.assertEquals(SECRET_TOKEN_SOURCES_ID, props.getSecretTokenSourcesId(), "the ID of the secret token's secret from Sources doesn't match");
+        Assertions.assertEquals(BEARER_TOKEN_SOURCES_ID, props.getBearerAuthenticationSourcesId(), "the ID of the bearer secret from Sources doesn't match");
 
         // Assert that the underlying function was called exactly two times, since we are expecting that both the
```
MikelAlejoBR, line 578:
```diff
-// Assert that the underlying function was called exactly two times, since we are expecting that both the
+// Assert that the underlying function was called exactly three times, since we are expecting that both the
```

backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTest.java
```diff
@@ -88,11 +96,15 @@ void getSecretsForEndpointTest() {
         final String secretToken = props.getSecretToken();
         Assertions.assertEquals(SECRET_TOKEN, secretToken, "the secret token doesn't match");
 
+        final String secretBearer = props.getBearerAuthentication();
+        Assertions.assertEquals(BEARER_TOKEN, secretBearer, "the secret bearer doesn't match");
+
         // Assert that the underlying function was called exactly two times, since we are expecting that both the
         // "basic authentication" and the "secret token" secrets were fetched.
```
MikelAlejoBR, line 103:
```diff
-        // Assert that the underlying function was called exactly two times, since we are expecting that both the
-        // "basic authentication" and the "secret token" secrets were fetched.
+        // Assert that the underlying function was called exactly three times,
+        // since we are expecting that the "basic authentication", "secret
+        // token" and "bearer authentication" secrets were fetched.
```

common/src/main/java/com/redhat/cloud/notifications/routers/sources/Secret.java
```diff
@@ -17,6 +17,7 @@ public final class Secret {
 
     public static final String TYPE_BASIC_AUTH = "notifications-basic-authentication";
     public static final String TYPE_SECRET_TOKEN = "notifications-secret-token";
+    public static final String TYPE_BEARER_TOKEN = "notifications-bearer-token";
```
MikelAlejoBR, line 20: What do you think of renaming this one to `TYPE_BEARER_AUTHENTICATION`?

backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTest.java
```diff
@@ -528,10 +573,11 @@ void updateSecretsForEndpointCreateTest() {
         // Assert the results.
         Assertions.assertEquals(BASIC_AUTH_SOURCES_ID, props.getBasicAuthenticationSourcesId(), "the ID of the basic authentication secret from Sources doesn't match");
         Assertions.assertEquals(SECRET_TOKEN_SOURCES_ID, props.getSecretTokenSourcesId(), "the ID of the secret token's secret from Sources doesn't match");
+        Assertions.assertEquals(BEARER_TOKEN_SOURCES_ID, props.getBearerAuthenticationSourcesId(), "the ID of the bearer secret from Sources doesn't match");
 
         // Assert that the underlying function was called exactly two times, since we are expecting that both the
-        // "basic authentication" and the "secret token" secrets were created.
-        final int wantedNumberOfInvocations = 2;
+        // "basic authentication", "secret token" and "bearer" secrets were created.
```
MikelAlejoBR, line 579:
```diff
-        // Assert that the underlying function was called exactly three, since we are expecting that both the
-        // "basic authentication", "secret token" and "bearer" secrets were created.
+        // Assert that the underlying function was called exactly three times,
+        // since we are expecting that the "basic authentication", "secret
+        // token" and "bearer authentication" secrets were created.
```

backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTest.java
```diff
@@ -371,6 +395,10 @@ void updateSecretsForEndpointTest() {
         final Secret secretTokenMock = new Secret();
         secretTokenMock.password = SECRET_TOKEN;
 
+        // Create a "bearer" secret mock.
```
MikelAlejoBR, line 398:
```diff
-       // Create a "bearer" secret mock.
+       // Create a "bearer authentication" secret mock.
```

backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTest.java
```diff
@@ -202,10 +225,11 @@ void createSecretsForEndpointBasicAuthNullTest() {
 
         // Assert the results.
         Assertions.assertEquals(SECRET_TOKEN_SOURCES_ID, props.getSecretTokenSourcesId(), "the ID of the secret token's secret from Sources doesn't match");
+        Assertions.assertEquals(BEARER_TOKEN_SOURCES_ID, props.getBearerAuthenticationSourcesId(), "the ID of the bearer secret from Sources doesn't match");
 
         // Assert that the underlying function was called exactly one time: just for the "secret token"'s secret
         // creation.
```
MikelAlejoBR, line 231:
```diff
-       // Assert that the underlying function was called exactly one time: just for the "secret token"'s secret
-        // creation.
+        // Assert that the underlying function was called exactly two times:
+        // one for the "secret token"'s secret creation and another one for the
+        // "bearer authentication"'s secret creation.

File: backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTest.java [hunk 12]
```diff
@@ -175,19 +193,24 @@ void createSecretsForEndpointBasicAuthNullTest() {
...
         Endpoint endpoint = new Endpoint();
         WebhookProperties webhookProperties = new WebhookProperties();
         webhookProperties.setSecretToken(SECRET_TOKEN);
+        webhookProperties.setBearerAuthentication(BEARER_TOKEN);
 
         endpoint.setProperties(webhookProperties);
         endpoint.setOrgId(orgId);
 
         // Set up the mock call for the "create" call from the REST Client. In this case, since the basic
         // authentication is null, only the secret token should be created.
```
MikelAlejoBR, line 212:
```diff
-      // Set up the mock call for the "create" call from the REST Client. In this case, since the basic
-      // authentication is null, only the secret token should be created.
+        // Set up the mock call for the "create" call from the REST Client. In
+        // this case, since the basic authentication is null, only the secret
+        // token should be created.

File: backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTest.java [hunk 13]
```diff
@@ -157,10 +174,11 @@ void createSecretsForEndpointTest() {
         // Assert the results.
         Assertions.assertEquals(BASIC_AUTH_SOURCES_ID, props.getBasicAuthenticationSourcesId(), "the ID of the basic authentication secret from Sources doesn't match");
         Assertions.assertEquals(SECRET_TOKEN_SOURCES_ID, props.getSecretTokenSourcesId(), "the ID of the secret token's secret from Sources doesn't match");
+        Assertions.assertEquals(BEARER_TOKEN_SOURCES_ID, props.getBearerAuthenticationSourcesId(), "the ID of the secret bearer from Sources doesn't match");
 
         // Assert that the underlying function was called exactly two times, since we are expecting that both the
         // "basic authentication" and the "secret token" secrets were created.
```
MikelAlejoBR, line 180:
```diff
-        // Assert that the underlying function was called exactly two times, since we are expecting that both the
-        // "basic authentication" and the "secret token" secrets were created.
+        // Assert that the underlying function was called exactly three times,
+        // since we are expecting that the "basic authentication", "secret
+        // token" and "bearer authentication" secrets were created.
```

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
```diff
@@ -2466,6 +2473,8 @@ public void testUpdateEndpointCreateSecrets() {
 
             Assertions.assertEquals(25L, webhookProperties.getBasicAuthenticationSourcesId(), "Sources was called to create the basic authentication secret, but the secret's ID wasn't present in the database");
             Assertions.assertEquals(50L, webhookProperties.getSecretTokenSourcesId(), "Sources was called to create the secret token secret, but the secret's ID wasn't present in the database");
+            Assertions.assertEquals(75L, webhookProperties.getBearerAuthenticationSourcesId(), "Sources was called to create the secret token secret, but the secret's ID wasn't present in the database");
```
MikelAlejoBR, line 2476:
```diff
-            Assertions.assertEquals(75L, webhookProperties.getBearerAuthenticationSourcesId(), "Sources was called to create the secret token secret, but the secret's ID wasn't present in the database");
+            Assertions.assertEquals(75L, webhookProperties.getBearerAuthenticationSourcesId(), "Sources was called to create the bearer authentication token secret, but the secret's ID wasn't present in the database");
```

backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTest.java
```diff
@@ -53,6 +55,10 @@ void getSecretsForEndpointTest() {
         final Secret secretTokenMock = new Secret();
         secretTokenMock.password = SECRET_TOKEN;
 
+        // Create a "bearer secret" secret mock.
```
MikelAlejoBR, line 58:
```diff
-         // Create a "bearer secret" secret mock.
+         // Create a "bearer authentication" secret mock.

File: backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTest.java [hunk 16]
```diff
@@ -125,6 +137,10 @@ void createSecretsForEndpointTest() {
         final Secret secretTokenMock = new Secret();
         secretTokenMock.id = SECRET_TOKEN_SOURCES_ID;
 
+        // Create a "bearer secret" secret mock.
```
MikelAlejoBR, line 140:
```diff
- // Create a "bearer secret" secret mock.
+ // Create a "bearer authentication" secret mock.
```

database/src/main/resources/db/migration/V1.81.0__RHCLOUD-27618_add_bearer_authentication_on_webhooks.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE endpoint_webhooks ADD COLUMN bearer_authentication_id BIGINT;
+ALTER TABLE endpoint_webhooks ADD COLUMN bearer_authentication text;
```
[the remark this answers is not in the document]
g-duval: Right

common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
g-duval: For me it's already confusing to have to send a Bearer or a token on a field named `password` ;)
This is why I prefer havind dinstinct method to build the right kind of secret.

josejulio:
Could you update the migration name?
`Caused by: org.flywaydb.core.api.FlywayException: Found more than one migration with version 1.81.0`

g-duval:
> Could you update the migration name? `Caused by: org.flywaydb.core.api.FlywayException: Found more than one migration with version 1.81.0`

Sure, thanks @josejulio

common/src/main/java/com/redhat/cloud/notifications/routers/sources/SecretUtils.java
MikelAlejoBR: Yeah, I have to admit it is a bit weird... Good, the only reason I was suggesting a change is because we are reusing the `createSecretTokenSecret` name, but for now is okay

backend/src/test/java/com/redhat/cloud/notifications/routers/sources/SecretUtilsTest.java
```diff
@@ -505,14 +554,15 @@ void updateSecretsForEndpointCreateTest() {
...
         webhookProperties.setBasicAuthentication(basicAuth);
         webhookProperties.setSecretToken(SECRET_TOKEN);
+        webhookProperties.setBearerAuthentication(BEARER_AUTHENTICATION);
 
         endpoint.setProperties(webhookProperties);
         endpoint.setOrgId(orgId);
 
         // Set up the mock calls for the "create" calls from the REST Client. Make sure we return the basic
         // authentication's ID first, and the secret token's ID second, since we are expecting a successful create
         // operation.
```
MikelAlejoBR, line 564:
```suggestion
        // Set up the mock calls for the "create" calls from the REST Client.
        // Make sure we return the basic authentication's ID first, the secret
        // token's ID second and the bearer authentication's ID third, since we
        // are expecting a successful create operation.
```

MikelAlejoBR (approved):
LGTM!

MikelAlejoBR (approved):

PR #2113, opened 2023-08-14
Remove Splunk and SNow migration filters

gwenneg:
The Splunk and SNow apps have been undeployed from the eventing namespace in stage and prod. The migration is over, we don't need the filters anymore.


MikelAlejoBR (approved):
LGTM

PR #2120, opened 2023-08-21
Fix futures check in aggregator job

gwenneg:
There is no point checking if `futures` are completed on each iteration of the `for` loop. Doing it once at the end of the loop is more than enough.


josejulio (approved):
lgtm

MikelAlejoBR (approved):

PR #2145, opened 2023-09-08
Hide user-preferences API

josejulio:
- These APIs are used internally, they are not really user friendly in any way. We should probably create other APIs if customers want to use these.


g-duval (approved):
LGTM

PR #2155, opened 2023-09-11
RHCLOUD-28235 Unify how org ID is passed to connectors

gwenneg:
This is phase 1 of a two-steps change:
- Phase 1: we're passing the top-level `orgId` field to all connectors from the `engine`
- Phase 2: after the `engine` has been deployed in production, we'll use the top-level `orgId` field from all connectors and stop using other sources of org ID

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelNotification.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelNotification.java
@@ -2,14 +2,7 @@ package com.redhat.cloud.notifications.processors.camel;
 
 public class CamelNotification {
 
-    public String orgId;
-
     public String webhookUrl;
 
     public String message;
-
-    @Override
-    public String toString() {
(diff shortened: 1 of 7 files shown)


engine/src/main/java/com/redhat/cloud/notifications/processors/camel/CamelNotification.java
```diff
@@ -2,14 +2,7 @@
...
 public class CamelNotification {
 
-    public String orgId;
-
     public String webhookUrl;
 
     public String message;
-
-    @Override
-    public String toString() {
```
gwenneg, line 12: This stopped being used a while ago.

engine/src/main/java/com/redhat/cloud/notifications/processors/camel/slack/SlackNotification.java
```diff
@@ -5,9 +5,4 @@
 public class SlackNotification extends CamelNotification {
 
     public String channel;
-
-    @Override
-    public String toString() {
```
gwenneg, line 10: Same.

g-duval (approved):
LGTM

PR #2163, opened 2023-09-14
RHCLOUD-28244 | refactor: rename the email connector

MikelAlejoBR:
Setting the "subtype" in the result endpoint causes the history stub to be stored with an "email_subscription" endpoint type and an "email" subtype, which causes the user interface in the event log page to break, since they do not rightfully expect a subtype when fetching a history element that is associated to an email endpoint.

## Links


gwenneg (approved):
LGTM

PR #2164, opened 2023-09-14
RHCLOUD-28244 | refactor: include the org ID in the payload for the email connector

MikelAlejoBR:
The email connector was not being able to extract the organization ID
from the payload because the engine was not sending it. That caused the
email connector not to send the organization ID to the RBAC service,
and that service ended up responding with a 401 instead of the response
we were expecting.

## Depends on
* https://github.com/RedHatInsights/notifications-backend/pull/2163

## Links


gwenneg (approved):
LGTM

PR #2175, opened 2023-09-19
RHCLOUD-28393 Lower time spent processing RBAC events

gwenneg:
This PR introduces caching for DB data that rarely changes such as event types and templates.

It also suppresses unneeded calls to IT/RBAC/BOP when users from an org do not need to be retrieved.

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/recipients/RecipientResolver.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/recipients/RecipientResolver.java
@@ -6,6 +6,7 @@ import jakarta.annotation.PostConstruct;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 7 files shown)


josejulio (approved):
lgtm

engine/src/main/java/com/redhat/cloud/notifications/recipients/RecipientResolver.java
```diff
@@ -40,6 +41,16 @@ public Set<User> recipientUsers(String orgId, Set<RecipientSettings> requests, S
     }
 
     private Set<User> recipientUsers(String orgId, RecipientSettings request, Set<String> subscribers, boolean isOptIn) {
+        /*
+         If the subscription type is opt-in, the user preferences should NOT be ignored and the subscribers Set is empty,
+         then we don't need to retrieve the users from RBAC/IT/BOP because we'll return an empty Set anyway.
+         */
+        // TODO Report this change in the email connector ASAP.
+        if (isOptIn && !request.isIgnoreUserPreferences() && subscribers.isEmpty()) {
```
josejulio, line 49: Considering that this won't be true for the drawer, I wonder if we should have something more for fetching/caching the user's :thinking:

gwenneg: I think the only reasonable approach in the long run is to make the whole users retrieval an async process. That's something I've had in mind for a while but now I'm pretty sure we'll have to implement it soon.

PR #2181, opened 2023-09-20
[RHCLOUD-28395] send daily digest with email connector

g-duval:


engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregator.java
```diff
@@ -101,16 +101,26 @@ public Map<User, Map<String, Object>> getAggregated(EmailAggregationKey aggregat
...
-                    getSubscribers(eventType, subscribersByEventType)
-                );
+                Set<User> users;
+
+                if (featureFlipper.isEmailConnectorEnabled()) {
+                    users = getSubscribers(eventType, subscribersByEventType).stream().map(username -> {
+                        User usr = new User();
+                        usr.setUsername(username);
+                        return usr;
+                    }).collect(Collectors.toSet());
```
gwenneg, line 111: Even if we discussed it in Meet, I still dont feel completely comfortable with this change. Theorically, we could skip the aggregation of payloads for some users if user preferences are ignored. Let's keep the old code in this class for now if that's fine with you. It may mean calling IT/RBAC a few extra times but that's not so bad as a temporary solution (until the users service is available).

g-duval: Ok, let's rollback it

gwenneg (approved):
LGTM

PR #2187, opened 2023-09-21
Bump Quarkus to 3.4.1

gwenneg:
```
./mvnw io.quarkus.platform:quarkus-maven-plugin:3.4.1:update -N -Dstream=3.4
```

### DIFF
--- a/connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/config/Environment.java
+++ b/connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/config/Environment.java
@@ -1,0 +1,20 @@
+package com.redhat.cloud.notifications.connector.email.config;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+@ApplicationScoped
+public class Environment {
+    public static final String DEVELOPMENT = "development";
+
+    @ConfigProperty(name = "env.name", defaultValue = DEVELOPMENT)
+    String environment;
+
+    /**
+     * Checks if the "ENV_NAME" environment variable is set to "local-dev".
+     * @return true if the "ENV_NAME" environment variable is set to "local-dev".
+     */
+    public boolean isDevelopmentEnvironment() {
+        return DEVELOPMENT.equals(this.environment);
+    }
+}
--- a/connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/config/Environment.java
+++ b/connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/config/Environment.java
@@ -11,8 +11,8 @@ public class Environment {
     String environment;
 
     /**
-     * Checks if the "ENV_NAME" environment variable is set to "local-dev".
-     * @return true if the "ENV_NAME" environment variable is set to "local-dev".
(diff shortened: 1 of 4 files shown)


gwenneg:
Tests are broken, maybe because of [this](https://github.com/quarkusio/quarkus/wiki/Migration-Guide-3.4#lifecycle). I'll take a closer look later.

gwenneg:
@MikelAlejoBR This bump is failing because some tests from `connector-email` no longer work with Quarkus 3.4.1. Could you please take a look when you have time? This is not urgent.

MikelAlejoBR:
Hey Gwenneg,

Apparently the tests are failing because of the IT endpoint: it is trying to load the "changeme" JKS key store, and makes everything fail. I wondered which solution was better: check if the environment is development or testing and create the IT endpoint without the certificate, or provide a fake JKS file.

I decided on the second approach, but I am starting to wonder if it is better to do the first thing. Since otherwise it might silently fail in stage and production and we wouldn't notice, because of that fallback fake certificate key store.

gwenneg:
> I decided on the second approach, but I am starting to wonder if it is better to do the first thing. Since otherwise it might silently fail in stage and production and we wouldn't notice, because of that fallback fake certificate key store.

I'd prefer the first approach with an env check. The tests shouldn't have any kind of impact on the runtime code, stage or prod.

MikelAlejoBR:
> > I decided on the second approach, but I am starting to wonder if it is better to do the first thing. Since otherwise it might silently fail in stage and production and we wouldn't notice, because of that fallback fake certificate key store.
> 
> I'd prefer the first approach with an env check. The tests shouldn't have any kind of impact on the runtime code, stage or prod.

ACK, I'll change it then

MikelAlejoBR:
Changed!

gwenneg:
Instead of introducing a feature flag, could we use something similar to [Environment](https://github.com/RedHatInsights/notifications-backend/blob/master/common/src/main/java/com/redhat/cloud/notifications/models/Environment.java) in the connector?

MikelAlejoBR:
> Instead of introducing a feature flag, could we use something similar to [Environment](https://github.com/RedHatInsights/notifications-backend/blob/master/common/src/main/java/com/redhat/cloud/notifications/models/Environment.java) (only with the needed fields/methods) in the connector?

Done! Please let me know what you think!

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/config/Environment.java
```diff
@@ -0,0 +1,20 @@
...
+@ApplicationScoped
+public class Environment {
+    public static final String DEVELOPMENT = "development";
+
+    @ConfigProperty(name = "env.name", defaultValue = DEVELOPMENT)
+    String environment;
+
+    /**
+     * Checks if the "ENV_NAME" environment variable is set to "local-dev".
+     * @return true if the "ENV_NAME" environment variable is set to "local-dev".
```
gwenneg, line 15:
```suggestion
     * Checks if the "ENV_NAME" environment variable is set to "development".
     * @return true if the "ENV_NAME" environment variable is set to "development".
```

gwenneg:
Thanks! LGTM

PR #2188, opened 2023-09-21
RHCLOUD-28244 | fix: backoffice proxy's path

MikelAlejoBR:
The backoffice proxy is returning a 404 at the moment because the path of the endpoint to send emails is wrong.

The reason for using those environment variable names which don't have the "NOTIFICATIONS_CONNECTOR" prefix is that they get set in AppInterface, and therefore we must honor that naming so that they get properly replaced when the connector gets deployed to stage or production.

## Jira ticket


gwenneg (approved):
LGTM

PR #2189, opened 2023-09-21
RHCLOUD-28244 | fix: IT service's URL generation

MikelAlejoBR:
Similar to how the BOP endpoint gets built, the IT endpoint needs to strip the schema from the URL in order to be able to build it properly, since otherwise the endpoint throws an exception once it gets used.

## Jira ticket


gwenneg (approved):
LGTM

PR #2190, opened 2023-09-21
[RHCLOUD-28465] Remove old email subscription mechanism

g-duval:


gwenneg:
We'll need to check with QE if this PR will break some of their tests.

database/src/main/resources/db/migration/V1.83.0__RHCLOUD-28465_drop_old_email_subscription_table.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE endpoint_email_subscriptions;
```
gwenneg, line 1: I would wait for at least 2 weeks between the Java code removal and the DB table drop. Just in case we would need to rollback the Java changes for some reason 😃

If you agree with that, please move the SQL file to a different PR that will be merged later.

g-duval: Yes, good point

gwenneg (approved):
LGTM

PR #2197, opened 2023-09-25
RHCLOUD-27242 Move SEDA logic to connector-common module

gwenneg:
After this PR, each connector can be migrated to SEDA separately with almost no effort.


g-duval (commented):
LGTM

PR #2201, opened 2023-09-26
RHCLOUD-28138 | feature: cache the extracted usernames from IT and RBAC calls

MikelAlejoBR:
Calling to IT and RBAC services is a really expensive operation, and usually the responses are quite static. It makes sense to use caching for these usernames in order to avoid having to call for that data each time we need to send an email, specially when there is a huge batch of emails to be sent.

## Jira ticket
[[RHCLOUD-28138]](https://issues.redhat.com/browse/RHCLOUD-28138)

### DIFF
--- a/connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/EmailRouteBuilder.java
+++ b/connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/EmailRouteBuilder.java
@@ -21,8 +21,13 @@ import com.redhat.cloud.notifications.connector.email.processors.rbac.group.RBAC
 import com.redhat.cloud.notifications.connector.email.processors.recipients.RecipientsFilter;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
+import org.apache.camel.Expression;
 import org.apache.camel.builder.PredicateBuilder;
 import org.apache.camel.builder.endpoint.dsl.HttpEndpointBuilderFactory;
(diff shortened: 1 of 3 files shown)


connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/EmailRouteBuilder.java
```diff
@@ -284,6 +330,31 @@ public void configureRoute() throws Exception {
...
+    /**
+     * Sets a common configuration for all the Caffeine caches.
+     */
+    protected void configureCaffeineCache() {
+        final CaffeineCacheComponent caffeineCacheComponent = this.getCamelContext().getComponent("caffeine-cache", CaffeineCacheComponent.class);
+        final CaffeineConfiguration caffeineConfiguration = caffeineCacheComponent.getConfiguration();
+
+        // We explicitly set the eviction type as time based, since the default
+        // value is size based: https://camel.apache.org/components/4.0.x/caffeine-cache-component.html.
+        caffeineConfiguration.setEvictionType(EvictionType.TIME_BASED);
```
gwenneg, line 355: How can we configure the expiration delay of the cache entries? We'll need that in this PR.

gwenneg: Thanks, that's what I was looking for.

gwenneg (approved):
LGTM

PR #2220, opened 2023-10-04
Make users cache expiration delay configurable

gwenneg:


connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/model/Environment.java
gwenneg, line 1: @MikelAlejoBR This class doesn't seem to be needed. Could you please confirm that?

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/config/EmailConnectorConfig.java
```diff
@@ -93,6 +94,9 @@ public class EmailConnectorConfig extends ConnectorConfig {
     @ConfigProperty(name = RBAC_URL)
     String rbacURL;
 
+    @ConfigProperty(name = USER_PROVIDER_CACHE_EXPIRE_AFTER_WRITE, defaultValue = "600")
```
gwenneg, line 97: We're currently caching users for 10 minutes in prod. Let's do the same in the connector.

MikelAlejoBR (approved):
LGTM!

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/model/Environment.java
MikelAlejoBR: This is right, yes. We can delete that class.

PR #2222, opened 2023-10-04
RHCLOUD-28138 | feature: cache the group principals results

MikelAlejoBR:
The group principals results need to be cached as well to avoid calling RBAC too many times.

## Jira ticket
[[RHCLOUD-28138]](https://issues.redhat.com/browse/RHCLOUD-28138)


gwenneg (approved):
LGTM

PR #2229, opened 2023-10-06
RHCLOUD-28681 | fix: cache gets overwritten

MikelAlejoBR:
Basically the users cache was getting overwritten because we were modifying the users' collection directly when filtering it. Summarizing:

1) We created a collection for collecting the users.
2) Cached it.
3) Modified it in the recipients filter.

Since all this was made to the same collection, the cache was modified as a result.

Copying the collections fixes the issue.

## Jira ticket
[[RHCLOUD-28681]](https://issues.redhat.com/browse/RHCLOUD-28681)


gwenneg (approved):
LGTM

PR #2230, opened 2023-10-06
RHCLOUD-28631 | Fix: connector sends duplicate emails

MikelAlejoBR:
The connector was sending duplicate emails because we were processing
the recipient settings in a for loop and sending an email for each one
of them.

However, the legacy implementation reduced all the fetched users from
the different user providers into a single list, which the email
connector didn't do.

The strategy followed for the fix is to split the exchanges and then
aggregate them, which also aggregates the users, producing a single
exchange to take care of, and therefore, a single email if the
configuration option is enabled. If not, we still can send one email per
user, even though that option is deprecated.

## Dependency

* https://github.com/RedHatInsights/notifications-backend/pull/2229

## Jira ticket

[[RHCLOUD-28631]](https://issues.redhat.com/browse/RHCLOUD-28631)


gwenneg:
Rebased onto `master`.

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/EmailRouteBuilder.java
```diff
@@ -147,11 +156,25 @@ public void configureRoute() throws Exception {
 
         from(direct(ENGINE_TO_CONNECTOR))
             .routeId(this.connectorConfig.getConnectorName())
-            /*
-             * Dispatches the exchanges either to fetch the users directly, or, if
-             * a group ID is present, fetch the user from that group from RBAC.
-             */
-            .process(this.dispatcherProcessor);
```
gwenneg, line 154: Can `DispatcherProcessor` be deleted now?

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/EmailRouteBuilder.java
```diff
@@ -147,11 +156,25 @@ public void configureRoute() throws Exception {
...
+                // fetched users from the user providers.
+                .process(exchange -> exchange.setProperty(ExchangeProperty.USERNAMES, new HashSet<String>()))
+                // As the body of the exchange might change throughout the
+                // routes, save it in an exchange property.
+                .setProperty(ExchangeProperty.CURRENT_RECIPIENT_SETTINGS, body())
+                .choice()
+                    .when(simpleF("${exchangeProperty.%s.groupUUID} == null", ExchangeProperty.CURRENT_RECIPIENT_SETTINGS))
+                        .to(direct(Routes.FETCH_USERS))
+                    .otherwise()
+                        .to(direct(Routes.FETCH_GROUP))
```
gwenneg, line 172: Shouldn't `ExchangeProperty.GROUP_UUID` be set before executing this line?

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/EmailRouteBuilder.java
```diff
@@ -147,11 +156,25 @@ public void configureRoute() throws Exception {
...
-             * Dispatches the exchanges either to fetch the users directly, or, if
-             * a group ID is present, fetch the user from that group from RBAC.
-             */
-            .process(this.dispatcherProcessor);
+            // Split each recipient setting and aggregate the usernames to end
+            // up with a single exchange.
+            .split(simpleF("${exchangeProperty.%s}", ExchangeProperty.RECIPIENT_SETTINGS), this.userAggregationStrategy).stopOnException()
+                // Initialize the usernames hash set, where we will gather the
+                // fetched users from the user providers.
+                .process(exchange -> exchange.setProperty(ExchangeProperty.USERNAMES, new HashSet<String>()))
```
gwenneg, line 164: Shouldn't this be done before the `split` block begins?

MikelAlejoBR: I think it doesn't matter:

* If we set the `USERNAMES` property before the split, all the split exchanges will have it when the split block begins.
* The way it is right now, every exchange starts without it but we add it afterwards.

I put it there because the exchange property is only relevant to the split block, but maybe it is a good idea to reduce the operations need to be done.

gwenneg: You're right, it doesn't really matter.

MikelAlejoBR: I've changed it anyways, that way we reduce that operation to being executed once.

connector-email/src/test/java/com/redhat/cloud/notifications/connector/email/EmailRouteBuilderRHCLOUD28631SingleEmailPerUserTest.java
```diff
@@ -0,0 +1,145 @@
...
+        // usernames we are sending.
+        AdviceWith.adviceWith(this.context, Routes.SEND_EMAIL_BOP, a -> {
+            a.weaveByToUri(String.format("%s*", this.emailConnectorConfig.getBopURL())).replace().to("mock:bopendpoint");
+            a.mockEndpointsAndSkip(String.format("direct:%s", ConnectorToEngineRouteBuilder.SUCCESS));
+        });
+
+        // Set up the expected exchanges count that the success endpoint should
+        // receive at the end of the test.
+        final MockEndpoint successEndpoint = this.getMockEndpoint(String.format("mock:direct:%s", ConnectorToEngineRouteBuilder.SUCCESS));
+        successEndpoint.expectedMessageCount(4);
```
gwenneg, line 109: The `engine` should only receive one Kafka message in the `fromcamel` topic, so this should be:
```suggestion
        successEndpoint.expectedMessageCount(1);
```

gwenneg (approved):
LGTM

connector-email/src/test/java/com/redhat/cloud/notifications/connector/email/EmailRouteBuilderRHCLOUD28631SingleEmailPerUserTest.java
MikelAlejoBR: Yes!

PR #2236, opened 2023-10-10
RHCLOUD-28708 | fix: Floorist queries give index error

MikelAlejoBR:
The Floorist queries are giving indexing errors, apparently because some of the columns need to be converted to strings.

## Jira tickets
[[RHCLOUD-28708]](https://issues.redhat.com/browse/RHCLOUD-28708)

### DIFF
--- a/.rhcicd/clowdapp-backend.yaml
+++ b/.rhcicd/clowdapp-backend.yaml
@@ -188,10 +188,10 @@ objects:
     - prefix: insights/notifications/endpoint_types
       query: >-
         SELECT
-          sub."Endpoint type",
+          sub."Endpoint type"::TEXT,
           COUNT(*) AS "Count",
-          sub."Enabled",
-          sub."Actively used"
+          sub."Enabled"::TEXT,
+          sub."Actively used"::TEXT
         FROM
           (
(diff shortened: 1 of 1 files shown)


.rhcicd/clowdapp-backend.yaml
```diff
@@ -231,8 +231,8 @@ objects:
         SELECT
           b.display_name AS "Bundle",
           a.display_name AS "Application",
-          et.display_name AS "Event type",
-          es.subscription_type AS "Subscription type",
+          et.display_name::TEXT AS "Event type",
```
gwenneg, line 234: Why would this display name require a transformation to text and not the previous ones?

MikelAlejoBR: Good question. You are right, let's be safe and double down on casting everything to text, except the counts...

MikelAlejoBR: Yes, yesterday they showed us in the meeting that the counts are getting read properly... So I think it is safe for us to leave them like this.

gwenneg (approved):
LGTM

PR #2251, opened 2023-10-17
Simplify recipients-resolver logic

gwenneg:
- Shared retry policy declaration for all external services
- Shared failure counter var for all external services
- Shared configuration values for all external services
- `User#isActive` is gone
- `User#isAdmin` is primitive now (and renamed)
- ClowdApp template cleanup
- The `test` dependency on `notifications-common` is gone


g-duval (approved):
LGTM

PR #2271, opened 2023-10-24
RHCLOUD-28935 Create dummy templates for OCM Service Logs

gwenneg:
These templates will be used to perform a first end-to-end test.


MikelAlejoBR (approved):
LGTM!

PR #2282, opened 2023-10-26
Use ClowdApp templates from stage in ephemeral

gwenneg:
Until now, whenever we introduced a breaking change in a ClowdApp template and it broke some deployment in ephemeral, we had to deploy that template up to production before ephemeral could be fixed.

Starting now, deploying the new template in stage will be enough to fix ephemeral.


gwenneg:
The PR check tests failures are unrelated and already being fixed by the QE team.

PR #2283, opened 2023-10-26
Check if setuptools is still affected by a CVE

gwenneg:


gwenneg:
The answer to the PR title is yes :cry:

PR #2286, opened 2023-10-27
Introduce SEDA in the email connector

gwenneg:
With SEDA, emails will be processed/sent in parallel, increasing the connector throughput.

This PR does not enable the parallel processing yet though. We'll decide later (from env vars) if and when we want to use SEDA.


MikelAlejoBR (approved):
LGTM!

PR #2287, opened 2023-10-27
Use Maven's --also-make option in Docker files

gwenneg:
When `-am` is used, Maven will automatically build modules that are dependencies of the current module.


gwenneg:
The PR check tests failures are unrelated.

PR #2289, opened 2023-10-30
RHCLOUD-28942 Skip BOP users resolution while sending emails

gwenneg:
I recently modified BOP to make this possible. We'll need to make sure the extra field in the request sent to MBOP doesn't cause any issue with MBOP specifically.

The feature flags from this PR will be enabled when BOP/MBOP are ready in each environment.


MikelAlejoBR (approved):
LGTM!

g-duval (approved):
LGTM, just a minor comment ;)

josejulio:
> We'll need to make sure the extra field in the request sent to MBOP doesn't cause any issue with MBOP specifically

Should we add the same change to MBOP for feature parity ?

gwenneg:
> Should we add the same change to MBOP for feature parity ?

AFAIU, MBOP doesn't require the same change I made in BOP because MBOP won't try to resolve incoming email addresses through the IT Users Service. This needs to be confirmed though.

josejulio:
> AFAIU, MBOP doesn't require the same change I made in BOP because MBOP won't try to resolve incoming email addresses through the IT Users Service. This needs to be confirmed though.

It looks like it tries to resolve emails though: https://github.com/josejulio/mbop/blob/master/internal/handlers/send_email.go#L43

gwenneg:
From what I understand [here](https://github.com/josejulio/mbop/blob/959d00dea3e4f7c3ce20b295ff8f91b3f54ee72e/internal/service/mailer/helpers.go#L17-L39), anything that contains `@` is excluded from the resolution logic.

PR #2290, opened 2023-10-30
RHCLOUD-29084 Replace optIn/optOut with subscribedByDefault

gwenneg:
Although this PR does not contain anything directly related to OCM, this is preparation work for event types subscribed by default.

Most of the changes are related to the `EmailSubscriptionType` -> `SubscriptionType` renaming.


backend/src/main/java/com/redhat/cloud/notifications/routers/models/SettingsValuesByEventType.java
```diff
@@ -31,7 +31,7 @@ public static class ApplicationSettingsValue {
     public static class EventTypeSettingsValue {
         @JsonIgnore
         public String displayName;
-        public Map<EmailSubscriptionType, Boolean> emailSubscriptionTypes = new HashMap<>();
+        public Map<SubscriptionType, Boolean> emailSubscriptionTypes = new HashMap<>();
```
gwenneg, line 34: `emailSubscriptionTypes` can't be renamed without requiring a frontend migration, which I don't want to do now.

PR #2298, opened 2023-10-31
RHCLOUD-29101 Introduce recipients.emails field

gwenneg:
`recipients.emails` will only be used when the email connector is active and `NOTIFICATIONS_CONNECTOR_BOP_SKIP_USERS_RESOLUTION == true`. I'm not updating the old email logic in `engine`.


g-duval (approved):
LGTM, "emails" property may have to be renamed later to avoid misunderstanding ;)

PR #2305, opened 2023-11-02
 RHCLOUD-21638 | feature: include the roles when returning applications

MikelAlejoBR:
The front end requires visualizing the owner roles for the applications. These changes satisfy that requirement.

## Jira ticket.
[[RHCLOUD-21638]](https://issues.redhat.com/browse/RHCLOUD-21638)


backend/src/main/java/com/redhat/cloud/notifications/models/InternalRoleAccess.java
```diff
@@ -93,7 +93,9 @@ public boolean equals(Object o) {
         }
         if (o instanceof InternalRoleAccess) {
             InternalRoleAccess other = (InternalRoleAccess) o;
-            return Objects.equals(id, other.id);
+            return Objects.equals(id, other.id)
```
josejulio, line 96: Usually comparing only the id should be enough, it shouldn't be possible for 2 objects with the same id to be different

MikelAlejoBR: All right, reverted back! Thanks!

josejulio (approved):
lgtm

PR #2317, opened 2023-11-07
Fix one test instability when nanos ends with 0

g-duval:
Default LocalDateTime.toString() method keeps last digit when it ends with 0, but DateTimeFormatter.ISO_DATE_TIME (used on DTO converter) don't.


MikelAlejoBR (approved):
LGTM!

PR #2319, opened 2023-11-07
RHCLOUD-29084 Get rid of isOptIn for good

gwenneg:
- `isOptIn` is now gone and replaced with `subscribedByDefault` everywhere
- recipients-resolver queries are simpler (getters, setters, equals and hashCode were not really needed)
- recipients-resolver queries are now serialized using snake case

⚠️ Keep in mind while reviewing that `isOptIn == !subscribedByDefault`.

### DIFF
--- a/recipients-resolver/src/test/java/com/redhat/cloud/notifications/recipients/resolver/RecipientsResolverTest.java
+++ b/recipients-resolver/src/test/java/com/redhat/cloud/notifications/recipients/resolver/RecipientsResolverTest.java
@@ -95,13 +95,13 @@ public class RecipientsResolverTest {
 
     @Test
     public void withPersonalizedEmailOn() {
-        boolean isOptIn = true;
+        boolean subscribedByDefault = false;
         // Default request, all subscribed users
         List<User> users = recipientsResolver.findRecipients(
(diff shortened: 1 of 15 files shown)


recipients-resolver/src/test/java/com/redhat/cloud/notifications/recipients/resolver/RecipientsResolverTest.java
```diff
@@ -357,7 +357,7 @@ public void withPersonalizedEmailOn() {
 
     @Test
     public void withPersonalizedEmailOnOptOut() {
```
g-duval, line 359:
```suggestion
    public void withPersonalizedEmailOnSubscribedByDefault() {
```

g-duval (approved):
!!LGTM

recipients-resolver/src/test/java/com/redhat/cloud/notifications/recipients/resolver/RecipientsResolverTest.java
gwenneg: I'll fix that in the next PR!

PR #2324, opened 2023-11-08
Manually init DB data

gwenneg:
This PR will not be merged as is.


gwenneg:
This was needed for a special environment.

PR #2329, opened 2023-11-09
RHCLOUD-29294 Setup emails for RBAC request-access event type

josejulio:


josejulio:
@g-duval Do i need to manually setup these templates on the admin-ui?

g-duval:
Nop you don't need to setup manually the template from admin UI.
Because template migration step runs at every startup, it will ba added automaticaly on next deployment.
Ths template looks fine good to me, could you please consider to add a minimal test case for this new one on [TestRbacTemplate](https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/test/java/com/redhat/cloud/notifications/templates/TestRbacTemplate.java|TestRbacTemplate) ?
The purpose of this test is just to guarantee template rendering ;)

josejulio:
@g-duval Updated the tests. Moved a bit the setup to prevent repeating the event types multiple times

g-duval (approved):
LGTM

PR #2348, opened 2023-11-17
RHCLOUD-24734 | feature: disable unmigratable endpoints

MikelAlejoBR:
The team decided that we are going to run the migrations again but disabling endpoints this time, so that if clients want to keep using the, what now are invalid endpoints, will have to enable them manually and correct the bits that are incorrect.

## Jira ticket
[[RHCLOUD-24734]](https://issues.redhat.com/browse/RHCLOUD-24734)

PR #2350, opened 2023-11-17
RHCLOUD-29443 | Remove the secrets' migration code and the feature flag

MikelAlejoBR:
After migrating all the secrets to sources, we decided that it was time
to remove both the database columns for the secrets and the feature
flag, since storing secrets in Sources is going to be enabled by
default.

## Dependencies
* https://github.com/RedHatInsights/notifications-backend/pull/2348

## Jira ticket
[[RHCLOUD-29443]](https://issues.redhat.com/browse/RHCLOUD-29443)


database/src/main/resources/db/migration/V1.89.0__RHCLOUD-29443_remove_secrets_columns.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE
```
g-duval, line 1: I think we should move this Drop column script to another PR, and merge it a little bit after, just in case we have to rollback the prod delivery for any reason.

MikelAlejoBR: You are right. I have removed the migration but we have to coordinate to merge this one, because I need to run the migrations first before removing all the code!

gwenneg:
Where are we on this topic?

- Are we done with the migration of the DB data in all environments?
- Does this PR needs a rebase/update or can it be reviewed/merged as is?

MikelAlejoBR:
> Where are we on this topic?
> 
>     * Are we done with the migration of the DB data in all environments?

Not yet. My idea was to coordinate the data migration with the merge of this PR, and then do a release to stage/prod. So I would love to do that early on the week. 

>     * Does this PR needs a rebase/update or can it be reviewed/merged as is?

I think it can be reviewed because I did put the schema migration script on a different PR.

gwenneg:
Perfect, let's do that next week! Thanks for the update.

engine/src/test/java/com/redhat/cloud/notifications/events/LifecycleITest.java
gwenneg, line 1: Sources is no longer available in the engine. All changes made in this class need to be reverted.

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
```diff
@@ -1599,6 +1719,23 @@ void testConnectionCount() {
             ep.setEnabled(true);
             ep.setProperties(properties);
 
+            // Mock the Sources service calls.
+            final Secret secretTokenSecret = new Secret();
+            secretTokenSecret.id = new Random().nextLong(1, Long.MAX_VALUE);
+            secretTokenSecret.password = properties.getSecretToken();
```
gwenneg, line 1725: This and other changes from this PR are repeated several times. Could you please lower the amount of duplicated code?

gwenneg (approved):
LGTM

backend/src/test/java/com/redhat/cloud/notifications/routers/EndpointResourceTest.java
MikelAlejoBR: Thank you for taking care of this. I was finishing some work for the service accounts that got in the way and was planning to review this today.

PR #2381, opened 2023-12-06
[RHCLOUD-28935] Ocm email templates, use subjet value as title

g-duval:
Ocm email templates, use subjet value as title if title property is empty


gwenneg (approved):
LGTM

PR #2382, opened 2023-12-06
[RHCLOUD-29882] manage recipients list with email on failure

g-duval:


connector-common/src/main/java/com/redhat/cloud/notifications/connector/EngineToConnectorRouteBuilder.java
```diff
@@ -40,17 +43,26 @@ public void configure() throws Exception {
...
         onException(Throwable.class)
                 .onWhen(redeliveryPredicate::matches)
-                .handled(true)
+                .onWhen(continueOnErrorPredicate::matches)
+                .maximumRedeliveries(connectorConfig.getRedeliveryMaxAttempts())
+                .redeliveryDelay(connectorConfig.getRedeliveryDelay())
+                .retryAttemptedLogLevel(DEBUG)
+                .onRedelivery(redeliveryCounterProcessor)
+                .process(exceptionProcessor)
+                .continued(true);
```
g-duval, line 52: Camel `.split` usage creates a flow for each element of split loop.
If we want to be able to aggregate all results of a split loop execution, we have to go back and continue the main process even if an exception is throwed.

g-duval:
replaced by https://github.com/RedHatInsights/notifications-backend/pull/2413

PR #2397, opened 2023-12-14
RHCLOUD-29996 Introduce EventType#subscriptionLocked

gwenneg:


g-duval (approved):
LGTM

backend/src/main/java/com/redhat/cloud/notifications/routers/models/SettingsValueByEventTypeJsonForm.java
```diff
@@ -33,6 +33,7 @@ public static class Field {
         public String checkedWarning;
         @JsonInclude(Include.NON_NULL)
         public String infoMessage;
+        public boolean disabled;
```
MikelAlejoBR, line 36: Why is this field named disabled?

MikelAlejoBR (approved):
LGTM!

backend/src/main/java/com/redhat/cloud/notifications/routers/models/SettingsValueByEventTypeJsonForm.java
MikelAlejoBR: I mean, I know that the settings values code is complex but

gwenneg: This field will be used in the frontend to disable an HTML component (checkbox).

MikelAlejoBR: Ah, and I guess it has to be named like that. All right! Thanks!

gwenneg: I have no idea how it will be used in the frontend 😅 but `disabled` didn't sound like a bad name so I went with that one. We can easily change it to anything later.

PR #2403, opened 2023-12-20
RHCLOUD-30039 | fix: Temporarily disable recipients-resolver certificate validation

MikelAlejoBR:
Apparently BOP is contacting using the HTTPS protocol, and we need to support building a secured endpoint in Camel for that.

## Jira ticket
[[RHCLOUD-30039]](https://issues.redhat.com/browse/RHCLOUD-30039)


gwenneg:
Based on what we saw yesterday with Jacob, I suspect the `SSLHandshakeException` was thrown when the email connector tried to call `recipients-resolver` with the HTTPS scheme. If that is correct, we need to load the certificate Jacob showed us and use it for `recipients-resolver`. BOP/MBOP shouldn't be an issue for now and therefore shouldn't require any changes in the immediate future.

gwenneg:
The `SSLHandshakeException` was caused by an HTTPS call from `connector-email` to `recipients-resolver`. This PR works around the issue by disabling the `recipients-resolver` certificate check as a temporary measure. Then, in a subsequent PR we'll use a certificate provided by Clowder to load a certification authority that makes the `recipients-resolver` certificate valid, and we'll re-enable the certificate validation.

gwenneg (approved):

PR #2413, opened 2024-01-09
[RHCLOUD-29882] manage partial email reinjection

g-duval:
[Feature RHCLOUD-28888](https://github.com/RedHatInsights/notifications-backend/pull/2372) have to be merged first


connector-common/src/main/java/com/redhat/cloud/notifications/connector/EngineToConnectorRouteBuilder.java
```diff
@@ -55,17 +58,26 @@ public void configure() throws Exception {
...
         onException(Throwable.class)
                 .onWhen(redeliveryPredicate::matches)
-                .handled(true)
+                .onWhen(continueOnErrorPredicate::matches)
+                .maximumRedeliveries(connectorConfig.getRedeliveryMaxAttempts())
+                .redeliveryDelay(connectorConfig.getRedeliveryDelay())
+                .retryAttemptedLogLevel(DEBUG)
+                .onRedelivery(redeliveryCounterProcessor)
+                .process(exceptionProcessor)
+                .continued(true);
```
g-duval, line 67: Camel `.split` usage creates a flow for each element of split loop.
If we want to be able to aggregate all results of a split loop execution, we have to go back and continue the main process even if an exception is throwed.

connector-common/src/main/java/com/redhat/cloud/notifications/connector/EngineToConnectorRouteBuilder.java
```diff
@@ -55,17 +58,26 @@ public void configure() throws Exception {
...
                 .maximumRedeliveries(connectorConfig.getRedeliveryMaxAttempts())
                 .redeliveryDelay(connectorConfig.getRedeliveryDelay())
                 .retryAttemptedLogLevel(DEBUG)
                 .onRedelivery(redeliveryCounterProcessor)
-                .process(exceptionProcessor);
+                .process(exceptionProcessor)
+                .handled(true);
 
         onException(Throwable.class)
-                .onWhen(redeliveryPredicate::doesNotMatch)
```
g-duval, line 66: According https://camel.apache.org/manual/exception-clause.html#_using_fine_grained_selection_using_onwhen_predicate we don't need the `doesNotMatch` predicate, because the `OnException` without `onWhen` condition will be considered only if other `OnException` with `onWhen` conditions failed

PR #2416, opened 2024-01-10
[RHCLOUD-30263] Add support of Service Account auth type

g-duval:


MikelAlejoBR (commented):
Looking good so far!

backend/src/main/java/com/redhat/cloud/notifications/auth/principal/rhid/RhServiceAccountIdentity.java
```diff
@@ -0,0 +1,41 @@
+package com.redhat.cloud.notifications.auth.principal.rhid;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class RhServiceAccountIdentity extends RhIdentity {
```
gwenneg, line 7: Does `ConsoleIdentityProvider` work out of the box with this class?
https://github.com/RedHatInsights/notifications-backend/blob/ddfb7dffa9979ac027f8ac6d2a62d0bf56a048d0/backend/src/main/java/com/redhat/cloud/notifications/auth/ConsoleIdentityProvider.java#L122-L123

backend/src/main/java/com/redhat/cloud/notifications/auth/principal/rhid/RhServiceAccountIdentity.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public String getOrgId() {
+        return orgId;
+    }
+
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    public static class ServiceAccount {
+        @JsonProperty("username")
+        private String username;
```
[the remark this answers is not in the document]
g-duval: It's more the service account owner like `RhUserIdentity.user.username`

backend/src/main/java/com/redhat/cloud/notifications/auth/principal/rhid/RhServiceAccountIdentity.java
g-duval: Yes it have to according the ADR

backend/src/main/java/com/redhat/cloud/notifications/auth/principal/rhid/RhIdentity.java
```diff
@@ -1,44 +1,12 @@
...
-
-        public String getUsername() {
-            return username;
-        }
-    }
-
-    @Override
-    public String getName() {
-        return getUser().getUsername();
+        return null;
```
gwenneg, line 10: Is that method really needed in this abstract class?

backend/src/main/java/com/redhat/cloud/notifications/auth/principal/rhid/RhServiceAccountIdentity.java
```diff
@@ -0,0 +1,41 @@
...
+        private String clientId;
+
+        public String getUsername() {
+            return username;
+        }
+    }
+
+    @Override
+    public String getName() {
+        return getServiceAccount().getUsername();
```
gwenneg, line 39: The `username` of a service account looks like metadata to me, not something we'd use in the `ConsolePrincipal`. Should this return `getServiceAccount().getClientId()` instead?

backend/src/main/java/com/redhat/cloud/notifications/auth/principal/rhid/RhIdentity.java
g-duval: Yes because account number it's still use on some table on db (we insert it but we never read it).
I think it should be remove on another ticket, with all remaining code about account number (RHCLOUD-30644).

backend/src/main/java/com/redhat/cloud/notifications/routers/UserConfigResource.java
```diff
@@ -132,6 +135,7 @@ public Response saveSettingsByEventType(@Context SecurityContext sec, @NotNull @
     @Produces(APPLICATION_JSON)
     @Tag(name = OApiFilter.PRIVATE)
     public Response getSettingsSchemaByEventType(@Context SecurityContext sec) {
+        forbidAccessInCaseOfServiceAccountAuthentication(sec);
```
gwenneg, line 138: Shouldn't the other `GET` method (line 152) be forbidden as well with service accounts?

backend/src/main/java/com/redhat/cloud/notifications/routers/UserConfigResource.java
```diff
@@ -267,4 +271,10 @@ private SettingsValuesByEventType getSettingsValueForUserByEventType(List<EventT
         patchWithUserPreferencesIfExists(settingsValues, emailSubscriptions);
         return settingsValues;
     }
+
+    private static void forbidAccessInCaseOfServiceAccountAuthentication(SecurityContext sec) {
+        if (isServiceAccountAuthentication(sec)) {
+            throw new ForbiddenException("This api can't be used form a service account authentication");
```
gwenneg, line 277: There's a typo in `from` but since it'll be fixed, I would use `with` instead 😄 
```suggestion
            throw new ForbiddenException("This api can't be used with a service account authentication");
```

backend/src/main/java/com/redhat/cloud/notifications/routers/UserPreferencesForPolicy.java
```diff
@@ -43,6 +44,9 @@ public static class V1 extends UserPreferencesForPolicy {
     @Deprecated
     public UserConfigPreferences getPreferences(
         @Context SecurityContext sec, @PathParam("bundleName") String bundleName, @PathParam("applicationName") String applicationName) {
+        if (isServiceAccountAuthentication(sec)) {
+            throw new ForbiddenException("This api can't be used form a service account authentication");
```
gwenneg, line 48: There's a typo in `from` but since it'll be fixed, I would use `with` instead 😄 
```suggestion
            throw new ForbiddenException("This api can't be used with a service account authentication");
```

backend/src/main/java/com/redhat/cloud/notifications/routers/UserConfigResource.java
g-duval: Sure, good catch

gwenneg (approved):
LGTM

PR #2423, opened 2024-01-11
[RHCLOUD-30332] add internal backend endpoint

g-duval:


g-duval:
It requires a JDK

docker/Dockerfile.notifications-backend.jvm
```diff
@@ -17,6 +17,7 @@ USER root
 RUN microdnf update --refresh --nodocs && microdnf clean all
 # Temporary workaround - setuptools is affected by a CVE and it is not needed to run our app which is why we're removing it from the base image
 RUN rpm -e --nodeps platform-python-setuptools
+RUN rpm -e --nodeps platform-python-pip
```
gwenneg, line 20: Is that supposed to be in this PR?

g-duval: You are right, it shouldn't

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/HeapDumpResource.java
```diff
@@ -0,0 +1,63 @@
...
+
+@Path(API_INTERNAL + "/heap_dump")
+@RolesAllowed(RBAC_INTERNAL_ADMIN)
+public class HeapDumpResource {
+
+    @ConfigProperty(name = "quarkus.log.cloudwatch.log-stream-name", defaultValue = "localhost")
+    String hostname;
+
+    @GET
+    @Path("{hostname}")
```
gwenneg, line 30: Shouldn't this start with a `/`?
```suggestion
    @Path("/{hostname}")
```

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/HeapDumpResource.java
```diff
@@ -0,0 +1,63 @@
...
+    @Produces(APPLICATION_OCTET_STREAM)
+    public Response getHeapDump(@PathParam("hostname") String hostname) {
```
gwenneg, line 32:
```suggestion
    public Response getHeapDump(@RestPath String hostname) {
```

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/HeapDumpResource.java
```diff
@@ -0,0 +1,63 @@
...
+        if (!this.hostname.equals(hostname)) {
+            Log.infof("%s doesn't match with this host (%s)", hostname, this.hostname);
+            return Response.status(HttpStatus.SC_SEE_OTHER).build();
+        }
+
+        String tmpFileName = "notifications_dump.hprof";
+        String fullPathTmpFile = String.format("%s/%s", System.getProperty("java.io.tmpdir"), tmpFileName);
```
gwenneg, line 39: We could do this the [clowder-quarkus-config-source](https://github.com/RedHatInsights/clowder-quarkus-config-source/blob/498a0b1ebac6ac5a6524892ef09b3ffe1a9029d1/src/main/java/com/redhat/cloud/common/clowder/configsource/ClowderConfigSource.java#L645-L653) way:
```java
File file = File.createTempFile("notifications_dump", ".hprof");
try {
    file.deleteOnExit();
} catch (SecurityException e) {
    log.warn("Delete on exit of the notifications heap dump file denied by the security manager", e);
}
```

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/HeapDumpResource.java
```diff
@@ -0,0 +1,63 @@
...
+        response.header("Content-Disposition", "attachment; filename=\"" + tmpFileName + "\"");
+        return response.build();
+    }
+
+    public static void dumpHeap(String filePath, boolean live) throws IOException {
+        MBeanServer server = ManagementFactory.getPlatformMBeanServer();
+        HotSpotDiagnosticMXBean mxBean = ManagementFactory.newPlatformMXBeanProxy(
+            server, "com.sun.management:type=HotSpotDiagnostic", HotSpotDiagnosticMXBean.class);
+        mxBean.dumpHeap(filePath, live);
+    }
```
gwenneg, line 62: `live` could be removed:
```suggestion
    public static void dumpHeap(String filePath) throws IOException {
        MBeanServer server = ManagementFactory.getPlatformMBeanServer();
        HotSpotDiagnosticMXBean mxBean = ManagementFactory.newPlatformMXBeanProxy(
            server, "com.sun.management:type=HotSpotDiagnostic", HotSpotDiagnosticMXBean.class);
        mxBean.dumpHeap(filePath, true);
    }
```

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/HeapDumpResource.java
g-duval: Not necessary, it works with or without it, and because they didn't added this `/` en Quarkus doc samples, I didn't added it here.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/HeapDumpResource.java
g-duval: `deleteOnExit()` will delete file when the application will terminate, not only the method.
The idea is to be able to generate sevral heap dumps to check memory increasing before backends reboots.

gwenneg (approved):
LGTM

PR #2447, opened 2024-01-31
RHCLOUD-27431 | refactor: don't return secrets on responses | Endpoints and service implementation

MikelAlejoBR:
## Summary
This second part focuses on implementing the changes in order to support the new endpoints for the secrets' management, and the required functions in the `SecretUtils` class.

## Depends on

* https://github.com/RedHatInsights/notifications-backend/pull/2448

## Description
We are making an effort to stop returning secrets in responses, and at
the same time, signal if they are present or not, so that the UI can
effectively show feedback about that.
    
Since we need a transition period until the UI moves to the new way of
managing secrets, we still support both ways of managing them.

## Jira ticket
[[RHCLOUD-27431]](https://issues.redhat.com/browse/RHCLOUD-27431)


MikelAlejoBR:
Closing because we decided we are going to follow the DTO approach and make our lives easier.

PR #2454, opened 2024-02-05
RHCLOUD-27431 | refactor: return credentials redacted

MikelAlejoBR:
We are moving towards not returning the credentials to the users, and the first step we want to make is to simply remove them from the payloads we return when listing the endpoints.

## Jira ticket
RHCLOUD-27431

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
@@ -8,6 +8,7 @@ import com.redhat.cloud.notifications.config.FeatureFlipper;
 import com.redhat.cloud.notifications.db.Query;
 import com.redhat.cloud.notifications.db.repositories.EndpointRepository;
 import com.redhat.cloud.notifications.db.repositories.NotificationRepository;
+import com.redhat.cloud.notifications.models.BasicAuthentication;
 import com.redhat.cloud.notifications.models.CamelProperties;
 import com.redhat.cloud.notifications.models.CompositeEndpointType;
 import com.redhat.cloud.notifications.models.Endpoint;
@@ -286,7 +287,12 @@ public class EndpointResource {
             this.secretUtils.createSecretsForEndpoint(endpoint);
         }
 
-        return endpointRepository.createEndpoint(endpoint);
+        final Endpoint createdEndpoint = this.endpointRepository.createEndpoint(endpoint);
+        // Redact all the secrets for the given endpoint so that we don't
+        // return them to the clients.
+        this.redactSecretsForEndpoint(endpoint);
+
+        return createdEndpoint;
     }
 
     private String checkSlackChannel(CamelProperties camelProperties) {
@@ -364,6 +370,9 @@ public class EndpointResource {
             if (this.featureFlipper.isSourcesUsedAsSecretsBackend()) {
                 this.secretUtils.loadSecretsForEndpoint(endpoint);
             }
+
+            // Redact all the credentials from the endpoint's properties.
+            this.redactSecretsForEndpoint(endpoint);
 
             return endpoint;
         }
@@ -480,6 +489,10 @@ public class EndpointResource {
             }
         }
 
+        // Redact all the secrets for the given endpoint so that we don't
+        // return them to the clients.
+        this.redactSecretsForEndpoint(endpoint);
+
         return Response.ok().build();
     }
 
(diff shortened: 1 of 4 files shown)


backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -551,4 +564,30 @@ private static void checkSystemEndpoint(EndpointType endpointType) {
...
         return !featureFlipper.isEmailsOnlyMode() || endpointType.isSystemEndpointType;
     }
+
+    /**
+     * Removes the secrets from the endpoint's properties when returning them
+     * to the client.
+     * @param endpoint the endpoint to redact the secrets from.
+     */
+    @Deprecated(forRemoval = true)
+    protected void redactSecretsForEndpoint(final Endpoint endpoint) {
```
g-duval, line 574: It should be applyed only to user without write access, to keep existing behavior for org admins

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -551,4 +564,30 @@ private static void checkSystemEndpoint(EndpointType endpointType) {
...
+        if (endpoint.getProperties() instanceof SourcesSecretable sourcesSecretable) {
+            final BasicAuthentication basicAuthentication = sourcesSecretable.getBasicAuthentication();
+            if (basicAuthentication != null) {
+                basicAuthentication.setPassword("[REDACTED CREDENTIAL]");
```
g-duval, line 578: Returning string with `[ ]` looks like some kind of missing placeholder transformation to me. On this case I would prefer sending back sting like `********`, because it's a standard behavior to hide secrets.
But that's my personnal choice :), WDYT ?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -551,4 +564,30 @@ private static void checkSystemEndpoint(EndpointType endpointType) {
...
+            }
+
+            final String bearerToken = sourcesSecretable.getBearerAuthentication();
+            if (bearerToken != null) {
+                sourcesSecretable.setBearerAuthentication("[REDACTED CREDENTIAL]");
+            }
+
+            final String secretToken = sourcesSecretable.getSecretToken();
+            if (secretToken != null) {
+                sourcesSecretable.setSecretToken("[REDACTED CREDENTIAL]");
```
g-duval, line 589: replacement string should be a constant.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
MikelAlejoBR: Why, though? We are not going to allow even org admins to see the secrets, no?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
MikelAlejoBR: You are right, yes, the asterisks are kind of universal for redacted secrets.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
MikelAlejoBR: Done!

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
g-duval: Yes for Org admin on V2, but I understood we would like to keep the existing behavior on current version ?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -551,4 +560,36 @@ private static void checkSystemEndpoint(EndpointType endpointType) {
...
+     * to the client.
+     * @param endpoint the endpoint to redact the secrets from.
+     */
+    @Deprecated(forRemoval = true)
+    protected void redactSecretsForEndpoint(final SecurityContext securityContext, final Endpoint endpoint) {
+        // Only redact the secrets for those users who have read only permissions.
+        if (securityContext.isUserInRole(ConsoleIdentityProvider.RBAC_READ_INTEGRATIONS_ENDPOINTS)
+            || securityContext.isUserInRole(ConsoleIdentityProvider.RBAC_READ_NOTIFICATIONS)
+            || securityContext.isUserInRole(ConsoleIdentityProvider.RBAC_READ_NOTIFICATIONS_EVENTS)
+        ) {
```
gwenneg, line 575: Shouldn't this be
```suggestion
        if (!securityContext.isUserInRole(ConsoleIdentityProvider.RBAC_WRITE_INTEGRATIONS_ENDPOINTS)) {
```
?

gwenneg (approved):
LGTM

g-duval (approved):
LGTM

PR #2476, opened 2024-02-26
RHCLOUD-29443 | feature: remove the bearer authentication column from the database

MikelAlejoBR:
After marking the "bearer token" property as "@Transient", and deploying that change, it is time to remove the column from the database.

## Jira ticket
[[RHCLOUD-29443]](https://issues.redhat.com/browse/RHCLOUD-29443)

### DIFF
--- a/database/src/main/resources/db/migration/V1.96.0__RHCLOUD-29443_remove_secrets_columns.sql
+++ b/database/src/main/resources/db/migration/V1.96.0__RHCLOUD-29443_remove_secrets_columns.sql
@@ -1,0 +1,4 @@
+ALTER TABLE
+    "endpoint_webhooks"
+DROP COLUMN
+    "bearer_authentication";


database/src/main/resources/db/migration/V1.95.0__RHCLOUD-29443_remove_secrets_columns.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE
+    "camel_properties"
+DROP COLUMN
+    "bearer_authentication";
+
```
gwenneg, line 5: I don't think the column exists in `camel_properties`.
```suggestion
```

MikelAlejoBR: You are right, yes... Thank you!

gwenneg:
@MikelAlejoBR `V1.95.0` is used already. Could you please increase the version to 96?

MikelAlejoBR:
But of course!

gwenneg (approved):
LGTM

PR #2482, opened 2024-02-28
RHCLOUD-28460 Move engine feature flags to Unleash

gwenneg:
This PR moves most of the `engine` feature flags to Unleash. Some flags are not moved because they're no longer used in any environment.

We'll transition from env vars to Unleash in several steps because some special environments may not support Unleash yet. That's why `FeatureFlipper` is still present and, for now, enabled by default.

Unleash is not integrated in the unit tests yet. This will come in a subsequent PR.

### DIFF
--- a/.rhcicd/clowdapp-engine.yaml
+++ b/.rhcicd/clowdapp-engine.yaml
@@ -236,6 +236,10 @@ objects:
           value: ${NOTIFICATIONS_DRAWER_CONNECTOR_ENABLED}
         - name: NOTIFICATIONS_ASYNC_AGGREGATION_ENABLED
           value: ${NOTIFICATIONS_ASYNC_AGGREGATION_ENABLED}
+        - name: NOTIFICATIONS_UNLEASH_ENABLED
+          value: ${NOTIFICATIONS_UNLEASH_ENABLED}
+        - name: QUARKUS_UNLEASH_ACTIVE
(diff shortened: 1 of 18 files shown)


MikelAlejoBR:
I have found a couple more occurrences in the `KafkaCOnsumedTotalChecker` and the `RbacRecipientUsersProvider` classes, though I guess those are the ones you were referring to when you said:

> Some flags are not moved because they're no longer used in any environment.

I'll leave them here just in case.

https://github.com/RedHatInsights/notifications-backend/blob/65ebf9a924977186f64c954e703aebc675e28f89/engine/src/main/java/com/redhat/cloud/notifications/health/KafkaConsumedTotalChecker.java#L28-L38

https://github.com/RedHatInsights/notifications-backend/blob/65ebf9a924977186f64c954e703aebc675e28f89/engine/src/main/java/com/redhat/cloud/notifications/recipients/rbac/RbacRecipientUsersProvider.java#L174-L177

MikelAlejoBR (approved):
LGTM

gwenneg:
Thanks for double checking @MikelAlejoBR!

There are indeed several occurrences of `FeatureFlipper#set...` in the code. I'm keeping them as is for now because there won't be any setters once we switch to Unleash. The setters will all go away soon.

Regarding `featureFlipper.isUseRbacForFetchingUsers()`, it's part of a dead code section so I didn't migrate it.

g-duval (approved):
LGTM

PR #2483, opened 2024-02-28
RHCLOUD-31231 | feature: filter internal RH accounts from results

MikelAlejoBR:
The change has removing internal accounts from results as the goal.

## Jira ticket
[[RHCLOUD-31231]](https://issues.redhat.com/browse/RHCLOUD-31231)


g-duval (approved):
LGTM

MikelAlejoBR:
Thanks for the review. I've tested it with `bonfire` and the variable is replaced correctly. I'll set up the org IDs to filter in AppInterface and we should be golden.

PR #2496, opened 2024-03-05
feature: clean up Sources secrets when endpoint creation fails

MikelAlejoBR:
Until now, when the endpoint creation failed due to an internal error,
which wasn't very common, that endpoint's secrets were left in Sources
and were not cleaned up.

This happened because our endpoint creation resource handler was
annotated with the "@transactional" method, and there was no exception
handling that would catch any errors and would try to delete the
"non-saved" endpoint's secrets from Sources.

Therefore, firstly the secrets would be created, and then we would
attempt to save the endpoint in the database. So if any exception was
thrown in the process, those secrets would remain undeleted in Sources.

The introduced changes take care of this problem by making the "endpoint
creation" function call an internal endpoint creation function instead,
which gets wrapped in a "try/catch" block and allows us to clean up
Sources secrets when saving the endpoint fails. By applying the
"@transactional" method to the internal function, no current behavior is
modified, and on the flip side we gain the clean up functionality.

## Jira ticket
[[RHCLOUD-22971]](https://issues.redhat.com/browse/RHCLOUD-22971)


gwenneg (approved):
LGTM

PR #2525, opened 2024-03-12
 RHCLOUD-22071 | feature(query): count the number of events and group them

MikelAlejoBR:
Counts the number of events per bundle, application and their status, and also adds another query to group them by organization ID too.

## Jira ticket
[[RHCLOUD-22071]](https://issues.redhat.com/browse/RHCLOUD-22071)

### DIFF
--- a/.rhcicd/floorist/floorplan.yaml
+++ b/.rhcicd/floorist/floorplan.yaml
@@ -236,6 +236,32 @@ objects:
             et.display_name,
             es.subscription_type,
             es.subscribed
+      # Count the number of events grouped by bundle, application, status and
+      # the organization ID.
+      - prefix: insights/notifications/events_by_bundle_app_status_org_id
+        query: >-
+          SELECT
+            bun.display_name::TEXT AS bundle,
+            apps.display_name::TEXT AS application,
+            e.org_id::TEXT,
+            nh.status::TEXT,
+            count(nh.*)
+          FROM
+            notification_history AS nh
+          INNER JOIN
+            "event" AS e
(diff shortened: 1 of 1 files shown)


.rhcicd/floorist/floorplan.yaml
```diff
@@ -165,6 +165,57 @@ objects:
             a.display_name,
             et.display_name,
             es.subscription_type;
+      # Count the number of events grouped by bundle, application and status.
+      - prefix: insights/notifications/events_by_bundle_app_status
```
g-duval, line 169: If an event triggers an email and a webhook action, it will be counted twice, is it the expected behaviour ?

MikelAlejoBR: I was considering counting all the successful, unsuccessful and whatever deliveries, yes, because then we need to perform another query to just count the incoming events.

Like, this one counts every delivery, and the other one will count the volume of the incoming events per app, so that they can compare both.

Thank you for your comment and the review :+1:

g-duval: Ok fine for me then

g-duval (approved):
LGTM

MikelAlejoBR:
I am closing this due to the recent discussions we've had with the Business Insights team. We don't need to send aggregated results because that way they can see to which org id every element belongs and then filter it out if they are internal org IDs.

MikelAlejoBR:
Removed the total aggregated results and the filter by organization ID, since the Business Insights people will be the ones taking care of the filtering.

PR #2567, opened 2024-03-19
Do NOT disable faulty endpoints by default

gwenneg:


g-duval (commented):
LGTM, for the record :)

PR #2595, opened 2024-03-27
RHCLOUD-31648 | remove org id filtering

MikelAlejoBR:
Removes the org ID filtering because the Business Insights team needs to have them in order to filter the internal ones from the results.

It also sorts the query names by alphabetical order to make it easier to locate them.

## Depends on
* https://github.com/RedHatInsights/notifications-backend/pull/2525

## Jira ticket
[[RHCLOUD-31648]](https://issues.redhat.com/browse/RHCLOUD-31648)

### DIFF
--- a/.rhcicd/floorist/floorplan.yaml
+++ b/.rhcicd/floorist/floorplan.yaml
@@ -54,6 +54,37 @@ objects:
             et.display_name,
             bg.org_id,
             actively_used
+        # Count the number of email subscriptions, and group them by
+        # application, subscription type, whether they are subscribed or not
+        # and their org id.
+      - prefix: insights/notifications/email_subscriptions_by_org_id
+        query: >-
+          SELECT
+            bun.display_name::TEXT,
+            apps.display_name::TEXT,
+            es.org_id::TEXT,
+            et.display_name::TEXT,
+            es.subscription_type::TEXT,
+            es.subscribed::BOOLEAN,
+            count(es.*)::INTEGER AS "count"
+          FROM
+            email_subscriptions AS es
+          INNER JOIN
+            event_type AS et
+              ON et.id = es.event_type_id
+          INNER JOIN
+            applications AS apps
+              ON apps.id = et.application_id
+          INNER JOIN
+            bundles AS bun
+              ON bun.id = apps.bundle_id
+          GROUP BY
+            bun.display_name,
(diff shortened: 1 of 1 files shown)


.rhcicd/floorist/floorplan.yaml
```diff
@@ -38,107 +38,55 @@ objects:
...
+            bg.org_id,
             actively_used
-      # Count the number of behavior groups associated to event types per
-      # organization, and also show the bundle and the application they are
-      # associated to.
-      - prefix: insights/notifications/behavior_groups_event_types_org_id
+        # Count the number of email subscriptions, and group them by
+        # application, subscription type, whether they are subscribed or not
+        # and their org id.
+      - prefix: insights/notifications/email_subscriptions_by_org_id
```
g-duval, line 60: Email subscription by default from event type doesn't seems covered here, is it on purpose ?

MikelAlejoBR: I just moved one of the queries but you might have spotted something that I missed. I don't follow you, though. What do you mean by the email subscription by default? Are you referring to `endpoint_email_subscriptions`, maybe? :confused: 

Thanks!

g-duval: I'm refering to column `subscribed_by_default` from db table `event_type` that we use for OCM.
If this request should reflect email subscriptions **choosen** by users, then it's correct.

MikelAlejoBR: Hmmmmm... so we would have to check the notification history for those events that have an event type which is subscribed by default, no? However wouldn't that give us some duplicates?

Like if I am a new user, I will be subscribed by default to event A which has that option switched on. Therefore, there won't be a record in the `email_subscriptions` table, right?

What if that user disables that subscription manually? I guess we will create an unsubscription record. But if the user turns it back on... Do we actually remove the unsubscription?

g-duval: In a second thought, events subscribed by default can't be only our database data.
Because unless users unsubscribed from this event, it will be sent to all members of an org, even if they never went to Notifications.
Building those statistics will require to fetch all users of all orgs then removing users how unsubscribed explicitly.
I think (I hope) it's out of scope :)

PR #2596, opened 2024-03-27
RHCLOUD-31647 Reset loggers level when the variant payload is empty

gwenneg:
Before this PR, the loggers level was not reset if the variant payload was `[]`.

I'm also aligning the code with the blog post.

### DIFF
--- a/common-unleash/src/main/java/com/redhat/cloud/notifications/unleash/LogLevelManager.java
+++ b/common-unleash/src/main/java/com/redhat/cloud/notifications/unleash/LogLevelManager.java
@@ -42,7 +42,7 @@ public class LogLevelManager {
     @Inject
     ObjectMapper objectMapper;
 
-    private final Map<String, Level> previousLoggerLevels = new HashMap<>();
+    private final Map<String, Level> originalLoggerLevels = new HashMap<>();
 
     public void process(@Observes ToggleCollection toggleCollection) {
         LogConfig[] logConfigs = getLogConfigs();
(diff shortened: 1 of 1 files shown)


MikelAlejoBR:
Hmmmmm I was wondering if we need to make it so that when no log configs are specified, then the original value is set. Because if I understood correctly, if we change the log level for the same logger multiple times, and then we decide to revert it, we will stay with the previous log level but it will never go back to its original level?

For example, if we start with a logger `com.redhat.notifications.MyLoggerA` with a level of `INFO`, our `previosLoggersLevels` will start as follows:

| previousLoggersLevels.key | previousLoggersLevels.value |
|---|---|
| | |

Then we pass a payload to modify `com.redhat.notifications.MyLoggerA` to `DEBUG`, the `setLoggerLevel` function will call `previousLoggersLevels.put("com.redhat.notifications.MyLoggerA", "INFO")`, so we will have the following in our map:

| previousLoggersLevels.key | previousLoggersLevels.value |
|---|---|
|`com.redhat.notifications.MyLoggerA`|`INFO`|

Now we decide we want to change our level to `TRACE` again,  so the `setLoggerLevel` function will call `previousLoggersLevels.put("com.redhat.notifications.MyLoggerA", "DEBUG")`, and we will end up with the following in our map:

| previousLoggersLevels.key | previousLoggersLevels.value |
|---|---|
|`com.redhat.notifications.MyLoggerA`|`DEBUG`|

Now we decide that we have finished with our debugging, so we simply remove that logger entry from the variant. Since our last known logger level for that logger is `DEBUG`, won't the function revert the logger's log level to that instead of the `INFO` that we probably intended?

gwenneg:
You are absolutely right! I'll update the PR, thanks!

MikelAlejoBR (approved):
I'm glad to help! :smile: LGTM!

PR #2598, opened 2024-03-27
[RHCLOUD-31655] Increase JTA transactions timeout for mass update on …

g-duval:
…Event db table


gwenneg:
Could you please add a comment near each `@TransactionConfiguration` to explain why it's needed and that it is a temporary workaround for the problem the annotation solves?

gwenneg:
My comment includes the existing annotation on the `deleteBundle` method.

PR #2599, opened 2024-03-27
Add Event type visibility check box on internal admin ui

g-duval:


admin-console/src/main/webapp/src/components/EventTypes/CreateEditModal.tsx
```diff
@@ -109,6 +109,15 @@ export const CreateEditModal: React.FunctionComponent<CreateEditModalProps> = (p
                             onChange={ (isChecked) => setEventType(prev => ({...prev, subscriptionLocked: isChecked})) }
                         />
                     </FormGroup>
+                    <FormGroup label='is visible?' fieldId='visible'
```
gwenneg, line 112:
```suggestion
                    <FormGroup label='Is visible?' fieldId='visible'
```

admin-console/src/main/webapp/src/components/EventTypes/CreateEditModal.tsx
```diff
@@ -109,6 +109,15 @@ export const CreateEditModal: React.FunctionComponent<CreateEditModalProps> = (p
...
+                        helperText='Should the event type be visible by users?'>
```
gwenneg, line 113:
```suggestion
                        helperText='Should the event type be visible in the public UI?'>
```

admin-console/src/main/webapp/src/pages/ApplicationPage.tsx
```diff
@@ -107,7 +107,7 @@ export const ApplicationPage: React.FunctionComponent = () => {
     const createEventType = () => {
         setShowModal(true);
         setIsEdit(false);
-        setEventTypes({});
+        setEventTypes({visible:true});
```
gwenneg, line 110: What is the effect of that line?

g-duval: It defines the event type default value on creation flow.

gwenneg: So `visible` will be true by default and the other boolean values will be false?

PR #2604, opened 2024-03-29
RHCLOUD-31712 Improve integration-disabled emails in case of IOException

gwenneg:


engine/src/main/resources/templates/drawer/Integrations/failedIntegrationDrawerBody.html
gwenneg, line 1: This file should have been deleted in an older PR.

engine/src/main/java/com/redhat/cloud/notifications/events/EndpointErrorFromConnectorHelper.java
```diff
@@ -52,45 +63,60 @@ public void manageEndpointDisablingIfNeeded(Endpoint endpoint, JsonObject payloa
...
-                     * server failure, which is usually something temporary. Sending another notification to
-                     * the same endpoint may work in the future, so the endpoint is only disabled if the max
-                     * number of endpoint failures allowed from the configuration is exceeded.
-                     */
-                    boolean disabled = endpointRepository.incrementEndpointServerErrors(endpoint.getId(), maxServerErrors);
-                    if (disabled) {
-                        disabledWebhooksServerErrorCount.increment();
-                        Log.infof("Endpoint %s was disabled because we received too many 5xx status while calling it", endpoint.getId());
-                        integrationDisabledNotifier.tooManyServerErrors(endpoint, maxServerErrors);
+                    if (HTTP_SERVER_ERRORS.contains(httpErrorType.get())) {
```
g-duval, line 91: Shouldn't it be an `else if` ?

g-duval (approved):
One minor comment, LGTM

engine/src/main/java/com/redhat/cloud/notifications/events/EndpointErrorFromConnectorHelper.java
gwenneg: It should indeed. Fixed!

PR #2619, opened 2024-04-05
RHCLOUD-22071 | feature(query): sent emails per organization, bundle and application

MikelAlejoBR:
Queries for the number of sent emails and groups them by organization, bundle and application.

## Jira ticket
[[RHCLOUD-22071]](https://issues.redhat.com/browse/RHCLOUD-22071)

### DIFF
--- a/.rhcicd/floorist/floorplan.yaml
+++ b/.rhcicd/floorist/floorplan.yaml
@@ -54,6 +54,33 @@ objects:
             et.display_name,
             bg.org_id,
             actively_used
+      - prefix: insights/notifications/emails_sent_by_bundle_app_org_id
+        query: >-
+          SELECT
+            bundles.display_name AS bundle,
+            applications.display_name AS application,
+            "event".org_id,
+            COUNT(notification_history.*)
+          FROM
+            notification_history
+          INNER JOIN
+            "event"
+              ON "event".id = notification_history.event_id
+          INNER JOIN
+            event_type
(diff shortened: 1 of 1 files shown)


.rhcicd/floorist/floorplan.yaml
```diff
@@ -54,6 +54,33 @@ objects:
             et.display_name,
             bg.org_id,
             actively_used
+      - prefix: insights/notifications/emails_sent_by_bundle_app_org_id
+        query: >-
+          SELECT
```
gwenneg, line 59: This query will show more emails than the real number because depending on the recipients settings and subscriptions, some Kafka messages sent to the email connector will not result in emails being sent.

MikelAlejoBR: Hmmmm I see... because there might be a case where we attempt to Notify an org but all of their users might have the email notifications turned off, and even though we will register a success result no email might have been sent...

gwenneg: What do you want to do with this query? I'm fine with merging it if the inaccurate data is acceptable.

MikelAlejoBR: I am going to put it in draft until I talk to the Business Insights people to see if we can add a notice to the page explaining what the data represents, or if we can build data that comes from Kibana too.

MikelAlejoBR:
Closed until we receive more news about Floorist.

PR #2621, opened 2024-04-09
RHCLOUD-31927 | fix: Floorist queries giving error

MikelAlejoBR:
One of the queries was returning three "display_name" columns instead of having aliased names, and that was causing Floorist to break for having multiple columns in a data frame.

## Jira ticket
[[RHCLOUD-31927]](https://issues.redhat.com/browse/RHCLOUD-31927)


gwenneg (approved):
LGTM

PR #2662, opened 2024-04-17
Default  prometheus push gateway value.

shannon-donahue:
This PR changes the following:
- defaults the value for the prometheus push gateway. 
   - This ensures that there will not be deployment errors if prometheus push gateway is not set in a cluster's env. 

The following error occurs if the value is not set.
```
Caused by: io.quarkus.runtime.configuration.ConfigurationException: Failed to load config value of type class java.lang.String for: prometheus.pushgateway.url
```


gwenneg (approved):
LGTM

PR #2665, opened 2024-04-19
[RHCLOUD-32092] add a minimum delay before disabling endpoints

g-duval:


MikelAlejoBR (approved):
LGTM!

database/src/main/resources/db/migration/V1.97.0__RHCLOUD-32092_add_first_error_date_column_on_endpoint.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE endpoints ADD COLUMN server_errors_since TIMESTAMP DEFAULT null;
```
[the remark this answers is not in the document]
g-duval: We don't need to store the timezone here, because all timestamps will be stored as utc time.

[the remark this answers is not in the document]
MikelAlejoBR: Perfect-o!

MikelAlejoBR (approved):

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -125,10 +128,11 @@ public List<Endpoint> getTargetEmailSubscriptionEndpoints(String orgId, UUID eve
      * Increments the server errors counter of the endpoint identified by the given ID.
      * @param endpointId the endpoint ID
      * @param maxServerErrors the maximum server errors allowed from the configuration
+     * @param minDelaySinceFirstServerErrorBeforeDisabling the minimum delay since the first error before an endpoint could be disabled
      * @return {@code true} if the endpoint was disabled by this method, {@code false} otherwise
      */
     @Transactional
-    public boolean incrementEndpointServerErrors(UUID endpointId, int maxServerErrors, int currentServerErrors) {
+    public boolean incrementEndpointServerErrors(UUID endpointId, int maxServerErrors, int currentServerErrors, Duration minDelaySinceFirstServerErrorBeforeDisabling) {
```
gwenneg, line 135: `EndpointRepository` could inject the config value directly instead of passing it as a method arg here.

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -139,10 +143,18 @@ public boolean incrementEndpointServerErrors(UUID endpointId, int maxServerError
          * It may or may not have been disabled already from the frontend or because of a 4xx error.
          */
         if (endpoint.isPresent() && endpoint.get().isEnabled()) {
-            if (endpoint.get().getServerErrors() + currentServerErrors > maxServerErrors) {
+            LocalDateTime currentTime = LocalDateTime.now(ZoneId.of("UTC"));
+            boolean minDelaySinceFirstErrorRespected = false;
+            if (endpoint.get().getServerErrorsSince() != null) {
+                Duration spentDuration = Duration.between(endpoint.get().getServerErrorsSince(), currentTime);
+                minDelaySinceFirstErrorRespected = spentDuration.compareTo(minDelaySinceFirstServerErrorBeforeDisabling) > 0;
+            }
```
gwenneg, line 151: This whole block could be under the old `if (endpoint.get().getServerErrors() + currentServerErrors > maxServerErrors)` condition as it only needs to be executed when that condition is `true`.

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -139,28 +146,45 @@ public boolean incrementEndpointServerErrors(UUID endpointId, int maxServerError
...
-                 */
-                String hql = "UPDATE Endpoint SET enabled = FALSE WHERE id = :id AND enabled IS TRUE";
-                int updated = entityManager.createQuery(hql)
-                        .setParameter("id", endpointId)
-                        .executeUpdate();
-                return updated > 0;
+            LocalDateTime currentTime = LocalDateTime.now(ZoneId.of("UTC"));
+
+            if (endpoint.get().getServerErrors() + currentServerErrors > engineConfig.getMaxServerErrors()) {
+                if (endpoint.get().getServerErrorsSince() != null) {
```
gwenneg, line 152: With the current code, endpoints with `getServerErrors() > 0` before this is deployed will never be disabled because `getServerErrorsSince()` will be `null` and remain `null` forever.

g-duval: Thats right, thant you

gwenneg (approved):
LGTM

PR #2671, opened 2024-04-22
Enable single daily by default + update tests

g-duval:


aggregator/src/main/java/com/redhat/cloud/notifications/config/AggregatorConfig.java
```diff
@@ -34,7 +34,7 @@ private static String toggleName(String feature) {
     @Deprecated(forRemoval = true, since = "To be removed when we're done migrating to Unleash in all environments")
     boolean unleashEnabled;
 
-    @ConfigProperty(name = "notifications.bundle.level.digest.enabled", defaultValue = "false")
+    @ConfigProperty(name = "notifications.bundle.level.digest.enabled", defaultValue = "true")
```
gwenneg, line 37: Could you please update the ClowdApp template as well?

engine/src/main/java/com/redhat/cloud/notifications/config/EngineConfig.java
```diff
@@ -44,7 +44,7 @@ public class EngineConfig {
     @Deprecated(forRemoval = true, since = "To be removed when we're done migrating to Unleash in all environments")
     boolean asyncAggregation;
 
-    @ConfigProperty(name = "processor.email.aggregation.use-recipients-resolver-clowdapp.enabled", defaultValue = "false")
+    @ConfigProperty(name = "processor.email.aggregation.use-recipients-resolver-clowdapp.enabled", defaultValue = "true")
```
gwenneg, line 47: This looks unrelated and could be a breaking change in some environments.

aggregator/src/main/java/com/redhat/cloud/notifications/config/AggregatorConfig.java
g-duval: Sure

engine/src/main/java/com/redhat/cloud/notifications/config/EngineConfig.java
g-duval: That's the idea, but I suppose it have to be validated on dedicated stage env before pomoting it to prod. May by I'm expecting too much, let's rollback it ;)

gwenneg (approved):
LGTM

PR #2676, opened 2024-04-24
[RHCLOUD-32299] Clean aggregation dead code

g-duval:


engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailProcessor.java
```diff
@@ -37,7 +37,7 @@ public class EmailProcessor extends SystemEndpointTypeProcessor {
     EmailActorsResolver emailActorsResolver;
 
     @Inject
-    EmailSubscriptionTypeProcessor emailSubscriptionTypeProcessor;
+    EmailAggregationProcessor emailSubscriptionTypeProcessor;
```
MikelAlejoBR, line 40: What do you think about refactoring the variable to `emailAggregationProcessor` too?

engine/src/main/java/com/redhat/cloud/notifications/events/EndpointProcessor.java
```diff
@@ -53,7 +53,7 @@ public class EndpointProcessor {
     EmailProcessor emailConnectorProcessor;
 
     @Inject
-    EmailSubscriptionTypeProcessor emailProcessor;
+    EmailAggregationProcessor emailProcessor;
```
MikelAlejoBR, line 56: Same here

engine/src/main/java/com/redhat/cloud/notifications/processors/email/AsyncAggregation.java
```diff
@@ -12,7 +12,7 @@
 public class AsyncAggregation implements Runnable {
 
     @Inject
-    EmailSubscriptionTypeProcessor emailSubscriptionTypeProcessor;
+    EmailAggregationProcessor emailSubscriptionTypeProcessor;
```
MikelAlejoBR, line 15: And here?

MikelAlejoBR (approved):
LGTM!

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailProcessor.java
g-duval: That's right, thank you :)

MikelAlejoBR (approved):
LGTM!

aggregator/src/main/java/com/redhat/cloud/notifications/DailyEmailAggregationJob.java
```diff
@@ -147,7 +144,7 @@ private void sendIt(List<AggregationCommand> aggregationCommands) {
             .withTimestamp(LocalDateTime.now(UTC))
             .withEvents(eventList)
             .withContext(new Context.ContextBuilder()
-                .withAdditionalProperty("single_daily_digest_enabled", aggregatorConfig.isSingleDailyDigestEnabled())
+                .withAdditionalProperty("single_daily_digest_enabled", true)
```
gwenneg, line 147: Are we keeping or removing this? It is still set here, but no longer consumed in the engine.

gwenneg: A 2 steps change would probably be safer. When this PR is deployed, an old aggregator could produce something a new engine wouldn't understand, unless the engine can still deal with old aggregation commands.

g-duval: Engine uses single daily on prod for more than a month, so we don't plan to use old aggregation commands anymore.

.rhcicd/clowdapp-aggregator.yaml
```diff
@@ -66,8 +66,6 @@ objects:
           value: ${NOTIFICATIONS_UNLEASH_ENABLED}
         - name: PROMETHEUS_PUSHGATEWAY_URL
           value: ${PROMETHEUS_PUSHGATEWAY}
-        - name: NOTIFICATIONS_BUNDLE_LEVEL_DIGEST_ENABLED
```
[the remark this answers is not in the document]
g-duval: That's right, thank you :)

common-aggregator/pom.xml
```diff
@@ -1,63 +0,0 @@
-<?xml version="1.0"?>
-<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd"
-         xmlns="http://maven.apache.org/POM/4.0.0"
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
-    <modelVersion>4.0.0</modelVersion>
-
-    <artifactId>notifications-common-aggregator</artifactId>
```
gwenneg, line 7: What is the benefit of removing the `common-aggregator` module? Now, several classes including `AggregationCommand` are included in the `backend` module classpath while they're not needed.

g-duval: Soon Aggregator will use Events instead of EmailAggregation. So we can move all Event mapping classes related to `common-aggregator`, or we can remove `common-aggregator` module to depend of common directly. The second option seems more reasonable to me. WDYT?

gwenneg (approved):
LGTM

PR #2682, opened 2024-04-26
[RHCLOUD-32364] convert Drawer notifications from HTML into Markdown

g-duval:
Because we only used Html tags 'b' and 'i', to convert it into Markdown we have to replace:

`<b>` and `</b>` by `**`

`<i>` and `</i>` by `*`


gwenneg:
I don't remember which decision we made (if any) about links. Most drawer notifications are currently useless because they don't redirect the user to something helpful.

gwenneg (approved):
LGTM

PR #2695, opened 2024-05-03
RHCLOUD-32456 | feature: remove the grouppings and the counts from our queries

MikelAlejoBR:
We have been told by the Floorist team that we should not group or aggregate our queries whatsoever, because that hinders and limits the aggregations that the Business Insights people can perform with Tableau.

The prefixes of the queries have been changed too so that the Business Insights team can start fresh with the queries, instead of having to deal with multiple versions of them which would make them have to deal with different columns depending on the dates the data was collected in.

## Jira ticket
[[RHCLOUD-32456]](https://issues.redhat.com/browse/RHCLOUD-32456)


gwenneg (approved):
LGTM

PR #2700, opened 2024-05-06
Kessel POC

g-duval:


recipients-resolver/src/main/java/com/redhat/cloud/notifications/recipients/resolver/RecipientsResolver.java
```diff
@@ -30,20 +29,24 @@ public class RecipientsResolver {
...
-    @RestClient
-    RelationshipsApi kesselRelationshipApi;
+    KesselService kesselLookupService;
 
-    @CacheResult(cacheName = "find-recipients")
     public Set<User> findRecipients(String orgId, Set<RecipientSettings> recipientSettings, Set<String> subscribers, Set<String> unsubscribers, boolean subscribedByDefault) {
+        return findRecipients(orgId, recipientSettings, subscribers, unsubscribers, subscribedByDefault, null);
+    }
+
+    @CacheResult(cacheName = "find-recipients")
```
MikelAlejoBR, line 38: I wonder if we should stop caching users here, since AuthZ is in theory going to be faster than RBAC... otherwise I guess a user that gets revoked their permission would probably receive the email because we've got the results cached, no?

recipients-resolver/src/main/java/com/redhat/cloud/notifications/recipients/resolver/RecipientsResolver.java
```diff
@@ -134,16 +147,15 @@ private static Optional<Set<String>> extractRequestUsersIntersection(Set<Recipie
...
-    private Optional<Set<String>> callKessel() {
-        Optional<Set<String>> allowedUsers = Optional.empty();
-        try {
-            ApiRebacV1ReadRelationshipsResponse kesselResponse = kesselRelationshipApi.relationshipsReadRelationships("group", "bob_club", "member", null, null, null);
-            Log.infof("Kessel Open Api response: %s", kesselResponse);
-            allowedUsers = Optional.of(kesselResponse.getRelationships().stream().filter(rel -> "user".equals(rel.getSubject().getObject().getType())).map(rel -> rel.getSubject().getObject().getId().toLowerCase()).collect(Collectors.toSet()));
-            Log.infof("Allowed users from Kessel are: %s", allowedUsers);
-        } catch (Exception ex) {
-            Log.error("Error calling Kessel relationship Api");
+    private Set<String> callKessel(ExternalAuthorizationCriteria externalAuthorizationCriteria) {
```
MikelAlejoBR, line 150: What do you think about naming this function as `findUsersWithCriteria`?

g-duval: Ok

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregator.java
```diff
@@ -128,7 +135,8 @@ public Map<User, Map<String, Object>> getAggregated(UUID appId, EmailAggregation
                             ).collect(toSet()),
                             subscribers,
                             unsubscribers,
-                            eventType.isSubscribedByDefault()
+                            eventType.isSubscribedByDefault(),
+                            externalAuthorizationCriteria
```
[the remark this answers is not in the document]
MikelAlejoBR: All right, that's okay!

recipients-resolver/src/main/java/com/redhat/cloud/notifications/recipients/resolver/RecipientsResolver.java
MikelAlejoBR: All right then!

MikelAlejoBR (approved):

PR #2720, opened 2024-05-23
RHCLOUD-31165 | fix: endpoint subentities not properly serialized

MikelAlejoBR:
I removed all the annotations for the internal entities, because I assumed we had all the bases covered when I changed the response entity from "Endpoint" to "EndpointDTO" in all of the REST endpoints that we have.

However, there are some other endpoints that return some other entities like behavior groups, which also serialize endpoints. But they serialize the internal model's endpoint, not the DTO model's one. And since we are changing our model slowly by slowly, we decided not to create DTOs for our entire internal model in one single step, since it would be too much work and very hard to review.

Gwenneg suggested bringing the JSON serialization annotations back, and all the tests passed everywhere, but we faced another problem in the UI: it was breaking in certain calls where the endpoints where being listed.

I realized that even though I reverted the "Endpoint"'s entity's JSON serialization annotations, I did not do it for the "Endpoint"'s properties. That is what probably is causing the issue with the UI, since the properties are not being properly serialized.

Finally, I'm reverting the functions that change the case of the
"EndpointType"'s enum values, because we are supposed to return the
values in snake case, and that was not working for the
"EMAIL_SUBSCRIPTION" enum value.

## Jira ticket
[[RHCLOUD-31165]](https://issues.redhat.com/browse/RHCLOUD-31165)

### DIFF
--- a/common/src/main/java/com/redhat/cloud/notifications/models/EndpointType.java
+++ b/common/src/main/java/com/redhat/cloud/notifications/models/EndpointType.java
@@ -1,7 +1,9 @@
 package com.redhat.cloud.notifications.models;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
 
+@Schema(enumeration = { "ansible", "camel", "drawer", "email_subscription", "webhook" }) // TODO remove them once the transition to DTOs have been completed.
 public enum EndpointType {
     @JsonProperty("webhook") // TODO remove them once the transition to DTOs have been completed.
     WEBHOOK(false, false),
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/redhat/cloud/notifications/models/EndpointType.java
```diff
@@ -1,12 +1,17 @@
 package com.redhat.cloud.notifications.models;
 
-import com.fasterxml.jackson.annotation.JsonValue;
+import com.fasterxml.jackson.annotation.JsonProperty;
 
 public enum EndpointType {
```
gwenneg, line 5: Should `@Schema` be added back here?

MikelAlejoBR: Sure, yeah, let's add it back just in case.

gwenneg (approved):
LGTM

gwenneg (approved):

PR #2725, opened 2024-05-27
Remove pip from base image

g-duval:
We aleady removed it in the past (https://github.com/RedHatInsights/notifications-backend/pull/2725) then we restored it because the CVE was fixed on base image.
Unfortunately, a new cve on the same pip package was raised (https://github.com/RedHatInsights/notifications-backend/actions/runs/9256269567/job/25461895905?pr=2725)


MikelAlejoBR (approved):
LGTM!

PR #2726, opened 2024-05-28
Add SocketTimeoutException as endpoint auto-disabling cause

g-duval:


connector-common-http/src/main/java/com/redhat/cloud/notifications/connector/http/HttpExceptionProcessor.java
```diff
@@ -49,7 +50,7 @@ protected void process(Throwable t, Exchange exchange) {
             } else {
                 logHttpError(ERROR, e, exchange);
             }
-        } else if (t instanceof ConnectTimeoutException) {
+        } else if (t instanceof ConnectTimeoutException || t instanceof SocketTimeoutException) {
```
gwenneg, line 53: SocketTimeout happens when the connection has been established already and we're waiting too long for data. It's different from a connection timeout so introducing a new error type would make sense.

gwenneg: I'm not sure I'm clear so here's an additional comment: I would process the SocketTimeoutException in a new `else if (t instanceof SocketTimeoutException` block and use a new `HttpErrorType` value dedicated to that case.

engine/src/main/java/com/redhat/cloud/notifications/events/HttpErrorType.java
```diff
@@ -1,6 +1,7 @@
 package com.redhat.cloud.notifications.events;
 
 public enum HttpErrorType {
+    SOCKET_TIMEOUT("a timeout happened while waiting for server response"),
```
gwenneg, line 4: Small change to be consistent with other messages:
```suggestion
    SOCKET_TIMEOUT("a timeout happened while waiting for the HTTP server response"),
```

gwenneg (approved):
LGTM

PR #2733, opened 2024-05-30
RHCLOUD-32502 | Kessel: proper exception handling

MikelAlejoBR:
When gRPC throws exceptions Quarkus simply returns the entire body of
the exception in the response to the client. It usually it is not useful
because it's hard to follow, and the client should not really receive
this kind of information.
    
This change logs the exception with all kinds of information and returns
an "internal server error" response to the client omitting the details
of what caused the error.

## Dependencies

- https://github.com/RedHatInsights/notifications-backend/pull/2704

## Jira ticket
[[RHCLOUD-32502]](https://issues.redhat.com/browse/RHCLOUD-32502)


g-duval (approved):
LGTM

PR #2792, opened 2024-07-02
[RHCLOUD-31320] Increased number of connector threads

cam-rod:
## Jira issue

https://issues.redhat.com/browse/RHCLOUD-31320

## Changes

All connectors _(except email)_ now have between 40 and 70 threads available:

- 10x Slack threads, due to generous API limits and currently low count.
- 3x drawer threads (limit is unknown)
- 2.5x Splunk, ServiceNow, webhook threads
- 2x Gchat, Teams threads for lower API limits
- ~~7x email threads, due to high demand for aggregated reports~~ no change to email threads due to backend limits

How I decided on the new thread counts:

- Peak and average lag for the different connectors when sending notifications
- Current CPU/memory usage
- Current demand and expected near-term demand
- API rate limits
  - while research was done on looking into rate limits for the APIs, it isn't the main driver for determining whether threads should be increased or not

### DIFF
--- a/.rhcicd/clowdapp-connector-email.yaml
+++ b/.rhcicd/clowdapp-connector-email.yaml
@@ -85,7 +85,7 @@ objects:
         - name: NOTIFICATIONS_CONNECTOR_SEDA_CONCURRENT_CONSUMERS
           value: ${NOTIFICATIONS_CONNECTOR_SEDA_CONCURRENT_CONSUMERS}
         - name: NOTIFICATIONS_CONNECTOR_SEDA_QUEUE_SIZE
-          value: ${NOTIFICATIONS_CONNECTOR_SEDA_QUEUE_SIZE}
+          value: ${NOTIFICATIONS_CONNECTOR_SEDA_CONCURRENT_CONSUMERS}
         - name: NOTIFICATIONS_CONNECTOR_USER_PROVIDER_BOP_API_TOKEN
           valueFrom:
             secretKeyRef:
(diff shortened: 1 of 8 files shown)


.rhcicd/clowdapp-connector-email.yaml
```diff
@@ -201,11 +201,8 @@ parameters:
   description: Maximum number of redelivery attempts (initial call not included)
   value: "2"
 - name: NOTIFICATIONS_CONNECTOR_SEDA_CONCURRENT_CONSUMERS
-  description: Number of concurrent threads processing exchanges with SEDA
-  value: "10"
-- name: NOTIFICATIONS_CONNECTOR_SEDA_QUEUE_SIZE
-  description: Maximum capacity of the SEDA queue
-  value: "10"
+  description: Number of concurrent threads processing exchanges with SEDA; Maximum capacity of the SEDA queue
+  value: "70"
```
g-duval, line 205: Could you please detail where 70 comes from?
Because we have 3 pods on prod, it means increasing from 30 to 2100 requests in parallel. It's possible but  did you validated it with BOP team?

cam-rod: Not yet, although I think the increase would be 210, instead of 2100

MikelAlejoBR (approved):

.rhcicd/clowdapp-connector-email.yaml
g-duval: yes sorry I added an extra 0 :)

g-duval (approved):
LGTM

PR #2798, opened 2024-07-04
[RHCLOUD-28890] Test engine autoscaler

jessicarod7:
Fork of #2774 for testing various metrics. Will update description when PR is undrafted.


jessicarod7:
Ran some testing, support does not appear to be available on ephemeral environment. We'll have to gate the changes and test in stage.

jessicarod7:
Tested autoscaling on the ephemeral cluster, and it successfully scaled up! I wasn't able to do a proper test of this configuration, as Prometheus isn't available there.

I'm in favour of merging this to stage, and then monitoring in that environment, adjusting the triggers if needed.

PR #2804, opened 2024-07-09
RHCLOUD-33871 | fix: null pointer exception when the "extras" object is null

MikelAlejoBR:
We are no longer storing Slack's channel when setting up a Slack endpoint, and that means that the "extras" object might be null or empty. We forgot to add a null check for that object which is causing some payloads to not be sent because of it.

## Jira ticket
[[RHCLOUD-33871]](https://issues.redhat.com/browse/RHCLOUD-33871)

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/slack/SlackProcessor.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/camel/slack/SlackProcessor.java
@@ -7,6 +7,8 @@ import com.redhat.cloud.notifications.processors.camel.CamelNotification;
 import com.redhat.cloud.notifications.processors.camel.CamelProcessor;
 import jakarta.enterprise.context.ApplicationScoped;
 
+import java.util.Map;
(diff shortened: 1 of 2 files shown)


engine/src/main/java/com/redhat/cloud/notifications/processors/camel/slack/SlackProcessor.java
```diff
@@ -29,7 +31,12 @@ protected CamelNotification getCamelNotification(Event event, Endpoint endpoint)
 
         SlackNotification notification = new SlackNotification();
         notification.webhookUrl = properties.getUrl();
-        notification.channel = properties.getExtras().get("channel");
+
+        final Map<String, String> extras = properties.getExtras();
+        if (null != extras) {
+            notification.channel = properties.getExtras().get("channel");
```
g-duval, line 37:
```suggestion
            notification.channel = extras.get("channel");
```

g-duval (approved):

PR #2807, opened 2024-07-10
RHCLOUD-33636 | feature: migrate duplicated endpoint and behavior groups' names

MikelAlejoBR:
The problem with using the migration endpoint for this is that we don't know when the Fedramp environment will be available to run that migration. And if we push the database constraints before the migration has run in Fedramp, Notifications might crash. So the way to ensure that this does not happen is by running the data migrations as a database migration.

## Jira ticket
[[RHCLOUD-33636]](https://issues.redhat.com/browse/RHCLOUD-33636)


g-duval (approved):
LGTM

PR #2808, opened 2024-07-10
RHINENG-11314: Collect metrics to HMS bucket

vkrizan:
Adds a second FloorPlan that would collect data for building metrics on endpoint and event usage and store those to HMS S3 bucket.

The aggregated counts are group by a day, org, event types and endpoint types, and status. 

7 day window should be enough to cover any possible collection outages.

Note, that technically we can collect all notifications without the aggregation, but that could possibly generate a lot of data.

RHINENG-11314

### DIFF
--- a/.rhcicd/floorist/floorplan.yaml
+++ b/.rhcicd/floorist/floorplan.yaml
@@ -216,6 +216,56 @@ objects:
           INNER JOIN
             bundles
               ON bundles.id = applications.bundle_id
+- apiVersion: metrics.console.redhat.com/v1alpha1
+  kind: FloorPlan
+  metadata:
+    name: notifications-hms
+  spec:
+    database:
+      secretName: ${FLOORIST_DB_SECRET_NAME}
+    objectStore:
+      secretName: ${FLOORIST_HMS_BUCKET_SECRET_NAME}
+    logLevel: ${FLOORIST_LOGLEVEL}
+    suspend: ${{FLOORIST_SUSPEND}}
+    queries:
+    - prefix: hms_analytics/notifications/${FLOORIST_ENV_PREFIX}/notifications_deliveries
+      chunksize: 1000
+      query: >-
+        SELECT
+          bun.display_name::TEXT AS bundle,
+          apps.display_name::TEXT AS application,
+          e.org_id::TEXT,
+          CASE
+            WHEN
+              nh.endpoint_type_v2 = 'CAMEL'
+            THEN
+              LOWER(nh.endpoint_sub_type)
+            ELSE
+              LOWER(nh.endpoint_type_v2)
+          END AS endpoint_type,
+          et.name::TEXT AS event_type,
+          nh.status::TEXT,
+          e.created::DATE,
+          COUNT(*) AS count
+        FROM
+          notification_history AS nh
+        INNER JOIN
+          "event" AS e
+            ON e.id = nh.event_id
+        INNER JOIN
+          event_type AS et
+            ON et.id = e.event_type_id
+        INNER JOIN
+          applications AS apps
+            ON apps.id = e.application_id
+        INNER JOIN
+          bundles AS bun
+            ON bun.id = apps.bundle_id
+        WHERE e.created > (CURRENT_DATE - INTERVAL '7 days')
+        GROUP BY
+          bun.display_name, apps.display_name, e.org_id, nh.endpoint_type_v2,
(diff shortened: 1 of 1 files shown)


vkrizan:
Looks like the error is unrelated to these changes:
```
Caused by: org.flywaydb.core.api.FlywayException: Found more than one migration with version 1.98.0
Offenders:
-> /home/runner/work/notifications-backend/notifications-backend/database/target/classes/db/migration/V1.98.0__RHCLOUD-33142_add_payload_database_holder.sql (SQL)
-> /home/runner/work/notifications-backend/notifications-backend/database/target/classes/db/migration/V1.98.0__RHCLOUD-33636-remove-duplicate-names-endpoints-behaviorgroups.sql (SQL)
```

vkrizan:
@MikelAlejoBR @gwenneg please review :pray: thanks

.rhcicd/floorist/floorplan.yaml
```diff
@@ -216,11 +216,68 @@ objects:
...
+  metadata:
+    name: notifications-hms
+  spec:
+    database:
+      secretName: ${FLOORIST_DB_SECRET_NAME}
+    objectStore:
+      secretName: ${FLOORIST_HMS_BUCKET_SECRET_NAME}
+    logLevel: ${FLOORIST_LOGLEVEL}
+    suspend: ${{FLOORIST_SUSPEND}}
+    disabled: ${{FLOORIST_DISABLED}}
```
MikelAlejoBR, line 230: We don't have the `FLOORIST_DISABLED` parameter in our template. Feel free to either remove it from this FloorPlan or to add it in our parameters below.

.rhcicd/floorist/floorplan.yaml
```diff
@@ -216,11 +216,68 @@ objects:
...
 - name: FLOORIST_BUCKET_SECRET_NAME
   description: Floorist's S3 bucket's secret name
   required: true
   value: dummy-secret
+- name: FLOORIST_HMS_BUCKET_SECRET_NAME
+  description: HMS bucket secret name
+  value: floorist-bucket
+- name: FLOORIST_ENV_PREFIX
+  description: Used to split data across environments on a single (HMS) bucket.
+  value: stage
```
MikelAlejoBR, line 280: Very nitpicky but could you please place these parameters alphabetically by their name so that we end up with:

- FLOORIST_BUCKET_SECRET_NAME
- FLOORIST_DB_SECRET_NAME
- FLOORIST_ENV_PREFIX
- FLOORIST_HMS_BUCKET_SECRET_NAME
- FLOORIST_LOGLEVEL
- FLOORIST_SUSPEND

.rhcicd/floorist/floorplan.yaml
```diff
@@ -216,11 +216,68 @@ objects:
...
+              nh.endpoint_type_v2 = 'CAMEL'
+            THEN
+              LOWER(nh.endpoint_sub_type)
+            ELSE
+              LOWER(nh.endpoint_type_v2)
+          END AS endpoint_type,
+          et.name::TEXT AS event_type,
+          nh.status::TEXT,
+          e.created::DATE,
+          COUNT(*) AS count
```
MikelAlejoBR, line 250: Just a comment just in case! Here you would be getting the events from our internal customers too, since the aggregation occurs in the database itself. I guess that this is fine with your use case, but just wanted to mention it in case you want to filter out the internal customers from the external ones.

This is why we had to have our queries not aggregated...

MikelAlejoBR (commented):
Looking good! Just a couple of very nitpicky things...

.rhcicd/floorist/floorplan.yaml
vkrizan: It should aggregate it by org_id as well. I understand the reason for NOT aggregating, and I'd personally prefer it. The issue is that it would create lots of unnecessary data for us.

.rhcicd/floorist/floorplan.yaml
vkrizan: done

.rhcicd/floorist/floorplan.yaml
vkrizan: Thank you. Removed!

.rhcicd/floorist/floorplan.yaml
MikelAlejoBR: Understood! Thanks for the explanation!

PR #2813, opened 2024-07-12
 RHCLOUD-33637 | refactor: remove toggle for the faulty endpoints

MikelAlejoBR:
The feature has been enabled in production and it was worked flawlessly. Therefore, we can leave it as a permanent enabled feature in our application.

## Jira application
[[RHCLOUD-33637]](https://issues.redhat.com/browse/RHCLOUD-33637)

### DIFF
--- a/.rhcicd/clowdapp-connector-google-chat.yaml
+++ b/.rhcicd/clowdapp-connector-google-chat.yaml
@@ -169,9 +169,6 @@ parameters:
 - name: NOTIFICATIONS_LOG_LEVEL
   description: Log level of Notifications
   value: INFO
-- name: NOTIFICATIONS_CONNECTOR_HTTP_DISABLE_FAULTY_ENDPOINTS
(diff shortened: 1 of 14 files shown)


g-duval:
@MikelAlejoBR If we just remove those checks, system endpoints such as Email and Drawer will be disabled in case of BOP or IT service temporary incident. Not cool dude! :)

WDYT about adding an early check/return in  [incrementEndpointServerErrors(...)](https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java#L138) and [disableEndpoint(...)](https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java#L275) if `EndpontType.isSystemEndpointType` is true?

g-duval (approved):
LGTM

PR #2820, opened 2024-07-16
[RHCLOUD-31596] Fix filtering for muted eventTypes

cam-rod:
## Jira ticket

https://issues.redhat.com/browse/RHCLOUD-31596

## Description

This fixes issues with the original version of the `excludeMutedTypes` param on the `notifications/eventTypes` endpoint. The new version works as follows:

- Endpoint captures security context, and if the parameter is set to `excludeMutedTypes=true`, retrieves a list of event type IDs tied to an organization or default behavior group
- IDs are passed into the main query, and if parameter is set to true, will filter the list

## Compatibility

The previous implementation listed event types linked to any behavior group globally, making it (almost) a no-op. This fix restores the intended functionality.

## Testing

- `testEventFetchingAndExcludeMutedTypes` now also adds event types to a default behavior group, and to a different organization. Only event types assigned to behaviour groups in the organization, or default behavior groups, are accepted.
- `testEventFetchingByBundleApplicationEventTypeNameAndExcludeMutedTypes` tests the same, showing that even when matching by all parameters, it will only filter if the organization ID matches

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/ApplicationRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/ApplicationRepository.java
@@ -246,12 +246,12 @@ public class ApplicationRepository {
         return rowCount > 0;
     }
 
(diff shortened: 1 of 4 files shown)


MikelAlejoBR (approved):
LGTM! Nice work!

PR #2829, opened 2024-07-19
RHCLOUD-34047 | feature: incrase transaction timeouts for the bundles operations

MikelAlejoBR:
We are suffering from timeouts in production when we attempt to modify or delete an existing bundle. This is due to the "events" table being huge, and the "bundle_id" column not being indexed.

Adding an index to that column for this purpose seems overkill for now, especially because these endpoints don't get used very often.

## Jira ticket
[[RHCLOUD-34047]](https://issues.redhat.com/browse/RHCLOUD-34047)


g-duval (approved):
LGTM

gwenneg (approved):

PR #2841, opened 2024-07-25
[RHCLOUD-31596] Exclude eventType for default behavior groups

cam-rod:
## Jira ticket

https://issues.redhat.com/browse/RHCLOUD-31596

## Description

The `api/notifications/v1.0/notifications/eventTypes?excludeMutedTypes=true` query parameter is slightly modified, so that event types associated with default behavior groups are no longer returned. This change is slightly inconsistent with other endpoints, which typically do consider default behavior groups. However, this allows for the endpoint to return a list of event types which the user is able to mute.

##  Compatibility

**Breaks compatibility** with the previous implementation (#2820), as event types associated with default behaviour groups will no longer be returned.

## Testing

- Cases `testEventFetchingAndExcludeMutedTypes` and `testEventFetchingByBundleApplicationEventTypeNameAndExcludeMutedTypes` now verify that event types which are only associated with a default behavior group are not returned when `excludeMutedTypes` is true


gwenneg (approved):
LGTM

PR #2845, opened 2024-07-26
Change wording from systems to subscriptions

Tomboyo:
Update the wording of subscription errata notifications to say that subscriptions are impacted rather than systems.l


gwenneg (approved):
LGTM

PR #2868, opened 2024-08-07
RHCLOUd-34317 | refactor: correct the high volume Kafka topic names

MikelAlejoBR:
We have seen that things don't break if the topic name is not correct. The email connector shows lots of warnings, but it keeps processing the events. The engine does not complain about it.

This means that until the high volume topic is enabled in other environments, we should not suffer from errors or crashloops.

## Jira ticket
[[RHCLOUD-34317]](https://issues.redhat.com/browse/RHCLOUD-34317)


g-duval (approved):
LGTM

PR #2869, opened 2024-08-07
[RHCLOUD-33769] Create notifications-connector-pagerduty

jessicarod7:
## Jira issue

https://issues.redhat.com/browse/RHCLOUD-33769

## Description

This branch implements the `connector-pagerduty` module. It does not include relevant changes in the `backend` or `common` modules, or the addition of the CI/CD components, which will be included in later PRs.

Highlights:

- A new connector for sending [alert events](https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTgx-send-an-alert-event) to PagerDuty.
- The PagerDuty connector is adapted from both the webhook and ServiceNow connectors.
- The connector authenticates using a SECRET_TOKEN for the integration key (called a `routing_key` in the payload), pulled from Secrets.
- Messages are sent as a JSON payload via HTTP POSTs payloads to a common endpoint (PagerDuty uses the integration key to determine the recipient on their end.

### Payload schema

Based on the required [PD-CEF payload format](https://support.pagerduty.com/main/docs/pd-cef), I've laid out the following JSON schema for mapping new notifications, while retaining as much data as possible. The full implementation will be part of [RHCLOUD-33788](https://issues.redhat.com/browse/RHCLOUD-33788), but I'm open to feedback now, as any change would affect this branch as well.

```json5
{
  "payload": {
    "summary": "<event_type.name>",
    "timestamp": "2024-08-07T17:34:07.000+0000", // required timestamp format, must be UTC timezone
    "severity": "warning", // default to "warning" for now, may allow tenant or user to modify from ConsoleDot (TBD)
    "source": "<application.name>",
    "component": null,
    "group": "<bundle.name>",
    "class": null,
    "custom_details": {
      "account_id": "<account_id>",
      "org_id": "<org_id>",
      "source_names": {
        "application": "<application.display_name>",
        "bundle": "<bundle.display_name>",
        "event_type": "<event_type.display_name>"
      }
    }
  },
  "routing_key": "<integration_key from Secrets>",
  "event_action": "trigger", // also supports "acknowledge" and "resolve", not planning to add support for these
  "dedup_key": null, // not planning to support, interested tenants could always provide their own and inquire about us filling this field at a later date
  "client": "Open Hybrid Cloud Console",
  "client_url": "https://console.redhat.com/",
  "links": null,
  "images": null
}
```

The PagerDutyCloudEventDataExtractor will also extract metadata from the folllowing top-level JSON object:

```json5
{
  "payload": {
    // BaseTransformer.toJsonObject(Event);
  },
  "url": "<properties.url>", // this is always https://events.pagerduty.com/v2/enqueue
  "authentication": {
    "type": "SECRET_TOKEN", // SECRET_TOKEN authn must be used, and no other methods are accepted
    "secretId": "<properties.secretToken>"
  },
  "environment_url": "<environment.url>" // Used to set `client` and `client_url`
}
```

## Compatibility

No impact - at the current stage, the connector will not create any resources, and does not modify any existing code.

## Testing

- `PagerDutyCloudEventDataExtractorTest` validates that the extractor successfully retrieves the endpoint and authentication data from the payload and validates the URL
- `PagerDutyConnectorRoutesTest` checks that exchange properties are set, and that all required payload fields (`.event_action`, `.payload.summary`, `.payload.severity`, and `.payload.source`) are present.

## Other questions

The integration key must always be provided for PagerDuty alert events — is it appropriate to raise an exception in `AuthenticationProcessor` if it is not present, or should we let that be reported by the API? If yes, is there a way to inject a placeholder token into Secrets for the test cases?


connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyCloudEventDataExtractor.java
```diff
@@ -0,0 +1,90 @@
...
+        exchange.setProperty(TARGET_URL_NO_SCHEME, exchange.getProperty(TARGET_URL, String.class).replace("https://", ""));
+        exchange.setProperty(TRUST_ALL, Boolean.valueOf(metadata.getString("trustAll")));
+
+        JsonObject authentication = metadata.getJsonObject(AUTHENTICATION);
+        authenticationDataExtractor.extract(exchange, authentication);
+
+        cloudEventData.remove(NOTIF_METADATA);
+
+        exchange.getIn().setBody(cloudEventData.encode());
+        exchange.getIn().setHeader(Exchange.HTTP_METHOD, HttpMethods.valueOf(metadata.getString(METHOD)));
```
MikelAlejoBR, line 59: Since there is no other way to send requests to PagerDuty but using the `POST` method, I think we can use the `HttpMethods.POST` constant here.

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyCloudEventDataExtractor.java
```diff
@@ -0,0 +1,90 @@
...
+    public static final String METHOD = "method";
+    public static final String TRUST_ALL = "trustAll";
+    public static final String URL = "url";
+
+    public static final String PAYLOAD = "payload";
+    public static final String CUSTOM_DETAILS = "custom_details";
+    public static final String EVENT_ACTION = "event_action";
+
+    private static final UrlValidator HTTPS_URL_VALIDATOR = new UrlValidator(new String[]{"https"}, ALLOW_LOCAL_URLS);
+    private static final UrlValidator HTTP_URL_VALIDATOR = new UrlValidator(new String[]{"http"}, ALLOW_LOCAL_URLS);
```
MikelAlejoBR, line 34: For the rest of the connectors we make sure that the target URLs don't point to an internal IP, so I think we can remove the `ALLOW_LOCAL_URLS` setting from here.

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyRouteBuilder.java
```diff
@@ -0,0 +1,74 @@
...
+            .to(PAGERDUTY_RESPONSE_TIME_METRIC + TIMER_ACTION_START)
+                // SSL certificates may or may not be verified depending on the integration settings.
+                .choice()
+                .when(exchangeProperty(TRUST_ALL))
+                    .toD(buildPagerDutyEndpoint(true), connectorConfig.getEndpointCacheMaxSize())
+                .endChoice()
+                // TODO is this a valid option for PagerDuty?
+                .otherwise()
+                    .toD(buildPagerDutyEndpoint(false), connectorConfig.getEndpointCacheMaxSize())
+                .end()
```
MikelAlejoBR, line 57: Maybe we could ask AppSre to check if they think that having PagerDuty installations on an HTTP endpoint is common, or if we should simply force everyone to use HTTPS for the integration :thinking: Can you please verify that?

MikelAlejoBR (commented):
Good first pass! Just a few comments.

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyRouteBuilder.java
cam-rod: From my research, PagerDuty isn't available as a self-hosted option, which means that all requests must go to <https://events.pagerduty.com/v2/enqueue>. I can still follow up with them, but I think we can just drop it entirely.

This applies to a lot of the variables for PagerDuty, which I mentioned a bit in the **Payload schema** section. I've started working on that in [the other branch](https://github.com/cam-rod/notifications-backend/blob/fcc0ef6507771699a9cb290bebcc7675ebc314af/common/src/main/java/com/redhat/cloud/notifications/models/PagerDutyProperties.java#L27-L37) if you're curious.

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyCloudEventDataExtractor.java
cam-rod: Would it be preferable to set the constant here, or to set it in the Endpoint entry (linked at [the end of my other comment](https://github.com/RedHatInsights/notifications-backend/pull/2869#discussion_r1709508392))

MikelAlejoBR: If we are not going to use any other HTTP method I think you can definitely leave it as a constant in the connector. We have something similar for Splunk, where we directly specify the HTTP method for the call:

https://github.com/RedHatInsights/notifications-backend/blob/fab15dcff93f61b1fb8f940e2abe1826e073144e/connector-splunk/src/main/java/com/redhat/cloud/notifications/connector/splunk/SplunkRouteBuilder.java#L66-L74

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyRouteBuilder.java
MikelAlejoBR: Perfect, thank you for checking that out! Then yes, let's remove that branch.

cam-rod: Removed, and clarifying that the branch kept is `.toD(buildPagerDutyEndpoint(false), connectorConfig.getEndpointCacheMaxSize())` since it doesn't disable TLS verification

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyCloudEventDataExtractor.java
cam-rod: so fun story, removing this actually breaks 6 of the test cases in `PagerDutyConnectorRoutesTest`, because the mock server URL is `localhost` 🙃. Replacing the host with an IP address also fails because TLS certs can't be issued to IP addresses...

Adding an additional `Exchange` param to check if a test case is being run would work, but that feels icky so I was wondering what you think.

MikelAlejoBR: Yeah, now that I think about it I guess this should be validated whenever the user creates a new Pager Duty integration through the UI. That's what we do for the webhooks, for example.

cam-rod: I'll add it back in for now. I don't think users will be setting the URL themselves anyways since the same endpoint is used for everyone.

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyRouteBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+                "(orgId ${exchangeProperty." + ORG_ID + "} account ${exchangeProperty." + ACCOUNT_ID + "}) " +
+                "to ${exchangeProperty." + TARGET_URL + "}")
+            .to(direct(SUCCESS));
+    }
+
+    private HttpEndpointBuilder buildPagerDutyEndpoint() {
+        return https("${exchangeProperty." + TARGET_URL_NO_SCHEME + "}").httpMethod("POST")
+                .sslContextParameters(getSslContextParameters())
+                .x509HostnameVerifier(NoopHostnameVerifier.INSTANCE);
+    }
```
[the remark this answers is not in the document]
cam-rod: that's a mistake, I removed the wrong part of the method when I took out the TRUST_ALL property

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/ExchangeProperty.java
```diff
@@ -0,0 +1,8 @@
+package com.redhat.cloud.notifications.connector.pagerduty;
+
+// TODO: rewrite for PagerDuty
+public class ExchangeProperty {
+    public static final String ACCOUNT_ID = "accountId";
```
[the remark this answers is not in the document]
cam-rod: to clarify, is `ACCOUNT_ID` referring to the Red Hat account that creates the integration, or an account associated with the integration (ex. ServiceNow account ID)?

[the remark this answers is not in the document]
gwenneg: Yes, that is referring to the Red Hat account. The account ID used to be the only identifier we had for a Red Hat account, but then we introduced the org ID and made it the primary way to identify accounts. You'll find both account ID and org ID side by side in many places, but in most cases we only need the org ID.

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyCloudEventDataExtractor.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Inject
+    AuthenticationDataExtractor authenticationDataExtractor;
+
+    @Override
+    public void extract(Exchange exchange, JsonObject cloudEventData) throws IllegalArgumentException {
+
+        validatePayload(cloudEventData);
+
+        JsonObject customDetails = cloudEventData.getJsonObject(PAYLOAD).getJsonObject(CUSTOM_DETAILS);
```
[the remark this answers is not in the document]
cam-rod: sure, I can switch that over. I wrote an early version of the placeholder [in the other branch](https://github.com/cam-rod/notifications-backend/blob/fcc0ef6507771699a9cb290bebcc7675ebc314af/engine/src/main/java/com/redhat/cloud/notifications/transformers/PagerDutyTransformer.java), so I'll move it over.

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java
```diff
@@ -0,0 +1,108 @@
...
+        messagePayload.put(CUSTOM_DETAILS, customDetails);
+        message.put(PAYLOAD, messagePayload);
+
+        exchange.getIn().setBody(message.encode());
+    }
+
+    /**
+     * Adapted from CamelProcessor template for Teams
+     * <br>
+     * TODO should this be a template in notifications-engine for maintainability?
```
jessicarod7, line 84: [Here is the template](https://github.com/RedHatInsights/notifications-backend/blob/71bc5427d96431aa6928444890dde24c1b6d30a8/engine/src/test/java/com/redhat/cloud/notifications/processors/camel/teams/TeamsProcessorTest.java#L17-L22) I used to get the display text and URL.

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java
```diff
@@ -0,0 +1,108 @@
...
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Constructs a <a href="https://support.pagerduty.com/main/docs/pd-cef">PD-CEF</a> alert event.
+ * <br>
+ * The severity is set to {@link PagerDutySeverity#WARNING}, and the action to {@link PagerDutyEventAction#TRIGGER} for
+ * now. The following optional fields are not set (in jq format): <code>.payload.component, .payload.class, .dedup_key, .links[], .trigger[]</code>
+ * <br>
+ * <ul>
+ *     <li>TODO determine which details to include/remove</li>
```
jessicarod7, line 18: If there are any details we'd rather not include in the message format in the PR description, I can remove them now.

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java
```diff
@@ -0,0 +1,108 @@
...
+
+        JsonObject cloudEventPayload = cloudEventData.getJsonObject(PAYLOAD);
+        JsonObject message = new JsonObject();
+        message.put(EVENT_ACTION, PagerDutyEventAction.TRIGGER);
+        message.mergeIn(getClientLink(cloudEventPayload, cloudEventData.getString(ENVIRONMENT_URL)));
+
+        JsonObject messagePayload = new JsonObject();
+        messagePayload.put(SUMMARY, cloudEventPayload.getString(EVENT_TYPE));
+        messagePayload.put(TIMESTAMP, LocalDateTime.parse(cloudEventPayload.getString(TIMESTAMP)).format(PD_DATE_TIME_FORMATTER));
+        // TODO is it possible for tenant apps to indicate what severity to use?
```
jessicarod7, line 60:
```suggestion
```

This might be a question for PM, so we could leave it like this for now.

jessicarod7 (commented):
Still need to push some updates to the test case, but the transformer has been moved to the connector. There are a few points I'm a bit unsure about that I tagged with TODOs.

MikelAlejoBR (approved):

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyCloudEventDataExtractor.java
```diff
@@ -0,0 +1,53 @@
...
+import static org.apache.commons.validator.routines.UrlValidator.ALLOW_LOCAL_URLS;
+
+@ApplicationScoped
+public class PagerDutyCloudEventDataExtractor extends CloudEventDataExtractor {
+
+    public static final String AUTHENTICATION = "authentication";
+    public static final String URL = "url";
+
+    // HTTP URLs (or disabling SSL verification) must be permitted to run test cases
+    private static final UrlValidator URL_VALIDATOR = new UrlValidator(new String[]{"http", "https"}, ALLOW_LOCAL_URLS);
```
gwenneg, line 23: If we're talking about unit tests, it should be possible to use `https` only. See this:
https://github.com/RedHatInsights/notifications-backend/blob/f313e719e473be0e79db9c93e210fee845d95224/connector-splunk/src/test/java/com/redhat/cloud/notifications/connector/splunk/SplunkConnectorRoutesTest.java#L100-L103
The tests from connector-pagerduty can be improved later, no need to change everything in this PR.

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyCloudEventDataExtractor.java
```diff
@@ -0,0 +1,53 @@
...
+
+        exchange.setProperty(TARGET_URL, cloudEventData.getString(URL));
+
+        JsonObject authentication = cloudEventData.getJsonObject(AUTHENTICATION);
+        authenticationDataExtractor.extract(exchange, authentication);
+
+        cloudEventData.remove(URL);
+        cloudEventData.remove(AUTHENTICATION);
+
+        exchange.getIn().setBody(cloudEventData.encode());
```
gwenneg, line 41: Instead of doing this, we could set the `exchange` body to `cloudEventData.getJsonObject(PAYLOAD)`. We wouldn't need to remove `URL` and `AUTHENTICATION`. `PagerDutyTransformer` would also be able to consume `exchange.getIn().getBody(JsonObject.class)` as the payload.

The current code is fine for now. It can be improved in a subsequent PR.

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java
gwenneg: Here's a template used in prod:
```
{#if data.context.display_name??}
<{data.environment_url}{#if data.bundle == "openshift" and data.application == "advisor"}/openshift/insights/advisor/clusters/{data.context.display_name}{#else}/insights/inventory/{#if data.context.inventory_id??}{data.context.inventory_id}{#else}?hostname_or_id={data.context.display_name}{/if}{/if}|{data.context.display_name}> triggered {data.events.size()} event{#if data.events.size() > 1}s{/if}{#else}
{data.events.size()} event{#if data.events.size() > 1}s{/if} triggered{/if} from {data.source.application.display_name} - {data.source.bundle.display_name}. <{data.environment_url}/{#if data.bundle == "openshift"}openshift/{/if}{#if data.application == "integrations"}settings/{#else}insights/{/if}{data.application}|Open {data.source.application.display_name}>
```
The URL will need more work but that can be done in a later PR.

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java
gwenneg: Yes, let's discuss that in Slack.

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyCloudEventDataExtractor.java
jessicarod7: It is possible to run it with only HTTPS, but it also requires setting `ALLOW_LOCAL_URLS`. Discussed it with @MikelAlejoBR  in [this thread](https://github.com/RedHatInsights/notifications-backend/pull/2869#discussion_r1711860897), I'm happy to configure it either way since HTTP won't be supported by PagerDuty anyways.

gwenneg:
@jessicarod7 Could you please fix the test failures? Then we'll merge the PR.

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyCloudEventDataExtractor.java
jessicarod7: There's one other field in the cloud event that needs to be set, `.environment_url`. I might move it to the payload to try to fix test case issues.

jessicarod7: and it turns out the encoding was breaking my test cases, so thanks for the tip 😅

jessicarod7:
test cases resolved, codecov lines highlighted above look appropriately covered.

gwenneg (approved):
LGTM

All unresolved comments will be taken care of in a follow-up PR.

PR #2874, opened 2024-08-09
RHCLOUD-34485 | Integrate Notifications with the Kessel inventory

MikelAlejoBR:
## Jira ticket

[[RHCLOUD-34485]](https://issues.redhat.com/browse/RHCLOUD-34485)


backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -359,6 +363,8 @@ public EndpointDTO createEndpoint(
         @RequestBody(required = true)   EndpointDTO endpointDTO
     ) {
         if (this.backendConfig.isKesselBackendEnabled()) {
+            this.kesselAssets.createIntegration(sec, UUID.randomUUID().toString(), UUID.randomUUID().toString());
+
             this.kesselAuthorization.hasPermissionOnResource(sec, WorkspacePermission.INTEGRATIONS_CREATE, ResourceType.WORKSPACE, WORKSPACE_ID_PLACEHOLDER);
 
             return this.internalCreateEndpoint(sec, endpointDTO);
```
alechenninger, line 370: Couple of things to think about here:

- These are in the wrong order – you want to make sure the user is authorized before calling inventory.
- In light of this, we are thinking about having inventory perform this permissions check on clients' behalf. In that case you would only need to call createIntegration, and you would get a 403 if not allowed. You wouldn't need to make a second call (`hasPermissionOnResource`)

alechenninger: Actually, looks like you have a `createIntegration` call inside `internalCreateEndpoint`. This is a better spot for it IMO.

backend/src/main/java/com/redhat/cloud/notifications/auth/kessel/KesselAuthorization.java
```diff
@@ -113,7 +109,7 @@ public void hasPermissionOnResource(final SecurityContext securityContext, final
         }
 
         // Stop the timer.
-        permissionCheckTimer.stop(this.meterRegistry.timer(KESSEL_METRICS_PERMISSION_CHECK_TIMER_NAME, Tags.of(KESSEL_METRICS_TAG_PERMISSION_KEY, permission.getKesselPermissionName(), KESSEL_METRICS_TAG_RESOURCE_TYPE_KEY, resourceType.getKesselName())));
+        permissionCheckTimer.stop(this.meterRegistry.timer(KESSEL_METRICS_PERMISSION_CHECK_TIMER_NAME, Tags.of(KESSEL_METRICS_TAG_PERMISSION_KEY, permission.getKesselPermissionName(), Constants.KESSEL_METRICS_TAG_RESOURCE_TYPE_KEY, resourceType.getKesselName())));
```
alechenninger, line 112: Would you be interested in a client that integrated with micrometer out of the box? (so you didn't have to do your own wrapping timers all overal?)

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
MikelAlejoBR: Yes, the first one was a test to try the connection to the inventory. I've removed it and kept the call in the `internal` method. Once the inventory API does the authorization for us we can also remove the call to the relationships API.

backend/src/main/java/com/redhat/cloud/notifications/auth/kessel/KesselAuthorization.java
MikelAlejoBR: As stated in the meeting, it would be nice as long as it is configurable, but I don't think we mind having to implement these calls ourselves either. So for now let's keep them and we'll come back to them.

backend/src/main/java/com/redhat/cloud/notifications/auth/kessel/KesselAssets.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Creates an integration in Kessel's inventory.
+     * @param securityContext the security context which holds the subject of
+     *                        the operation.
+     * @param workspaceId the identifier for the workspace the integration will
+     *                    be created on.
+     * @param integrationId the identifier of the integration we created in
+     *                      our side.
+     */
+    public void createIntegration(final SecurityContext securityContext, final String workspaceId, final String integrationId) {
```
gwenneg, line 49: Unless I missed something, `securityContext` is only used for logging purposes in this method and the other one. What is the point of logging the identity if it's not part of the request sent to Kessel?

backend/src/main/java/com/redhat/cloud/notifications/auth/kessel/KesselAssets.java
```diff
@@ -0,0 +1,154 @@
...
+        Log.tracef("[identity: %s][workspace_id: %s][integration_id: %s] Payload for the integration creation in Kessel's inventory: %s", SecurityContextUtil.extractRhIdentity(securityContext), workspaceId, integrationId, request);
+
+        // Measure the time it takes to perform the operation with Kessel.
+        final Timer.Sample createIntegrationTimer = Timer.start(this.meterRegistry);
+
+        // Send the request to the inventory.
+        final CreateNotificationsIntegrationResponse response;
+        try {
+            response = this.notificationsIntegrationClient.CreateNotificationsIntegration(request);
+        } catch (final StatusRuntimeException e) {
```
gwenneg, line 62: We could catch `Exception` here.

backend/src/main/java/com/redhat/cloud/notifications/auth/kessel/KesselAssets.java
```diff
@@ -0,0 +1,154 @@
...
+        Log.tracef("[identity: %s][workspace_id: %s][integration_id: %s] Payload for deleting the integration in Kessel's inventory: %s", SecurityContextUtil.extractRhIdentity(securityContext), workspaceId, integrationId, request);
+
+        // Measure the time it takes to perform the operation with Kessel.
+        final Timer.Sample deleteIntegrationTimer = Timer.start(this.meterRegistry);
+
+        // Send the request to the inventory.
+        final DeleteNotificationsIntegrationResponse response;
+        try {
+            response = this.notificationsIntegrationClient.DeleteNotificationsIntegration(request);
+        } catch (final StatusRuntimeException e) {
```
gwenneg, line 101: We could catch `Exception` here.

gwenneg: I suppose `StatusRuntimeException` is used to let us know that something went wrong in Kessel while processing our request. Is that correct?

backend/src/main/resources/application.properties
```diff
@@ -102,4 +102,5 @@ quarkus.unleash.active=false
 quarkus.unleash.url=http://localhost:4242
 
 # Kessel
-notifications.kessel.target-url=localhost:9000
+notifications.kessel.inventory-api.target-url=localhost:9080
+notifications.kessel.relationships-api.target-url=localhost:9000
```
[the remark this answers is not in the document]
MikelAlejoBR: That's nice! The next PR also refactors this so that we use the client libraries' configuration values directly, because it's just easier to use their config for the providers.

backend/src/main/java/com/redhat/cloud/notifications/auth/kessel/producer/inventory/KesselInventoryClientsProducer.java
```diff
@@ -0,0 +1,81 @@
...
+import org.project_kessel.client.NotificationsIntegrationClient;
+
+/**
+ * We have implemented our own producer for the clients because we need our
+ * clients to be "@ApplicationScoped" so that we can mock them in our tests.
+ * The default producer from the library gives us non-mockable and dependent
+ * beans, which we cannot use in our tests.
+ */
+@ApplicationScoped
+@Priority(1)
```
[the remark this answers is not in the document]
MikelAlejoBR: Yes, if I don't add the priority the `@Inject` annotations seem to inject the beans from the libraries instead of our providers, and since we don't have the libraries' configuration properties set up, the tests break.

backend/src/main/java/com/redhat/cloud/notifications/auth/kessel/producer/relationships/KesselRelationshipsClientsProducer.java
```diff
@@ -13,43 +13,54 @@
...
 @ApplicationScoped
 @Priority(1)
-public class KesselClientsProducer {
-    private static final String KESSEL_TARGET_URL = "notifications.kessel.target-url";
+public class KesselRelationshipsClientsProducer {
+    /**
+     * Constant for the configuration key that defines the relationships API's
+     * target URL.
+     */
+    private static final String RELATIONSHIPS_API_TARGET_URL = "notifications.kessel.relationships-api.target-url";
```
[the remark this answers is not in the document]
MikelAlejoBR: In the next PR this line is removed in favor of the libraries' configuration options.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -431,7 +435,16 @@ protected Endpoint internalCreateEndpoint(
 
         this.secretUtils.createSecretsForEndpoint(endpoint);
 
-        return this.endpointRepository.createEndpoint(endpoint);
+        final Endpoint createdEndpoint = this.endpointRepository.createEndpoint(endpoint);
+
+        // Attempt creating the integration in Kessel's inventory. Any
+        // exception here would roll back the operation and our integration
+        // would not be created in our database either.
```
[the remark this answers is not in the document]
MikelAlejoBR: Yes, any exception thrown will be caught by this try catch block:

```java
try {
    return this.endpointMapper.toDTO(
        this.internalCreateEndpoint(securityContext, endpoint)
    );
} catch (final Exception e) {
    // Clean up the secrets from Sources if any were created.
    this.secretUtils.deleteSecretsForEndpoint(endpoint);

    throw e;
}
```

So:

- If Hibernate throws an exception, nothing has yet been created in Kessel so we're fine.
- If Kessel throws an exception, the endpoint's save operation is rolled back due to the `@Transactional` annotation.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -574,7 +587,14 @@ public Response deleteEndpoint(@Context SecurityContext sec, @PathParam("id") UU
         if (this.backendConfig.isKesselBackendEnabled()) {
             this.kesselAuthorization.hasPermissionOnResource(sec, IntegrationPermission.DELETE, ResourceType.INTEGRATION, id.toString());
 
-            return this.internalDeleteEndpoint(sec, id);
+            final Response noContentResponse = this.internalDeleteEndpoint(sec, id);
+
+            // Attempt deleting the integration from Kessel. If any exception
+            // is thrown the whole transaction will be rolled back and the
+            // integration will not be deleted from our database.
```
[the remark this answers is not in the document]
MikelAlejoBR: Here what we do is:

* Delete the secrets in Sources.
* Delete the endpoint in our database.
* Delete the integration from Kessel if no exception was thrown.

It is true that we might have to switch the order for the Sources' secrets deletion, so I've created a ticket for it: [[RHCLOUD-35158]](https://issues.redhat.com/browse/RHCLOUD-35158)

backend/src/main/java/com/redhat/cloud/notifications/auth/kessel/KesselAssets.java
MikelAlejoBR: This is a great question. Actually I think we are going to be creating the resources on the user's behalf, but I am unsure if we will need to specify those details in the future. Let me ask the Kessel team.

backend/src/main/java/com/redhat/cloud/notifications/auth/kessel/KesselAssets.java
MikelAlejoBR: That is correct, yes. I think we can modify that and set the generic `Exception` catch because exception mapper will catch the particular `StatusRuntimeException` and handle it properly anyway.

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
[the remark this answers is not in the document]
gwenneg: Perfect, thanks! :+1:

backend/src/main/java/com/redhat/cloud/notifications/auth/kessel/KesselAssets.java
gwenneg: Let's make the changes (if any) in a follow-up PR.

gwenneg (approved):
LGTM

PR #2877, opened 2024-08-13
RHCLOUD-34459 | feature: migration resource for Errata subscriptions

MikelAlejoBR:
We need to migrate all the customer's subscriptions from the Errata's legacy system to our database, and for that, we are going to use a resource that allows us to upload either a CSV or a JSON file with the subscription contents.

We will subscribe every given user to all the Errata event types.

## Jira ticket

[[RHCLOUD-34459]](https://issues.redhat.com/browse/RHCLOUD-34459)


backend/pom.xml
```diff
@@ -158,6 +158,13 @@
             <version>${kessel.version}</version>
         </dependency>
 
+        <!-- Temporary dependency for the Errata subscriptions' migration service -->
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-csv</artifactId>
```
gwenneg, line 164: We talked about using JSON in Slack, so I guess this won't be used.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataMigrationRepository.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * Saves the given Errata subscriptions in the database.
+     * @param errataSubscriptions the list of the Errata subscriptions to save
+     *                            in the database.
+     */
+    @Transactional
+    protected void saveErrataSubscriptions(final List<ErrataSubscription> errataSubscriptions) {
+        Log.infof("Errata subscriptions' migration begins with a batch size of %s", INSERTION_BATCH_SIZE);
```
gwenneg, line 58: Should we empty the table first in case we need to run the migration more than once?

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataUserPreferencesMigrationResource.java
```diff
@@ -0,0 +1,96 @@
...
+public class ErrataUserPreferencesMigrationResource {
+    private static final String USERNAME_KEY = "username";
+    private static final String ORG_ID_KEY = "org_id";
+    private static final String[] CSV_HEADERS = {USERNAME_KEY, ORG_ID_KEY};
+
+    @Inject
+    ErrataMigrationRepository errataMigrationRepository;
+
+    @Consumes(MediaType.MULTIPART_FORM_DATA)
+    @Path("/migrate/csv")
```
gwenneg, line 39: As discussed in Slack with Teamnado, it seems we won't need this endpoint.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataUserPreferencesMigrationResource.java
```diff
@@ -0,0 +1,96 @@
...
+        } catch (final Exception e) {
+            Log.error("Unable to migrate errata subscriptions due to an exception. The insertions were rolled back", e);
+        }
+    }
+
+    @Consumes(MediaType.MULTIPART_FORM_DATA)
+    @Path("/migrate/json")
+    @POST
+    public void migrateTeamNadoUserPreferencesJSON(@NotNull @RestForm("jsonFile") InputStream jsonFile) throws IOException {
+        final byte[] contents = jsonFile.readAllBytes();
```
gwenneg, line 75: Do we have an estimate of the file size already? Loading all bytes in memory could be an issue if it's too big.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataMigrationRepository.java
```diff
@@ -0,0 +1,104 @@
...
+     */
+    private static final int INSERTION_BATCH_SIZE = 500;
+
+    /**
+     * The internal name of the Errata application in our database.
+     */
+    public static final String ERRATA_APPLICATION_NAME = "errata-notifications";
+
+    @Inject
+    EntityManager entityManager;
```
gwenneg, line 30: We could use a `StatelessSession` in this class:
```java
@Inject
StatelessSession statelessSession;
```
https://thorben-janssen.com/hibernates-statelesssession/

backend/pom.xml
MikelAlejoBR: Removed!

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataMigrationRepository.java
MikelAlejoBR: Hmmm... If we clean up the table wouldn't we delete the email subscriptions that existed before the migration? I was thinking that maybe we can do an "ON CONFLICT DO NOTHING" insert so that we can safely run it multiple times. Or manually check for that in the code and log a "duplicated email subscription found. Doing nothing" kind of thing too.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataUserPreferencesMigrationResource.java
MikelAlejoBR: Removed!

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataUserPreferencesMigrationResource.java
MikelAlejoBR: No, we could be talking about 20,000 subscriptions, so it might be a few megs of size, I guess. The thing is that if we use a JSON file I don't think we can simply read line by line...

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataMigrationRepository.java
MikelAlejoBR: I have tried that but when executing the code the `this.statelessSession.begin()` statement gives me the following error:

```
java.lang.IllegalStateException: Not supported for JTA entity managers
```

I have been trying to find information about this but with no luck. Does it complain because we are using `ResourceHelpers` that are creating stuff via the `EntityManager`? Do I have to manually create stuff for this instead? :thinking:

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataUserPreferencesMigrationResource.java
gwenneg: Good point. I wrote my comment with the CSV file in mind, but we don't have a choice with JSON.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataMigrationRepository.java
gwenneg: Emptying the table is indeed not an option. Why did I even suggest that? :sweat_smile:

Anyway, being able to run the migration more than once would be great. I'm fine with ignoring records that are already in the DB.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataUserPreferencesMigrationResource.java
```diff
@@ -0,0 +1,58 @@
...
+
+        // Extract the subscriptions from the JSON file.
+        final String stringContents = new String(contents, StandardCharsets.UTF_8);
+        final JsonArray jsonArray = new JsonArray(stringContents);
+
+        final List<ErrataSubscription> errataSubscriptions = new ArrayList<>();
+        for (final Object rawJsonObject : jsonArray) {
+            final JsonObject teamNadoSubscription = (JsonObject) rawJsonObject;
+
+            errataSubscriptions.add(new ErrataSubscription(teamNadoSubscription.getString(USERNAME_KEY), teamNadoSubscription.getString(ORG_ID_KEY)));
```
g-duval, line 47: according https://issues.redhat.com/browse/RHCLOUD-30812, the current portal allow user to select the equivalent of or event type, I think we must honor those choices.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataMigrationRepository.java
```diff
@@ -0,0 +1,100 @@
+package com.redhat.cloud.notifications.routers.internal.errata;
```
g-duval, line 1: Even if it's temporary, this bean should be part of `com.redhat.cloud.notifications.db.repositories` package.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataUserPreferencesMigrationResource.java
Tomboyo: If it helps, we could intentionally break the data dump into multiple files. (It's also possible to stream over the elements of a json array if you have jackson on your class path [[example](https://stackoverflow.com/a/24840488)].)

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataMigrationRepository.java
MikelAlejoBR: Perfect, changed that!

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataMigrationRepository.java
MikelAlejoBR: Done!

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataMigrationRepository.java
MikelAlejoBR: Moved!

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataUserPreferencesMigrationResource.java
MikelAlejoBR: Done too!

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataUserPreferencesMigrationResource.java
MikelAlejoBR: All right, refactored! Now we process the JSON file in batches and we create the corresponding objects at that time, by cleaning them up afterwards so that we don't use as much memory.

g-duval (approved):
LGTM

backend/src/test/java/com/redhat/cloud/notifications/routers/internal/errata/ErrataUserPreferencesMigrationResourceTest.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+
+     /**
+     * Tests that uploading a JSON file with the Errata subscriptions creates
+     * the corresponding subscriptions in our database.
+     * @throws URISyntaxException if the JSON file's URL is not properly built.
+     */
+    @Test
+    void testJsonMigration() throws URISyntaxException {
```
gwenneg, line 96: Could you please set `BATCH_SIZE` to `2` from the test so that the whole `while` loop from `ErrataUserPreferencesMigrationResource` is tested?

MikelAlejoBR: Yes, and I've added an environment variable so that we can configure it in case we need to reduce the batch size due to memory constraints.

gwenneg (approved):
LGTM

PR #2887, opened 2024-08-21
RHCLOUD-34728 | WIP | feature: authentication on the Kessel clients

MikelAlejoBR:
Adds authentication to the Kessel clients so that our calls are authenticated.

## Dependency

- https://github.com/RedHatInsights/notifications-backend/pull/2874

## Jira ticket
[[RHCLOUD-34728]](https://issues.redhat.com/browse/RHCLOUD-34728)


backend/pom.xml
```diff
@@ -163,6 +163,14 @@
             <version>${kessel.relationships-client.version}</version>
         </dependency>
 
+        <!-- oAuth library for the gRPC clients -->
+        <dependency>
+            <groupId>com.nimbusds</groupId>
+            <artifactId>oauth2-oidc-sdk</artifactId>
+            <version>${kessel.oauth-client.version}</version>
+            <scope>provided</scope>
```
gwenneg, line 171: Why is it `provided`?

.rhcicd/clowdapp-backend.yaml
```diff
@@ -142,6 +138,18 @@ objects:
           value: ${INTERNAL_RBAC_ENABLED}
         - name: INTERNAL_ADMIN_ROLE
           value: ${INTERNAL_RBAC_ADMIN}
+        - name: INVENTORY_API_AUTHN_CLIENT_ID
+          valueFrom:
+            secretKeyRef:
+              name: kessel-authentication
+              key: inventory.client.id
```
gwenneg, line 145: `inventory.client.id` / `relations-api.client.id`

Could make this consistent? Either `-api` in both or in none.

.rhcicd/clowdapp-backend.yaml
```diff
@@ -142,6 +138,18 @@ objects:
...
+        - name: INVENTORY_API_AUTHN_CLIENT_ISSUER
+          value: ${NOTIFICATIONS_KESSEL_OIDC_ISSUER}
+        - name: INVENTORY_API_AUTHN_CLIENT_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: kessel-authentication
+              key: inventory.client.secret
```
gwenneg, line 152: Same comment about consistency.

backend/src/main/resources/application.properties
```diff
@@ -103,5 +103,16 @@ quarkus.unleash.active=false
 quarkus.unleash.url=http://localhost:4242
 
 # Kessel
-notifications.kessel.inventory-api.target-url=localhost:9080
-notifications.kessel.relationships-api.target-url=localhost:9000
+inventory-api.authn.client.id=insights-notifications
+inventory-api.authn.client.issuer=http://localhost:8084/realms/redhat-external
```
gwenneg, line 107: Does this URL help with local testing? Or could we use `https://redhat.com/realms/redhat-external` here?

backend/pom.xml
MikelAlejoBR: I think I copied and pasted what the team gave us for the library. I see that [they mark the library as `provided`](https://github.com/project-kessel/relations-client-java/pull/17/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R91-R96) in their code because they expect us to provide it, so I assume that we could use the default one or `runtime`. I went for the default one, but I'm open to changing it.

.rhcicd/clowdapp-backend.yaml
MikelAlejoBR: Yes, changed!

.rhcicd/clowdapp-backend.yaml
MikelAlejoBR: Changed!

backend/src/main/resources/application.properties
MikelAlejoBR: Yes, it does help with local testing if you set up Keycloak locally, but... I'm not sure if we should change it to the URL you pointed :thinking:

gwenneg: If it helps with local testing, then the current value is fine.

gwenneg (approved):
LGTM

PR #2892, opened 2024-08-26
RHCLOUD-34808 | feature: reduce event retention days

MikelAlejoBR:
We are reducing the event retention to 15 days to be able to handle the huge incoming load that we are about to face due to the TeamNado integration. The UI does not show more than 14 days, so we can safely reduce the event retention to 15 days. This way, we can be sure that all the timezones across the globe will have the 14 days of events available to be able to check them from the UI.

## Jira ticket

[[RHCLOUD-34808]](https://issues.redhat.com/browse/RHCLOUD-34808)


gwenneg (approved):
LGTM

PR #2903, opened 2024-08-28
[RHCLOUD-34233] Disable integration in case of SSL Exception

g-duval:


connector-common-http/src/main/java/com/redhat/cloud/notifications/connector/http/HttpExceptionProcessor.java
```diff
@@ -61,6 +63,8 @@ protected void process(Throwable t, Exchange exchange) {
             exchange.setProperty(HTTP_ERROR_TYPE, SSL_HANDSHAKE);
         } else if (t instanceof UnknownHostException) {
             exchange.setProperty(HTTP_ERROR_TYPE, UNKNOWN_HOST);
+        } else if (t instanceof SSLException) {
```
gwenneg, line 66: Can `SSLException` be thrown for another reason than `Unsupported or unrecognized SSL message`? Should we change the condition to this?
```suggestion
        } else if (t instanceof SSLException && "Unsupported or unrecognized SSL message".equals(t.getMessage())) {
```

gwenneg (approved):
LGTM

PR #2916, opened 2024-09-03
RHCLOUD-34317 | refactor: move the high volume topic env var to Unleash

MikelAlejoBR:
The reason for the change is that this way we will be able to manually control when we enable and disable the production and consumption of messages in that topic.

## Jira ticket
[[RHCLOUD-34317]](https://issues.redhat.com/browse)


g-duval (approved):
LGTM (build error is not related to this PR)

PR #2919, opened 2024-09-04
RHCLOUD-28271 - Add backend as optional dependecy for testing

Fynardo:
Backend set as an optional dependency will make clowder fill its details on the IQE pod while not modifying the deployment


g-duval (approved):
LGTM

MikelAlejoBR (approved):

PR #2928, opened 2024-09-09
Fine-tune microdnf commands used while building images

gwenneg:
The images build is currently unstable (timeouts). I'm trying to fix that.

- `microdnf update` is an alias for `microdnf upgrade` which is the actual command
- `--setopt=install_weak_deps=0`: prevents microdnf from installing additional packages (weak dependencies) we don't need
- `-y`: automatically answer yes for all questions


g-duval (commented):
LGTM 
/konflux

PR #2934, opened 2024-09-11
RHCLOUD-35145 | feature: Disable microprofile fault tolerance metrics

MikelAlejoBR:
AppSre has pinged us because apparently some of those metrics are making the production instance of Prometheus to crash. In order to avoid it, they've asked us to remove them.

According to the [Microprofile docs](https://download.eclipse.org/microprofile/microprofile-fault-tolerance-3.0/microprofile-fault-tolerance-spec-3.0.html#_configuring_metrics_integration), there is no fine tuning the metrics. Either we enable them all or we disable them.

Since we are not using those metrics, we can simply disable them for now.

## Jira ticket
[[RHCLOUD-35145]](https://issues.redhat.com/browse/RHCLOUD-35145)

### DIFF
--- a/backend/src/main/resources/application.properties
+++ b/backend/src/main/resources/application.properties
@@ -103,3 +103,7 @@ quarkus.unleash.url=http://localhost:4242
 
 # Kessel
 notifications.kessel.target-url=localhost:9000
+
+# Disable the fault tolerance metrics for the MicroProfile REST Clients. The
+# reason for this can be found in the following ticket: https://issues.redhat.com/browse/RHCLOUD-35145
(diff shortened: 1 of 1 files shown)


gwenneg (changes requested):
The analysis from App SRE looks wrong to me. See the Slack thread for more details.

gwenneg:
As discussed in Slack, this PR shouldn't be needed so I'm closing it.

Thanks @MikelAlejoBR for quickly providing a fix for the issue reported by App SRE!

MikelAlejoBR:
Thank you for investigating it! It was a nice catch :+1:

PR #2943, opened 2024-09-13
RHCLOUD-34459 | fix username json key

MikelAlejoBR:
* The JSON key for the username is "user_name", and we were attempting to
parse the "username" key, which was causing exceptions.
* The exceptions are supposed to rollback the entire operation, so it is
simply safe to reuse the "try with resources" statement's "catch" block
to roll back the exception if any error occurs.

## Jira ticket
[[RHCLOUD-34459 ]](https://issues.redhat.com/browse/RHCLOUD-34459)


g-duval (approved):
LGTM

PR #2981, opened 2024-09-20
[RHCLOUD-34490] Auto manage bg when endpoint to event type api is use

g-duval:


backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
```diff
@@ -498,6 +499,29 @@ public void appendBehaviorGroupToEventType(final String orgId, final UUID behavi
...
 
+    @Transactional
+    public void appendActionToBehaviorGroup(final String orgId, final UUID behaviorGroupUuid, final UUID endpointUuid, final int position) {
+        final String appendBehaviorGroupQuery =
+            "INSERT INTO " +
+                "behavior_group_action(behavior_group_id, endpoint_id, created, position) " +
+                "SELECT bg.id, ep.id, :created, :position " +
+                "FROM behavior_group AS bg, endpoints AS ep " +
+                "WHERE ep.id = :endpointUuid AND ep.org_id = :orgId AND bg.id = :behaviorGroupUuid AND bg.org_id = :orgId " +
+                "ON CONFLICT (behavior_group_id, endpoint_id) DO NOTHING";
```
gwenneg, line 510: Does the position need to be updated in case of conflict?
```suggestion
                "ON CONFLICT (behavior_group_id, endpoint_id) DO UPDATE SET position = :position";
```

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
```diff
@@ -498,6 +499,29 @@ public void appendBehaviorGroupToEventType(final String orgId, final UUID behavi
...
     }
 
+    @Transactional
+    public void appendActionToBehaviorGroup(final String orgId, final UUID behaviorGroupUuid, final UUID endpointUuid, final int position) {
+        final String appendBehaviorGroupQuery =
+            "INSERT INTO " +
+                "behavior_group_action(behavior_group_id, endpoint_id, created, position) " +
+                "SELECT bg.id, ep.id, :created, :position " +
+                "FROM behavior_group AS bg, endpoints AS ep " +
+                "WHERE ep.id = :endpointUuid AND ep.org_id = :orgId AND bg.id = :behaviorGroupUuid AND bg.org_id = :orgId " +
```
gwenneg, line 509: Will that be used with default behavior groups? If so, we may need additional conditions with `org_id IS NULL`.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
```diff
@@ -498,6 +499,29 @@ public void appendBehaviorGroupToEventType(final String orgId, final UUID behavi
         }
     }
 
+    @Transactional
+    public void appendActionToBehaviorGroup(final String orgId, final UUID behaviorGroupUuid, final UUID endpointUuid, final int position) {
+        final String appendBehaviorGroupQuery =
+            "INSERT INTO " +
+                "behavior_group_action(behavior_group_id, endpoint_id, created, position) " +
+                "SELECT bg.id, ep.id, :created, :position " +
+                "FROM behavior_group AS bg, endpoints AS ep " +
```
gwenneg, line 508: This query creates a [cartesian product](https://www.tutorialspoint.com/sql/sql-cross-join.htm) of the `behavior_group` and `endpoints` tables. This would solve the issue:
```sql
INSERT INTO behavior_group_action(behavior_group_id, endpoint_id, created, position)
SELECT :behaviorGroupUuid, :endpointUuid, :created, :position
WHERE EXISTS (SELECT 1 FROM behavior_group bg WHERE bg.id = :behaviorGroupUuid AND bg.org_id = :orgId)
AND EXISTS (SELECT 1 FROM endpoints ep WHERE ep.id = :endpointUuid AND ep.org_id = :orgId)
ON CONFLICT (behavior_group_id, endpoint_id) DO NOTHING
```

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -938,6 +952,18 @@ protected void legacyRBACDeleteEventTypeFromEndpoint(final SecurityContext secur
     private void internalDeleteEventTypeFromEndpoint(final SecurityContext securityContext, final UUID eventTypeId, final UUID endpointId) {
         final String orgId = getOrgId(securityContext);
         endpointEventTypeRepository.deleteEndpointFromEventType(eventTypeId, endpointId, orgId);
+
+        // Sync behavior group model
+        List<BehaviorGroup> behaviorGroupsLinkedToThisEndpoint = behaviorGroupRepository.findBehaviorGroupsByEndpointId(orgId, endpointId);
+        for (BehaviorGroup behaviorGroup : behaviorGroupsLinkedToThisEndpoint) {
+            if (!behaviorGroup.getBehaviors().isEmpty() && behaviorGroup.getBehaviors().stream().anyMatch(bg -> bg.getId().eventTypeId.equals(eventTypeId))) {
```
gwenneg, line 959: This isn't really needed:
```suggestion
            if (behaviorGroup.getBehaviors().stream().anyMatch(bg -> bg.getId().eventTypeId.equals(eventTypeId))) {
```

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -995,6 +1027,71 @@ public void legacyRbacUpdateEventTypesLinkedToEndpoint(final SecurityContext sec
...
+
+        // delete endpoint from existing behavior group
+        List<BehaviorGroup> behaviorGroupsLinkedToThisEndpoint = behaviorGroupRepository.findBehaviorGroupsByEndpointId(orgId, endpointId);
+        for (BehaviorGroup behaviorGroup : behaviorGroupsLinkedToThisEndpoint) {
+            if (behaviorGroup.getActions().size() == 1) {
+                behaviorGroupRepository.delete(orgId, behaviorGroup.getId());
+            } else {
+                behaviorGroupRepository.deleteEndpointFromBehaviorGroup(behaviorGroup.getId(), endpointId, orgId);
+            }
+        }
```
gwenneg, line 1042: This will affect behavior groups created by customers. Should we have a way (boolean field?) to distinguish them from the behavior groups that are internally created because we're transitioning away from behavior groups?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -995,6 +1027,71 @@ public void legacyRbacUpdateEventTypesLinkedToEndpoint(final SecurityContext sec
...
+
+        // Create or update relevant behavior groups
+        createOrUpdateLinkedBehaviorGroup(eventTypeIds, endpointId, updatedEndpoint.getName(), orgId);
+    }
+
+    private void createOrUpdateLinkedBehaviorGroup(Set<UUID> eventTypeIds, UUID endpointId, String endpointName, String orgId) {
+        String behaviorGroupName = String.format("Integration \"%s\" behaviour group", endpointName);
```
gwenneg, line 1049: Behavior group names have to be unique in prod (the feature toggle is on). Is there a chance this would collide with an existing name?

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -995,6 +1027,71 @@ public void legacyRbacUpdateEventTypesLinkedToEndpoint(final SecurityContext sec
...
+        for (UUID eventTypeId : eventTypeIds) {
+            Optional<Bundle> bundle = eventTypeRepository.findBundleByEventTypeId(eventTypeId);
+            if (bundle.isPresent()) {
+                if (eventTypesGroupedByBundle.containsKey(bundle.get().getId())) {
+                    eventTypesGroupedByBundle.get(bundle.get().getId()).add(eventTypeId);
+                } else {
+                    Set<UUID> eventTypes = new HashSet<>();
+                    eventTypes.add(eventTypeId);
+                    eventTypesGroupedByBundle.put(bundle.get().getId(), eventTypes);
+                }
```
gwenneg, line 1062:
```suggestion
                eventTypesGroupedByBundle.computeIfAbsent(bundle.get().getId(), ignored -> new HashSet<>())
                        .add(eventTypeId);
```

backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -995,6 +1027,71 @@ public void legacyRbacUpdateEventTypesLinkedToEndpoint(final SecurityContext sec
...
+                    }
+                }
+            } else {
+                // Create or update legacy behavior group structure
+                BehaviorGroup behaviorGroup = new BehaviorGroup();
+                behaviorGroup.setBundleId(bundleId);
+                behaviorGroup.setDisplayName(behaviorGroupName);
+
+                behaviorGroupRepository.createFull(
+                    null,
```
gwenneg, line 1088: I think we should provide the account ID value here until we decide to remove it from the DB table for everyone.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
g-duval: I refactored this query because it wasn't relevant.
I also added a unit test case for it.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
```diff
@@ -498,6 +499,26 @@ public void appendBehaviorGroupToEventType(final String orgId, final UUID behavi
         }
     }
 
+    @Transactional
+    public void appendActionToBehaviorGroup(final UUID behaviorGroupUuid, final UUID endpointUuid, final int position) {
+        final String appendBehaviorGroupQuery =
+            "INSERT INTO " +
```
gwenneg, line 505: We should check (here or in an upper layer) that `endpointUuid` is owned by the current `orgId`.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
```diff
@@ -537,6 +558,27 @@ public void deleteBehaviorGroupFromEventType(final UUID eventTypeUuid, final UUI
         }
     }
 
+    @Transactional
+    public void deleteEndpointFromBehaviorGroup(final UUID behaviorGroupUuid, final UUID endpointUuid, final String orgId) {
+        final var deleteFromEventTypeQuery =
+            "DELETE FROM BehaviorGroupAction AS bga " +
```
gwenneg, line 564: We should check (here or in an upper layer) that `endpointUuid` is owned by the current `orgId`.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
g-duval: This check is akready done by a sub method called by an upper layer who will throw and exception, but it's not obious from a mantenance perspective. You are right, lets check it here to avoid confusion.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/BehaviorGroupRepository.java
g-duval: This check is akready done by a sub method called by an upper layer who will throw and exception, but it's not obious from a mantenance perspective. You are right, lets check it here to avoid confusion.

gwenneg (approved):
LGTM

PR #2998, opened 2024-09-26
RHCLOUD-35435 | refactor: remove Kessel's Relations bean producer

MikelAlejoBR:
The Kessel team has updated the Relations' bean producer and it now produces "@ApplicationScoped" beans which are mockable in tests, so we no longer need to have our own producer for that particular bean.

## Jira ticket
[[RHCLOUD-35435]](https://issues.redhat.com/browse/RHCLOUD-35435)


gwenneg (approved):
LGTM

PR #3015, opened 2024-10-02
[RHCLOUD-34735] Add linked event types into endpoint get methods

g-duval:


common/src/main/java/com/redhat/cloud/notifications/models/dto/v1/endpoint/EndpointDTO.java
```diff
@@ -89,6 +91,9 @@ private boolean isSubTypeNotPresentWhenNotRequired() {
         return this.type.requiresSubType || this.subType == null;
     }
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    private Set<BundleDTO> eventTypes;
```
gwenneg, line 95: image

g-duval: UI team requested to have those event types grouped by bundle and app. But the name could be updated to something like `eventTypesByBundle` ?

g-duval:
handled by https://github.com/RedHatInsights/notifications-backend/pull/3028

PR #3016, opened 2024-10-02
RHCLOUD-35479 | feature: fetch workspace IDs from RBAC

MikelAlejoBR:
In order to be able to check on users' permissions with Kessel, we need to know which workspace they are trying to act upon. And for now, RBAC is going to provide that information.

## Jira ticket
[[RHCLOUD-35479]](https://issues.redhat.com/browse/RHCLOUD-35479)


g-duval (approved):
LGTM

PR #3019, opened 2024-10-03
[RHCLOUD-35533] don't include empty patch daily digest section

g-duval:
In some cases, Patch daily email section could be empty:
image

In those cases, it should not be included in the daily digest.


engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/PatchEmailPayloadAggregator.java
```diff
@@ -63,8 +63,6 @@ void processEmailAggregation(EmailAggregation notification) {
             return;
         }
 
-        JsonObject payloadContext = notificationPayload.getJsonObject(CONTEXT_KEY);
```
gwenneg, line 66: Was that unused?

gwenneg (approved):
LGTM

engine/src/main/java/com/redhat/cloud/notifications/processors/email/aggregators/PatchEmailPayloadAggregator.java
g-duval: No so I cleaned it :)

PR #3020, opened 2024-10-04
Downgrade base images to ubi8

g-duval:


gwenneg:
`Dockerfile.notifications-connector-pagerduty` will need to be downgraded too.

gwenneg:
What about the instructions such as
```
RUN rpm -e --nodeps platform-python-setuptools
```
which were removed in 2906. Do we need them back?

g-duval:
> What about the instructions such as
> 
> ```
> RUN rpm -e --nodeps platform-python-setuptools
> ```
> 
> which were removed in #2906. Do we need them back?

No it seems has been fixed on recent ubi8 images

gwenneg (approved):
LGTM

gwenneg:
Failures in Konflux are unrelated.

PR #3036, opened 2024-10-10
Add additional validation for PagerDutyTransformer

jessicarod7:
In testing I discovered that the `PagerDutyTransformer` (part of `notifications-connector-pagerduty`) failed to handle cases where optional fields were not provided. This PR fixes the handling of missing fields, and also adds a new `PagerDutyTransformerTest` class to validate it's handled gracefully.

### DIFF
--- a/connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java
+++ b/connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java
@@ -58,16 +58,12 @@ public class PagerDutyTransformer implements Processor {
         messagePayload.put(SUMMARY, cloudEventPayload.getString(EVENT_TYPE));
 
         String timestamp = cloudEventPayload.getString(TIMESTAMP);
-        if (timestamp != null) {
-            try {
-                messagePayload.put(TIMESTAMP, LocalDateTime.parse(timestamp).format(PD_DATE_TIME_FORMATTER));
-            } catch (DateTimeParseException e) {
-                Log.warnf(e, "Unable to parse timestamp %s, dropped from payload", timestamp);
-            } catch (DateTimeException e) {
-                Log.warnf(e, "Timestamp %s was successfully parsed, but could not be reformatted for PagerDuty, dropped from payload", timestamp);
-            }
-        } else {
-            Log.warn("Timestamp not provided for payload, ignored");
+        try {
+            messagePayload.put(TIMESTAMP, LocalDateTime.parse(timestamp).format(PD_DATE_TIME_FORMATTER));
+        } catch (DateTimeParseException e) {
+            Log.warnf(e, "Unable to parse timestamp %s, dropped from payload", timestamp);
+        } catch (DateTimeException e) {
+            Log.warnf(e, "Timestamp %s was successfully parsed, but could not be reformatted for PagerDuty, dropped from payload", timestamp);
         }
 
         messagePayload.put(SEVERITY, PagerDutySeverity.fromJson(cloudEventPayload.getString(SEVERITY)));
@@ -82,7 +78,7 @@ public class PagerDutyTransformer implements Processor {
         // Add source names, if provided
         JsonObject cloudSource = getSourceNames(cloudEventPayload.getJsonObject(SOURCE));
         if (cloudSource != null) {
-            customDetails.put(SOURCE, cloudSource);
+            customDetails.put(SOURCE_NAMES, cloudSource);
         }
 
         // Keep events, if provided
--- a/connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java
+++ b/connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java
@@ -1,12 +1,15 @@
 package com.redhat.cloud.notifications.connector.pagerduty;
 
+import io.quarkus.logging.Log;
 import io.vertx.core.json.JsonObject;
 import jakarta.enterprise.context.ApplicationScoped;
 import org.apache.camel.Exchange;
 import org.apache.camel.Processor;
 
+import java.time.DateTimeException;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
 
 /**
  * Constructs a <a href="https://support.pagerduty.com/main/docs/pd-cef">PD-CEF</a> alert event.
@@ -45,6 +48,7 @@ public class PagerDutyTransformer implements Processor {
     @Override
     public void process(Exchange exchange) {
         JsonObject cloudEventPayload = exchange.getIn().getBody(JsonObject.class);
+        validatePayload(cloudEventPayload);
 
         JsonObject message = new JsonObject();
         message.put(EVENT_ACTION, PagerDutyEventAction.TRIGGER);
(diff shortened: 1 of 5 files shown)


connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java
```diff
@@ -45,30 +48,43 @@ public class PagerDutyTransformer implements Processor {
...
+                messagePayload.put(TIMESTAMP, LocalDateTime.parse(timestamp).format(PD_DATE_TIME_FORMATTER));
+            } catch (DateTimeParseException e) {
+                Log.warnf(e, "Unable to parse timestamp %s, dropped from payload", timestamp);
+            } catch (DateTimeException e) {
+                Log.warnf(e, "Timestamp %s was successfully parsed, but could not be reformatted for PagerDuty, dropped from payload", timestamp);
+            }
+        } else {
+            Log.warn("Timestamp not provided for payload, ignored");
+        }
+
```
MikelAlejoBR, line 72: It seems like the `BaseTransfomer` class will always add the `Timestamp` field of the action. Have you found any edge case where we might not be adding the timestamp? Because in that case we should fix it for all the processors in the engine :eyes:

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java
```diff
@@ -45,30 +48,43 @@ public class PagerDutyTransformer implements Processor {
...
         customDetails.put(ORG_ID, cloudEventPayload.getString(ORG_ID));
-        customDetails.put(CONTEXT, JsonObject.mapFrom(cloudEventPayload.getJsonObject(CONTEXT)));
-        customDetails.put(SOURCE_NAMES, sourceNames);
+        customDetails.put(CONTEXT, cloudEventPayload.getJsonObject(CONTEXT));
+
+        // Add source names, if provided
+        JsonObject cloudSource = getSourceNames(cloudEventPayload.getJsonObject(SOURCE));
+        if (cloudSource != null) {
+            customDetails.put(SOURCE, cloudSource);
+        }
```
MikelAlejoBR, line 86: Same thing about the source. The bundle, application and event type are fields that will always be present when coming from the engine, so I would say it's safe to assume that they'll be there for the connector.

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java
MikelAlejoBR: Whoops, my bad. Misread the code :facepalm:

MikelAlejoBR: Actually not, I think we can spare the null check for the timestamp value. Sorry for the confusion :sweat:

jessicarod7: No it should be safe to take them out, I'll update this.

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java
jessicarod7: Will update as well

jessicarod7: Actually I think this one has to stay, because the display names (`.payload.source.bundle.display_name`) are not a required field. For example, [test notifications don't include them](https://github.com/RedHatInsights/notifications-backend/blob/01b015725aaf585764872a9311e56a19835ca628/connector-pagerduty/src/test/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformerTest.java#L89-L116).

jessicarod7: The actual names (`.payload.bundle`) are required, so they don't need a check ([like here](https://github.com/RedHatInsights/notifications-backend/blob/a384554c31289549afea76d9dbca9cd81061d279/connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java#L75)).

jessicarod7: turns out I accidentally changed a field name here (meant to be `source_names` in the final payload, not `source`), that might have caused some of the confusion.

MikelAlejoBR: My point was that the `BaseTransformer` always adds those fields to the `source` key:

https://github.com/RedHatInsights/notifications-backend/blob/73af1df3cfdd8b7ab057a48069e0eff78099f232/engine/src/main/java/com/redhat/cloud/notifications/transformers/BaseTransformer.java#L62-L76

And that the `PagerDutyProcessor` uses that transformer too before sending the events through Kafka to the connector:

https://github.com/RedHatInsights/notifications-backend/blob/73af1df3cfdd8b7ab057a48069e0eff78099f232/engine/src/main/java/com/redhat/cloud/notifications/processors/pagerduty/PagerDutyProcessor.java#L75

That's why I think I'm missing how the connector would receive a payload without that data :thinking:

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java
```diff
@@ -80,14 +92,38 @@ public void process(Exchange exchange) {
...
+    private void validatePayload(final JsonObject cloudEventPayload) {
+        String summary = cloudEventPayload.getString(EVENT_TYPE);
+        if (summary == null || summary.isEmpty()) {
+            throw new IllegalArgumentException("Event type must be specified for PagerDuty payload summary");
+        }
+
+        String source = cloudEventPayload.getString(APPLICATION);
+        if (source == null || source.isEmpty()) {
+            throw new IllegalArgumentException("Application must be specified for PagerDuty payload source");
+        }
```
MikelAlejoBR, line 105: Both the Event Type and the Application are always present on the Cloud Event, so I think we can spare these checks

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java
jessicarod7: That's what I thought, but I discovered that test notifications (sent by `POST /endpoints/{uuid}/test`) don't include those fields:

https://github.com/RedHatInsights/notifications-backend/blob/01b015725aaf585764872a9311e56a19835ca628/connector-pagerduty/src/test/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformerTest.java#L92-L112

It looks like the test endpoint bypasses the `PagerDutyProcessor` completely, and just injects its own cloud event into the Kafka topic?

https://github.com/RedHatInsights/notifications-backend/blob/45eef7f93689ad3615b6a2cfa229dc98f072edc1/engine/src/main/java/com/redhat/cloud/notifications/routers/EndpointTestResource.java#L34-L56

https://github.com/RedHatInsights/notifications-backend/blob/45eef7f93689ad3615b6a2cfa229dc98f072edc1/common/src/main/java/com/redhat/cloud/notifications/models/event/TestEventHelper.java#L60-L93

That would explain why the fields are missing...

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyTransformer.java
jessicarod7: yeah this is an outdated check, will remove

MikelAlejoBR (approved):
LGTM

PR #3039, opened 2024-10-14
RHCLOUD-35454 | feature: Kessel bulk import

MikelAlejoBR:
Implements the required migration endpoint so that Kessel can import the
integrations from Notifications.

## Dependency

- https://github.com/RedHatInsights/notifications-backend/pull/3016

## Jira ticket

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
@@ -391,7 +391,8 @@ public class EndpointRepository {
     }
 
     /**
-     * Retrieves a list of endpoints.
+     * Retrieves a list of endpoints ordered by their creadtion date in
+     * ascending order —oldest first—.
      * @param limit the size of the list to retrieve.
      * @param offset the offset we should apply to skip the already read
      *               endpoints.
@@ -399,7 +400,7 @@ public class EndpointRepository {
      */
     public List<Endpoint> getNonSystemEndpointsWithLimitAndOffset(final int limit, final int offset) {
         return this.entityManager
-            .createQuery("FROM Endpoint WHERE orgId IS NOT NULL", Endpoint.class)
+            .createQuery("FROM Endpoint WHERE orgId IS NOT NULL ORDER BY created ASC", Endpoint.class)
             .setFirstResult(offset)
             .setMaxResults(limit)
             .getResultList();
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
@@ -397,9 +397,9 @@ public class EndpointRepository {
      *               endpoints.
      * @return a list of endpoints.
      */
-    public List<Endpoint> getEndpointsWithLimitAndOffset(final int limit, final int offset) {
+    public List<Endpoint> getNonSystemEndpointsWithLimitAndOffset(final int limit, final int offset) {
         return this.entityManager
-            .createQuery("FROM Endpoint", Endpoint.class)
+            .createQuery("FROM Endpoint WHERE orgId IS NOT NULL", Endpoint.class)
             .setFirstResult(offset)
             .setMaxResults(limit)
             .getResultList();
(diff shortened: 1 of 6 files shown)


backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -389,4 +389,19 @@ public Optional<Endpoint> getEndpointWithLinkedEventTypes(String orgId, UUID id)
...
+    /**
+     * Retrieves a list of endpoints.
+     * @param limit the size of the list to retrieve.
+     * @param offset the offset we should apply to skip the already read
+     *               endpoints.
+     * @return a list of endpoints.
+     */
+    public List<Endpoint> getEndpointsWithLimitAndOffset(final int limit, final int offset) {
+        return this.entityManager
+            .createQuery("FROM Endpoint", Endpoint.class)
```
g-duval, line 402: I think we shoudl remove system endpoints ( when orgId is null)

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/kessel/KesselAssetsMigrationService.java
```diff
@@ -0,0 +1,144 @@
...
+            // If for some reason we have fetched full pages from the database
+            // all the time, the last one might be empty, so there is no need
+            // to attempt calling Kessel.
+            if (fetchedEndpoints.isEmpty()) {
+                break;
+            }
+
+            final CreateTuplesRequest request = this.createTuplesRequest(fetchedEndpoints);
+            final int finalOffset = offset;
+            this.relationTuplesClient.createTuples(request, new StreamObserver<>() {
```
g-duval, line 82: Is this service replayable? Will it be ok if we send the same endpoint multiple times?

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
MikelAlejoBR: You are right, yes. I updated the query!

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/kessel/KesselAssetsMigrationService.java
MikelAlejoBR: I have been told that in principle yes, but that they will tell us for sure soon.

g-duval (approved):
LGTM

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/kessel/KesselAssetsMigrationService.java
MikelAlejoBR: It is not replayable, so we will only be able to run the migration once. We will have to make sure we disable "creates" and "deletes" for existing integrations.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/kessel/KesselAssetsMigrationService.java
```diff
@@ -0,0 +1,144 @@
...
+@ApplicationScoped
+@RolesAllowed(ConsoleIdentityProvider.RBAC_INTERNAL_ADMIN)
+@Path(API_INTERNAL)
+public class KesselAssetsMigrationService {
+    /**
+     * Defines the relation between the subject and the resource when passing
+     * the tuple to Kessel. <a href="https://github.com/RedHatInsights/rbac-config/blob/a806fb03c95959391eceb0b42c7eefd8ae2350ae/configs/prod/schemas/schema.zed#L3">
+     * Reference</a>
+     */
+    public static final String RELATION = "t_workspace";
```
alechenninger, line 41: The relation should be "workspace" – the t_ prefix is an internal implementation detail in how we are mapping to SpiceDB's schema constructs. Go by the relation that's in KSL – be careful going by what you see in the underlying SpiceDB schema.

alechenninger: i.e. this is the relevant line to link to: https://github.com/RedHatInsights/rbac-config/blob/bb2d840a33705e84e0819d5fe697aaa7c87c378d/configs/prod/schemas/src/notifications.ksl#L13

MikelAlejoBR: Got it, thank you!

PR #3043, opened 2024-10-15
fix(CI): use jq in checkimage

petrblaho:
Use jq to query for image digest to remove the need to rely on skopeo inspect output having json keys in particular order. RHINENG-13679


MikelAlejoBR (approved):
LGTM

gwenneg:
Thank you for this @petrblaho!

PR #3045, opened 2024-10-16
RHCLOUD-35435 | refactor: remove the Inventory beans' provider

MikelAlejoBR:
The new client's version includes the "@ApplicationScoped" beans which allows us to mock them for tests, so we no longer need our custom providers.

I tested the integration with the Kessel back end again and adjusted the requests to the latest model changes, by including the required fields that will be expected once the integration goes live in stage and production environments.

Since integration removals in the Kessel inventory are still not working as expected, I gated them behind a feature flag so that we don't suffer from errors until the feature is completed on their end.

## Jira ticket
[[RHCLOUD-35435]](https://issues.redhat.com/browse/RHCLOUD-35435)


g-duval (approved):
LGTM

PR #3047, opened 2024-10-17
RHCLOUD-35763 | feature: return proper authorization responses

MikelAlejoBR:
In order to return consistent responses when an unauthorized principal
calls our APIs, we have decided as a team that we will return "not
found" responses.
    
The reasoning is that many existing services will not let you know
whether the resource exists or not if you are not authorized, so we
thought that following this pattern was a good idea.

## Jira ticket
[[RHCLOUD-35763]](https://issues.redhat.com/browse/RHCLOUd-35763)


g-duval (approved):
LGTM, could you please rebase it, then I could have a last look before merging it?

MikelAlejoBR:
Sure thing, done!

g-duval (approved):
LGTM

PR #3049, opened 2024-10-18
RHCLOUD-35769 | feature: Kessel inventory tests

MikelAlejoBR:
feature: tests for the Kessel Inventory integration

Tests that creating and deleting integrations also notifies Kessel
Inventory about it. We also test if the "@Transactional" methods work as
expected and that we will not be creating a desynchronization between
our database and the Kessel Inventory if we suffer from any kind of
errors. More precisely:

- When saving the integration in our database fails, we check that
  Kessel does not get an "integration created" request.
- When we succesfully save an integration in our database but sending
  the "integration created" request to Kessel fails, we check that the
  whole transaction gets rolled back and no integration gets saved in
  our database.
- When deleting the integration in our database fails, we check that
  Kessel does not get an "integration deleted" request.
- Finally, when we successfully save the integration in our database but
  sending the "integration deleted" request to Kessel fails, we check
  that the whole transaction gets rolled back and that the integration
  does not get deleted from our database.

## Dependency

- https://github.com/RedHatInsights/notifications-backend/pull/3045

## Jira ticket
[[RHCLOUD-35769]](https://issues.redhat.com/browse/RHCLOUD-35769)

### DIFF
--- a/backend/src/test/java/com/redhat/cloud/notifications/routers/EmailsOnlyModeTest.java
+++ b/backend/src/test/java/com/redhat/cloud/notifications/routers/EmailsOnlyModeTest.java
@@ -82,8 +82,8 @@ public class EmailsOnlyModeTest extends DbIsolatedTest {
 
     @MethodSource("argumentsProvider")
     @ParameterizedTest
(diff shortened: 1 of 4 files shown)


g-duval (approved):
LGTM

PR #3051, opened 2024-10-18
Add PagerDutyProperties to EndpointRepository

jessicarod7:
This was missing and caused a crash when trying to send notifications to PagerDuty.


gwenneg (approved):
LGTM

PR #3054, opened 2024-10-22
[RHCLOUD-34488] Fix sync between Behavior group and endpoint to event  types data models

g-duval:
We didn't handle properly the new data model sync when:
Some event types are removed from existing behavior groups using` /eventTypes/{eventTypeId}/behaviorGroups` api.


MikelAlejoBR (approved):
LGTM!

PR #3060, opened 2024-10-23
[RHCLOUD-35863] Add orgId filtering capacity on Kessel feature flags

g-duval:


backend/src/main/java/com/redhat/cloud/notifications/config/BackendConfig.java
```diff
@@ -106,9 +107,9 @@ void logConfigAtStartup(@Observes Startup event) {
         config.put(drawerToggle, isDrawerEnabled());
         config.put(EMAILS_ONLY_MODE, isEmailsOnlyModeEnabled());
         config.put(ERRATA_MIGRATION_BATCH_SIZE, getErrataMigrationBatchSize());
-        config.put(KESSEL_INVENTORY_ENABLED, isKesselInventoryEnabled());
-        config.put(KESSEL_INVENTORY_INTEGRATIONS_REMOVAL_ENABLED, areKesselInventoryIntegrationRemovalsEnabled());
-        config.put(KESSEL_RELATIONS_ENABLED, isKesselRelationsEnabled());
+        config.put(KESSEL_INVENTORY_ENABLED, isKesselInventoryEnabled(null));
+        config.put(KESSEL_INVENTORY_INTEGRATIONS_REMOVAL_ENABLED, areKesselInventoryIntegrationRemovalsEnabled(null));
+        config.put(KESSEL_RELATIONS_ENABLED, isKesselRelationsEnabled(null));
```
MikelAlejoBR, line 112: I feel like we should maybe not even report these if we can't specify an accurate value at startup, what do you think?

MikelAlejoBR (approved):
LGTM! I think we can remove the `areKesselInventoryIntegrationRemovalsEnabled` feature flag too. I reported the problem to the Kessel team and after testing it locally, the deletions seem to work too. However if you don't want to mess with the tests just leave it and I'll open another quick PR :smiley: :+1:

backend/src/main/java/com/redhat/cloud/notifications/config/BackendConfig.java
```diff
@@ -152,17 +153,19 @@ public boolean isInstantEmailsEnabled() {
...
 
-    public boolean isKesselInventoryEnabled() {
+    public boolean isKesselInventoryEnabled(String orgId) {
         if (unleashEnabled) {
-            return unleash.isEnabled(kesselInventoryToggle, false);
+            UnleashContext unleashContext = buildUnleashContextWithOrgId(orgId);
+            return unleash.isEnabled(kesselInventoryToggle, unleashContext, false);
         } else {
             return kesselInventoryEnabled;
         }
```
MikelAlejoBR, line 171: Also I am not sure if we can disable the inventory per organization ID :disappointed: , because if we migrate all of our integrations to Kessel, and we enable the inventory integration for the org id `acme`, we would only be creating the integrations for them...

g-duval:
> areKesselInventoryIntegrationRemovalsEnabled

Ok good to know :), yes I prefer removing it on a dedicated PR just after this one then.

MikelAlejoBR (approved):

MikelAlejoBR (approved):

PR #3061, opened 2024-10-23
RHCLOUD-35813 | feature: implement Kessel permissions for the NotificationsResource

MikelAlejoBR:
These permissions are needed for when the integration with the Kessel Relations API is enabled. The Kessel permissions are derived from the existing ones, so in theory the old RBAC permissions should be mapped with the new ones we have specified here.

## Jira ticket
[[RHCLOUD-35813]](https://issues.redhat.com/browse/RHCLOUD-35813)


g-duval (approved):
LGTM, awesome work!

PR #3065, opened 2024-10-25
 RHCLOUD-35924 | feature: read and intercept the Kessel clients' URLs

MikelAlejoBR:
The previous configuration was pointing to the `localhost` URLs, so we needed to make sure that when using Clowder and the JSON configuration file, the proper URLs were being fed to build the Kessel clients.

However, the `clowder-quarkus-config-source` extension we use to resolve the properties always appends the `http` or `https` schemes to the URLs, and since the gRPC clients are expecting a `${HOSTNAME}:${PORT}` string, we had to implement a configuration interceptor to be able to fed the clients with the proper URLs.

## Jira ticket
[[RHCLOUD-35924]](https://issues.redhat.com/browse/RHCLOUD-35924)


g-duval (approved):
LGTM

PR #3073, opened 2024-10-28
[RHCLOUD-35901] Add fine grain maintenance mode

g-duval:
We can now use Unleash to manage Notifications "maintenance mode".
Unleash can honor rollout rules like 100% or if a particular property value start with a particular value.
Thanks to than we can isolate only some APIs.


MikelAlejoBR (approved):
LGTM! Although if we're removing the `StatusResource` and `StatusRepository`, maybe we need a migration to remove the database table too? :thinking:

PR #3075, opened 2024-10-29
[RHCLOUD-35969] drop orphans system endpoints

g-duval:


MikelAlejoBR (approved):
LGTM! One question: we don't need to check the `endpoint_event_type` table, right?

gwenneg (approved):
LGTM

gwenneg (changes requested):
The SQL file needs to be renamed to `V1.106`.

gwenneg (approved):

g-duval:
> LGTM! One question: we don't need to check the `endpoint_event_type` table, right?

No we don't have to @MikelAlejoBR, because those endpoints were generated automatically by aggregation process which skips event_type associations.

PR #3076, opened 2024-10-29
RHCLOUD-35970 | refactor: add "hasPermissionOnWorkspace" method

MikelAlejoBR:
The goal of this method is to make it easier to use workspace permissions and reduce the potential mistakes when checking for permissions: since the method only takes the "WorkspacePermissions" enum, accidentally feeding it with the "IntegrationPermissions" enum will tell you that you might be assigning the wrong permission to the wrong resource.

## Jira ticket
[[RHCLOUD-35970]](https://issues.redhat.com/browse/RHCLOUD-35970)


g-duval (approved):
LGTM

PR #3111, opened 2024-11-12
Update kessel criterion for email integration

g-duval:
Name and structure updates from ADR-53 reviews

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/ExternalAuthorizationCriterion.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/ExternalAuthorizationCriterion.java
@@ -3,7 +3,6 @@ package com.redhat.cloud.notifications.processors;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
-import io.vertx.core.json.JsonObject;
 import jakarta.validation.constraints.NotNull;
 import java.util.Objects;
 
@@ -12,7 +11,7 @@ import java.util.Objects;
 public class ExternalAuthorizationCriterion {
 
     @NotNull
-    private JsonObject type;
+    private Type type;
 
     @NotNull
     private String id;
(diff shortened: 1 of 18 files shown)


engine/src/main/java/com/redhat/cloud/notifications/processors/ExternalAuthorizationCriterion.java
```diff
@@ -3,34 +3,35 @@
...
 import java.util.Objects;
 
 @JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
 @JsonIgnoreProperties(ignoreUnknown = true)
-public class ExternalAuthorizationCriteria {
+public class ExternalAuthorizationCriterion {
 
     @NotNull
-    private String assetType;
+    private JsonObject type;
```
MikelAlejoBR, line 15: Why not have a dedicated object for the type? I think we will have to modify the `getLookupSubjectsRequest` to adjust the `.setName()` and the `.setNamespace()` calls.

MikelAlejoBR:
Also, I think we need to change the subject type to `rbac/principal` too: https://github.com/RedHatInsights/notifications-backend/blob/ae2f54ca6df5de2a79377c0e1ffa52065d9e126b/recipients-resolver/src/main/java/com/redhat/cloud/notifications/recipients/resolver/kessel/KesselService.java#L60

engine/src/main/java/com/redhat/cloud/notifications/processors/ExternalAuthorizationCriterion.java
g-duval: I was still dreaming to be agnostic of Kessel structure, but you're right, we have to be realistic :)

recipients-resolver/src/main/java/com/redhat/cloud/notifications/recipients/resolver/kessel/KesselService.java
```diff
@@ -21,6 +21,8 @@ public class KesselService {
 
     static final String SUBJECT_TYPE_USER = "user";
```
MikelAlejoBR, line 22: This one needs to be `principal`: https://github.com/RedHatInsights/rbac-config/blob/97d7501a55cff790dc44162a6ede8412be2528a8/configs/prod/schemas/schema.zed#L13

MikelAlejoBR (approved):
LGTM!

PR #3124, opened 2024-11-14
RHCLOUD-34485 | feature: update API for Kessel Inventory's client

MikelAlejoBR:
The back end requested a "workspace_id" field that was not present in the client's API. It turns out they needed to update the client, and in turn, we needed to update our implementation too.

## Jira ticket
[[RHCLOUD-34485]](https://issues.redhat.com/browse/RHCLOUD-34485)


g-duval (approved):
LGTM

PR #3128, opened 2024-11-17
RHCLOUD-36297 | refactor: handlers' versioning

MikelAlejoBR:
The goal is to improve readability by providing specific classes for each version of the endpoints that we serve. By having them grouped by class and file, it is easy to check what the V1 or V2 paths have.

## Jira ticket
[[RHCLOUD-36297]](https://issues.redhat.com/browse/RHCLOUD-36297)


g-duval (approved):
LGTM, awesome, thank you Mikel.
Nevertheless I would split those only for apis requesting a V2, WDYT?

g-duval (commented):
Couldn't we also keep only EndpointResourceV2 & NotificationsV2?
Then we could add some *resourceV2 when we will be needed for other APIs?

g-duval (approved):
LGTM

PR #3131, opened 2024-11-18
[RHCLOUD-36316] Add Kessel relations config for recipients-resolver app

g-duval:


.rhcicd/clowdapp-recipients-resolver.yaml
```diff
@@ -14,6 +14,7 @@ objects:
     envName: ${ENV_NAME}
     dependencies:
     - rbac
+    - kessel-relations
```
gwenneg, line 17: Let's do this until Kessel is stable in prod:
```suggestion
    optionalDependencies:
    - kessel-relations
```

.rhcicd/clowdapp-recipients-resolver.yaml
```diff
@@ -263,3 +278,14 @@ parameters:
   value: "https"
 - name: NOTIFICATIONS_RECIPIENTS_RESOLVER_USE_KESSEL_ENABLED
   value: "false"
+- name: KESSEL_DOMAIN
```
gwenneg, line 281: This looks unused.

.rhcicd/clowdapp-recipients-resolver.yaml
```diff
@@ -263,3 +278,14 @@ parameters:
...
+  value: "redhat"
+- name: NOTIFICATIONS_KESSEL_RELATIONS_ENABLED
```
gwenneg, line 283: This looks unused.

recipients-resolver/src/main/resources/application.properties
```diff
@@ -54,6 +54,9 @@ quarkus.unleash.active=false
...
 
 # Kessel integration gRpc
-notifications.recipients-resolver.kessel.target-url=localhost:9000
+notifications.recipients-resolver.kessel.target-url=${clowder.endpoints.kessel-relations-api:localhost:9000}
 notifications.recipients-resolver.kessel.secure-client=false
-
+notifications.recipients-resolver.kessel.client.id=insights-notifications
+notifications.recipients-resolver.kessel.client.issuer=http://localhost:8084/realms/redhat-external
+notifications.recipients-resolver.kessel.client.secret=development-value-123
+notifications.recipients-resolver.kessel.mode=oidc-client-credentials
```
gwenneg, line 62: The keys are different from what we have here: https://github.com/RedHatInsights/notifications-backend/blob/caa59a6f6239d0f2e0f56fa7bd8ede30d9947280/backend/src/main/resources/application.properties#L117-L122

@MikelAlejoBR The keys from `backend` should work in this project, right?

g-duval: Yes it should work with the default `relations-api` format which I personally don't like because it's not obvious that it is about Kessel.
If we had the opportunity to create our own client, we probably would have identified it more clearly, as we always have.
That's why I chosen to rename those keys to be more consistent with other integrations we already have.
But of course it could be reverted if you think it's a bad idea :)

MikelAlejoBR: When we had to use our custom bean producers I also used some custom configuration options, but in order to avoid having to deal with implementing the configuration interface and what not, I simply decided to use Kessel's configuration options and then use a configuration override to fix the URL.

I personally prefer overriding the configuration instead of having to do the config implementation and having to produce the clients ourselves, but in theory this should work too.

.rhcicd/clowdapp-recipients-resolver.yaml
g-duval: That's right, thanks

.rhcicd/clowdapp-recipients-resolver.yaml
g-duval: That's right, thanks

recipients-resolver/src/main/java/com/redhat/cloud/notifications/recipients/config/RecipientsResolverConfig.java
```diff
@@ -179,13 +199,22 @@ public String getMbopEnv() {
...
-        return kesselTargetUrl;
+        try {
+            final URL url = new URI(kesselTargetUrl).toURL();
+            final String newKesselUrl = url.getHost() + ":9000";
+
+            Log.debugf("Kessel URL changed from \"%s\" to \"%s\"", kesselTargetUrl, newKesselUrl);
+
+            return newKesselUrl;
+        } catch (final IllegalArgumentException | MalformedURLException | URISyntaxException e) {
+            Log.debugf(e, "Unable to create a URL from value \"%s\"", kesselTargetUrl);
```
gwenneg, line 215: Shouldn't this be logged at `warn` level?

recipients-resolver/src/main/resources/application.properties
gwenneg: @g-duval Makes sense. I agree that the `relations-api*` config keys from Kessel could be better.

I'm not sure which approach I prefer so I won't request any changes in this PR.

gwenneg (approved):
LGTM

PR #3136, opened 2024-11-20
Add severity column

Tomboyo:
This change adds a severity column for RHSA email notifications (not for RHEA or RHBAs), per CHAINSAW-6. Please refer to that ticket for a preview of the rendered HTML.

There is a corresponding back-end change to Errata Notifications (hosted on gitlab.cee) which updates the json messages we send to the notifications API. This changes in this MR rely on those back-end changes, so we need to coordinate deployment to production (and preferably stage as well, but that's less important.)

- [x] check this box when errata is dpeloyed

(In one test case I changed the severity from "Low" to "Important" to see how the larger word (and largest of the options) affected layout. This doesn't impact the tests, but I left it in as a convenience for when we inspect the layout in the future.)

### DIFF
--- a/engine/src/main/resources/templates/Errata/securityEmailBodyV2.html
+++ b/engine/src/main/resources/templates/Errata/securityEmailBodyV2.html
@@ -15,7 +15,8 @@
         <thead>
             <tr>
               <th style="width: 30%">Security update</th>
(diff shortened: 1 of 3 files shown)


gwenneg:
image

Unless you got a green light from Teamnado.

PR #3142, opened 2024-11-25
[ADR-46] Parametrize CPU and Memory resource usage in clowdapps

Fynardo:


gwenneg:
Will other ClowdApp templates be updated in follow-up PRs?

Fynardo:
> Will other ClowdApp templates be updated in follow-up PRs?

The idea was to create a commit per each clowdapp, but later I found that all of them were already properly parametrized

gwenneg (approved):
Sorry, I didn't notice this only changed the DB cleaner cronjob resources 😅

LGTM, thanks!

PR #3160, opened 2024-12-02
[RHCLOUD-36527] Add eventType restrictions on event types

g-duval:


backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
```diff
@@ -1068,6 +1058,7 @@ protected void redactSecretsForEndpoint(final SecurityContext securityContext, f
         @APIResponse(responseCode = "404", content = @Content(mediaType = TEXT_PLAIN,  schema = @Schema(type = SchemaType.STRING)),
             description = "No event type or endpoint found with the passed id.")
     })
+    @Tag(name = OApiFilter.PRIVATE)
```
g-duval, line 1061: After a talk with UI team, it seems those two services will not be used.
I hidded them, then they could be removed it still not relevant after new UI workflow release (Q1Y25?).

PR #3161, opened 2024-12-02
RHCLOUD-36553 | feature: log RBAC requests and responses

MikelAlejoBR:
The goal is to be able to log the key requests and responses that we send, so that we can figure out what the problem is when calling RBAC.

The reason for implementing custom logging instead of using the [REST
Client's logging](https://quarkus.io/guides/rest-client#logging-traffic) is that we want to be able to log the requests only
when we receive a non "2xx" response, since otherwise it would produce
too much noise with all the successful requests that we send.
## Jira ticket
[[RHCLOUD-36553]](https://issues.redhat.com/browse/RHCLOUD-36553)


g-duval (approved):
LGTM

PR #3171, opened 2024-12-04
RHCLOUD-35454 | fix: RBAC workspace exceptions stop the bulk import for Kessel

MikelAlejoBR:
When RBAC returned a non-200 response, and the "RbacService" class threw a "ClientWebApplicationException", the exception was not properly handled by the migration tool and it stopped the migration's execution.

The intended behavior from the beginning was to continue with the migration even when we faced these kind of errors.

## Jira ticket


g-duval (approved):
LGTM

PR #3174, opened 2024-12-05
Try to fix MSTeams post-deploy CJI as its name could be longer than permitted

Fynardo:
There is a limitation on CJI names that they need to be shorter than 64 characters. 
Based on how CJI names are created, the Teams one gets up to 65 characters


MikelAlejoBR (approved):
LGTM!

gwenneg (approved):

PR #3194, opened 2024-12-10
RHCLOUD-35454 | feature: run the Kessel migration in a virtual thread

MikelAlejoBR:
We want to avoid blocking Quarkus' event thread, and since the migration uses blocking calls, we want to make sure it does not get killed by Vert.x.

## Jira ticket


g-duval (approved):
LGTM

PR #3195, opened 2024-12-10
[RHCLOUD-36688] Fix excessive drawer endpoint creations

g-duval:


engine/src/main/java/com/redhat/cloud/notifications/processors/drawer/DrawerProcessor.java
```diff
@@ -111,7 +104,7 @@ public void process(Event event, List<Endpoint> endpoints) {
             externalAuthorizationCriteriaExtractor.extract(event)
         );
 
-        connectorSender.send(event, endpoint, JsonObject.mapFrom(drawerNotificationToConnector));
+        connectorSender.send(event, endpoints.getFirst(), JsonObject.mapFrom(drawerNotificationToConnector));
```
g-duval, line 107: Probably need to be revisited later, with a larger reflexion about how to handle enpoints union properly on for Email and Drawer integrations, when multiple of those are linked to the same event type.

MikelAlejoBR (approved):
LGTM!

PR #3207, opened 2024-12-12
[RHCLOUD-36892] Forbid inconsistent user subscriptions

g-duval:
This PR adds consistency checks on email and drawer subscriptions.
It forbid user subscription it the linked event type don't have a template for the selected subscription type.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
@@ -157,7 +157,7 @@ public class TemplateRepository {
     }
 
     public void checkIfExistAggregationEmailTemplatesByEventType(UUID eventTypeId) throws NotFoundException {
-        String hql = "select count(t) FROM AggregationEmailTemplate t JOIN t.application a JOIN a.eventTypes ev WHERE ev.id = :eventTypeId";
+        String hql = "select count(t) FROM AggregationEmailTemplate t WHERE t.eventType.id = :eventTypeId";
         Long count = entityManager.createQuery(hql, Long.class)
             .setParameter("eventTypeId", eventTypeId)
             .getSingleResult();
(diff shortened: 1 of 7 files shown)


backend/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
```diff
@@ -142,7 +143,29 @@ public InstantEmailTemplate findInstantEmailTemplateByEventType(UUID eventTypeId
 
             return instantEmailTemplate;
         } catch (NoResultException e) {
-            throw new NotFoundException("Instant email template not found");
+            throw new NotFoundException(String.format("Event type '%s' doesn't support instant email subscription", eventTypeId));
+        }
+    }
+
+    public AggregationEmailTemplate findAggregationEmailTemplatesByEventType(UUID eventTypeId) {
```
gwenneg, line 150: The returned object is never used. This method could return a boolean and `FETCH` could be avoided to lower the pressure on the DB. Unless you have a plan to use the method from elsewhere and use the returned data in the near future.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
```diff
@@ -142,7 +143,29 @@ public InstantEmailTemplate findInstantEmailTemplateByEventType(UUID eventTypeId
...
+        try {
+            return entityManager.createQuery(hql, AggregationEmailTemplate.class)
+                .setParameter("eventTypeId", eventTypeId)
+                .getSingleResult();
+        } catch (NoResultException e) {
+            throw new NotFoundException(String.format("Event type '%s' doesn't support daily email subscription", eventTypeId));
+        }
+    }
+
+    public IntegrationTemplate findDrawerTemplateByEventType(UUID eventTypeId) {
```
gwenneg, line 161: Same comment about the pressure on the DB.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/TemplateRepository.java
g-duval: sure, I refactored those methods to not fetch results anymore

gwenneg (approved):
LGTM

PR #3211, opened 2024-12-13
RHCLOUD-35158 | feature: improve resilience when deleting the secrets from Sources

MikelAlejoBR:
Before this patch, if reporting the integration deletion to the Kessel Inventory failed, the secrets were deleted from Sources but the integration deletion would be rolled back in the database, which left it in an invalid state unable to fetch the associated secrets from Sources.

With this patch:

- If deleting the integration from the database goes wrong, the transaction gets rolled back and nothing gets changed.
- If deleting the integration from Kessel goes wrong, the transaction is rolled back and the above point applies.
- If deleting the integration's secrets' from Sources goes wrong, the first point still applies and we make sure that the integration gets recreated in Inventory.

This should give us enough resiliency to avoid invalid states with integrations in Notifications and the services we depend upon.

## Jira ticket
[[RHCLOUD-35158]](https://issues.redhat.com/browse/RHCLOUD-35158)

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/EndpointResource.java
@@ -728,7 +728,6 @@ public class EndpointResource {
 
         // Clean up the secrets in Sources.
         final Endpoint endpoint = endpointRepository.getEndpoint(orgId, id);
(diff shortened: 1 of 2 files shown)


g-duval (approved):
LGTM

gwenneg (approved):
LGTM

PR #3220, opened 2024-12-19
Configure post-deploy tests in Drawer connector

Fynardo:
Configures the iqePlugin so we can run tests in post-deploy jobs
Also adds the backend app as an optional dependency to avoid configuration problems


gwenneg (approved):
LGTM

PR #3222, opened 2024-12-19
[RHCLOUD-28663] Add a query parameter to links in Events emitted by Notifications

jessicarod7:
## Jira issue

https://issues.redhat.com/browse/RHCLOUD-28663

## Description

> [!IMPORTANT]
> 
> Merge after #3221 and #3229.

This builds on RHCLOUD-22813, and adds a new query parameter `?from=notifications&integration=[integration_type]` to the two URLs generated by `InsightsUrlBuilder`:

- `data.inventory_url`
- `data.application_url`

This change will affects URLs in the following integrations:

| Integration | query parameter (`integration=`) | Modified via |
|------------------|----------------------------------|--------------------|
| Slack| `slack`             | SlackProcessor |
| Microsoft Teams | `teams`             | TeamsProcessor |
| Google Chat | `google_chat`       | GoogleChatProcessor |
| PagerDuty | `pagerduty`         | PagerDutyProcessor |
| ServiceNow | `servicenow`        | EventingProcessor |
| Webhook | `webhook`           | WebhookTypeProcessor |
| Event-driven Ansible | `ansible`           | WebhookTypeProcessor |
| Instant emails | `instant_email`     | EmailProcessor and email templates |
| Aggregated emails | `daily_digest`      | EmailAggregationProcessor and email templates |

New support for the drawer and Splunk will be added in a separate PR.

## Testing

- Add test cases for `PagerDutyTransformer`, all `ProcessorTest` classes, and `TestInventoryTemplate` to ensure that the query parameter is added correctly

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregationProcessor.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregationProcessor.java
@@ -22,6 +22,7 @@ import com.redhat.cloud.notifications.models.EventType;
 import com.redhat.cloud.notifications.models.SubscriptionType;
 import com.redhat.cloud.notifications.models.Template;
 import com.redhat.cloud.notifications.processors.ConnectorSender;
+import com.redhat.cloud.notifications.processors.InsightsUrlsBuilder;
 import com.redhat.cloud.notifications.processors.SystemEndpointTypeProcessor;
 import com.redhat.cloud.notifications.processors.email.connector.dto.EmailNotification;
 import com.redhat.cloud.notifications.processors.email.connector.dto.RecipientSettings;
@@ -75,6 +76,8 @@ public class EmailAggregationProcessor extends SystemEndpointTypeProcessor {
     protected static final String TAG_KEY_APPLICATION = "application";
     protected static final String TAG_KEY_ORG_ID = "orgid";
 
+    public static final String DAILY_DIGEST_QUERY_PARAM_TYPE = "daily_digest";
+
     @Inject
     EmailAggregationRepository emailAggregationRepository;
 
@@ -104,6 +107,9 @@ public class EmailAggregationProcessor extends SystemEndpointTypeProcessor {
 
     @Inject
     ActionParser actionParser;
+
+    @Inject
+    InsightsUrlsBuilder insightsUrlsBuilder;
 
     @Inject
     EndpointRepository endpointRepository;
@@ -339,9 +345,10 @@ public class EmailAggregationProcessor extends SystemEndpointTypeProcessor {
 
                 Map<String, Object> action = Map.of("context", Map.of("title", emailTitle, "items", result),
                     "bundle", bundle);
+                String insightsQueryParams = insightsUrlsBuilder.buildQueryParams(List.of(), DAILY_DIGEST_QUERY_PARAM_TYPE);
 
                 // build final body
-                String bodyStr = templateService.renderTemplate(action, SingleBodyTemplate);
+                String bodyStr = templateService.renderTemplate(action, SingleBodyTemplate, insightsQueryParams);
 
                 // Format data to send to the connector.
                 Set<String> recipientsUsernames = listApplicationWithUserCollection.getValue().stream().map(User::getUsername).collect(Collectors.toSet());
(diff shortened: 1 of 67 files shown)


g-duval:
@jessicarod7 , according Jira's comments, we could also add the integration type, I think "just" updating qute templates when they are available should be fine.
For others like PagerDuty, Splunk and Snow, may have to be hardcoded.

jessicarod7:
Ah, I missed that comment. I'll see what I can do.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregationProcessor.java
```diff
@@ -75,6 +76,8 @@ public class EmailAggregationProcessor extends SystemEndpointTypeProcessor {
     protected static final String TAG_KEY_APPLICATION = "application";
     protected static final String TAG_KEY_ORG_ID = "orgid";
 
+    public static final String DAILY_DIGEST_QUERY_PARAM_TYPE = "daily_digest";
```
g-duval, line 79: Because the trackinq query parameter is a constant, it would be better to add only it in Qute templates only.
Based on what you did, you can just add:
`{@String query_params = 'from=notification_daily_digest'}` as first line of `insightsEmailBodyLight.html` 
`{@String query_params = 'from=notification_instant_email'}` as first line of `insightsEmailBody.html`
Since the java part will remain unchanged.

g-duval: Note: in unit tests, daily digests will have `notification_instant_email` as parameter except for `TestSingleDailyTemplate` which should have the correct value: `notification_daily_digest`.
The 'fix' will come with the removal of deprecated code about daily digest by app.

engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
```diff
@@ -71,6 +75,8 @@ private void process(Event event, Endpoint endpoint) {
         WebhookProperties properties = endpoint.getProperties(WebhookProperties.class);
 
         final JsonObject payload = transformer.toJsonObject(event);
+        insightsUrlsBuilder.buildInventoryUrl(payload, endpoint.getType().name()).ifPresent(url -> payload.put("inventory_url", url));
```
g-duval, line 78: Did you validated with PM we whant to add a new property named "application_url" on webhook (also used by EDA)?

g-duval (commented):
Just added a first review regarding emails management, could avoid to update few java classes ;)

engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
jessicarod7: Not specifically, this was to maintain consistency with the formatting of other classes that now include the field (PagerDuty and ServiceNow). I'll double-check with this one.

engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregationProcessor.java
jessicarod7: Ok, looked through `TestSingleDailyTemplate` and what you said makes sense, I'll make the change.

engine/src/test/java/com/redhat/cloud/notifications/templates/TestInventoryTemplate.java
```diff
@@ -130,6 +130,8 @@ public void testDailyEmailBody(final String groupToValidte) {
         assertTrue(result.contains("Host Name"), "Body should contain 'Host Name' header");
         assertTrue(result.contains("Error"), "Body should contain 'Error' header");
         assertTrue(result.contains(TestHelpers.HCC_LOGO_TARGET));
+        // TODO set boolean to false once app-specific daily digest code is removed
```
jessicarod7, line 133: Based on [this comment](https://github.com/RedHatInsights/notifications-backend/pull/3222#discussion_r2028934059), this should be set to true for now.

engine/src/main/java/com/redhat/cloud/notifications/processors/webhooks/WebhookTypeProcessor.java
jessicarod7: Removed based on BU feedback

g-duval:
Thank you for this PR @jessicarod7 :)
Almost looks good to me, could you please update the query parameter format to somthing like `?from=notifications&integration=instant_email` ?
Since it would be easier to aggregate results by app and/or by integration type.

PR #3233, opened 2025-01-10
[RHCLOUD-37252] Update new aggregation logic

g-duval:
[RHCLOUD-37252] Update new aggregation logic to be closer to legacy aggregation, to be able to compare result during migration phase

### DIFF
--- a/aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationRepository.java
+++ b/aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationRepository.java
@@ -59,7 +59,7 @@ public class EmailAggregationRepository {
             "EXISTS (SELECT 1 FROM AggregationEmailTemplate WHERE application.id = ev.applicationId and subscriptionType = 'DAILY') " +
             // After prod validation phase, the previous AggregationEmailTemplate check should be remove in favor of:
             // check than at least one user of the org subscribed for daily digest with this event type
-            //"EXISTS (SELECT 1 FROM EventTypeEmailSubscription es where ev.orgId = es.id.orgId and es.id.subscriptionType='DAILY' and es.eventType.id = ev.eventType.id and es.subscribed = true) " + warning: need an new index on EventTypeEmailSubscription before being enabled
+            //"EXISTS (SELECT 1 FROM EventTypeEmailSubscription es where ev.orgId = es.id.orgId and es.id.subscriptionType='DAILY' and es.eventType = ev.eventType and es.subscribed is true) " + warning: need an new index on EventTypeEmailSubscription before being enabled
 
             // check for linked email integration linked to this event type (to honor legacy mechanism)
             "AND EXISTS (SELECT 1 FROM Endpoint ep where ev.orgId = ep.orgId and ep.compositeType.type = 'EMAIL_SUBSCRIPTION' and EXISTS (select 1 from EndpointEventType where eventType.id = ev.eventType.id and endpoint.id = ep.id)) " +
--- a/aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationRepository.java
+++ b/aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationRepository.java
@@ -53,9 +53,19 @@ public class EmailAggregationRepository {
         LocalDateTime currentTimeTwoDaysAgo = now.minusDays(2);
         String query = "SELECT DISTINCT ev.orgId, ev.bundleId, ev.applicationId, acp.lastRun, bu.name, ap.name FROM Event ev " +
             "join Application ap on ev.applicationId = ap.id join Bundle bu on ev.bundleId = bu.id " +
-            "join AggregationOrgConfig acp on ev.orgId = acp.orgId " +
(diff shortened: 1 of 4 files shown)


aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationRepository.java
```diff
@@ -53,9 +53,19 @@ public List<AggregationCommand> getApplicationsWithPendingAggregationAccordingOr
...
             "join Application ap on ev.applicationId = ap.id join Bundle bu on ev.bundleId = bu.id " +
-            "join AggregationOrgConfig acp on ev.orgId = acp.orgId " +
-            "WHERE EXISTS (SELECT 1 FROM EventTypeEmailSubscription es where ev.orgId = es.id.orgId and es.id.subscriptionType='DAILY' and es.subscribed = true) " +
+            "join AggregationOrgConfig acp on ev.orgId = acp.orgId WHERE " +
+
+            // check than a aggregation template exists for the event application
+            "EXISTS (SELECT 1 FROM AggregationEmailTemplate WHERE application.id = ev.applicationId and subscriptionType = 'DAILY') " +
+            // After prod validation phase, the previous AggregationEmailTemplate check should be remove in favor of:
+            // check than at least one user of the org subscribed for daily digest with this event type
+            //"EXISTS (SELECT 1 FROM EventTypeEmailSubscription es where ev.orgId = es.id.orgId and es.id.subscriptionType='DAILY' and es.eventType.id = ev.eventType.id and es.subscribed = true) " + warning: need an new index on EventTypeEmailSubscription before being enabled
```
gwenneg, line 62:
```suggestion
            //"EXISTS (SELECT 1 FROM EventTypeEmailSubscription es where ev.orgId = es.id.orgId and es.id.subscriptionType='DAILY' and es.eventType = ev.eventType and es.subscribed is true) " + warning: need an new index on EventTypeEmailSubscription before being enabled
```

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationRepository.java
```diff
@@ -53,9 +53,19 @@ public List<AggregationCommand> getApplicationsWithPendingAggregationAccordingOr
         LocalDateTime currentTimeTwoDaysAgo = now.minusDays(2);
         String query = "SELECT DISTINCT ev.orgId, ev.bundleId, ev.applicationId, acp.lastRun, bu.name, ap.name FROM Event ev " +
             "join Application ap on ev.applicationId = ap.id join Bundle bu on ev.bundleId = bu.id " +
-            "join AggregationOrgConfig acp on ev.orgId = acp.orgId " +
-            "WHERE EXISTS (SELECT 1 FROM EventTypeEmailSubscription es where ev.orgId = es.id.orgId and es.id.subscriptionType='DAILY' and es.subscribed = true) " +
+            "join AggregationOrgConfig acp on ev.orgId = acp.orgId WHERE " +
+
+            // check than a aggregation template exists for the event application
+            "EXISTS (SELECT 1 FROM AggregationEmailTemplate WHERE application.id = ev.applicationId and subscriptionType = 'DAILY') " +
+            // After prod validation phase, the previous AggregationEmailTemplate check should be remove in favor of:
```
gwenneg, line 60: Can't we keep both conditions after the validation phase?

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationRepository.java
```diff
@@ -53,9 +53,19 @@ public List<AggregationCommand> getApplicationsWithPendingAggregationAccordingOr
...
+            // check than at least one user of the org subscribed for daily digest with this event type
+            //"EXISTS (SELECT 1 FROM EventTypeEmailSubscription es where ev.orgId = es.id.orgId and es.id.subscriptionType='DAILY' and es.eventType.id = ev.eventType.id and es.subscribed = true) " + warning: need an new index on EventTypeEmailSubscription before being enabled
+
+            // check for linked email integration linked to this event type (to honor legacy mechanism)
+            "AND EXISTS (SELECT 1 FROM Endpoint ep where ev.orgId = ep.orgId and ep.compositeType.type = 'EMAIL_SUBSCRIPTION' and EXISTS (select 1 from EndpointEventType where eventType.id = ev.eventType.id and endpoint.id = ep.id)) " +
```
gwenneg, line 65:
```suggestion
            "AND EXISTS (SELECT 1 FROM Endpoint ep, EndpointEventType eet where ev.orgId = ep.orgId and ep.compositeType.type = 'EMAIL_SUBSCRIPTION' and eet.eventType = ev.eventType and eet.endpoint = ep) " +
```

aggregator/src/main/java/com/redhat/cloud/notifications/db/EmailAggregationRepository.java
g-duval: To be ISO with the existing behavior, we check if an daily email template exists for the current app.
It generates a payload entry that will be send to engine, even if nobody subscribed to any daily digest for current app and org.

The future condition checks the same thing (because it's not possible to subscribe to a daily digest if the relevant template doesn't exist), in addition it checks if someone subscribed to it, to be sure to send a useful data to engine.

This is why I don't think we will need to keep both.

gwenneg (approved):
LGTM

PR #3237, opened 2025-01-14
RHCLOUD-37310 | fix: rename email subscription user IDs to lowercase

MikelAlejoBR:
We had a customer report where they complained that they were still receiving Errata notifications even after unsubscribing from them.

After investigating, we found out that we had multiple and even overlapping email subscriptions for almost the same username. By "almost", I mean that the username was essentially the same, but with different case. For example: UserA and usera.

Notifications takes all the email subscriptions' usernames and makes them lowercase to make a case insensitive comparison. Also, when user preferences are saved, the email subscriptions that are generated are generated with lowercase user IDs.

When we migrated the Errata email subscriptions, some of them contained mixed case user IDs, and we did not realize about that.

What ended up happening was:

- A customer was initially subscribed to errata Notifications, and therefore we migrated their email subscription from Errata with a mixed case.
- The customer unsubscribed, which generated a lowercase email unsubscription.
- The recipients resolver took both subscriptions and since it still found the mixed case subscription, it was sending the email.

## Jira ticket
[[RHCLOUD-37310]](https://issues.redhat.com/browse/RHCLOUD-37310)

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/SubscriptionRepository.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/db/repositories/SubscriptionRepository.java
@@ -9,8 +9,14 @@ import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import jakarta.persistence.EntityManager;
 import jakarta.transaction.Transactional;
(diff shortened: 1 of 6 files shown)


g-duval (approved):
LGTM

PR #3242, opened 2025-01-16
[RHCLOUD-37354] add bulk resources for Notifications GW

g-duval:
Curently, Notifications GW need to check baet (Bundle + Application + Event Type) existence to valid messages, and also to check certificate CN for tenants who could use their stage environment with Notifications Prod instance.

To get the GW resilient to a database outage/maintenance, we need to cache those data at GW startup, then refresh those periodically when Notifications Backend is available. (If it's not it will have to keep and work with initially loaded  data set)


MikelAlejoBR (approved):
LGTM!

PR #3248, opened 2025-01-22
RHCLOUD-37430 | feature: log statements when fetching integrations for debugging purposes

MikelAlejoBR:
In order to identify the strange behavior when fetching the endpoints in the stage environmen, we need to see what we are issuing to the database, with which parameters and what is being returned.

## Jira ticket
[[RHCLOUD-37430]](https://issues.redhat.com/browse/RHCLOUD-37430)

### DIFF
--- a/backend/src/main/resources/application.properties
+++ b/backend/src/main/resources/application.properties
@@ -122,3 +122,7 @@ relations-api.authn.client.secret=development-value-123
 relations-api.authn.mode=oidc-client-credentials
 relations-api.is-secure-clients=false
 relations-api.target-url=${clowder.endpoints.kessel-relations-api:localhost:9000}
+
+# Enabling "trace" logging level for debugging purposes.
+quarkus.log.category."com.redhat.cloud.notifications.db.builder.Parameters".min-level=TRACE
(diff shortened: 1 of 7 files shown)


backend/src/main/resources/application.properties
```diff
@@ -122,3 +122,7 @@ relations-api.authn.client.secret=development-value-123
 relations-api.authn.mode=oidc-client-credentials
 relations-api.is-secure-clients=false
 relations-api.target-url=${clowder.endpoints.kessel-relations-api:localhost:9000}
+
+# Enabling "trace" logging level for debugging purposes.
+quarkus.log.category."com.redhat.cloud.notifications.db.builder.Parameters".min-level=TRACE
+quarkus.log.category."com.redhat.cloud.notifications.routers.EndpointResource".min-level=TRACE
```
gwenneg, line 128: Do you plan on reverting changes from this PR after the debugging is done?

gwenneg (approved):
LGTM

backend/src/main/resources/application.properties
MikelAlejoBR: Yes, that was my idea. Let me leave a comment there just in case...

PR #3251, opened 2025-01-23
[RHCLOUD-37376] Replace Kafka lag metrics with AWS MSK metrics

jessicarod7:
## Jira issue

https://issues.redhat.com/browse/RHCLOUD-37376

## Description

This replaces old Kafka metrics in Grafana, with new values pulled directly from AWS MSK, as per the migration.

Panel changes:
- _Kafka lag (messages):_ replaces instances of `kafka_consumergroup_group_lag_seconds` with `aws_kafka_sum_offset_lag_sum`, pulled from the CloudWatch Exporter clusters (based on `SumOffsetLag`)
- _Kafka lag (seconds):_ **REMOVE** panel, as AWS does not provide a suitable replacement metric ([relevant Slack message](https://redhat-internal.slack.com/archives/C07FQ00BJV8/p1737623526313699?thread_ts=1737567199.373179&cid=C07FQ00BJV8))

## Compatibility

- _Kafka lag (seconds)_ panel is removed

## Testing

Fun story, because Grafana is provisioned, I'm unable to reupload the JSON file to check if my changes are correct. I was able to get most of the changes done in one session, but some of them had to be done afterwards, and I really don't want to go about recreating all the changes in the UI.

It should Just Work. At worst, I'd only expect that one of the variables, `$dbinstanceidentifier`, _may_ be affected.

### DIFF
--- a/.rhcicd/grafana/grafana-dashboard-notifications-general.configmap.yaml
+++ b/.rhcicd/grafana/grafana-dashboard-notifications-general.configmap.yaml
@@ -5231,7 +5231,7 @@ data:
         {
           "datasource": {
             "type": "prometheus",
-            "uid": "$datasource"
+            "uid": "${aws_resources_exporter}"
           },
           "description": "",
           "fieldConfig": {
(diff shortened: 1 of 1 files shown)


gwenneg (approved):
LGTM

PR #3253, opened 2025-01-24
RHCLOUD-37430 | fix: integrations are not being returned sometimes

MikelAlejoBR:
We were getting reports of integrations not showing up in stage when Kessel Relations was enabled. More specifically, the issue was that after trying to filter integrations by name, some of them did simply not show up in the results even though they were recently created.

After investigating the issue, it turned out that the problem was that we were getting an incomplete set of authorized integration IDs from Kessel, which in turn made us filter the database query results incorrectly, which again, in turn, made us return an inaccurate list of integrations for an organization.

The QE organization which reported the issue had around 2,500 integrations created, and when a request was made to look up for a very specific integration with name "X", Notifications first had to ask Kessel for the principal's authorized integrations. Since Kessel only returned a batch of a 1,000 we would use just those to apply a "integration ID in" filter in the query, so if the specified name in the request was in that batch, the user was lucky and the integration was returned.

In order to avoid this problem, we are switching from pre-filtering the integrations to post-filtering them. This has a couple of benefits:

1. We will not have to wait to fetch all the authorized integrations from Kessel.
2. We will specifically verify the authorization for the integrations that we fetched from the database.

The downside is that we might actually fetch more integrations from the database than the principal can see, to then discard them. However right now, since we come from a "principal can see every integration from the organization", it directly translates to a "principal can see every integration from the workspace" model. If in the future we see that this becomes an issue we can always check first whether the principal is authorized to access the workspace in the first place.

## Jira ticket
[[RHCLOUD-37430]](https://issues.redhat.com/browse/RHCLOUD-37430)


backend/src/main/resources/application.properties
```diff
@@ -109,16 +109,16 @@ quarkus.unleash.url=http://localhost:4242
...
+inventory-api.authn.client.secret=h91qw8bPiDj9R6VSORsI5TYbceGU5PMH
 inventory-api.authn.mode=oidc-client-credentials
 inventory-api.is-secure-clients=false
 inventory-api.target-url=${clowder.endpoints.kessel-inventory-api:localhost:9081}
 
-relations-api.authn.client.id=insights-notifications
+relations-api.authn.client.id=svc-test
 relations-api.authn.client.issuer=http://localhost:8084/realms/redhat-external
-relations-api.authn.client.secret=development-value-123
+relations-api.authn.client.secret=h91qw8bPiDj9R6VSORsI5TYbceGU5PMH
```
MikelAlejoBR, line 121: These are the default values when launching the Kessel's Inventory API with the authentication enabled: https://github.com/project-kessel/inventory-api/blob/e0fc00b00af6af4977f8b4c66230a34f850da4a2/scripts/get-token.sh

In order to simplify things, I think it's just better to leave the credentials like this by default. It eases development and debugging, and since the credentials are public facing, I think it is just safe to have them here.

MikelAlejoBR:
Closing in favor of https://github.com/RedHatInsights/notifications-backend/pull/3256.

PR #3258, opened 2025-01-29
[RHCLOUD-37537] Identify events with authorization criterion

g-duval:


database/src/main/resources/db/migration/V1.111.0__RHCLOUD-37537_flag_event_with_authorization_criterion.sql
```diff
@@ -0,0 +1,5 @@
+
+ALTER TABLE event ADD COLUMN has_authorization_criterion BOOLEAN NOT NULL DEFAULT FALSE;
+
+CREATE INDEX ix_event_org_id_created_authorization_criterion ON event (org_id, created, has_authorization_criterion, id);
+DROP INDEX ix_event_org_id;
```
MikelAlejoBR, line 5: Why are we dropping the org_id index?

MikelAlejoBR (approved):
Just a question, other than that LGTM!

database/src/main/resources/db/migration/V1.111.0__RHCLOUD-37537_flag_event_with_authorization_criterion.sql
MikelAlejoBR: Aaaaah, I see. Thank you for the explanation!

PR #3268, opened 2025-02-04
 RHCLOUD-37680 | feature: bypass the maximum behavior group creation limit with Unleash

MikelAlejoBR:
There are very specific organizations that, for reasons that make sense, might want to have more than the currently established limit of behavior groups.

For that purpose, we are simply adding an Unleash toggle that will be enabled by default for everybody, and that will allow us to selectively disable it for certain organizations.

## Jira ticket
[[RHCLOUD-37680]](https://issues.redhat.com/browse/RHCLOUD-37680)


g-duval (approved):
LGTM

PR #3273, opened 2025-02-07
GetGroupUsers adding query param admin_only

dscervantes:
rbac added a new query parameter for getgroupusers called admin_only. Now we don't have to use our own filtering for adminonly.

### DIFF
--- a/recipients-resolver/src/main/java/com/redhat/cloud/notifications/recipients/resolver/FetchUsersFromExternalServices.java
+++ b/recipients-resolver/src/main/java/com/redhat/cloud/notifications/recipients/resolver/FetchUsersFromExternalServices.java
@@ -312,6 +312,7 @@ public class FetchUsersFromExternalServices {
         } else {
             users = getWithPagination(page -> {
                 Timer.Sample getGroupUsersPageTimer = Timer.start(meterRegistry);
+                // or do i have to add it here 
                 Page<RbacUser> rbacUsers = retryOnError(() ->
                     rbacServiceToService.getGroupUsers(orgId, groupId, page * recipientsResolverConfig.getMaxResultsPerPage(), recipientsResolverConfig.getMaxResultsPerPage()));
                 // Micrometer doesn't like when tags are null and throws a NPE.
@@ -322,6 +323,7 @@ public class FetchUsersFromExternalServices {
 
             // getGroupUsers doesn't have an adminOnly param.
             if (adminOnly) {
+                //users = getGroupUsers(adminOnly)
                 users = users.stream().filter(User::isAdmin).collect(Collectors.toList());
             }
         }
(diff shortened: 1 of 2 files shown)


recipients-resolver/src/main/java/com/redhat/cloud/notifications/recipients/resolver/rbac/RbacServiceToService.java
```diff
@@ -40,6 +40,7 @@ RbacGroup getGroup(
     Page<RbacUser> getGroupUsers(
             @HeaderParam("x-rh-rbac-org-id") String orgId,
             @PathParam("groupId") UUID groupId,
+            @QueryParam("admin_only") Boolean adminOnly,
```
gwenneg, line 43: They made the field a `string` in RBAC so I'm not sure we can use a `Boolean` here.

Related question: https://github.com/RedHatInsights/insights-rbac/pull/857#discussion_r1950455502

recipients-resolver/src/main/java/com/redhat/cloud/notifications/recipients/resolver/FetchUsersFromExternalServices.java
```diff
@@ -312,6 +312,7 @@ public List<User> getGroupUsers(String orgId, boolean adminOnly, UUID groupId) {
         } else {
             users = getWithPagination(page -> {
                 Timer.Sample getGroupUsersPageTimer = Timer.start(meterRegistry);
+                // or do i have to add it here
```
gwenneg, line 315: Yes, you need to pass the `adminOnly` argument from line 289 to the `rbacServiceToService.getGroupUsers` method. Then lines 324 to 328 will no longer be needed.

recipients-resolver/src/main/java/com/redhat/cloud/notifications/recipients/resolver/rbac/RbacServiceToService.java
gwenneg: Nevermind, they changed it to `boolean` after the initial commit.

MikelAlejoBR:
I was wondering if we need to change this, because now the "admins" are not just the org admins themselves, but also the "Notifications admins". Shouldn't we also call RBAC to fetch those? What do you think @gwenneg?

gwenneg:
Good question @MikelAlejoBR.

The frontend explicitely states that if `Admins` is selected as an option for email/drawer notifs, then the `Organization administrators for your accout` will be notified. This PR doesn't change that, so I think we can safely continue the work here and have it merged.

Then, we could discuss with PM whether notifying org admins still makes sense vs notifying Notifications admins.

MikelAlejoBR:
That's fine, yeah. Thanks for confirming it!

MikelAlejoBR (changes requested):
LGTM! Although you will to remove the `import java.util.stream.Collectors;` import so that the tests pass.

MikelAlejoBR:
Also you will have to add another `Mockito.anyBoolean()` calls in the `FetchUsersExternalServicesTest` class too, so that test can compile and pass.

gwenneg (approved):
LGTM, thanks!

MikelAlejoBR (approved):

PR #3275, opened 2025-02-07
refactor: remove unnecessary log

MikelAlejoBR:
The log statement was part of an old debug session and it is not used anymore.


g-duval (approved):
LGTM

PR #3283, opened 2025-02-10
[RHCLOUD-37777] evaluate authz critera to render event log

g-duval:


backend/src/main/java/com/redhat/cloud/notifications/routers/handlers/event/EventResource.java
```diff
@@ -113,8 +115,32 @@ public Page<EventLogEntry> getInternalEvents(final SecurityContext securityConte
...
+            }
+            if (uuidToExclude.isEmpty()) {
+                uuidToExclude = null;
+            }
+            events = eventRepository.getEvents(orgId, bundleIds, appIds, eventTypeDisplayName, startDate, endDate, basicTypes, compositeTypes, invocationResults, includeActions, notificationStatusSet, query, Optional.ofNullable(uuidToExclude), true);
+            count = eventRepository.count(orgId, bundleIds, appIds, eventTypeDisplayName, startDate, endDate, basicTypes, compositeTypes, invocationResults, notificationStatusSet, Optional.ofNullable(uuidToExclude), true);
+        } else {
+            events = eventRepository.getEvents(orgId, bundleIds, appIds, eventTypeDisplayName, startDate, endDate, basicTypes, compositeTypes, invocationResults, includeActions, notificationStatusSet, query, Optional.empty(), false);
+            count = eventRepository.count(orgId, bundleIds, appIds, eventTypeDisplayName, startDate, endDate, basicTypes, compositeTypes, invocationResults, notificationStatusSet, Optional.empty(), false);
+        }
```
MikelAlejoBR, line 143: I have a couple of questions:

* When Kessel is enabled with fetch the events for an organization twice: one with the criteria and then a second time excluding those events that the principal has no authorization to see. Why did you follow a double fetch approach for this instead of editing the `getEvents` function?
* For the `count` function, did you take into account the events outside the page we are fetching but the user might not have authorization to see? Let me clarify the question with an example.

| Event | Is principal authorized to see? | Date |
|---|---|---|
| A | Yes | 2025-01-01 |
| B | Yes | 2025-01-01 |
| C | No | 2025-01-01 |
| D | No | 2025-01-02 |
| E | Yes | 2025-01-02 |
| F | Yes | 2025-01-02 |

Imagine we fetch all the events with a `LIMIT 3 ORDER BY date ASC`. Our function should just fetch the `A`, `B` and `C` events, but the `count` should also take into account the events the principal has no authorization for in the unfetched pages. However, aren't we just checking the authorization for the page of events we fetched? I am worried that the `count` might return a total count of 5, instead of 4, which is what the principal is authorized to see.

g-duval: Good questions:
Let's take the case where 2000 events complies to user criterias but 3 of those are not accessible for the user because of specific authz criterion.
 - First fetch (`getEventsWithCriterion`) will return only Id + payload of events with criterion (3)
 - In our case, those 3 events will be added to uuid to be excluded list, used in `getEvents` and `count` methods
 Since the returned event list and count should be correct if I'm not mistaken?

MikelAlejoBR: Oh, my bad. I missed that the `limit` and `offset` are not passed to the `getEventsWithCriterion`, so we should be good, and getting all the UUIDs that need to be excluded every time!

I also assumed that most of the events would come with the authorization criteria, that's why I was worried that we'd fetch everything twice, but I guess we'll cross that bridge when it happens.

MikelAlejoBR (approved):

PR #3300, opened 2025-02-17
RHCLOUD-37937 | feature: send general notifications to customers

MikelAlejoBR:
The goal of this feature is to be able to send general communications from Red Hat to the customers. We idea is to have all the parts ready and to just modify the template accordingly whenever it is necessary.

## Implementation details

- An internal REST service is available now in the back end.
- This service calls to another REST service in the engine, which injects an event of the `general-communication` type in the "ingress" Kafka channel.

## Sample email

image

## Jira ticket
[[RHCLOUD-37937]](https://issues.redhat.com/browse/RHCLOUD-37937)

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
@@ -2,6 +2,7 @@ package com.redhat.cloud.notifications.db.repositories;
 
 import com.redhat.cloud.notifications.config.EngineConfig;
 import com.redhat.cloud.notifications.models.CamelProperties;
+import com.redhat.cloud.notifications.models.CompositeEndpointType;
 import com.redhat.cloud.notifications.models.Endpoint;
 import com.redhat.cloud.notifications.models.EndpointProperties;
 import com.redhat.cloud.notifications.models.EndpointType;
@@ -20,6 +21,8 @@ import jakarta.transaction.Transactional;
 import java.time.Duration;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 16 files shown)


engine/src/main/java/com/redhat/cloud/notifications/events/GeneralCommunicationsHelper.java
```diff
@@ -0,0 +1,70 @@
...
+        generalCommunicationAction.setApplication(GENERAL_COMMUNICATIONS_APPLICATION);
+        generalCommunicationAction.setEventType(GENERAL_COMMUNICATIONS_EVENT_TYPE);
+        generalCommunicationAction.setOrgId(orgId);
+        generalCommunicationAction.setTimestamp(LocalDateTime.now(Clock.systemUTC()));
+
+        // Override the recipients' settings for this action, since we want
+        // everyone to receive this action.
+        final Recipient recipient = new Recipient();
+        recipient.setIgnoreUserPreferences(true);
+        generalCommunicationAction.setRecipients(List.of(recipient));
```
MikelAlejoBR, line 40: As a note, I created a singe recipient object which ignores the user preferences for this event, just in case.

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -333,6 +336,42 @@ public boolean disableEndpoint(Endpoint endpoint) {
...
 
+    /**
+     * Finds all the integration names of the given type and groups them by
+     * the organization identifier.
+     * @param integrationType the type of the integration we want to find.
+     * @return a map with the organization ID as the key, and a list of
+     * integration names as the value.
+     */
+    public Map<String, List<String>> findIntegrationNamesByTypeGroupedByOrganizationId(final CompositeEndpointType integrationType) {
+        final String findByTypeQuery =
```
g-duval, line 347: Shouldn't we restrict results to only used teams integrations (linked to an event type) and also to those without the new teams url pattern?

gwenneg: We're not 100% sure about the new URL pattern, so it's risky to exclude some integrations based on a guess.

gwenneg: About the used/unused integrations, that's probably a question for PM.

gwenneg: This was addressed in Slack. Resolving.

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EndpointRepository.java
```diff
@@ -333,6 +336,42 @@ public boolean disableEndpoint(Endpoint endpoint) {
...
+            "SELECT " +
+                "e " +
+            "FROM " +
+                "Endpoint AS e " +
+            "WHERE " +
+                "e.compositeType = :type " +
+            "ORDER BY " +
+                "orgId";
```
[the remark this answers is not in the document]
MikelAlejoBR: Removed!

gwenneg (approved):
LGTM

PR #3303, opened 2025-02-19
[RHCLOUD-37943] Add test coverage on Kessel Inventory call during system endpoint creation

g-duval:
Add test coverage on Kessel Inventory call during system endpoint creation


MikelAlejoBR (approved):
LGTM!

PR #3307, opened 2025-02-20
[RHCLOUD-38198] Fix endpoint deletion from behavior group when other endpoints are linked to the same behavior group

g-duval:


gwenneg (approved):
LGTM

PR #3316, opened 2025-02-26
RHCLOUD-38452 | fix: template rendering errors are not returned in the response

MikelAlejoBR:
Apparently the exception that gets thrown does not contain the useful error message that gets returned back to the back end, so we needed to go one level deeper to get it.

## Jira ticket
[[RHCLOUD-38452]](https://issues.redhat.com/browse/RHCLOUD-38452)


g-duval:
IQE tests failure is not link to this PR

PR #3326, opened 2025-03-05
[RHCLOUD-38590] Add a Kessel integrations consistency check service

g-duval:
this feature has been designed as a one shot test service, that's why no tests were implemented.


MikelAlejoBR (approved):
LGTM!

PR #3354, opened 2025-03-21
RHCLOUD-32168 | fix: webhook properties can be created with any HTTP method

MikelAlejoBR:
We only send "POST" requests to clients with the details of the notifications, so it does not make sense to allow customers to create webhook integrations with any other method in their properties.

## Jira ticket
[[RHCLOUD-32168]](https://issues.redhat.com/browse/RHCLOUD-32168)


g-duval (approved):
LGTM

g-duval:
@MikelAlejoBR could you please have a look to test failure?

PR #3367, opened 2025-03-28
[RHCLOUD-39099] New Aggregation logic can't parse misformated payloads

g-duval:
Some tenants send `context` and `events[].payload` data as Json string instead of Json object.
Json transformer used by new aggregation logic must be more permissive to support this format.

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregator.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregator.java
@@ -27,6 +27,7 @@ import com.redhat.cloud.notifications.recipients.User;
 import com.redhat.cloud.notifications.recipients.recipientsresolver.ExternalRecipientsResolver;
 import com.redhat.cloud.notifications.recipients.request.ActionRecipientSettings;
 import com.redhat.cloud.notifications.recipients.request.EndpointRecipientSettings;
+import com.redhat.cloud.notifications.transformers.BaseTransformer;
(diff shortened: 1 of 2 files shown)


engine/src/main/java/com/redhat/cloud/notifications/processors/email/EmailAggregator.java
```diff
@@ -235,7 +239,7 @@ private void aggregationBasedOnEvent(EventAggregationCriteria eventAggregationCr
                     AbstractEmailPayloadAggregator aggregator = aggregated
                         .computeIfAbsent(recipient, ignored -> EmailPayloadAggregatorFactory.by(eventAggregationCriteria, start, end));
                     // It's aggregation time!
-                    EmailAggregation eventDataToAggregate = new EmailAggregation(aggregation.getOrgId(), eventAggregationCriteria.getBundle(), eventAggregationCriteria.getApplication(), new JsonObject(aggregation.getPayload()));
+                    EmailAggregation eventDataToAggregate = new EmailAggregation(aggregation.getOrgId(), eventAggregationCriteria.getBundle(), eventAggregationCriteria.getApplication(), baseTransformer.toJsonObject(aggregation));
```
gwenneg, line 242: We were transforming `aggregation.getPayload()` (which is actually the return value of `Event#getPayload`) into a `JsonObject` in the old code. Now, we're transforming `aggregation` (the `Event` object) into a `JsonObject`.

I would expect that change to break the implementation logic in several aggregators. Is it a safe change?

gwenneg (approved):
LGTM

PR #3371, opened 2025-03-31
[RHCLOUD-39127] Migrate Kessel Relations to Kessel Inventory

g-duval:


backend/src/main/java/com/redhat/cloud/notifications/auth/kessel/KesselAuthorization.java
jessicarod7, line 1: I saw that there's one method which wasn't brought into the new class, `listWorkspaceIntegrations(final UUID workspaceID)`, called by [`checkMigratedAssets()`](https://github.com/RedHatInsights/notifications-backend/blob/786eef86ac99e7948f7ac54a95a8861569f419b7/backend/src/main/java/com/redhat/cloud/notifications/routers/internal/kessel/KesselAssetsMigrationService.java#L151) in the REST API. Are we planning to remove that endpoint entirely, or are we waiting on the Kessel team to add support to the `NotificationsIntegrationClient`?

jessicarod7 (commented):
Understood what I could, and these changes seem to make sense. I'm not familiar enough with Kessel to evaluate the Authorization test cases, so I'll let someone else review that.

Just left a question on a deprecated method I noticed.

backend/src/main/java/com/redhat/cloud/notifications/auth/kessel/KesselInventoryAuthorization.java
```diff
@@ -0,0 +1,490 @@
...
+                    -> CheckOperation.UPDATE;
+            };
+        } else if ((resourceType == KesselInventoryResourceType.INTEGRATION) && (permission instanceof IntegrationPermission integrationPermission)) {
+            return switch (integrationPermission) {
+                case VIEW, VIEW_HISTORY -> CheckOperation.CHECK;
+                case DELETE, DISABLE, EDIT, ENABLE, TEST -> CheckOperation.UPDATE;
+            };
+        } else {
+            Log.errorf("Resource/permission pair unsupported for Kessel check: %s/%s", resourceType, permission);
+            return CheckOperation.UPDATE;
```
MikelAlejoBR, line 487: Maybe throw an `IllegalArgumentException` here to avoid calling Kessel if we are not able to identify the resource / permission combination.

MikelAlejoBR (approved):
LGTM!

backend/src/main/java/com/redhat/cloud/notifications/auth/kessel/KesselAuthorization.java
g-duval: That's a good question, Kessel team didn't implemented yet the equivalent of this service into the new API.
That's why we keep it as it is for now.

PR #3377, opened 2025-04-02
OCM-14738 | fix: add new template for rosa hcp nodepool upgrades

davidleerh:


engine/src/main/resources/templates/OCM/clusterUpdateInstantEmailBodyV2.html
```diff
@@ -41,6 +41,16 @@
...
         </ul>
 
+        <strong>For more information, refer to</strong> our <a href="https://access.redhat.com/documentation/en-us/red_hat_openshift_service_on_aws/4/html/upgrading/rosa-hcp-upgrading" target="_blank" title="Upgrading ROSA with HCP">Documentation</a>.
+    </p>
+{#else if global_var.template_sub_type == 'upgrade-scheduled-template-rosa-hcp-nodepool'}
+    <p>
+        <strong>What can you expect?</strong>
+        <ul>
+            <li>The machine pools hosting the applications will be upgraded.</li>
+            <li>The worker node for the machine pools will be operational during the upgrade.</li>
```
bchandra-ocp, line 51: What do we mean by "The worker node for the machine pools will be operational during the upgrade"? Isn't this about node pool upgrade notification? How would the worker node will be operational during the upgrade?

engine/src/main/resources/templates/OCM/clusterUpdateInstantEmailBodyV2.html
```diff
@@ -41,6 +41,16 @@
             <li>The machine pools hosting the applications will not be upgraded. The machine pools must be upgraded separately.</li>
         </ul>
 
+        <strong>For more information, refer to</strong> our <a href="https://access.redhat.com/documentation/en-us/red_hat_openshift_service_on_aws/4/html/upgrading/rosa-hcp-upgrading" target="_blank" title="Upgrading ROSA with HCP">Documentation</a>.
+    </p>
+{#else if global_var.template_sub_type == 'upgrade-scheduled-template-rosa-hcp-nodepool'}
+    <p>
+        <strong>What can you expect?</strong>
+        <ul>
+            <li>The machine pools hosting the applications will be upgraded.</li>
```
bchandra-ocp, line 50: "machine pool**s**"? 
Is this a notification for single machine pool or you combine notification for 1+ machine pool upgrades based on all the on-going upgrades?

davidleerh: This is in combination. 
So it should encompass both single and multiple mp/np upgrades (for HCP only).
Should we modify the text to something like 
"machine pool(s)"?

bchandra-ocp: > This is in combination.
So, at 9:00 AM customer starts an upgrade of machine pool MP1. 
And, at 9:05 AM customer starts an upgrade of machine pool MP2.
When will this notification be sent?

engine/src/main/resources/templates/OCM/clusterUpdateInstantEmailBodyV2.html
davidleerh: Based on documentation it seems nodes are replaced in rolling basis:
https://docs.redhat.com/en/documentation/red_hat_openshift_service_on_aws/4/epub/upgrading/rosa-deleting-cluster-upgrade-ocm_rosa-hcp-upgrading?utm_source=chatgpt.com#rosa-upgrade-options_rosa-hcp-upgrading:~:text=This%20initiates%20a%20rolling%20replacement%20of%20nodes%20in%20the%20specified%20machine%20pool%2C%20and%20temporarily%20impacts%20the%20worker%20nodes%20on%20that%20machine%20pool.%20You%20can%20also%20upgrade%20multiple%20machine%20pools%20concurrently.

Please correct me if i'm wrong but doesn't this imply that the service running as part of the nodepool be operational? would it be better to specify below instead:

"Nodepool upgrade initiates a rolling replacement of nodes in the specified machine pool, and temporarily impacts the worker nodes on that machine pool."

engine/src/main/resources/templates/OCM/clusterUpdateInstantEmailBodyV2.html
davidleerh: Sorry for the confusion here;
The template is specifically used per individual node pool upgrade scheduled.
The expectation here is that the email is triggered when the individual upgrade nodepool event is triggered.
The user will first receive the email notification for mp1 soon after they create the schedule.
Separately, the user will receive another email notification for mp2 soon after they create the second schedule.

**From my understanding per HCP, nodepools are upgraded individually and not all at once based on OCM API.
Please correct me if I am wrong.**

FYI, to avoid confusion:
This template is not used for HCP control plane node upgrades, triggered via "rosa upgrade cluster" command or POST /clusters/{id}/upgrade_policies

davidleerh: These will be sent individually as such:
image

bchandra-ocp: Correct, machine pools are upgraded independently and also concurrently (i.e., >1 machine pool can be upgraded at the same time). 

A notification should go for every node pool being upgraded. Similarly, a notification must go out when the control plane is upgraded.

engine/src/main/resources/templates/OCM/clusterUpdateInstantEmailBodyV2.html
bchandra-ocp: Why don't we keep this notification to focus on one thing - i.e, notify of event. 

Other things like explaining what happens can be left to product documentation? In the notification, refer to documentation so that in future if the upgrade procedure evolves, we don't need a code change/deployment?

engine/src/main/resources/templates/OCM/clusterUpdateInstantEmailBodyV2.html
davidleerh: Ad indicated, nodepools scheduled to be upgraded at said same time, will all have individual emails sent (on a per node pool basis) as indicated.

In addition, it seems that we are utilizing the following email body:
https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/main/resources/templates/OCM/clusterUpdateInstantEmailBodyV2.html#L35-L44
Which is separate from the node pool upgrade email body.

I went ahead and adjusted the node pool upgrade body to be more generic.
Would you kindly provide final review and approval? @bchandra-ocp

bchandra-ocp: @davidleerh 
Can we use the documentation for HCP upgrades to convey the behavior of upgrade that way we can work on the documentation easily w/o requiring code changes? https://docs.redhat.com/en/documentation/red_hat_openshift_service_on_aws/4/html/upgrading/rosa-hcp-upgrading#rosa-hcp-upgrading-cli-machinepool_rosa-hcp-upgrading

What is the template for machine pool upgrade?
What is the template for HCP upgrade? [This](https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/main/resources/templates/OCM/clusterUpdateInstantEmailBodyV2.html#L35-L44)? 

I'm not 100% sure what we mean by this statement. What is 'cluster' here? 
> Your cluster will remain operational during the upgrade.

davidleerh: Hi @bchandra-ocp ,
== Cluster upgrade template for non-hcp: (**existing template** -> currently sent to customers)
https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/main/resources/templates/OCM/clusterUpdateInstantEmailBodyV2.html#L18-L34

== Control plane upgrade for hcp rosa: (**existing template** -> currently sent to customers)
https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/main/resources/templates/OCM/clusterUpdateInstantEmailBodyV2.html#L35-L46

++ Nodepool upgrade for hcp rosa: (To implement -> changes suggested in this Pull Request)

From the existing email body / template being sent to customers, the statement "cluster will remain operational",
I assume that it means that the data plane (where user applications run) will remain active and responsive.

Do you wish to edit the existing templates (first two links) I have posted above in accordance to simplifying the overall email body? Thus referring users to the documentation only. Do you mean to imply that i should remove all bullet points and keep the user facing doc link only in the email body?
If not, would you be able to provide me the expected email content and body should look like for each type?
(1. classic cluster upgrade, 2. hcp control plane upgrade, 3. hcp node pool upgrade)

Please kindly advise.

bchandra-ocp:
@davidleerh Been OOTO last 4 days. I'll take a look today and get back. thanks for your patience.

engine/src/main/resources/templates/OCM/clusterUpdateInstantEmailBodyV2.html
bchandra-ocp: > I'm not 100% sure what we mean by this statement. What is 'cluster' here?

Refer to what is a cluster in hosted control planes in the context/scope of upgrades: https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/hosted_control_planes/updating-hosted-control-planes 

In ROSA HCP, referring cluster must be disambiguated by calling explicitly 'Hosted Control Plane' or 'Machine Pool(s)'. In OCM-CS also there's no single API call to upgrade either 'HCP and machine pool(s) together' or 'multiple machine pool(s) together'. 

So to me it looks like:
1. We can't use the same message for both ROSA Classic/OSD and ROSA HCP.
2. We can't use the same message for both HCP and Machine pool upgrades. 

Whether this needs different templates or not is implementation detail I'd leave it up to you.

bchandra-ocp:
@davidleerh are there any pending questions? 
What is the next step on this?

PR #3378, opened 2025-04-02
RHCLOUD-39238 Lowercase all user_id fields from email_subscriptions table

gwenneg:
This PR applies the `LOWER` function to the `user_id` field from the `email_subscriptions` records when:
- the `user_id` field does not exclusively contain lowercase characters
- and there is no existing record in the same table with a `(org_id, LOWER(user_id), event_type_id, subscription_type)` key

In case of conflict, the newer record (user preferences edit from the UI) is kept and the older record (import data) is deleted.


gwenneg:
@g-duval @MikelAlejoBR Please review this PR very carefully. I want both of your reviews before merging it.

gwenneg:
We fixed this another way. If we ever run into the same issue in the future, this PR is still a good solution.

PR #3394, opened 2025-04-14
[RHCLOUD-38890] Update chat template for Errata

jessicarod7:
## Jira ticket

https://issues.redhat.com/browse/RHCLOUD-38890

## Description

This ticket uses the new `common-template` module to create an updated Google Chat/Teams/Slack/Drawer message template for errata notifications. These are placed within a new `SubscriptionServices` bundle mapping.

## Compatibility

No impact, as `common-template` has not been enabled yet.

## Testing

Added Errata test cases for all chat and drawer integrations.


common-template/src/test/java/drawer/TestErrataNotificationsTemplate.java
```diff
@@ -0,0 +1,33 @@
...
+@QuarkusTest
+class TestErrataNotificationsTemplate {
+
+    static final String BUGFIX_ERRATA = "new-subscription-bugfix-errata";
+    private static final Action ACTION = ErrataTestHelpers.createErrataAction();
+
+    @Inject
+    TemplateService templateService;
+
+    @Test
```
g-duval, line 23: For those tests, you can use `@ParameterizedTest` annotation to test each errata templates for each integration type like what we have [here](https://github.com/RedHatInsights/notifications-backend/blob/master/common-template/src/test/java/drawer/TestRbacTemplate.java).
Because rendered templates for those three event types are almost identical, the test should be quite easy to understand/maintain.

g-duval (changes requested):
Great work, I added some minor comments mostly to extend unit test coverage to all of them.

PR #3411, opened 2025-04-24
[RHCLOUD-38780] Link to OCM cluster in application_url

jessicarod7:
## Jira issue

https://issues.redhat.com/browse/RHCLOUD-38780

## Description

Modify the `application_url` generated by `InsightsUrlsBuilder` to link to a specific OCM cluster, if the subscription ID is provided by the incoming payload. This logic is adapted from OCM email templates.

These URLs will be used in "chat" integrations (Slack, MS Teams, Google Chat), PagerDuty, ServiceNow, and Splunk (pending [RHCLOUD-28663](https://issues.redhat.com/browse/RHCLOUD-28663)).

Example MS Teams message:

```markdown
 1 event triggered from Cluster Manager - OpenShift. [Open Cluster Manager](https://console.redhat.com/openshift/details/s/64503ec1-a365-4a1b-8c8b-0a6c519ec5fb?from=notifications&integration=teams)
```

## Compatibility

For incoming `openshift/cluster-manager` payloads:
- If `.events[0].payload.global_vars.subscription_id` is present, a new format URL is created
- If the cluster ID is not present, links to path `/openshift/cluster-list` (this is also the fallback for an invalid cluster ID)

## Testing

Added a test case to `TeamsProcessorTest`.


engine/src/main/java/com/redhat/cloud/notifications/processors/InsightsUrlsBuilder.java
```diff
@@ -105,7 +106,15 @@ public String buildApplicationUrl(JsonObject data, String integration_type) {
             // Settings paths
             case "integrations", "notifications" -> path + "settings/" + application;
             // OpenShift path override
-            case "cluster-manager" -> path;
+            case "cluster-manager" -> {
```
g-duval, line 109: Since we are no longer relying on a single common template for chat notifications, it should be great to have some templates dedicated for OCM with their own logic to build links.

For the current requirement, we can apply the same model for all their event types.

jessicarod7: Changes implemented with a new `OcmClusterUrlExtension` in common-template. If the environment_url is not found within `data.context`, it will fallback to using the `application_url`.

common-template/src/main/resources/templates/google_chat/OCM/ocmDefault.json
```diff
@@ -0,0 +1 @@
+{"text":"{data.events.size()} event{#if data.events.size() > 1}s{/if} triggered from {data.source.application.display_name} - {data.source.bundle.display_name}. <{#if data.context.environment_url??}{data.context.environment_url}/{data.ocm_cluster_path}?from=notifications&integration=google_chat{#else}{data.application_url}{/if}|Open {data.source.application.display_name}>"}
```
g-duval, line 1: Because those templates are dedicated to OCM and `subscription_id` is always present in their messages,  we should keep it simple as email logic like:
`{"text":"{data.events.size()} event{#if data.events.size() > 1}s{/if} triggered from {data.source.application.display_name} - {data.source.bundle.display_name}. <https://cloud.redhat.com/openshift/details/s/{action.events[0].payload.global_vars.subscription_id}#overview?from=notifications&integration=google_chat|Open {data.source.application.display_name}>"}`

We may add complexity later regarding next enhancement to add more details on those notifications.
(same comment for Teams and Slack)

g-duval (changes requested):

common-template/src/main/resources/templates/google_chat/OCM/ocmDefault.json
jessicarod7: I can make that change, I was just concerned that if the format is ever changed, the alerts wouldn't send. But if it's stable, I'll make this change.

common-template/pom.xml
```diff
@@ -15,6 +15,10 @@
 
     <dependencies>
         <!-- notifications modules -->
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-rest-jackson</artifactId>
+        </dependency>
```
g-duval, line 21:
```suggestion
            <artifactId>quarkus-rest-jackson</artifactId>
            <scope>test</scope>
        </dependency>
```

g-duval (approved):
LGTM with adding scope test fo Jackson dep

PR #3460, opened 2025-05-14
[RHCLOUD-40052] Include account Id on daily digest events

g-duval:


g-duval:
Team decision: until account Id will not be not always present in every tenant payload, we should not try to include it.

PR #3461, opened 2025-05-14
[RHCLOUD-38779] Add event type name in OCM emails subjects

g-duval:


gwenneg (approved):
LGTM

IQE failures are unrelated.

PR #3492, opened 2025-06-04
[RHCLOUD-40288] Use new template module to build user preferences page

g-duval:


engine/src/test/java/com/redhat/cloud/notifications/EmailTemplatesRendererHelper.java
```diff
@@ -101,7 +102,7 @@ protected String generateAggregatedEmailBody(Map<String, Object> originalContext
             Arrays.stream(sections).filter(e -> !e.isBlank()).toList());
 
         Map<String, Object> mapData = Map.of("title", "Daily digest - Red Hat Enterprise Linux", "items", List.of(dailyDigestSection), "orgId", DEFAULT_ORG_ID);
-        TemplateDefinition globalDailyTemplateDefinition = new TemplateDefinition(IntegrationType.EMAIL_DAILY_DIGEST_BODY, null, null, null);
+        TemplateDefinition globalDailyTemplateDefinition = new TemplateDefinition(IntegrationType.EMAIL_DAILY_DIGEST_BODY, INTERNAL_DAILY_DIGEST_BUNDLE, null, null);
```
g-duval, line 105: We need to add a fake bundle to avoid this internal parent template to be considered as a daily digest template by default.

backend/src/test/java/com/redhat/cloud/notifications/routers/handlers/userconfig/UserConfigResourceTemplateModuleTest.java
```diff
@@ -0,0 +1,648 @@
+package com.redhat.cloud.notifications.routers.handlers.userconfig;
```
g-duval, line 1: Almost a copy of `UserConfigTest` with some adaptation accorting test cases, to test on app and event type without daily diest template, or event type without instant template at all.

backend/src/main/java/com/redhat/cloud/notifications/config/BackendConfig.java
```diff
@@ -130,6 +131,7 @@ void postConstruct() {
         bypassBehaviorGroupMaxCreationLimitToggle = toggleRegistry.register("bypass-behavior-group-max-creation-limit", true);
         ignoreSourcesErrorOnEndpointDeleteToggle = toggleRegistry.register("ignore-sources-error-on-endpoint-delete", true);
         kesselInventoryUseForPermissionsChecksToggle = toggleRegistry.register("kessel-inventory-permissions-checks", true);
+        useCommonTemplateModuleForUserPrefApisToggle = toggleRegistry.register("use-common-template-module-for-user-pref-apis", true);
```
gwenneg, line 134: Should we log it at startup?

common-template/src/main/java/com/redhat/cloud/notifications/qute/templates/TemplateService.java
```diff
@@ -166,4 +166,13 @@ public Map<String, Object> convertActionToContextMap(final Action action) {
         return objectMapper
             .convertValue(action.getContext(), new TypeReference<Map<String, Object>>() { });
     }
+
+    public boolean isValidTemplateDefinition(final TemplateDefinition config) {
+        try {
+            compileTemplate(config);
+        } catch (TemplateNotFoundException e) {
```
gwenneg, line 173: What if the template exists but doesn't compile for some reason? Does that make the template definition valid? Or should we catch additional exceptions here or an exception with a larger scope?

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/SubscriptionRepository.java
```diff
@@ -51,7 +51,7 @@ public void unsubscribe(String orgId, String username, UUID eventTypeId, Subscri
     @Transactional
     void updateSubscription(String orgId, String username, UUID eventTypeId, SubscriptionType subscriptionType, boolean subscribed) {
 
-        if (subscribed) {
+        if (subscribed && !backendConfig.isUseCommonTemplateModuleForUserPrefApisToggle()) {
```
gwenneg, line 54: Is it okay to remove this check? Can't we run it with the files instead of the DB?

backend/src/main/java/com/redhat/cloud/notifications/routers/handlers/userconfig/UserConfigResource.java
```diff
@@ -103,6 +112,11 @@ public Response saveSettingsByEventType(@Context SecurityContext sec, @NotNull @
                         if (eventType.isPresent() && !eventType.get().isSubscriptionLocked()) {
                             // for each email subscription
                             eventTypeValue.emailSubscriptionTypes.forEach((subscriptionType, subscribed) -> {
+                                boolean supported = isTemplateSupported(bundleName, applicationName, eventType.get(), subscriptionType);
+
+                                if (!supported) {
+                                    throw new NotFoundException(String.format("Event type '%s' doesn't support '%s' subscription", eventType.get().getId(), subscriptionType.name()));
```
gwenneg, line 118: This will likely never be thrown because the frontend will not make it look possible, right?

PR #3495, opened 2025-06-06
[RHCLOUD-40478] Fix splunk and Snow connectors dependency lib vulnerability

g-duval:
Also implements RHCLOUD-40555

### DIFF
--- a/connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyCloudEventDataExtractor.java
+++ b/connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyCloudEventDataExtractor.java
@@ -2,7 +2,7 @@ package com.redhat.cloud.notifications.connector.pagerduty;
 
 import com.redhat.cloud.notifications.connector.CloudEventDataExtractor;
 import com.redhat.cloud.notifications.connector.authentication.AuthenticationDataExtractor;
-import com.redhat.cloud.notifications.connector.http.UrlValidator;
+import com.redhat.cloud.notifications.connector.pagerduty.config.PagerDutyConnectorConfig;
 import io.vertx.core.json.JsonObject;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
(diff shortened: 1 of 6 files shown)


connector-servicenow/src/main/java/com/redhat/cloud/notifications/connector/servicenow/ServiceNowCloudEventDataExtractor.java
```diff
@@ -6,23 +6,21 @@
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import org.apache.camel.Exchange;
-import org.apache.commons.validator.routines.UrlValidator;
+import org.apache.hc.core5.http.URIScheme;
```
gwenneg, line 9: Gentle reminder: we talked about replacing this lib with a simple constant. Is that still your plan?

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyCloudEventDataExtractor.java
```diff
@@ -37,16 +31,11 @@ public void extract(Exchange exchange, JsonObject cloudEventData) throws Illegal
             exchange.setProperty(TARGET_URL, providedUrl);
         }
 
+        UrlValidationHelper.validateTargetUrl(exchange);
```
gwenneg, line 34: We did allow `http` before with PagerDuty and now we're forbidding it. Did you check in the DB that no one is using an `http` URL?

connector-common-http/src/main/java/com/redhat/cloud/notifications/connector/http/UrlValidationHelper.java
```diff
@@ -0,0 +1,32 @@
...
+
+
+import org.apache.camel.Exchange;
+import org.apache.http.ProtocolException;
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import static com.redhat.cloud.notifications.connector.ExchangeProperty.TARGET_URL;
+
+public class UrlValidationHelper {
```
gwenneg, line 11: Nitpick:
```suggestion
public class UrlValidator {
```

connector-common-http/src/main/java/com/redhat/cloud/notifications/connector/http/UrlValidationHelper.java
```diff
@@ -0,0 +1,32 @@
...
+    public static void validateTargetUrl(Exchange exchange) throws Exception {
+        String targetUrl = exchange.getProperty(TARGET_URL, String.class);
+        try {
+            String scheme = (new URI(targetUrl)).getScheme();
+
+            if (HTTP_SCHEME.equalsIgnoreCase(scheme)) {
+                throw new ProtocolException("HTTP protocol is not supported");
+            } else if (!HTTPS_SCHEME.equalsIgnoreCase(scheme)) {
+                throw new IllegalArgumentException("URL validation failed");
+            }
```
gwenneg, line 25: This could be simplified:
```suggestion
            if (!HTTPS_SCHEME.equalsIgnoreCase(scheme)) {
                throw new IllegalArgumentException("Illegal URL scheme: " + scheme + ". Please use https.");
            }
```

connector-pagerduty/src/main/java/com/redhat/cloud/notifications/connector/pagerduty/PagerDutyCloudEventDataExtractor.java
jessicarod7: I think the only requirement was that the test cases needed to use non-TLS URLs (they gave a cert error otherwise) but maybe that's not the case anymore

PR #3511, opened 2025-06-16
[RHCLOUD-39778] Handle integer format for total recipients counter

g-duval:
## Summary by Sourcery

Handle integer-formatted recipient counts by treating Integer details directly and updating tests and helpers accordingly

Enhancements:
- Support direct Integer values for the totalRecipients field in event histories
- Allow test helpers to accept Object-typed totalRecipients rather than only strings

Tests:
- Update assertion logic to handle both Integer and String formats for recipients count
- Modify tests to pass integer values for totalRecipients


gwenneg (approved):
LGTM

PR #3602, opened 2025-07-20
RHCLOUD-40807 Authenticate with RBAC using OIDC (backend)

gwenneg:
Assisted-by: Claude-4-Sonnet (via Cursor)

## Summary by Sourcery

Enable RBAC authentication using OIDC alongside the existing PSK option by introducing dedicated RestClient interfaces, wiring in a feature toggle, configuring necessary OIDC properties, and covering both flows with integration tests and mock servers.

New Features:
- Add OIDC-based RestClient (RbacWorkspacesOidcClient) for RBAC workspace retrieval
- Introduce PSK-based RestClient (RbacWorkspacesPskClient) to replace the deprecated RbacServer#getWorkspaces method
- Implement feature toggle (rbac-oidc-auth) to switch between OIDC and PSK clients
- Add integration tests and MockServer resources for both OIDC and PSK authentication scenarios

Enhancements:
- Update WorkspaceUtils to choose between OIDC and PSK clients based on the feature toggle
- Remove the deprecated RbacServer#getWorkspaces method and related code

Build:
- Add Quarkus OIDC client filter dependency and OIDC configuration properties to application.properties

CI:
- Add OIDC environment variables (issuer URL, client ID/secret, enabled flag) to clowdapp-backend.yaml

Tests:
- Include RbacWorkspacesOidcClientTest and setup mock lifecycles for OIDC and RBAC servers

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/workspace/WorkspaceUtils.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/auth/rbac/workspace/WorkspaceUtils.java
@@ -76,7 +76,7 @@ public class WorkspaceUtils {
     public UUID getDefaultWorkspaceId(final String orgId) {
         // Call RBAC.
         Page<RbacWorkspace> workspacePage;
-        if (backendConfig.isRbacOidcAuthEnabled()) {
+        if (backendConfig.isRbacOidcAuthEnabled(orgId)) {
             workspacePage = rbacWorkspacesOidcClient.getWorkspaces(
                 orgId,
                 WorkspaceType.DEFAULT.toString().toLowerCase(),
--- a/backend/src/main/java/com/redhat/cloud/notifications/config/BackendConfig.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/config/BackendConfig.java
@@ -152,7 +152,7 @@ public class BackendConfig {
         config.put(KESSEL_DOMAIN, getKesselDomain());
         config.put(RBAC_ENABLED, isRBACEnabled());
         config.put(UNLEASH, unleashEnabled);
-        config.put(rbacOidcAuthToggle, isRbacOidcAuthEnabled());
+        config.put(rbacOidcAuthToggle, isRbacOidcAuthEnabled(null));
 
         Log.info("=== Startup configuration ===");
         config.forEach((key, value) -> {
@@ -314,9 +314,10 @@ public class BackendConfig {
         }
     }
 
-    public boolean isRbacOidcAuthEnabled() {
+    public boolean isRbacOidcAuthEnabled(String orgId) {
         if (unleashEnabled) {
-            return unleash.isEnabled(rbacOidcAuthToggle, false);
+            UnleashContext unleashContext = buildUnleashContextWithOrgId(orgId);
+            return unleash.isEnabled(rbacOidcAuthToggle, unleashContext, false);
         } else {
             return false;
         }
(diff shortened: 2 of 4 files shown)


.rhcicd/clowdapp-backend.yaml
```diff
@@ -343,3 +357,9 @@ parameters:
   value: "40"
 - name: LIVENESS_INITIAL_DELAY
   value: "40"
+- name: OIDC_CLIENT_ENABLED
+  description: If true, the OIDC client will be enabled.
+  value: "true"
```
g-duval, line 362: Shouldn't it be false by default? Event if not used in runtime code because of unleash FF, we could avoid to init this on startup in env who can't support it yet?

.rhcicd/clowdapp-backend.yaml
```diff
@@ -215,6 +215,20 @@ objects:
...
         - name: KESSEL_DOMAIN
           value: ${KESSEL_DOMAIN}
+        - name: QUARKUS_OIDC_CLIENT_AUTH_SERVER_URL
+          value: ${OIDC_ISSUER}
+        - name: QUARKUS_OIDC_CLIENT_CLIENT_ENABLED
+          value: ${OIDC_CLIENT_ENABLED}
+        - name: QUARKUS_OIDC_CLIENT_CLIENT_ID
+          valueFrom:
+            secretKeyRef:
+              name: kessel-authentication
```
g-duval, line 225: Shoudn't we rename those keys to something more generic like `user-access-authentcation` (because it's the name of Rbac app in the console and it could be more stable than the internal team name?).
Specially because `relations-api` should be remove in a near future.

backend/src/test/java/com/redhat/cloud/notifications/auth/rbac/workspace/RbacWorkspacesOidcClientTest.java
```diff
@@ -0,0 +1,39 @@
...
+    @RestClient
+    RbacWorkspacesOidcClient rbacClient;
+
+    private static final String TEST_ORG_ID = "test-org-id";
+
+    @Test
+    @DisplayName("Should successfully call getWorkspaces with OIDC authentication")
+    void shouldSuccessfullyCallGetWorkspaces() {
+        // If OIDC headers are missing, the mock server will return 401 and this will throw an exception
+        // If OIDC headers are present and correct, the mock server returns 200 and this succeeds
```
g-duval, line 34: Do we have some tests somewhere for incorrect OIDC header, OIDC issuer timeout or OIDC issuer errors like 500, 401, 403?

g-duval: My concern here is more to check how Notifications will handle different kind of errors/exceptions that Quarkus OIDC extention could return, will notifiations will return a 404 or 403 if unable get a proper RBAC response because of an OIDC issue?

.rhcicd/clowdapp-backend.yaml
g-duval: great, thank you :)

backend/src/test/java/com/redhat/cloud/notifications/auth/rbac/workspace/RbacWorkspacesOidcClientTest.java
gwenneg: I created https://issues.redhat.com/browse/RHCLOUD-41298 to further investigate this.

PR #3606, opened 2025-07-21
[RHCLOUD-34917] Add config to load, build autorenewed certificate

jessicarod7:
## Jira ticket

https://issues.redhat.com/browse/RHCLOUD-34917

## Description

Adds new configuration to use the autorenewing certificate for IT services:

1. Loads generated certificate into volume mount `/mnt/keystores`
2. Runs an init container with a mounted script to assemble the certificate into a JKS. This could not be done with the `ubi9/openjdk-21` image, as it does not include the required `openssl-devel` package.

## Compatibility

Changes certificate used to communicate with IT services. This will not take effect until the parameters specifying which JKS file to use are updated in the backend.

## Testing

- Validated script commands on local machine
- Currently testing in ephemeral environment


.rhcicd/clowdapp-recipients-resolver.yaml
```diff
@@ -191,6 +201,44 @@ objects:
...
+    - name: it-services-jks-builder
+      image: ${IT_SERVICES_JKS_BUILDER_IMAGE}:${IT_SERVICES_JKS_BUILDER_IMAGE_TAG}
+      command: ["bash", "/mnt/it-services-jks-builder/build-it-services-jks.sh"]
+      env:
+      - name: IT_SERVICES_KEYSTORE_PASSWORD
+        valueFrom:
+          secretKeyRef:
+            name: it-services
+            key: it-services-keystorepassword
+            optional: true
```
jessicarod7, line 214: There's also the option to inherit all environment variables from the main container, but I only needed this one so I decided to make it explicit.

recipients-resolver/src/test/java/com/redhat/cloud/notifications/recipients/startup/StartupUtilsTest.java
```diff
@@ -25,7 +25,7 @@ public class StartupUtilsTest {
 
     @Test
     void readKeystoreTest() {
-        String resourceName = "testKeystore.jks";
+        String resourceName = "testKeystore.p12";
```
g-duval, line 28: We also need to add `quarkus.rest-client.it-s2s.key-store-type=P12` in `application.properties` file ([JKS is the defaut type](https://quarkus.io/guides/all-config#quarkus-rest-client-config_quarkus-rest-client-client-key-store-type))

.rhcicd/clowdapp-recipients-resolver.yaml
```diff
@@ -191,6 +201,44 @@ objects:
...
+  data:
+    build-it-services-jks.sh: |
+      #!/bin/bash -e
+      # Install OpenJDK 21 and OpenSSL
+      microdnf -y install java-21-openjdk-devel openssl-devel
+
+      # Extract account name (subject UID)
+      ACCOUNT_NAME=$(openssl x509 -in /mnt/keystores/tls.crt -noout -subject -nameopt rfc2253 |\
+          sed 's/^subject=//' |\
+          awk -F ',' '{for (i=1;i<=NF;i++) if ($i ~ /^UID/) print $i;}' | awk -F '=' '{print $2}')
```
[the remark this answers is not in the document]
g-duval: Not sure to understand why do we need to extract the account name? In previous implementation, it was just an alias manually added, wasn't it?

g-duval (changes requested):
looks already great :)

g-duval (changes requested):

g-duval (approved):
LGTM

recipients-resolver/src/test/java/com/redhat/cloud/notifications/recipients/startup/StartupUtilsTest.java
jessicarod7: Replacing config in `application.properties` with parameter/env var to allow for testing with the existing JKS in ephemeral. See [!153239](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/153239) for more context.

PR #3659, opened 2025-08-27
Reapply certificate changes

jessicarod7:
Minor change to try to get a more usable error message out

## Summary by Sourcery

Reapply and refine certificate handling by restructuring secret mounts into dedicated volumes, updating the init script to use the new paths for PKCS#12 generation, and improve error reporting in the RecipientsResolverResource.

Bug Fixes:
- Preserve the full exception when throwing WebApplicationException in RecipientsResolverResource to surface more detailed error messages

Enhancements:
- Add keystore.p12, tls.crt, tls.key, and ca.crt items to secrets and mount them under separate `it-services-cert` and `p12-archive` volumes to bypass read-only restrictions
- Update the init script to reference `/mnt/it-services-cert` for certificate files and generate the PKCS#12 archive on the new volume


g-duval:
Done with another way

PR #3664, opened 2025-08-29
SWATCH-3874: Add usage-threshold-exceeded instant email for subscription watch

kflahert:
https://issues.redhat.com/browse/SWATCH-3874

Creating a placeholder email template to test SWATCH integration.

### DIFF
--- a/common-template/src/test/java/helpers/SubscriptionWatchTestHelpers.java
+++ b/common-template/src/test/java/helpers/SubscriptionWatchTestHelpers.java
@@ -1,0 +1,49 @@
+package helpers;
+
+import com.redhat.cloud.notifications.ingress.Action;
+import com.redhat.cloud.notifications.ingress.Context;
+import com.redhat.cloud.notifications.ingress.Event;
+import com.redhat.cloud.notifications.ingress.Metadata;
+import com.redhat.cloud.notifications.ingress.Payload;
+
+import java.time.LocalDateTime;
+import java.util.List;
+
+import static helpers.TestHelpers.DEFAULT_ORG_ID;
+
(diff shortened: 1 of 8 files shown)


common-template/src/test/java/helpers/SubscriptionWatchTestHelpers.java
```diff
@@ -0,0 +1,49 @@
...
+    public static Action createSubscriptionWatchAction() {
+        Action emailActionMessage = new Action();
+        emailActionMessage.setBundle("subscription-services");
+        emailActionMessage.setApplication("subscription-watch");
+        emailActionMessage.setTimestamp(LocalDateTime.of(2020, 10, 3, 15, 22, 13, 25));
+        emailActionMessage.setEventType("usage-threshold-exceeded");
+        emailActionMessage.setRecipients(List.of());
+
+        emailActionMessage.setContext(
+                new Context.ContextBuilder()
```
g-duval, line 25: Usually the `Context` part is used to define global parameters that can be applied to all events, such as an host id if all events of this message are related to the same host.

common-template/src/test/java/helpers/SubscriptionWatchTestHelpers.java
```diff
@@ -0,0 +1,49 @@
...
+                        .withAdditionalProperty("org_id", "123456789")
```
g-duval, line 26: The org Id is already present at the higher level, as you defined it at the end of this method  (`emailActionMessage.setOrgId(DEFAULT_ORG_ID);`)

g-duval (commented):
Looks already great, dropped few comments if you can have a look?

PR #4150, opened 2026-02-19
[RHCLOUD-45311] Implement engine gracefull shutdown

g-duval:
## Why?
Currently, some Kafka messages could be lost during engine pod shutdown, because Quarkus may close database connection before ending processing Kafka events.

## Summary by Sourcery

Implement a graceful shutdown mechanism to ensure Kafka consumers and asynchronous event processing complete before engine resources are closed during pod termination.

Enhancements:
- Introduce a GracefulShutdownManager to coordinate pausing Kafka consumers and draining async executors during application shutdown.
- Expose the EventConsumer executor to allow controlled draining of in-flight tasks on shutdown.
- Configure a Quarkus shutdown timeout to align application shutdown behavior with Kubernetes pod termination timing.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced graceful shutdown: consumer channels are paused and a short drain period allows in‑flight messages to complete, reducing potential message loss.
  * Added configurable shutdown timing and delay controls, providing a longer, adjustable shutdown window.
  * Improved shutdown logging to surface pause outcomes and warn on pause failures.

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/GracefulShutdownManager.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/GracefulShutdownManager.java
@@ -56,11 +56,9 @@ public class GracefulShutdownManager {
      * and BEFORE Quarkus infrastructure (EntityManager, datasource) shutdown.
      *
      * The priority value of 100 ensures this executes early in the shutdown
-     * sequence (lower numbers = higher priority), giving us control over the
-     * shutdown order. Default is 2000 (APPLICATION level).
+     * sequence (lower numbers = higher priority). Default is 2500.
      */
-    @Priority(100)
-    void onShutdown(@Observes ShutdownEvent event) {
+    void onShutdown(@Observes @Priority(100) ShutdownEvent event) {
         Log.info("=== Starting graceful shutdown sequence ===");
 
         // Step 1: Stop all Kafka consumers from fetching new messages
--- a/engine/src/main/java/com/redhat/cloud/notifications/GracefulShutdownManager.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/GracefulShutdownManager.java
@@ -1,0 +1,191 @@
+package com.redhat.cloud.notifications;
+
+import com.redhat.cloud.notifications.config.EngineConfig;
+import com.redhat.cloud.notifications.events.ConnectorReceiver;
+import com.redhat.cloud.notifications.events.EventConsumer;
+import com.redhat.cloud.notifications.events.ReplayEventConsumer;
+import com.redhat.cloud.notifications.exports.ExportEventListener;
+import io.quarkus.logging.Log;
+import io.quarkus.runtime.ShutdownEvent;
+import io.smallrye.reactive.messaging.ChannelRegistry;
+import io.smallrye.reactive.messaging.PausableChannel;
+import jakarta.annotation.Priority;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.event.Observes;
+import jakarta.inject.Inject;
+
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Manages graceful shutdown of Kafka consumers and async processing to ensure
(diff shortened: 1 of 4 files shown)


engine/src/main/java/com/redhat/cloud/notifications/GracefulShutdownManager.java
```diff
@@ -0,0 +1,189 @@
...
+
+    @Inject
+    EngineConfig engineConfig;
+
+    /**
+     * High-priority shutdown observer that runs BEFORE @PreDestroy methods
+     * and BEFORE Quarkus infrastructure (EntityManager, datasource) shutdown.
+     *
+     * The priority value of 100 ensures this executes early in the shutdown
+     * sequence (lower numbers = higher priority). Default is 2500.
```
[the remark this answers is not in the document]
g-duval: Default is 2500 (https://github.com/quarkusio/quarkus/discussions/38028)

engine/src/main/resources/application.properties
```diff
@@ -184,4 +186,9 @@ quarkus.cache.caffeine.aggregation-target-email-subscription-endpoints.expire-af
 quarkus.unleash.active=false
 quarkus.unleash.url=http://localhost:4242
 
+# Graceful shutdown configuration for pod rollouts
+# Ensures all Kafka messages are processed before EntityManager closes
+# Must be less than terminationGracePeriodSeconds value of clowdapp-engine.yaml (default value is 30s, current is 75s)
+quarkus.shutdown.timeout=60S
```
gwenneg: Is this a valid Duration value, or do we need to transform it to `PT60S`?

engine/src/main/java/com/redhat/cloud/notifications/GracefulShutdownManager.java
```diff
@@ -0,0 +1,88 @@
...
+@ApplicationScoped
+public class GracefulShutdownManager {
+
+    @Inject
+    ChannelRegistry channelRegistry;
+
+    @Inject
+    KafkaClientService kafkaClientService;
+
+    void onShutdown(@Observes ShutdownDelayInitiatedEvent event) {
```
gwenneg:
```suggestion
    @ShutdownDelayInitiated
    void onShutdown() {
```

engine/src/main/java/com/redhat/cloud/notifications/GracefulShutdownManager.java
```diff
@@ -0,0 +1,88 @@
...
+                Log.warnf(e, "  ✗ Failed to pause channel: %s - Messages may be lost during shutdown!", channelName);
+            }
+        }
+
+        Log.infof("Successfully paused %d Kafka consumer(s), skipped %d non-pausable channel(s)", pausedCount, skippedCount);
+
+        // Give Kafka a brief moment to process in-flight fetch requests
+        try {
+            Log.infof("Waiting 5s to process already pulled messages", pausedCount, skippedCount);
+            Thread.sleep(5000);
```
gwenneg, line 83: There's no need for a `wait` instruction here. Regardless of how long the execution of the `void onShutdown` method needs, Quarkus will wait for the duration specified in the `quarkus.shutdown.delay` config property before proceeding with the shutdown.

engine/src/main/java/com/redhat/cloud/notifications/GracefulShutdownManager.java
gwenneg: https://quarkus.io/guides/lifecycle#pre_shutdown_annotation

engine/src/main/java/com/redhat/cloud/notifications/GracefulShutdownManager.java
g-duval: Actually I re-tested it locally and we need to keep it.
The `quarkus.shutdown.delay` seems to apply only on the content of the `onShutdown` method.
`channel.pause()` method returns when it stops to consume new messages, not when already pulled messages processing ends.
Without it, the database connection shutdowns immediatly after pausing Kafka consumers , then we ccould lost some messages because not able to persist them.

gwenneg (approved):
LGTM

PR #4287, opened 2026-04-01
[RHCLOUD-35790] Re-add remote cache event deduplication 

jessicarod7:
## Jira work item

https://redhat.atlassian.net/browse/RHCLOUD-35790

## Description

This re-adds support for event deduplication via remote caches, implemented with Valkey. It works identically to the PostgreSQL database table, and depends on two parameters:
- `IN_MEMORY_DB_ENABLED`: instructs Clowder to create a Valkey instance to store the data in
- `notifications-engine.valkey-event-deduplicator.enabled`: Unleash toggle to switch to using Valkey for event dedup

Deduplication is done with keys in the format

```
engine:event-deduplication:<eventTypeId>:<dedupKey>
```

And additionally stores `deleteAfter` as a value so it is internally similar to Postgres. If the event has not been seen before, the key is set to expire at this time.

**Improvements since last time:**
- Reimplemented using the lower-level Vert.x Mutiny API, meaning that it will not try to connect to a Valkey instance on startup. This means it will work where Valkey isn't available.
- Implements the new deduplication logic
- The test configuration actually works locally

## Compatibilty

No impact unless _both_ of the above feature flags are enabled.

## Testing

- Added test cases to check that event deduplication works with either Valkey or Postgres enabled
- Added ValkeyServiceTest to validate that duplicates are identified correctly
- Modified TestLifecycleManager to configure Valkey pod

## Summary by CodeRabbit

* **New Features**
  * Optional in-memory DB mode for event deduplication with a configurable toggle and Redis/Valkey-backed deduplicator; cross-checking between native and Valkey deduplication with mismatch logging and safer cleanup on errors.
  * Test profile support to enable Redis/Valkey provisioning for test runs.

* **Tests**
  * New and expanded tests for deduplication, duplicate detection, and expiry; test lifecycle now supports independent Postgres and Valkey containers.

### DIFF
--- a/engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
@@ -101,6 +101,9 @@ public class EventConsumer {
     @Inject
     SeverityTransformer severityTransformer;
 
+    @Inject
+    ValkeyService valkeyService;
+
     ConsoleCloudEventParser cloudEventParser = new ConsoleCloudEventParser();
 
     private Counter rejectedCounter;
@@ -262,7 +265,21 @@ public class EventConsumer {
              * Before we persist the event into the DB and process it, we need to check whether the event is
              * a duplicate using the custom event deduplication logic tenants might have implemented.
              */
-            if (!eventDeduplicator.isNew(event)) {
+            boolean isNewEvent;
+
+            try {
+                isNewEvent = eventDeduplicator.isNew(event);
+            } catch (Exception e) {
+                // If Valkey deduplication is enabled, remove any keys that may have been created
+                if (engineConfig.isInMemoryDbEnabled() && engineConfig.isValkeyEventDeduplicatorEnabled()) {
+                    Optional<String> dedupKey = eventDeduplicator.getEventDeduplicationConfig(event).getDeduplicationKey(event);
+                    dedupKey.ifPresent(key -> valkeyService.removeEventFromDeduplication(key));
+                }
+
+                throw e;
+            }
+
+            if (!isNewEvent) {
                 // The event is already known and should therefore be ignored.
                 Log.debug("Duplicated event ignored");
                 registry.counter(DUPLICATE_EVENT_COUNTER_NAME,
--- a/engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
@@ -273,7 +273,7 @@ public class EventConsumer {
                 // If Valkey deduplication is enabled, remove any keys that may have been created
                 if (engineConfig.isInMemoryDbEnabled() && engineConfig.isValkeyEventDeduplicatorEnabled()) {
                     Optional<String> dedupKey = eventDeduplicator.getEventDeduplicationConfig(event).getDeduplicationKey(event);
-                    dedupKey.ifPresent(key -> valkeyService.removeEventFromDeduplication(key));
+                    dedupKey.ifPresent(key -> valkeyService.removeEventFromDeduplication(event.getEventType().getId(), key));
                 }
 
                 throw e;
--- a/engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
@@ -173,6 +173,8 @@ public class EventConsumer {
         Timer.Sample consumedTimer = Timer.start(registry);
         String payload = message.getPayload();
         Map<String, String> tags = new HashMap<>();
+        Event event = null;
+
         // The two following variables have to be final or effectively final. That why their type is String[] instead of String.
         /*
          * Step 1
@@ -258,27 +260,14 @@ public class EventConsumer {
              * The EventType was found. It's time to create an Event from the current message.
              */
             Optional<String> sourceEnvironmentHeader = kafkaHeaders.get(SOURCE_ENVIRONMENT_HEADER);
-            Event event = new Event(eventType, payload, eventWrapperToProcess, sourceEnvironmentHeader, messageId);
+            event = new Event(eventType, payload, eventWrapperToProcess, sourceEnvironmentHeader, messageId);
 
             /*
              * Step 5
              * Before we persist the event into the DB and process it, we need to check whether the event is
              * a duplicate using the custom event deduplication logic tenants might have implemented.
              */
-            boolean isNewEvent;
-
-            try {
-                isNewEvent = eventDeduplicator.isNew(event);
-            } catch (Exception e) {
-                // If Valkey deduplication is enabled, remove any keys that may have been created
-                if (engineConfig.isInMemoryDbEnabled() && engineConfig.isValkeyEventDeduplicatorEnabled()) {
-                    Optional<String> dedupKey = eventDeduplicator.getEventDeduplicationConfig(event).getDeduplicationKey(event);
-                    dedupKey.ifPresent(key -> valkeyService.removeEventFromDeduplication(event.getEventType().getId(), key));
-                }
-
-                throw e;
-            }
-
+            boolean isNewEvent = eventDeduplicator.isNew(event);
             if (!isNewEvent) {
                 // The event is already known and should therefore be ignored.
                 Log.debug("Duplicated event ignored");
@@ -315,10 +304,16 @@ public class EventConsumer {
         } catch (Exception e) {
             /*
              * An exception was thrown at some point during the Kafka message processing,
-             * it is logged and added to the exception counter metric.
+             * it is logged and added to the exception counter metric. Any deduplication
+             * events added in Valkey will be rolled back.
              */
             processingExceptionCounter.increment();
             Log.infof(e, "Could not process the payload: %s", payload);
+            if (event != null && engineConfig.isInMemoryDbEnabled() && engineConfig.isValkeyEventDeduplicatorEnabled()) {
+                Event finalEvent = event;
+                Optional<String> dedupKey = eventDeduplicator.getEventDeduplicationConfig(event).getDeduplicationKey(event);
+                dedupKey.ifPresent(key -> valkeyService.removeEventFromDeduplication(finalEvent.getEventType().getId(), key));
+            }
         } finally {
             consumedTimer.stop(registry.timer(
                     CONSUMED_TIMER_NAME,
--- a/engine/src/main/java/com/redhat/cloud/notifications/events/ValkeyService.java
+++ b/engine/src/main/java/com/redhat/cloud/notifications/events/ValkeyService.java
@@ -1,0 +1,44 @@
+package com.redhat.cloud.notifications.events;
+
+import io.quarkus.redis.datasource.RedisDataSource;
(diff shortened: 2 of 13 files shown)


engine/src/main/java/com/redhat/cloud/notifications/events/ValkeyService.java
```diff
@@ -0,0 +1,91 @@
...
+    /** Implementation of the Redis/Valkey API, using {@link #valkeyClient} */
+    private RedisAPI valkey;
+
+    @PostConstruct
+    void initialize() {
+        if (config.isInMemoryDbEnabled()) {
+            if (valkeyHost.isEmpty() || valkeyHost.get().isEmpty()) {
+                throw new IllegalStateException("In-memory DB enabled, but Valkey connection string was not provided");
+            } else {
+                RedisOptions valkeyOptions = new RedisOptions().setConnectionString(valkeyHost.get());
```
jessicarod7: There should be a mapping here to go from `valkey://` to `redis://`, like in TestLifecycleManager

engine/src/main/java/com/redhat/cloud/notifications/events/deduplication/EventDeduplicator.java
```diff
@@ -43,15 +53,22 @@ public boolean isNew(Event event) {
...
         }
 
-        String sql = "INSERT INTO event_deduplication(event_type_id, deduplication_key, delete_after) " +
-                "VALUES (:eventTypeId, :deduplicationKey, :deleteAfter) " +
-                "ON CONFLICT (event_type_id, deduplication_key) DO NOTHING";
+        UUID eventTypeId = event.getEventType().getId();
+        LocalDateTime deleteAfter = eventDeduplicationConfig.getDeleteAfter(event);
+
+        if (engineConfig.isInMemoryDbEnabled() && engineConfig.isValkeyEventDeduplicatorEnabled()) {
+            return valkeyService.isNewEvent(eventTypeId, deduplicationKey.get(), deleteAfter, event.getId());
```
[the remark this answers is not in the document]
jessicarod7: I'll add the key to the catch block.

jessicarod7:
this build will probably fail, figuring out why

engine/src/main/java/com/redhat/cloud/notifications/events/ValkeyService.java
```diff
@@ -0,0 +1,99 @@
...
+    public boolean isNewEvent(UUID eventTypeId, String deduplicationKey, LocalDateTime deleteAfter, UUID eventId) {
+        String key = formatDeduplicationKey(eventTypeId, deduplicationKey);
+        String deleteAfterIso = deleteAfter.format(DateTimeFormatter.ISO_DATE_TIME);
+
+        boolean isNew = valkey.setnxAndAwait(key, deleteAfterIso).toBoolean();
+        if (isNew) {
+            boolean expireSet = valkey.expireatAndAwait(List.of(
+                    key,
+                    String.valueOf(deleteAfter.toEpochSecond(ZoneOffset.UTC))
+            )).toBoolean();
```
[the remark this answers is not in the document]
jessicarod7: I wasn't sure about the arguments to pass and the RESP3 replies, but the bot seems relatively confident so I'll give it a try.

engine/src/main/java/com/redhat/cloud/notifications/events/ValkeyService.java
```diff
@@ -0,0 +1,109 @@
...
+        } catch (Exception ignored) {
+            // Nil response
+            isNew = false;
+        }
+
+        if (isNew) {
+            // dedup key may include private information, so other fields are used
+            Log.warnf("unable to set expiry for Valkey event deduplication [event_type_id=%s, event_id=%s, delete_after=%s]",
+                    eventTypeId, eventId, deleteAfterIso);
+        }
```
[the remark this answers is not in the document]
jessicarod7: oops, mixed this up with the previous call to `SETEX`

engine/src/main/java/com/redhat/cloud/notifications/events/ValkeyService.java
```diff
@@ -0,0 +1,110 @@
...
+        try {
+            if (valkeyResp == null) {
+                isNew = false;
+            } else {
+                isNew = valkeyResp.toString().equals("OK");
+            }
+        } catch (Exception ignored) {
+            // Invalid response could not be mapped to string. Assume event is new
+            // dedup key may include private information, so other fields are used
+            Log.warnf("unable to check for duplicate event in Valkey [event_type_id=%s, event_id=%s, delete_after=%s]",
```
g-duval: In most cases, the eventId is empty at the step, I would recommand to add the `deduplicationKey` here.
I don't think those dedup keys could include sensible data (like security password or token).

engine/src/main/java/com/redhat/cloud/notifications/events/ValkeyService.java
```diff
@@ -0,0 +1,110 @@
...
+                "NX",
+                "EXAT",
+                String.valueOf(deleteAfter.toEpochSecond(ZoneOffset.UTC))
+        ));
+
+        try {
+            if (valkeyResp == null) {
+                isNew = false;
+            } else {
+                isNew = valkeyResp.toString().equals("OK");
```
g-duval, line 100: Shoudlnt we only surround line 93 by this try catch?

engine/src/main/java/com/redhat/cloud/notifications/events/deduplication/EventDeduplicator.java
```diff
@@ -43,14 +54,35 @@ public boolean isNew(Event event) {
...
+        LocalDateTime deleteAfter = eventDeduplicationConfig.getDeleteAfter(event);
+
+        if (engineConfig.isInMemoryDbEnabled() && engineConfig.isValkeyEventDeduplicatorEnabled()) {
+            boolean isNewEvent = postgresEventDeduplication(eventTypeId, deduplicationKey, deleteAfter);
+            boolean valkeyIsNewEvent = valkeyService.isNewEvent(eventTypeId, deduplicationKey.get(),
+                    deleteAfter, event.getId());
+            if (valkeyIsNewEvent != isNewEvent) {
+                Log.warnf(
+                        "Valkey event deduplication (isNewEvent=%s) does not align with Postgres result (isNewEvent=%s) [event_type_id=%s, event_id=%s]",
+                        valkeyIsNewEvent, isNewEvent, eventTypeId, event.getId());
```
g-duval: Same as Valkey service, eventId is null in most cases at this step. We need to log the deduplicationKey instead.

engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
```diff
@@ -262,7 +265,21 @@ public void process(Message<String> message) {
              * Before we persist the event into the DB and process it, we need to check whether the event is
              * a duplicate using the custom event deduplication logic tenants might have implemented.
              */
-            if (!eventDeduplicator.isNew(event)) {
+            boolean isNewEvent;
+
+            try {
+                isNewEvent = eventDeduplicator.isNew(event);
+            } catch (Exception e) {
+                // If Valkey deduplication is enabled, remove any keys that may have been created
```
g-duval: This should move into the global catch of this method, near `processingExceptionCounter.increment();`

g-duval (commented):
Looks great, I added some small adjustment requests.

engine/src/main/java/com/redhat/cloud/notifications/events/ValkeyService.java
jessicarod7: I wasn't thinking of security, but more privacy. [I saw those fields in SubWatch](https://github.com/RedHatInsights/notifications-backend/blob/28510d2fd5f96dc8ee0af348c72856b6b4fe954e/engine/src/main/java/com/redhat/cloud/notifications/events/deduplication/SubscriptionsDeduplicationConfig.java#L53-L59) include stuff like billing account ID, usage, and SLAs. I can make the switch if you think it's alright, or I could create a new abstract method in `EventDeduplicationConfig` to output a less sensitive version of the dedup key, and use that.

engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
jessicarod7: Deduplication key is derived from `event`, which is declared within the `try` body, so I can't use it in the global `catch`. There isn't a way to guarantee that `event` will be included in the Exception (for example, if [`postgresEventDeduplication()](https://github.com/RedHatInsights/notifications-backend/blob/9c92ccdc4a53866827dd712a049c1676f2011897/engine/src/main/java/com/redhat/cloud/notifications/events/deduplication/EventDeduplicator.java#L77) fails), so I have to use a nested try statement.

engine/src/main/java/com/redhat/cloud/notifications/events/ValkeyService.java
g-duval: Ok, don't worry, those data are not confidential from log/Sentry/database perspective. They could be safely added in the warn message.

engine/src/main/java/com/redhat/cloud/notifications/events/EventConsumer.java
g-duval: I think we could declare `Event event=null;` before the global catch, and keep `event = new Event(eventType, payload, eventWrapperToProcess, sourceEnvironmentHeader, messageId);` just before calling `eventDeduplicator.isNew(event);` and add `event != null` to the exiting `if (engineConfig.isInMemoryDbEnabled() && engineConfig.isValkeyEventDeduplicatorEnabled()) ` in the catch section.

My concern here is, without it, we will not be able to replay this message if an internal error, like simple npe happens during step 6 or 7.

engine/src/main/java/com/redhat/cloud/notifications/health/ValkeyHealthCheck.java
```diff
@@ -0,0 +1,41 @@
...
+import com.redhat.cloud.notifications.events.ValkeyService;
+import io.smallrye.mutiny.TimeoutException;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import org.eclipse.microprofile.health.HealthCheck;
+import org.eclipse.microprofile.health.HealthCheckResponse;
+import org.eclipse.microprofile.health.HealthCheckResponseBuilder;
+import org.eclipse.microprofile.health.Readiness;
+
+@Readiness
```
g-duval, line 14: Shouldn't it be also tagged as `@Liveness` to be used after the initial startup?

jessicarod7: oh right, forgot that one

engine/src/main/java/com/redhat/cloud/notifications/events/ValkeyService.java
jessicarod7: alright, will make this change

g-duval (approved):
LGTM

jessicarod7:
Only failing IQE test is `TestdrawerConnector.test_drawer_included_in_drawer_field_controls_visibility` which seems to be an unrelated error about a missing event type. Thank you @g-duval, merging!

PR #4310, opened 2026-04-08
[RHCLOUD-44983] Remove Camel from Splunk connector

g-duval:
Remove Camel from Splunk connector

## Summary by CodeRabbit

* **Refactor**
  * Migrated Splunk connector from Camel routes/processors to a handler-based architecture.

* **New Features**
  * Added Splunk message handler, REST client with automatic retries, and Splunk notification model.
  * Introduced shared JSON UTF‑8 content-type constant.

* **Bug Fixes**
  * Improved target URL normalization/validation and HTTPS mock selection in tests.

* **Tests**
  * Added extensive handler and exception tests; updated test messaging, TLS, and in-memory test config.

* **Chores**
  * Updated connector dependencies and runtime messaging/configuration.


connector-splunk/src/main/java/com/redhat/cloud/notifications/connector/splunk/SplunkMessageHandler.java
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    private String fetchAndCheckSplunkAuthorizationHeader(SplunkNotification notification) {
+        JsonObject authentication = notification.metadata.getJsonObject(AUTHENTICATION_KEY);
+
+        final Optional<AuthenticationResult> authResult;
+        try {
+            authResult = authenticationLoader.fetchAuthenticationData(notification.getOrgId(), authentication);
+        } catch (WebApplicationException | IllegalStateException e) {
+            throw new RuntimeException("Error fetching authentication data", e);
```
gwenneg, line 113: Should we log before throwing, like in [WebhookMessageHandler](https://github.com/RedHatInsights/notifications-backend/blob/6b365d23825282da9141c08a30c850cc3f1507ea/connector-webhook/src/main/java/com/redhat/cloud/notifications/connector/webhook/WebhookMessageHandler.java#L49)?

connector-splunk/src/main/java/com/redhat/cloud/notifications/connector/splunk/SplunkMessageHandler.java
```diff
@@ -0,0 +1,192 @@
...
+    static String buildSplunkPayload(SplunkNotification notification) {
+        JsonObject base = new JsonObject();
+        base.put("org_id", notification.getOrgId());
+        base.put("account_id", notification.accountId);
+
+        JsonArray events = notification.events;
+
+        if (events == null || events.isEmpty() || events.size() == 1) {
+            base.put("events", events);
+            return wrapPayload(base).encode();
```
gwenneg, line 173: In the new code, we might return an empty table of events while in the Camel implementation we didn't do that. Claude is flagging this as an inconsistency between the two implementations. Is that intentional?

g-duval: The legacy Camel implementation used to crash in such case, with an obscure message like:
`No body available of type: java.io.InputStream but has type: io.vertx.core.json.JsonObject on: Message. Caused by: No type converter available to convert from type: io.vertx.core.json.JsonObject to the required type: java.io.InputStream. Exchange[94C619...`
The new implementation give it a chance, but you are right, I need to check the Splunk plugin behavior if the events list is empty.
Few tenants don't use events at all, that's also why I dident throw an IllegalArgumentException yet. But if the Splunk plugin can't support it, I will.

gwenneg (approved):
LGTM

PR #4338, opened 2026-04-13
[RHCLOUD-46735] add severity as event log sortable field

g-duval:
[RHCLOUD-46735] add severity as event log sortable field

[RHCLOUD-46735]: https://redhat.atlassian.net/browse/RHCLOUD-46735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added `severity` as a supported sort option for event listings (ascending/descending), ensuring correct ordering across `CRITICAL` through `NONE` and `UNDEFINED`.
  * Introduced an internal maintenance action to backfill stored severity ordering to keep sorting consistent.
* **Documentation**
  * Updated API documentation to reflect `severity` as an allowed sort field for events.
* **Tests**
  * Expanded automated coverage to validate severity ordering behavior, including null/undefined scenarios.


database/src/main/resources/db/migration/V1.130.0__RHCLOUD-46735_add_severity_order_column_to_event.sql
```diff
@@ -0,0 +1,27 @@
+-- Step 1: Add nullable column (instant, no table rewrite)
+ALTER TABLE event ADD COLUMN severity_order SMALLINT;
```
gwenneg: This script needs to be pushed to prod before the new field is added to the JPA entity.

gwenneg (changes requested):

database/src/main/resources/db/migration/V1.130.0__RHCLOUD-46735_add_severity_order_column_to_event.sql
g-duval: That's right, added in https://github.com/RedHatInsights/notifications-backend/pull/4455

common/src/main/java/com/redhat/cloud/notifications/models/Event.java
```diff
@@ -34,18 +34,29 @@
 @Table(name = "event")
 public class Event {
 
+    public static final String SEVERITY_SORT_EXPRESSION = "CASE e.severity"
```
gwenneg: 🤖 **Claude (automated review):** This CASE expression computes severity order live per row and never references the `severity_order` column, so it can't use the `ix_event_org_id_severity_order_created` index or the data this PR's backfill endpoint populates. Consider sorting by a column reference (e.g. `e.severityOrder`) instead of a live CASE, so the index and backfill actually pay off.

common/src/main/java/com/redhat/cloud/notifications/models/Event.java
```diff
@@ -34,18 +34,29 @@
...
+            "severity", SEVERITY_SORT_EXPRESSION
     );
 
     private static final Map<String, String> SORT_FIELDS_DENORMALIZED = Map.of(
             "bundle", "e.bundleDisplayName",
             "application", "e.applicationDisplayName",
             "event", "e.eventTypeDisplayName",
-            "created", "e.created"
+            "created", "e.created",
+            "severity", SEVERITY_SORT_EXPRESSION
```
gwenneg, line 53: 🤖 **Claude (automated review):** This denormalized map is also used by `DrawerNotificationRepository.getNotifications()` (`Sort.getSort(query, "created:DESC", Event.getSortFields(true))`), so drawer notifications silently gain `sort_by=severity` too. `DrawerResource`'s docs still only list `bundle, application, event, created`, and no drawer test covers this. Either document it there or exclude `severity` from this map if it's not meant to apply to the drawer endpoint.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EventRepository.java
```diff
@@ -433,4 +433,21 @@ public void cleanupInventoryEvents(int limit) {
...
+
+    @Transactional
+    public int backfillSeverityOrder(int limit) {
+        String backfillQuery = "UPDATE event SET severity_order = CASE severity" +
+            " WHEN 'CRITICAL' THEN 10" +
+            " WHEN 'IMPORTANT' THEN 20" +
+            " WHEN 'MODERATE' THEN 30" +
+            " WHEN 'LOW' THEN 40" +
+            " WHEN 'NONE' THEN 50" +
+            " WHEN 'UNDEFINED' THEN 60" +
```
gwenneg: 🤖 **Claude (automated review):** This `ELSE 70` (mirroring the DB trigger) buckets legacy pre-severity events (`severity IS NULL`) together with any unrecognized value, sorting them differently from explicit `UNDEFINED` (60). Consider `CASE COALESCE(severity, 'UNDEFINED') WHEN ... END` so legacy NULL rows sort the same as `UNDEFINED`.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EventRepository.java
```diff
@@ -433,4 +433,21 @@ public void cleanupInventoryEvents(int limit) {
             .setParameter("limit", limit)
             .executeUpdate();
     }
+
+    @Transactional
+    public int backfillSeverityOrder(int limit) {
+        String backfillQuery = "UPDATE event SET severity_order = CASE severity" +
```
gwenneg: 🤖 **Claude (automated review):** This severity→order mapping now exists in three places with two different scales: this query and the DB trigger `set_severity_order()` use 10-70, while `Event.SEVERITY_SORT_EXPRESSION` uses 0-6. Consider centralizing the mapping so a future `Severity` value can't silently desync them.

gwenneg (commented):
🤖 **Automated review by Claude** — the comments below were generated by an AI code review assistant (Claude), not a human reviewer. Please treat them as suggestions to verify, not authoritative blockers.

common/src/main/java/com/redhat/cloud/notifications/models/Event.java
```diff
@@ -101,6 +112,9 @@ public static Map<String, String> getSortFields(boolean useNormalized) {
     @Column(length = 20)
     private Severity severity;
 
+    @Column(name = "severity_order", insertable = false, updatable = false)
+    private Short severityOrder;
```
gwenneg, line 110: 🤖 **Claude (automated review):** `severityOrder` has no getter/setter and is never read in Java code — only written by the DB trigger and the native backfill query. Since the live sort above doesn't use it either, consider whether the entity needs this field, or wire it into the sort query it was built for.

gwenneg (approved):
LGTM

PR #4356, opened 2026-04-16
fix(notifications-mcp): fix clowdapp.yaml to expose service port

florkbr:
We are unable to communicate with the new notifications MCP service from our hcc-ai-assistant pod. This update should resolve the issues.

## Release Notes

* **Configuration**
  * Enabled public web service access for the notifications service with configured API path routing.

### DIFF
--- a/.rhcicd/clowdapp-mcp.yaml
+++ b/.rhcicd/clowdapp-mcp.yaml
@@ -20,7 +20,8 @@ objects:
       minReplicas: ${{MIN_REPLICAS}}
       webServices:
         public:
-          enabled: false
+          enabled: true
+          apiPath: notifications-mcp
       podSpec:
         image: ${IMAGE}:${IMAGE_TAG}
         resources:


.rhcicd/clowdapp-mcp.yaml
```diff
@@ -20,7 +20,8 @@ objects:
       minReplicas: ${{MIN_REPLICAS}}
       webServices:
         public:
-          enabled: false
+          enabled: true
```
florkbr, line 23: Note - even though this is set to "public" changes would still need to be made to 3-scale to expose this service outside the cluster. So AFAIK this is still inter-cluster coms.

Hyperkid123: @florkbr there should also be a private entry for the endpoints in the clowder config. Care to see if the host is in there? We usually only ever needed to read from the public entries.

https://github.com/RedHatInsights/app-common-js/blob/master/index.ts#L100

Here is TS type. I think it's worth checking.

florkbr: @Hyperkid123 @gwenneg 

Yes, private service is an option, however we still need to explicitly set `private.enabled: true`. There's no automatic fallback AFAIK?

But note that `private` uses port `10000` (Clowder's default), not `8000`. So either:                                                                                                                                                                                                         
  - Their Quarkus app needs to listen on 10000, or
  - They just use public.enabled: true to keep port 8000   

  The Clowder code has explicit gates for both:

  - Public endpoint: only created if `webServices.public.enabled: true`
  - Private endpoint: only created if `webServices.private.enabled: true`
  
  Omitting or setting both to false means no Service ports, no endpoints in the ACG config — which is exactly the state notifications-mcp is in right now.
                                                                                                                                                                                                                                                                                          
  So either way we'll need to enable at least one.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
                                                        
On our side we'd need notifications-mcp in our optionalDependencies so the endpoint appears in the ACG config.

g-duval (approved):
LGTM

PR #4363, opened 2026-04-17
[RHCLOUD-46948] Update user preferences Instant label to monthly for Planning team

g-duval:
## Summary by CodeRabbit

* **New Features**
  * Subscription UI now shows dynamic labels, descriptions, and opt-in warnings tailored to each event type.
  * RHEL lifecycle retiring events and RHEL roadmap monthly reports display "Monthly notification" with appropriate frequency guidance; other events show "Instant notification" and an opt-in warning when applicable.

* **Tests**
  * Added tests validating mappings from event types to notification labels, descriptions, and warnings.

### DIFF
--- a/common-template/src/main/java/com/redhat/cloud/notifications/qute/templates/mapping/InstantLabelMapping.java
+++ b/common-template/src/main/java/com/redhat/cloud/notifications/qute/templates/mapping/InstantLabelMapping.java
@@ -17,14 +17,10 @@ public class InstantLabelMapping {
     public static InstantTemplateDetails buildInstantNotificationDescription(final String bundleName, final String applicationName, final String eventTypeName) {
         InstantTemplateDetails instantLabels = INSTANT_NOTIFICATION;
         if (Rhel.BUNDLE_NAME.equals(bundleName)) {
-            if (Rhel.LIFECYCLE_APP_NAME.equals(applicationName)) {
-                if (Rhel.RETIRING_LIFECYCLE.equals(eventTypeName)) {
-                    instantLabels = MONTHLY_NOTIFICATION;
-                }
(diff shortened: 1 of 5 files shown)


common-template/src/main/java/com/redhat/cloud/notifications/qute/templates/mapping/InstantLabelMapping.java
```diff
@@ -0,0 +1,33 @@
...
+
+    static final InstantTemplateDetails INSTANT_NOTIFICATION = new InstantTemplateDetails(
+        "Instant notification",
+        "Immediate email for each triggered application event.",
+        "Opting into this notification may result in a large number of emails");
+
+    public static InstantTemplateDetails buildInstantNotificationDescription(final String bundleName, final String applicationName, final String eventTypeName) {
+        InstantTemplateDetails instantLabels = INSTANT_NOTIFICATION;
+        if (Rhel.BUNDLE_NAME.equals(bundleName)) {
+            if (Rhel.LIFECYCLE_APP_NAME.equals(applicationName)) {
```
bonscji1: Just out of interest, is there a reason why we are not using if (A && B) here instead of If(A){ if(B) }?

bonscji1 (approved):
Looks good to me.

common-template/src/main/java/com/redhat/cloud/notifications/qute/templates/mapping/InstantLabelMapping.java
g-duval: It was to forcast if another instant Event type will be create in the future for this same app with the same  naming update (from instant to monthly).
You are right, I should keep it simple for now.
I'll update this.

PR #4386, opened 2026-04-21
RHCLOUD-44774 improve drawer api response - logic changes

bonscji1:
Logic changes to switch to on demand DrawerNotification logic with read tracking. 

## Summary by CodeRabbit

* **New Features**
  * Drawer read state now persists in a dedicated store; bulk mark-read/unread supported.
  * Drawer entries honor subscribe-by-default and authorization-based visibility exclusions.

* **Improvements**
  * Listing/counting now derive read state from the new store, improving consistency.
  * Enhanced pagination, filtering (bundle/app/event type/date/read) and default sort vocabulary.
  * Stronger org isolation and feature-flag behavior.


bonscji1:
@coderabbitai review

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
```diff
@@ -27,149 +36,269 @@ public class DrawerNotificationRepository {
...
+            return entityManager.createNativeQuery(deleteSql)
+                .setParameter("orgId", orgId)
+                .setParameter("userId", username)
+                .setParameter("notificationIds", notificationIds.toArray(new UUID[0]))
+                .executeUpdate();
+        }
     }
 
-    public List<DrawerEntryPayload> getNotifications(String orgId, String username, Set<UUID> bundleIds, Set<UUID> appIds, Set<UUID> eventTypeIds,
+    public List<DrawerEntryPayload> getNotifications(SecurityContext securityContext, String orgId, String username, Set<UUID> bundleIds, Set<UUID> appIds, Set<UUID> eventTypeIds,
```
g-duval: `SecurityContext securityContext` should stay at the resources level to split application layers and responsabilities.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
```diff
@@ -27,149 +36,269 @@ public class DrawerNotificationRepository {
...
                                               LocalDateTime startDate, LocalDateTime endDate, Boolean readStatus, Query query) {
 
         boolean useNormalized = backendConfig.isNormalizedQueriesEnabled(orgId);
-        Optional<Sort> sort = Sort.getSort(query, "created:DESC", DrawerNotification.getSortFields(useNormalized));
+        Set<UUID> subscribedEventTypes = GetSubscribedEventTypes(orgId, username);
```
g-duval: I think we only need to check if `renderedDrawerNotification` is not null.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
```diff
@@ -27,149 +36,269 @@ public class DrawerNotificationRepository {
...
+
+        if (subscribedEventTypes.isEmpty()) {
+            return new ArrayList<>();
+        }
+
+        LocalDate startLocalDate = startDate != null ? startDate.toLocalDate() : null;
+        LocalDate endLocalDate = endDate != null ? endDate.toLocalDate() : null;
+
+        List<UUID> uuidToExclude = getExcludedEventIds(securityContext, useNormalized, orgId, subscribedEventTypes, startLocalDate, endLocalDate);
```
g-duval: Like above, it should be done at the resource level, reuse the same results in `count` and `getNotifications` methods.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
```diff
@@ -27,149 +36,269 @@ public class DrawerNotificationRepository {
...
+            } else {
+                return "SELECT COUNT(e.id) FROM Event e " +
+                    "JOIN Bundle bundle ON e.bundleId = bundle.id " +
+                    "LEFT JOIN DrawerReadStatus drs ON drs.id.eventId = e.id AND drs.id.orgId = :orgId AND drs.id.userId = :userid " +
+                    "WHERE e.orgId = :orgId AND e.eventType.includedInDrawer = true AND e.eventType.id IN (:subscribedEventTypes)";
+            }
+        } else {
+            if (useNormalized) {
+                return "SELECT e.id, " +
+                    "CASE WHEN drs.readAt IS NOT NULL THEN true ELSE false END, " +
```
g-duval: We can just check if `drs.eventId` exists.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
```diff
@@ -27,149 +36,269 @@ public class DrawerNotificationRepository {
...
+                    "bundle.displayName, app.displayName, et.displayName, e.created, e.renderedDrawerNotification, bundle.name, e.severity " +
+                    "FROM Event e " +
+                    "JOIN e.eventType et " +
+                    "JOIN et.application app " +
+                    "JOIN app.bundle bundle " +
+                    "LEFT JOIN DrawerReadStatus drs ON drs.id.eventId = e.id AND drs.id.orgId = :orgId AND drs.id.userId = :userid " +
```
g-duval: en event is always assighed to a single org, so you can skip `AND drs.id.orgId = :orgId` condition.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
```diff
@@ -27,149 +36,269 @@ public class DrawerNotificationRepository {
...
+                    "WHERE e.orgId = :orgId AND et.includedInDrawer = true AND et.id IN (:subscribedEventTypes)";
```
g-duval: You can replace `et.includedInDrawer` by `et.renderedDrawerNotification is not null` by

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
```diff
@@ -27,149 +36,269 @@ public class DrawerNotificationRepository {
...
+            } else {
+                return "SELECT e.id, " +
```
g-duval: Because the drawer is not in prod yet, we can skip/remove the not normalized case.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
```diff
@@ -27,149 +36,269 @@ public class DrawerNotificationRepository {
...
                       LocalDateTime startDate, LocalDateTime endDate, Boolean readStatus) {
+
         boolean useNormalized = backendConfig.isNormalizedQueriesEnabled(orgId);
+        Set<UUID> subscribedEventTypes = GetSubscribedEventTypes(orgId, username);
+
+        if (subscribedEventTypes.isEmpty()) {
+            return 0L;
+        }
+
+        LocalDate startLocalDate = startDate != null ? startDate.toLocalDate() : null;
```
g-duval: Why converting it in `LocalDate`? `addHqlConditions(...)` method will cast those back to LocalDateTime, did I missed something?

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EventRepository.java
```diff
@@ -76,6 +76,89 @@ public List<EventAuthorizationCriterion> getEventsWithCriterion(String orgId, bo
...
+            typedQuery.setParameter("startDate", Timestamp.valueOf(startDate.atStartOfDay()));
+        }
+        if (endDate != null) {
+            typedQuery.setParameter("endDate", Timestamp.valueOf(endDate.atTime(LocalTime.MAX)));
+        }
+
+        List<Event> eventsWithAuthorizationCriterion = typedQuery.getResultList();
+
+        // Populate denormalized display name fields from joined entities
+        if (useNormalized && !eventsWithAuthorizationCriterion.isEmpty()) {
```
g-duval: Since we will use only UUIDs, I don't think we need to update bundle/application/eventType display names

backend/src/test/java/com/redhat/cloud/notifications/db/DbCleaner.java
```diff
@@ -68,6 +68,10 @@ public void clean() {
         for (Class<?> entity : ENTITIES) {
             entityManager.createQuery("DELETE FROM " + entity.getSimpleName()).executeUpdate();
         }
+        // Delete non-DEFAULT IntegrationTemplates (test-created ones)
+        entityManager.createQuery("DELETE FROM IntegrationTemplate it WHERE it.templateKind IS NULL OR it.templateKind != :default")
```
g-duval, line 72: fyi IntegrationTemplate are not used anymore in prod for at least 8 months 😅
I've to create a tech dept card to et rid of it.

database/src/main/resources/db/migration/V1.131.0__RHCLOUD-44774_create_drawer_read_status.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE IF NOT EXISTS drawer_read_status (
+    org_id TEXT NOT NULL,
+    user_id TEXT NOT NULL,
+    event_id UUID NOT NULL,
+    read_at TIMESTAMP NOT NULL DEFAULT NOW(),
```
g-duval: We just need to know it the user read a drawer notif or not, we don't have to know when.

database/src/main/resources/db/migration/V1.132.0__RHCLOUD-44774_drop_drawer_notification_table.sql
```diff
@@ -0,0 +1,6 @@
+DROP PROCEDURE IF EXISTS insert_drawer_notifications(text, text, uuid, timestamp);
```
g-duval: While `V1.131.0__RHCLOUD-44774_create_drawer_read_status.sql`could be icluded in this PR, `V1.132.0__RHCLOUD-44774_drop_drawer_notification_table.sql‎` can't because it affects table and procedure still in use until pods rollout will be completed.

g-duval (commented):
Great work, I think we could simplify it a bit 🤔

database/src/main/resources/db/migration/V1.132.0__RHCLOUD-44774_drop_drawer_notification_table.sql
g-duval: Second thought: since it is not yet use in Prod, we can keep it in the same PR.

bonscji1: V1.131.0 has its own PR, since it created new table that logic uses, so if logic was up before pods, we could get errors. i will then also have to move V1.132.0 into its own PR, to cleanup db after the logic changes has been implemented.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
bonscji1: can i actually do that? renderedDrawerNotification means that something can be rendered, includedInDrawer is flag that show that something should be rendered. do they truly mean the same thing in our program? if we want to hide eventType due to testing, we will be setting includedInDrawer to false no?

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
bonscji1: Likely my and claude mistake, since everything in EventRepository uses LocalDate and i validate methods against that, it sliped into my method and this is on of the places where i did not catch it and fix it. thank you for noticing it. This conversion vas only passed into getExcludedEventIds, everything else used localDatetime.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EventRepository.java
bonscji1: Good catch

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
bonscji1: I think i will need explanation, since i do not understand.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EventRepository.java
```diff
@@ -76,6 +77,80 @@ public List<EventAuthorizationCriterion> getEventsWithCriterion(String orgId, bo
...
+        if (eventTypeIdsNotEmpty) {
+            typedQuery.setParameter("eventTypeIds", eventTypeIds);
+        }
+
+        if (startDate != null) {
+            typedQuery.setParameter("startDate", Timestamp.valueOf(startDate.toLocalDate().atStartOfDay()));
+        }
+        if (endDate != null) {
+            typedQuery.setParameter("endDate", Timestamp.valueOf(endDate.toLocalDate().atTime(LocalTime.MAX)));
+        }
```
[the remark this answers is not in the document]
bonscji1: I believe this widening comes from and applies to checking against resource-based authorization and is also used for normal events.

database/src/main/resources/db/migration/V1.132.0__RHCLOUD-44774_drop_drawer_notification_table.sql
g-duval: The case you described with 131 and logic updates only applies when part of the logic updates are located in another pod than the one in charge of appliing sql updates.
Because sql logic update to use the new structure are only located into notifications-backend pod, which is also in charge of SQL deployment, we can push it in the same PR.

I think you are right for 132, even if drawer's tables and stored procedure are not used, Hibernate may complains if a table is missing even after the initial sql validation check. Because I'm not 100% sure of Hibernate behavior in this case, I think we should move 132 in another PR as you mentioned.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
g-duval: I'm not sure I put this comment in the right place 😅.
My point is, because we already check the `includedInDrawer` flag to fill the `renderedDrawerNotification` or not, we don't need to check it one more time to retreive drawer notifications from the api.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
g-duval: Currently, the renderedDrawerNotification field is fill only when the includeInDrawer is true in [DrawerProcessor](https://github.com/RedHatInsights/notifications-backend/blob/master/engine/src/main/java/com/redhat/cloud/notifications/processors/drawer/DrawerProcessor.java#L87), we are doing twice the same check.

database/src/main/resources/db/migration/V1.132.0__RHCLOUD-44774_drop_drawer_notification_table.sql
bonscji1: I will move 132 to another PR.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
bonscji1: hmmm, does this structure then create lag? we populate renderedDrawerNotification in event when includedInDrawer for eventype is true, if we then change includedInDrawer to false, any new events wil not be show in the drawer, but old ones will since they would have renderedDrawerNotification. in the same way, if we change include in drawer from false to true, only new  events will be shown since they have renderedDrawerNotification but old ones will not be?

bonscji1: ok, this was explained to me, i will fix the code.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
```diff
@@ -25,151 +26,194 @@ public class DrawerNotificationRepository {
...
+            return " " + sort.getSortQuery();
+        }
     }
 
-    private static String addHqlConditions(String hql, boolean useNormalized,
+    private static String addHqlConditions(String hql,
                                            boolean bundlesNotEmpty, boolean applicationsNotEmpty, boolean eventTypesNotEmpty,
-                                           LocalDateTime startDate, LocalDateTime endDate, Boolean readStatus) {
+                                           boolean excludeNotEmpty, LocalDateTime startDate, LocalDateTime endDate, Boolean readStatus) {
+        // Exclusion list from RBAC check
```
g-duval:
```suggestion
        // Exclusion list from Kessel check
```

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/DrawerNotificationRepository.java
```diff
@@ -27,149 +36,269 @@ public class DrawerNotificationRepository {
...
+                hql += " AND e.applicationId IN (:appIds)";
             }
         } else if (bundlesNotEmpty) {
             if (useNormalized) {
                 hql += " AND bundle.id IN (:bundleIds)";
             } else {
-                hql += " AND dn.event.bundleId IN (:bundleIds)";
+                hql += " AND e.bundleId IN (:bundleIds)";
             }
         }
```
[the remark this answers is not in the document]
g-duval: No it can't since eventType have only one application, and an application have only one bundle.

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EventTypeRepository.java
```diff
@@ -53,4 +53,17 @@ public List<EventType> findByIds(final Set<UUID> eventTypeIds) {
...
             .getResultList();
     }
+
+    /**
+     * Get all event type IDs where includedInDrawer = true.
+     * Used for drawer subscribe-by-default logic.
+     *
+     * @return Set of event type UUIDs that are included in drawer
+     */
+    public Set<UUID> getEventTypeIdsByIncludedInDrawer() {
```
g-duval, line 63: Is it still used?

g-duval (commented):
Almost great, just two tiny last comments

backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EventTypeRepository.java
bonscji1: Its used in getSubscribedEventTypes in DrawerNotificationRepository. Its the initial list of ids which is then filtered with unsubscribed types for each user.

g-duval (approved):
LGTM

PR #4420, opened 2026-04-28
[RHCLOUD-46778] Add Application services user preferences migration service

g-duval:
We need to migrate Application services user preferences from legacy tool to Notifications.
Application services (aka Jboss csp) will probide us an export file with about 500 users. We need to load, parse and create requested user preferences from it.

## Summary by CodeRabbit

* **New Features**
  * Added internal admin endpoint to import user subscription preferences from JSON files
  * Supports bulk import of notification subscriptions with automatic validation and conflict handling

* **Tests**
  * Added test coverage for import operations and subscription persistence
  * Includes tests for handling unsupported event types during import

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/internal/userpreferencesmigration/ApplicationServicesSubscriptionInput.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/internal/userpreferencesmigration/ApplicationServicesSubscriptionInput.java
@@ -1,0 +1,10 @@
+package com.redhat.cloud.notifications.routers.internal.userpreferencesmigration;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public record ApplicationServicesSubscriptionInput(
+    String username,
+    @JsonProperty("org-id") String orgId,
+    @JsonProperty("notification_category") String eventType
+) {
+}
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/internal/userpreferencesmigration/ApplicationServicesUserPreferencesMigrationResource.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/internal/userpreferencesmigration/ApplicationServicesUserPreferencesMigrationResource.java
@@ -1,0 +1,87 @@
+package com.redhat.cloud.notifications.routers.internal.userpreferencesmigration;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.redhat.cloud.notifications.Constants;
+import com.redhat.cloud.notifications.auth.ConsoleIdentityProvider;
+import com.redhat.cloud.notifications.db.repositories.ApplicationServicesMigrationRepository;
+import com.redhat.cloud.notifications.models.EventType;
+import io.quarkus.logging.Log;
+import jakarta.annotation.security.RolesAllowed;
+import jakarta.inject.Inject;
+import jakarta.transaction.Transactional;
+import jakarta.validation.constraints.NotNull;
+import jakarta.ws.rs.BadRequestException;
+import jakarta.ws.rs.Consumes;
+import jakarta.ws.rs.POST;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.MediaType;
+import org.jboss.resteasy.reactive.RestForm;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+@Path(Constants.API_INTERNAL + "/application-services")
+@RolesAllowed(ConsoleIdentityProvider.RBAC_INTERNAL_ADMIN)
+public class ApplicationServicesUserPreferencesMigrationResource {
+
+    @Inject
+    ApplicationServicesMigrationRepository applicationServicesMigrationRepository;
+
+    @Inject
+    ObjectMapper objectMapper;
+
+    @Consumes(MediaType.MULTIPART_FORM_DATA)
+    @Path("/migrate/json")
+    @POST
+    @Produces(MediaType.TEXT_PLAIN)
+    @Transactional
+    public void migrateApplicationServicesUserPreferencesJSON(@NotNull @RestForm("jsonFile") InputStream jsonFile) {
+        Log.info("Start migrateApplicationServicesUserPreferencesJSON");
+
+        final List<EventType> applicationServicesEventTypes = applicationServicesMigrationRepository.findApplicationServicesEventTypes();
+        final Map<String, EventType> mappedEventTypes = applicationServicesEventTypes
+            .stream()
+            .collect(
+                Collectors.toMap(
+                    EventType::getName,
+                    Function.identity()
+                )
+            );
+
+        List<ApplicationServicesSubscriptionInput> inputs;
+        try {
+            inputs = objectMapper.readValue(
+                jsonFile.readAllBytes(),
+                objectMapper.getTypeFactory().constructCollectionType(List.class, ApplicationServicesSubscriptionInput.class)
+            );
+        } catch (final IOException e) {
+            Log.error("Failed to parse JSON migration file", e);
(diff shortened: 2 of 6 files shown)


backend/src/main/java/com/redhat/cloud/notifications/routers/internal/userpreferencesmigration/ApplicationServicesUserPreferencesMigrationResource.java
```diff
@@ -0,0 +1,87 @@
...
+        Map<String, Long> skippedByEventType = new HashMap<>();
+        for (final ApplicationServicesSubscriptionInput input : inputs) {
+            final EventType eventType = mappedEventTypes.get(input.eventType());
+            if (eventType == null) {
+                Log.warnf("Unknown event type \"%s\" for user \"%s\" (org %s), skipping", input.eventType(), input.username(), input.orgId());
+                skippedByEventType.merge(input.eventType(), 1L, Long::sum);
+                continue;
+            }
+
+            applicationServicesMigrationRepository.saveApplicationServicesSubscription(input.username(), input.orgId(), eventType);
```
bonscji1, line 78: if i understand right, you insert into db one record at a time. depending on number of record batch insert may be better.

bonscji1, line 78: What happens if one insert fails, does the error cause the whole process to exit? depending on how we want the code to handle, we may need to use try and catch here.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/userpreferencesmigration/ApplicationServicesUserPreferencesMigrationResource.java
```diff
@@ -0,0 +1,87 @@
...
+            .collect(
+                Collectors.toMap(
+                    EventType::getName,
+                    Function.identity()
+                )
+            );
+
+        List<ApplicationServicesSubscriptionInput> inputs;
+        try {
+            inputs = objectMapper.readValue(
```
bonscji1, line 59: Is there a reason why we want to parse the json file first to then iterate over it with for-each instead of parsing item by and immediately processing them? something like:

try {                                                 
          // Create streaming parser                                                                                                                                                                                                                                                                                
          JsonParser parser = objectMapper.getFactory().createParser(jsonFile);                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                    
          // Expect start of array: [                                                                                                                                                                                                                                                                               
          if (parser.nextToken() != JsonToken.START_ARRAY) {                                                                                                                                                                                                                                                        
              throw new BadRequestException("Expected JSON array");                                                                                                                                                                                                                                                 
          }
                                                                                                                                                                                                                                                                                                                    
          // Read and process one object at a time                                                                                                                                                                                                                                                                  
          while (parser.nextToken() != JsonToken.END_ARRAY) {
              // Deserialize single object from current position                                                                                                                                                                                                                                                    
              ApplicationServicesSubscriptionInput input = parser.readValueAs(ApplicationServicesSubscriptionInput.class);                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                    
              // Process immediately                                                                                                                                                                                                                                                                                
              final EventType eventType = mappedEventTypes.get(input.eventType());                                                                                                                                                                                                                                  
              if (eventType == null) {                                                                                                                                                                                                                                                                              
                  Log.warnf("Unknown event type \"%s\" for user \"%s\" (org %s), skipping",
                      input.eventType(), input.username(), input.orgId());                                                                                                                                                                                                                                          
                  skippedByEventType.merge(input.eventType(), 1L, Long::sum);                                                                                                                                                                                                                                       
                  continue;                                                                                                                                                                                                                                                                                         
              }                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                    
              applicationServicesMigrationRepository.saveApplicationServicesSubscription(                                                                                                                                                                                                                           
                  input.username(), input.orgId(), eventType
              );                                                                                                                                                                                                                                                                                                    
              totalInsertedElements++;                      
          }                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                    
      } catch (JsonProcessingException e) {                                                                                                                                                                                                                                                                         
          Log.error("Failed to parse JSON migration file", e);                                                                                                                                                                                                                                                      
          throw new BadRequestException("Invalid JSON format in migration file: " + e.getMessage());
      } catch (IOException e) {                                                                                                                                                                                                                                                                                     
          Log.error("IO error reading migration file", e);
          throw new InternalServerErrorException("Failed to read migration file");                                                                                                                                                                                                                                  
      }

backend/src/test/java/com/redhat/cloud/notifications/routers/internal/applicationservices/ApplicationServicesUserPreferencesMigrationResourceTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+    @Inject
+    ResourceHelpers resourceHelpers;
+
+    @ConfigProperty(name = "internal.admin-role")
+    String adminRole;
+
+    @Inject
+    SubscriptionRepository subscriptionRepository;
+
```
bonscji1, line 50: We could also include malformed JSON test to validate BadRequestException "Invalid JSON format in migration file"

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/userpreferencesmigration/ApplicationServicesSubscriptionInput.java
```diff
@@ -0,0 +1,10 @@
+package com.redhat.cloud.notifications.routers.internal.userpreferencesmigration;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public record ApplicationServicesSubscriptionInput(
```
bonscji1, line 5: Just a question, do we trust the data provided for migration? parser and ApplicationServicesSubscriptionInput would allow nulls but db will then throw PSQLException.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/userpreferencesmigration/ApplicationServicesUserPreferencesMigrationResource.java
g-duval: Yes, the file will not be large enought to generate memory issues. 
We had this kind of approch with `ErrataUserPreferencesMigrationResource` but it brings complexity for no value.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/userpreferencesmigration/ApplicationServicesUserPreferencesMigrationResource.java
g-duval: Exact, but we have too few records to justify bach approach complexity.

g-duval: Yes in case of failure, it will rollback the entire transaction.

backend/src/test/java/com/redhat/cloud/notifications/routers/internal/applicationservices/ApplicationServicesUserPreferencesMigrationResourceTest.java
g-duval: We could but since it is a one time job and the file will be validated locally first, I don't think we will need it.

backend/src/main/java/com/redhat/cloud/notifications/routers/internal/userpreferencesmigration/ApplicationServicesSubscriptionInput.java
g-duval: Yes, it will be validated locally first, then the entire transaction will be rollbacked.

bonscji1 (approved):
LGTM

PR #4430, opened 2026-04-30
 [RHCOUD-44733] Include event type drawer visibility check in user preferences API

g-duval:
## Summary by CodeRabbit

* **Bug Fixes**
  * Drawer notification option now appears only for event types explicitly marked eligible, even when drawer support is enabled for the account. Instant notifications remain available. This prevents showing an unavailable drawer option for ineligible event types and ensures subscription choices reflect actual capabilities.

### DIFF
--- a/backend/src/main/java/com/redhat/cloud/notifications/routers/handlers/userconfig/UserConfigResource.java
+++ b/backend/src/main/java/com/redhat/cloud/notifications/routers/handlers/userconfig/UserConfigResource.java
@@ -304,6 +304,8 @@ public class UserConfigResource {
         if (backendConfig.isUseCommonTemplateModuleForUserPrefApisToggle()) {
             if (!backendConfig.isDrawerEnabled(orgId) && subscriptionType == DRAWER) {
                 supported = false;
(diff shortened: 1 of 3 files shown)


backend/src/main/java/com/redhat/cloud/notifications/routers/handlers/userconfig/UserConfigResource.java
```diff
@@ -304,6 +304,8 @@ private boolean isTemplateSupported(String bundleName, String applicationName, E
         if (backendConfig.isUseCommonTemplateModuleForUserPrefApisToggle()) {
             if (!backendConfig.isDrawerEnabled(orgId) && subscriptionType == DRAWER) {
                 supported = false;
+            } else if (subscriptionType == DRAWER && !eventType.isIncludedInDrawer()) {
+                return false;
```
bonscji1: this is not consistent with the rest of the class, should probably be : supported = false;

g-duval: that's right, updated

bonscji1 (approved):
LGTM

PR #4461, opened 2026-05-06
[RHCLOUD-47531] Switch maven central repo for Konflux builds

g-duval:
## Summary by CodeRabbit

* **Chores**
  * Standardized Maven configuration to use a Google-hosted mirror for dependency downloads and updated the Maven wrapper to point to the same mirror, streamlining builds and improving dependency resolution across build environments.

### DIFF
--- a/.mvn/settings.xml
+++ b/.mvn/settings.xml
@@ -1,0 +1,10 @@
+<settings>
+  <mirrors>
+    <mirror>
+      <id>google-maven-central</id>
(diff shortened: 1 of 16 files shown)


gwenneg (commented):
What happens if the Google mirror goes down? Should we add additional settings to use Maven Central as a backup solution?

gwenneg (approved):
LGTM

PR #4524, opened 2026-05-20
Update let block conditions for lifecycle notification

pr-watson:
### Changes
- **Template logic**: Sections now display based on payload data presence rather than `systems_count` values
  - RHEL section: Shows when `rhel_retired` exists in payload
  - RHEL 8/9/10 appstream sections: Show when respective version data exists in `appstream_retired` payload (using `rhel{X}Data.count.or(-1) >= 0` check)
- **Bug fix**: Fixed missing `/` in RHEL near-retirement URL on line 35
- **Tests**: Updated 6 tests to verify payload-based visibility behavior

### Behavior
With this change, sections appear when payload data exists, even if all counts are 0. Sections are hidden only when the corresponding payload data is completely absent.

**Example**: Payload with only `rhel_retired` and `rhel8` → displays only RHEL and RHEL 8 sections (no RHEL 9 or 10)

## Summary by CodeRabbit

* **Bug Fixes**
  * Lifecycle retirement emails now render the RHEL life cycle section whenever retirement counts are supplied (even if overall system count is zero). App-stream subsections render only when app-stream data is present; a clear message appears when none is detected.
* **Tests**
  * Email template tests updated to check section visibility and presence of per-RHEL count elements instead of exact numeric values.


g-duval (approved):
LGTM

hosekadam:
https://redhat.atlassian.net/browse/RSPEED-3091

PR #4539, opened 2026-05-21
[RHCLOUD-44976] Remove Camel dependency from email connector

g-duval:
## Summary

Migrates the email connector off Apache Camel to the \`connector-common-v2\` framework, and includes a few improvements to the shared v2 framework that unblocked the migration.

### Email connector — Camel removal

Removed Camel-specific classes:
- \`EmailRouteBuilder\` — replaced by \`EmailMessageHandler\` (extends \`MessageHandler\` from v2)
- \`EmailManagementProcessor\` — logic folded into \`EmailMessageHandler\`
- \`EmailCloudEventDataExtractor\` — no longer needed; v2 framework handles cloud event extraction
- \`ExchangeProperty\` — Camel \`Exchange\` property bag replaced by typed model classes

Added/updated:
- \`EmailMessageHandler\`: single entry point for processing a notification; handles recipients resolution, template rendering, and BOP submission; includes Micrometer timers for BOP and recipients resolver call durations
- \`HighVolumeMessageConsumer\`: dedicated \`@Incoming("highvolumemessages")\` consumer on its own thread pool; delegates to \`MessageConsumer.handleMessage()\` after an enabled-flag guard
- \`EmailConnectorConfig\`: switched from v1 \`HttpConnectorConfig\` to the v2 one; removed the now-redundant manual Kafka high-volume poll properties (\`max-poll-interval-ms\`, \`max-poll-records\`, \`poll-on-error\`) since they are configured directly via SmallRye \`mp.messaging.incoming.highvolumemessages.*\` properties

### connector-common-v2 — \`MessageConsumer\` refactoring

- Moved the connector-header filtering (early ack) **before** cloud event metadata extraction — avoids unnecessary work and aligns with the logical message lifecycle
- Extracted \`handleMessage()\` as a package-private method so \`HighVolumeMessageConsumer\` can delegate to it without going through the \`@Incoming\`-annotated entry point

### connector-common-http-v2 — HTTP exception handling improvements

- Changed \`HttpExceptionHandler.process()\` return type from the generic \`HandledExceptionDetails\` to the more specific \`HandledHttpExceptionDetails\`, enabling subclasses to access HTTP-specific fields directly
- Added response body capture (truncated to 1 000 characters) when a \`ClientWebApplicationException\` is caught; stored in \`HandledHttpExceptionDetails.responseBody\` and surfaced in the cloud event history for easier debugging
- Fixed response resource leak by using try-with-resources on the JAX-RS \`Response\`

### Drawer connector — type alignment

Updated \`DrawerExceptionHandler\` and \`HandledDrawerExceptionDetails\` to use \`HandledHttpExceptionDetails\` instead of the now-less-specific \`HandledExceptionDetails\`, following the updated base class hierarchy.

## Test plan
- [x] \`EmailMessageHandlerTest\` — new unit tests covering message handling, recipient batching, and template rendering paths
- [x] \`EmailExceptionProcessorTest\` — new unit tests for exception processing
- [x] \`HighVolumeMessageConsumerTest\` — new unit tests for the enabled-flag guard and delegation
- [x] \`CloudEventHistoryBuilderTest\` — updated to cover new fields
- [x] \`EmailConnectorIntegrationTest\` — updated end-to-end integration test (renamed from the Camel-era test)
- [x] \`HttpExceptionHandlerTest\` — updated to cover response body capture
- [x] \`MessageConsumerTest\` — updated to cover header-filtering-before-extraction and \`handleMessage()\`

## Summary by CodeRabbit

* **New Features**
  * Email notifications support high-volume event processing, recipient partitioning, daily digests, and inline or remotely retrieved payloads.
  * Email delivery failures include HTTP response details and associated payload identifiers when available.
  * Engine requests automatically retry transient failures.
* **Bug Fixes**
  * Unsupported or missing connector headers are acknowledged and discarded without recording processing failures.
  * HTTP error responses are safely captured and limited to 1,000 characters.
* **Tests**
  * Expanded coverage for email delivery, recipient handling, high-volume events, retries, error responses, and message filtering.


larrymou9:
❌ **Performance Tests failed**

| | |
|---|---|
| **PR** | #4539 |
| **Branch** | `RHCLOUD-44976_migrateEmailConnectorOutOfCamel` |
| **SHA** | `193eb56a352eed6d5a02bbc12dfff37b5eec60ff` |
| **Build** | [#227](https://jenkins-csb-perf-master.dno.corp.redhat.com/job/InsightsNotifications_PRTest_builder/227/) |
| **Image(s) tested** | `notifications-backend:on-pr-193eb56a352eed6d5a02bbc12dfff37b5eec60ff`<br>`notifications-engine:on-pr-193eb56a352eed6d5a02bbc12dfff37b5eec60ff`<br>`notifications-connector-email:on-pr-193eb56a352eed6d5a02bbc12dfff37b5eec60ff`<br>`notifications-connector-webhook:on-pr-193eb56a352eed6d5a02bbc12dfff37b5eec60ff`<br>`notifications-recipients-resolver:on-pr-193eb56a352eed6d5a02bbc12dfff37b5eec60ff` |

Tests run: InsightsNotificationswebhook_runner · InsightsNotificationsgw_runner · InsightsNotificationsemail_runner

<details>
<summary>📊 Performance Results</summary>

**InsightsNotificationswebhook_runner** — ⚠️ build `FAILURE`, no performance results uploaded ([console](https://jenkins-csb-perf-master.dno.corp.redhat.com/job/InsightsNotificationswebhook_runner/835/))

**InsightsNotificationsgw_runner**

| Metric | Result | Value | Min | Max |
|--------|--------|-------|-----|-----|
| `notifications-engine-service.restarts.sum` | ✅ PASS | 0.000 | 0.000 | 0.000 |
| `notifications-gw-service.restarts.sum` | ✅ PASS | 0.000 | 0.000 | 0.000 |
| `SUMMARY.total_rps` | ❌ FAIL | 13.983 | 14.152 | 14.590 |
| `SUMMARY.median_response_time` | ✅ PASS | 10.000 | 10.000 | 12.000 |
| `POST_notifications.avg_response_time` | ❌ FAIL | 30.191 | 24.948 | 28.899 |
| `SUMMARY.num_failures` | ✅ PASS | 0.000 | 0.000 | 0.000 |
| `SUMMARY.fail_ratio` | ✅ PASS | 0.000 | 0.000 | 0.000 |

**InsightsNotificationsemail_runner** — ⚠️ build `FAILURE`, no performance results uploaded ([console](https://jenkins-csb-perf-master.dno.corp.redhat.com/job/InsightsNotificationsemail_runner/843/))

</details>

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/CloudEventHistoryBuilder.java
```diff
@@ -1,57 +1,51 @@
...
-
-import static com.redhat.cloud.notifications.connector.email.constants.ExchangeProperty.ADDITIONAL_ERROR_DETAILS;
 
 @ApplicationScoped
 @Alternative
 @Priority(0) // The value doesn't matter.
 public class CloudEventHistoryBuilder extends HttpOutgoingCloudEventBuilder {
 
     public static final String TOTAL_RECIPIENTS_KEY = "total_recipients";
+    public static final String ADDITIONAL_ERROR_DETAILS = "additionalErrorDetails";
```
gwenneg: `ADDITIONAL_ERROR_DETAILS` is declared here but never referenced anywhere in this class or elsewhere in the email connector — `buildFailure()` only uses `TOTAL_RECIPIENTS_KEY` and `PAYLOAD_DETAILS_ID_KEY`. It also duplicates the identical constant already defined and used in the parent `HttpOutgoingCloudEventBuilder.ADDITIONAL_ERROR_DETAILS`, which is what actually surfaces `responseBody` in the failure JSON via `super.buildFailure()`. Suggest removing this unused/duplicate constant.

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/HighVolumeMessageConsumer.java
```diff
@@ -0,0 +1,36 @@
...
+    @Inject
+    MessageConsumer messageConsumer;
+
+    @Incoming("highvolumemessages")
+    @Blocking("high-volume-thread-pool")
+    @RunOnVirtualThread
+    public CompletionStage<Void> processMessage(Message<JsonObject> message) {
+        if (!emailConnectorConfig.isIncomingKafkaHighVolumeTopicEnabled()) {
+            Log.warnf("Kafka message rejected because high-volume topic is disabled on this connector");
+            return message.ack();
```
gwenneg: Before this change, the high-volume Kafka route only existed when `isIncomingKafkaHighVolumeTopicEnabled()` was true (see the old `EmailRouteBuilder`), so disabling the toggle meant the connector never subscribed and messages simply stayed queued in Kafka. Now `@Incoming("highvolumemessages")` is always subscribed, and when the toggle is off this handler acks the message here, permanently discarding it instead of leaving it unconsumed. Since the producer-side toggle (`EngineConfig.isOutgoingKafkaHighVolumeTopicEnabled`) is a separate, independently-configurable flag, an operator disabling only this connector-side toggle (e.g. during an incident) while the engine keeps producing to the high-volume topic would silently drop those notifications rather than deferring them. Worth considering not acking (or nacking) instead, or gating the `@Incoming` subscription itself the way the old Camel route did.

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/EmailMessageHandler.java
```diff
@@ -0,0 +1,237 @@
...
+        JsonObject dataToProcess = cloudEventData;
+
+        if (payloadId != null) {
+            try {
+                UUID.fromString(payloadId);
+            } catch (IllegalArgumentException e) {
+                throw new IllegalStateException("Invalid payload ID format (expected UUID): " + payloadId, e);
+            }
+            handledMessageDetails.payloadId = payloadId;
+            PayloadDetails payloadDetails = internalEngine.getPayloadDetails(payloadId);
```
gwenneg, line 143: For large payloads fetched here, `OutgoingCloudEventBuilder.buildMessage()` (connector-common-v2) logs `orgId`/`endpointId` by re-parsing `incomingCloudEvent.getData()` — the raw, unmodified Kafka message envelope — for both the success and failure log lines. `ConnectorSender.send()` (engine side) replaces the outgoing payload with just `{payload_details_id: ...}` before queuing it when it's offloaded to the database, so that envelope never contains `org_id`/`endpoint_id` for this path. The result: the shared "Notification sent [orgId=..., EndpointId=...]" log line will always print `orgId=null, EndpointId=null` here, even though the real org/endpoint IDs are resolved right here from the fetched payload. The previous Camel-based `EmailCloudEventDataExtractor` explicitly reset the exchange's `ORG_ID`/`ENDPOINT_ID` properties from the fetched payload for exactly this reason; there's no equivalent hook in the v2 framework, so org-scoped log tracing breaks for this path.

gwenneg (commented):
A few specific points from a review of this change.

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/EmailMessageHandler.java
```diff
@@ -0,0 +1,237 @@
...
+
+        final HandledEmailMessageDetails handledMessageDetails = new HandledEmailMessageDetails();
+
+        // Phase 1: Extract and resolve payload
+        final EmailNotification emailNotification = extractPayload(incomingCloudEvent.getData(), handledMessageDetails);
+
+        // Phase 2: Render templates
+        String renderedSubject;
+        String renderedBody;
+        final String orgId = emailNotification.orgId();
```
gwenneg, line 92: This drops the `oldOrgId()` fallback that the previous Camel-based `EmailCloudEventDataExtractor` applied for daily-digest events: `if (exchange.getProperty(ORG_ID) == null) { exchange.setProperty(ORG_ID, emailNotification.oldOrgId()); }`. `EmailNotification.oldOrgId()` is documented as "a typo that was introduced a long time ago ... We need to keep both for the next production rollout" — that comment is still in `EmailNotification.java` and the field is unchanged. If a daily-digest event still emits the legacy `orgId` key instead of `org_id`, `orgId` here resolves to `null`, which is passed straight into `EmailAggregationProcessor.aggregate()` and used for `isUseBetaTemplatesEnabled(orgId, ...)` — silently breaking beta-template targeting for that org. Worth confirming the legacy key is no longer produced anywhere before dropping this fallback.

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/EmailMessageHandler.java
```diff
@@ -0,0 +1,200 @@
...
+        String renderedSubject;
+        String renderedBody;
+        if (emailNotification.isDailyDigest()) {
+            EmailAggregation emailAggregation = objectMapper.convertValue(emailNotification.eventData(), EmailAggregation.class);
+            renderedSubject = renderEmailAggregationTitle(emailAggregation.bundleDisplayName());
+            renderedBody = emailAggregationProcessor.aggregate(emailAggregation, orgId, renderedSubject);
+        } else {
+            renderedBody = renderInstantEmail(emailNotification, IntegrationType.EMAIL_BODY);
+            renderedSubject = renderInstantEmail(emailNotification, IntegrationType.EMAIL_TITLE);
+        }
```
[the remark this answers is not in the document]
g-duval: It should be used like this.

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/EmailMessageHandler.java
```diff
@@ -0,0 +1,200 @@
...
+        if (emailConnectorConfig.isEmailsInternalOnlyEnabled()) {
+            Set<String> forbiddenEmail = emails.stream()
+                .filter(email -> !email.trim().toLowerCase().endsWith("@redhat.com"))
+                .collect(Collectors.toSet());
+            if (!forbiddenEmail.isEmpty()) {
+                Log.warnf(" %s emails are forbidden for message historyId: %s ", forbiddenEmail, incomingCloudEvent.getId());
+            }
+            emails.removeAll(forbiddenEmail);
+        }
+        recipientsList.addAll(emails);
```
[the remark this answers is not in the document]
g-duval: Yes that's on purpose, this restriction must not be applied to org recipients, only additional recipients.

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/EmailMessageHandler.java
```diff
@@ -0,0 +1,200 @@
...
+            .stream().map(User::getEmail).filter(email -> email != null && !email.isBlank()).collect(toSet());
+        recipientsResolverResponseTimeMetric.stop(meterRegistry.timer(RECIPIENTS_RESOLVER_RESPONSE_TIME_METRIC));
+        return recipientsList;
+    }
+
+    private void sendToBop(Set<String> recipientsList, String subject, String body,
+                           String sender, String orgId, String historyId) {
+        List<List<String>> packedRecipients = partition(recipientsList,
+            emailConnectorConfig.getMaxRecipientsPerEmail() - 1);
+
```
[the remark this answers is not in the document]
g-duval: it is controlled by admins only, it makes no sens to set `maxRecipientsPerEmail < 1`. I don't think adding syntactic sugar here will help.

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/HighVolumeMessageConsumer.java
```diff
@@ -0,0 +1,109 @@
...
+        try {
+            Log.debugf("Processing high-volume message %s", message.getPayload());
+            HandledMessageDetails additionalConnectorDetails = messageHandler.handle(cloudEventMetadata);
+            outgoingMessageSender.sendSuccess(cloudEventMetadata, additionalConnectorDetails, startTime);
+            success = true;
+        } catch (Exception e) {
+            Log.errorf(e, "Error processing high-volume message: %s", e.getMessage());
+            HandledExceptionDetails processedExceptionDetails = exceptionProcessor.processException(e, cloudEventMetadata);
+            outgoingMessageSender.sendFailure(cloudEventMetadata, processedExceptionDetails, startTime);
+        } finally {
```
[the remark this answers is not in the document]
g-duval: fixed

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/CloudEventHistoryBuilder.java
g-duval: Done

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/EmailMessageHandler.java
g-duval: fixed

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/EmailMessageHandler.java
g-duval: I removed oldOrgId because it is not provided by engine anymore, since few months

connector-email/src/main/java/com/redhat/cloud/notifications/connector/email/CloudEventHistoryBuilder.java
```diff
@@ -20,38 +19,32 @@ public class CloudEventHistoryBuilder extends HttpOutgoingCloudEventBuilder {
...
-            data.put(PayloadDetails.PAYLOAD_DETAILS_ID_KEY, payloadId);
-
-            exchange.removeProperty(ExchangeProperty.PAYLOAD_ID);
+    @Override
+    public JsonObject buildFailure(HandledExceptionDetails processedExceptionDetails) {
+        JsonObject data = super.buildFailure(processedExceptionDetails);
+        if (!data.containsKey("details")) {
+            data.put("details", new JsonObject());
         }
+        data.getJsonObject("details").put(TOTAL_RECIPIENTS_KEY, 0);
```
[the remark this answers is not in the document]
g-duval: If BOP fails, that means we didn't send email successfully to recipients, so we assume none of recipients received it, that's why we set it to 0.

gwenneg (approved):
LGTM

PR #4597, opened 2026-06-04
Add tls support for valkey integration

g-duval:
## Problem

In stage, Clowder provisions Valkey with TLS and sets `quarkus.redis.hosts` to a `valkeys://` URL. `ValkeyService` was building its `RedisOptions` with two bugs:

1. The scheme replacement only handled `valkey://` → `redis://`, leaving `valkeys://` untouched instead of mapping it to `rediss://`.
2. When the resulting connection string uses `rediss://` (TLS), Vert.x requires an explicit hostname verification algorithm on the net client options. Without it, the connection fails with:
   ```
   Missing hostname verification algorithm: you must set TCP client options host name verification algorithm
   ```

## Fix

- Added `valkeys://` → `rediss://` scheme mapping alongside the existing `valkey://` → `redis://` one.
- When the connection string starts with `rediss://`, set `hostnameVerificationAlgorithm("HTTPS")` on the `NetClientOptions`.


gwenneg (approved):
LGTM

PR #4602, opened 2026-06-05
[RHCLOUD-48189] Lower log level

g-duval:
When a connector fails to deliver a notification (e.g. DNS resolution failure, connection timeout, bad HTTP response), MessageConsumer.processMessage() catches the exception and unconditionally logs it at ERROR level with a full stack trace — before passing it to the ExceptionHandler.
  
  The problem is that these exceptions are already handled correctly: HttpExceptionHandler classifies them (e.g. UnknownHostException → UNKNOWN_HOST) and the failure is reported back to the engine. They represent expected delivery failures, often caused by user-misconfigured endpoints, not bugs in the connector itself.

  Logging them at ERROR causes noise in monitoring dashboards and can trigger false alerts. Lowering to INFO keeps the information visible in logs for debugging without polluting error-level observability signals.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated exception logging behavior in message processing to accurately reflect error severity.


gwenneg (approved):
LGTM

PR #4629, opened 2026-06-17
chore: set up togi

gwenneg:
Sets up [togi](https://github.com/gwenneg/togi). Every time Claude stumbles in this repo — a wrong assumption, a missing convention, a denied command — that's a gap in our context docs. Togi captures those moments and turns them into doc PRs, so the same stumble doesn't happen twice.

  **Is this safe to merge?** Yes — and you can verify it from the diff alone. The entire change is three inert text files: the adoption note `adopt-togi.md`, a pointer in `README.adoc`, and `.gitignore` entries. No settings, no hooks, no code: merging installs nothing and runs nothing on anyone's machine. If the diff shows anything beyond those files, reject this PR.

  **Trying it costs a minute.** Run the commands in `adopt-togi.md`, work normally, and check `.togi/friction/pending/` after a few sessions. Capture is opt-in per developer and runs inside your own session — the model writes a short note when it hits friction, nothing is sent to any third party, and there's no meaningful added cost. Haven't opted in? You'll see a one-time notice and nothing else will ever run. Leave any time with `/togi:disable`.


g-duval (approved):
LGTM

PR #4670, opened 2026-06-30
Update notifications-aggregator for cluster selection via unleash flag

bonscji1:
Add code that allows for cluster selection for aggregator from unleash, helping with aggregator cluster migrations. 

Assisted-by: Claude Sonnet 4.5 (via Claude Code)

## Summary by CodeRabbit

* **New Features**
  * Added optional cluster coordination configuration for the notifications aggregator (including a new cluster identifier setting).
  * Daily email aggregation now runs only on the matching active cluster, using a safe fallback when the active cluster can’t be determined.
  * Startup logging now includes the configured cluster and active cluster status when available.
* **Bug Fixes**
  * Prevents duplicate processing by skipping aggregation when the cluster identifiers don’t match.
* **Tests**
  * Added coverage for the daily email aggregation cluster-coordination scenarios.

### DIFF
--- a/aggregator/src/main/java/com/redhat/cloud/notifications/config/AggregatorConfig.java
+++ b/aggregator/src/main/java/com/redhat/cloud/notifications/config/AggregatorConfig.java
@@ -1,29 +1,95 @@
 package com.redhat.cloud.notifications.config;
 
+import com.redhat.cloud.notifications.unleash.ToggleRegistry;
+import io.getunleash.Unleash;
+import io.getunleash.variant.Payload;
+import io.getunleash.variant.Variant;
 import io.quarkus.logging.Log;
+import jakarta.annotation.PostConstruct;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.event.Observes;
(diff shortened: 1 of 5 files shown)


aggregator/src/main/java/com/redhat/cloud/notifications/config/AggregatorConfig.java
```diff
@@ -1,29 +1,95 @@
...
+
+    /**
+     * Gets the active cluster ID from Unleash variant payload.
+     * Returns Optional.empty() if:
+     * - Unleash is disabled (variant will be disabled)
+     * - Unleash is unreachable
+     * - Variant is not enabled
+     * - Payload is missing or invalid
+     */
+    public Optional<String> getActiveCluster() {
```
g-duval, line 59: Because it will not be available in all clusters, log levels should be Info only to avoid Prods alerting.

g-duval (commented):
Looks quite good, could you please lower startup log level?

g-duval (approved):
LGTM

PR #4677, opened 2026-07-02
RHCLOUD-46541: feat(event-log): Add hasAction filter to exclude events with no notification actions

Bannermanj:
## Summary by CodeRabbit

* **New Features**
  * Added a new `hasAction` query filter to the events endpoint to narrow results to events with related actions.
  * Event listings and counts now accept and apply the `hasAction` filter consistently.

* **Bug Fixes**
  * Improved filtering logic so action-based constraints are applied uniformly across both event retrieval and count responses.

* **Tests**
  * Added parameterized test coverage for `hasAction` behavior, validating default vs action-only results.


backend/src/main/java/com/redhat/cloud/notifications/db/repositories/EventRepository.java
```diff
@@ -381,6 +384,10 @@ private String addHqlConditions(String hql, boolean useNormalized, boolean bundl
             hql += " AND EXISTS (SELECT 1 FROM NotificationHistory nh WHERE nh.event = e AND " + String.join(" AND ", subQueryConditions) + ")";
         }
 
+        if (hasAction) {
```
bonscji1: I think changing this to:
`}else if (hasAction) {
            hql += " AND EXISTS (SELECT 1 FROM NotificationHistory nh WHERE nh.event = e)";
 }`
 would make sense as going into above path will return requested data without adding the same part into the sql again. what do you think?

gwenneg: I agree. We don't want an additional subquery from the new code while the previous subquery already filters out entries with no actions implicitely. That would be bad for performances.

bonscji1 (approved):

PR #4679, opened 2026-07-02
Fix init container write path to use Clowder-provided mount

g-duval:
Clowder ignores init container volumeMounts and copies main container's volumeMounts instead. The main container mounts it-services-cert-decoded at /mnt/it-services-cert-autorenew, so the init container must write to that path, not /mnt/it-services-cert-decoded.

Error was: /mnt/it-services-cert-decoded/keystore.p12: No such file or directory

The init container's volumeMounts definition in the ClowdApp is ignored by Clowder - it uses the main container's paths instead.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved certificate keystore handling during startup so the decoded certificate is stored in the correct location, helping renewals work more reliably.


dehort (approved):
Looks good to me!

PR #4722, opened 2026-07-16
scheduler templates - remove query parameters from export urls

dehort:
scheduler templates - remove query parameters from export urls

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated scheduled export download links in notifications and emails to use the direct export URL without unnecessary query parameters.
  * Improved link consistency across export completion messages.

* **Tests**
  * Updated validation to reflect the simplified export download URL.


g-duval:
@dehort , long time ago, it was requested to add those tracking parameters in all links generated by Notifications.
If you remove it from Scheduler templates, it will bring an inconsistency with all other templates from others services.
Are you sure you want to do that? 🧐

dehort:
@g-duval I see.  I think I made a mistake in my testing.  The export-service will ignore these extra parameters.  So this change is not needed.  Sorry for the trouble!

PR #4735, opened 2026-07-21
Revert --immutable on admin-console yarn install

g-duval:
Renovate/Mintmaker's bot environment runs global Yarn Classic 1.22.22 without corepack enabled, so it can no longer regenerate yarn.lock now that package.json pins packageManager to yarn@4.17.0. This leaves dependency-bump PRs (e.g. #4726) with a package.json/yarn.lock mismatch that --immutable then turns into a hard CI failure. Dropping --immutable restores the previous tolerant behavior so those PRs can build while the underlying Renovate/corepack incompatibility is addressed separately.

## Summary by CodeRabbit

* **Chores**
  * Updated the build process to allow Yarn dependencies to be installed without enforcing an immutable lockfile.

### DIFF
--- a/admin-console/pom.xml
+++ b/admin-console/pom.xml
@@ -37,7 +37,7 @@
                             <goal>corepack</goal>
                         </goals>
                         <configuration>
-                            <arguments>yarn install --immutable</arguments>
+                            <arguments>yarn install</arguments>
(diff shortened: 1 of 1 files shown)


gwenneg (approved):
LGTM

PR #4761, opened 2026-07-27
[RHCLOUD-49663] Paginate findEventsToExport and stream export output

g-duval:
## Summary

Exporting events via the export-service integration (CSV/JSON) is slow for orgs with large event
volumes: exporting a date range covering ~260,000 events took more than a minute.

`EventRepository#findEventsToExport(orgId, from, to)` ran a single unbounded query with no
`ORDER BY` / `LIMIT`, then loaded the *entire* matching result set (up to 260k `Event` objects)
into one `List<Event>` before handing it to the CSV/JSON transformers — which then built a
*second*, fully separate in-memory representation from it, so both the raw events and the
rendered output were held in memory at the same time.

- **Keyset (seek) pagination** instead of unbounded/offset pagination — `findEventsToExport` now
  loops over bounded pages using `WHERE (created, id) > (:lastCreated, :lastId) ORDER BY created,
  id LIMIT n`. Keyset was chosen over `OFFSET`/`LIMIT` because this is a one-time, front-to-back
  sequential export, and `OFFSET` would re-skip every prior row on each page.
- **Two-phase "find IDs, then fetch"** — reusing the pattern already used by the backend module's
  `GET /events` endpoint: a cheap, index-only query finds the next page of matching event IDs,
  then a bounded query fetches the full projection (display names) only for that page's IDs.
- **Streaming the output instead of double-materializing it** — `findEventsToExport` no longer
  returns a `List<Event>`; it streams each page to a caller-supplied `PageConsumer` as soon as
  it's fetched. `EventExporterService` wires that callback directly into the CSV/JSON transformer
  (now built around an incremental `addRecords(page)` / `finish()` API instead of a one-shot
  `transform(List)`), so events are appended straight into the CSV/JSON output as they arrive from
  the database, and only converted to a final `String` once, at the very end. The full event list
  and the fully rendered export output are never both held in memory at once — only one page
  (default 5,000, configurable via `notifications.events.export.page-size`) plus the growing
  output buffer.

## What changed (engine module)

- `EngineConfig` — new `notifications.events.export.page-size` config property (default `5000`).
- `EventRepository` — `findEventsToExport` rewritten as a keyset-paginated, two-phase, page-
  streaming method; signature changed from returning `List<Event>` to accepting a
  `PageConsumer<Event>` callback.
- New `PageConsumer<T>` functional interface.
- `ResultsTransformer<T>` — changed from `transform(List<T>)` to `addRecords(List<T>)` / `finish()`.
- `CSVEventTransformer` / `JSONEventTransformer` — build their output incrementally, page by page.
- `EventExporterService` — streams repository pages directly into the transformer.
- Tests updated accordingly, plus a new parameterized pagination test asserting events are
  returned correctly, in order, across multiple pages, for both normalized and denormalized query
  modes.

Jira: https://redhat.atlassian.net/browse/RHCLOUD-49663

## Test plan

- [x] `./mvnw validate -pl :notifications-engine` — checkstyle clean
- [x] `./mvnw clean verify -pl :notifications-engine -am` — full build green, all engine tests
      passing
- [ ] `EXPLAIN (ANALYZE, BUFFERS)` against a staging org with ~260k events (follow-up)

## Summary by CodeRabbit

* **New Features**
  * Event exports now process results in configurable pages, improving efficiency for large exports.
  * Added configuration for export page size, with a default of 5,000 events.
  * CSV and JSON exports now support combining multiple result pages into one complete file.

* **Bug Fixes**
  * Improved export ordering and consistency when events share the same creation timestamp.
  * Invalid page-size configurations now fail fast with a clear error.


larrymou9:
✅ **Performance Tests passed**

| | |
|---|---|
| **PR** | #4761 |
| **Branch** | `RHCLOUD-49663-events-export-pagination` |
| **SHA** | `705082dacf5982b0792c97abcffe7931d505b4e8` |
| **Build** | [#57](https://jenkins-csb-perf-master.dno.corp.redhat.com/job/InsightsNotifications_PRTest_builder/57/) |
| **Image tested** | `quay.io/redhat-user-workloads/hcc-integrations-tenant/notifications/notifications-backend:on-pr-705082dacf5982b0792c97abcffe7931d505b4e8` |

Tests run: InsightsNotificationswebhook_runner · InsightsNotificationsgw_runner · InsightsNotificationsemail_runner

engine/src/main/java/com/redhat/cloud/notifications/exports/transformers/event/CSVEventTransformer.java
```diff
@@ -15,34 +15,55 @@ public final class CSVEventTransformer implements ResultsTransformer<Event> {
...
+    public CSVEventTransformer() throws TransformationException {
         final CSVFormat csvFormat = CSVFormat.DEFAULT
             .builder()
             .setHeader(CSV_HEADERS)
             .setRecordSeparator(System.lineSeparator())
             .build();
 
-        final StringWriter stringWriter = new StringWriter();
+        try {
+            this.csvPrinter = new CSVPrinter(this.stringWriter, csvFormat);
```
bonscji1: can we clean out `this.` notation? i know its only cosmetic but it got me wondering if something is wrong and we have to use it. its not, its just redundant

engine/src/main/java/com/redhat/cloud/notifications/exports/transformers/event/JSONEventTransformer.java
```diff
@@ -27,9 +27,16 @@ public String transform(final List<Event> events) {
             jsonEvent.put("eventType", event.getEventTypeDisplayName());
             jsonEvent.put("created", event.getCreated().toInstant(ZoneOffset.UTC));
 
-            jsonEvents.add(jsonEvent);
+            this.jsonEvents.add(jsonEvent);
```
bonscji1: same `this.` here.

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EventRepository.java
```diff
@@ -32,19 +34,121 @@ public Event create(Event event) {
...
-     * @return a list of events that comply with the provided filters.
+     * @param pageConsumer the callback invoked with each page of events that
+     *                      comply with the provided filters, in ascending
+     *                      "created" order.
+     * @throws TransformationException if the page consumer fails to handle a
+     * page.
      */
-    public List<Event> findEventsToExport(final String orgId, final LocalDate from, final LocalDate to) {
-        final StringBuilder findEventsQuery = new StringBuilder();
+    public void findEventsToExport(final String orgId, final LocalDate from, final LocalDate to, final PageConsumer<Event> pageConsumer) throws TransformationException {
```
bonscji1: I dont really like this method as i feel it imports into the repository aspects that are not in line with just db data handling. 
How would you feel about using Iterator pattern instead? so something like:
```
public Iterator<List<Event>> findEventsToExport(final String orgId, final LocalDate from, final LocalDate to) {
      final Timestamp createdMin = from == null ? null : Timestamp.valueOf(from.atStartOfDay());
      final Timestamp createdMax = to == null ? null : Timestamp.valueOf(to.atTime(LocalTime.MAX));
      final boolean normalizedQueries = engineConfig.isNormalizedQueriesEnabled(orgId);
      final int pageSize = engineConfig.getEventsExportPageSize();

      return new Iterator<>() {
          Timestamp cursorCreated = null;
          UUID cursorId = null; 
          boolean exhausted = false;
          List<Event> nextPage = fetchNext();
 
          private List<Event> fetchNext() {
              if (exhausted) {
              }   
              final List<Object[]> idsPage = findEventIdsPage(orgId, createdMin, createdMax, cursorCreated, cursorId, pageSize);
              if (idsPage.isEmpty()) {
                  exhausted = true;
                  return null;
              }   
              final List<UUID> pageIds = idsPage.stream().map(row -> (UUID) row[0]).toList();
              final Object[] lastRow = idsPage.get(idsPage.size() - 1);
              cursorId = (UUID) lastRow[0];
              cursorCreated = (Timestamp) lastRow[1];
              if (idsPage.size() < pageSize) {
                  exhausted = true;
              }   
              return findEventsByIds(pageIds, normalizedQueries);
          }   
          
          @Override
          public boolean hasNext() {
              return nextPage != null;
          }   
          
          @Override
          public List<Event> next() {
              final List<Event> current = nextPage;
              nextPage = fetchNext();
              return current;
          }
      };
  }
```
and then on service side:
```
`final Iterator<List<Event>> pages = this.eventRepository.findEventsToExport(orgId, eventFilters.from(), eventFilters.to());
  while (pages.hasNext()) {
      transformer.addRecords(pages.next());
  }`
```

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EventRepository.java
```diff
@@ -32,19 +34,121 @@ public Event create(Event event) {
...
+        final Timestamp createdMin = from == null ? null : Timestamp.valueOf(from.atStartOfDay());
```
bonscji1, line 53: This is a nitpick 😅 
 I would personally move this to service logic and have the method accept timestamps instead of Localdate. i would also prefer to change all timestamp references to LocalDateTime since we use it everywhere else apart from engine and backend but that is discussion for another time.

engine/src/test/java/com/redhat/cloud/notifications/db/repositories/EventRepositoryTest.java
bonscji1, line 91: The new pagination test's fixture events all have distinct created timestamps, so the tie-break branch (e.created = :cursorCreated AND e.id > :cursorId) in the keyset cursor is never exercised. A bug there could duplicate or drop  events at a page boundary in production, undetected by tests.  

Should be easily fixable by new test specifically for this fallback.

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EventRepository.java
```diff
@@ -32,19 +34,121 @@ public Event create(Event event) {
...
+
+            final Object[] lastRow = idsPage.get(idsPage.size() - 1);
+            cursorId = (UUID) lastRow[0];
+            cursorCreated = (Timestamp) lastRow[1];
+
+            if (idsPage.size() < pageSize) {
+                break;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
g-duval: Agreed this is a real gap, but I don'\''t think a full fix is worth it here.

Guaranteeing a single consistent snapshot across the whole export would mean wrapping the entire multi-page loop in one long-lived transaction (Postgres `REPEATABLE READ`/`SERIALIZABLE`, since the default `READ COMMITTED` re-snapshots per statement). For a 260k-row export that's exactly the kind of long-held DB resource (open snapshot, potential vacuum bloat, held connection) this PR is trying to get rid of, and it'\''d diverge from the `GET /events` endpoint pattern this code intentionally mirrors, which has the same characteristic today.

The race is also narrow: it only bites when `to` is unbounded/covers "now", since `created` is assigned at object-construction time and a slow-committing transaction could in theory land behind an already-passed cursor. For a one-time, operator-triggered export tool, I'\''m going to accept that as a known limitation rather than pay the resource cost. Happy to revisit with a lighter mitigation (e.g. clamping the effective `to` to the export'\''s start time) if this becomes a real problem in practice.

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EventRepository.java
g-duval: Fair point, and I agree LocalDate/LocalDateTime everywhere would be more consistent, but since you flagged it as a broader, separate discussion I'\''m leaving `Timestamp`/`LocalDate` as-is here to keep this PR scoped to the pagination change. Happy to pick it up separately.

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EventRepository.java
g-duval: Good call, done. Switched to `Iterator<List<Event>>`, but implemented lazily (fetch happens inside `next()`, no read-ahead of a full page of events) so the "only one page of events in memory at a time" property from the PR description still holds. The only thing cached between calls is the small, index-only "next page of ids" lookahead used to answer `hasNext()` correctly, which is cheap and doesn't duplicate the full projection.

engine/src/main/java/com/redhat/cloud/notifications/exports/transformers/event/CSVEventTransformer.java
g-duval: Done.

engine/src/main/java/com/redhat/cloud/notifications/exports/transformers/event/JSONEventTransformer.java
g-duval: Done.

engine/src/main/java/com/redhat/cloud/notifications/db/repositories/EventRepository.java
```diff
@@ -32,19 +34,121 @@ public Event create(Event event) {
...
-    public List<Event> findEventsToExport(final String orgId, final LocalDate from, final LocalDate to) {
-        final StringBuilder findEventsQuery = new StringBuilder();
+    public void findEventsToExport(final String orgId, final LocalDate from, final LocalDate to, final PageConsumer<Event> pageConsumer) throws TransformationException {
+        final Timestamp createdMin = from == null ? null : Timestamp.valueOf(from.atStartOfDay());
+        final Timestamp createdMax = to == null ? null : Timestamp.valueOf(to.atTime(LocalTime.MAX));
 
-        if (engineConfig.isNormalizedQueriesEnabled(orgId)) {
-            findEventsQuery.append(
+        final boolean normalizedQueries = engineConfig.isNormalizedQueriesEnabled(orgId);
+        final int pageSize = engineConfig.getEventsExportPageSize();
```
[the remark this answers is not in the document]
g-duval: Done, now fails fast with an `IllegalStateException` if `notifications.events.export.page-size` is configured to a non-positive value.

engine/src/test/java/com/redhat/cloud/notifications/db/repositories/EventRepositoryTest.java
g-duval: Added `testGetAllPaginatedTiedCreatedTimestamps`: four fixtures sharing the exact same `created` timestamp, with the page size forced so a page boundary falls in the middle of the tied group. Asserts no duplicates/omissions and that the tied group comes back in ascending `id` order (had to account for Postgres comparing its `uuid` type byte-wise/unsigned, which differs from `UUID#compareTo`'\''s signed comparison).

bonscji1 (approved):

larrymou9:
❌ **Performance Tests failed**

| | |
|---|---|
| **PR** | #4761 |
| **Branch** | `RHCLOUD-49663-events-export-pagination` |
| **SHA** | `2730f48390b32ed793356348625b69e7ed4a4607` |
| **Build** | [#58](https://jenkins-csb-perf-master.dno.corp.redhat.com/job/InsightsNotifications_PRTest_builder/58/) |
| **Image tested** | `quay.io/redhat-user-workloads/hcc-integrations-tenant/notifications/notifications-backend:on-pr-2730f48390b32ed793356348625b69e7ed4a4607` |

Tests run: InsightsNotificationswebhook_runner · InsightsNotificationsgw_runner · InsightsNotificationsemail_runner

<details>
<summary>📊 Performance Results</summary>

| Metric | Result | Value | Min | Max |
|--------|--------|-------|-----|-----|
| `notifications-engine-service.restarts.sum` | ✅ PASS | 0.000 | 0.000 | 0.000 |
| `notifications-connector-email-service.restarts.sum` | ✅ PASS | 0.000 | 0.000 | 0.000 |
| `SUMMARY.total_rps` | ❌ FAIL | 14.635 | 13.966 | 14.498 |
| `SUMMARY.median_response_time` | ✅ PASS | 11.000 | 10.000 | 12.000 |
| `POST_notifications.avg_response_time` | ❌ FAIL | 24.351 | 25.356 | 30.688 |
| `SUMMARY.num_failures` | ✅ PASS | 0.000 | 0.000 | 0.000 |
| `SUMMARY.fail_ratio` | ✅ PASS | 0.000 | 0.000 | 0.000 |
| `notifications-engine-service.restarts.sum` | ✅ PASS | 0.000 | 0.000 | 0.000 |
| `notifications-connector-webhook-service.restarts.sum` | ✅ PASS | 0.000 | 0.000 | 0.000 |
| `results.created_at.duration_stats.mean` | ❌ FAIL | 160.217 | 141.021 | 154.384 |
| `results.created_at.duration_stats.median` | ❌ FAIL | 182.579 | 151.605 | 170.846 |
| `results.created_at.rps_stats.mean` | ✅ PASS | 3.475 | 3.240 | 3.679 |
| `results.created_at.rps_stats.median` | ❌ FAIL | 2.800 | 2.900 | 3.200 |

</details>

